PR #82, opened 2022-05-25
feat: enable TypeScript rule useUnknownInCatchVariables

n1ru4l:


B2o5T (approved):

packages/web/app/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "dev": "next dev",
     "start": "next start",
-    "build": "bob runify --single"
+    "build": "bob runify --single",
+    "check:ts": "tsc"
```
kamilkisiela, line 9: You can do `"typecheck": "tsc --noEmit"` and it will run as part of CI

n1ru4l:
```suggestion
    "typecheck": "tsc"
```

--noEmit is not necessary as it is already set within the tsconfig.json

kamilkisiela: okok

PR #84, opened 2022-05-25
feat: remove unused exports

n1ru4l:
This PR removes all the dead and unused code.
Also, it enables the eslint rule for avoiding committing dead code in the future.


packages/web/app/src/components/v2/icon.tsx
```diff
@@ -149,21 +149,6 @@ export const ArrowDownIcon: FC<IconProps> = ({ className }) => (
   </svg>
 );
 
-export const MailIcon: FC<IconProps> = ({ className, ...props }) => (
```
B2o5T, line 152: Let’s keep icons from this file, we’ll add storybook after with icons stories

B2o5T (changes requested):
Let’s just keep icons from v2/Icon.tsx, also you can setup this rule to explore unused exports https://github.com/kamilkisiela/graphql-config/blob/master/website/.eslintrc.js#L15

packages/web/app/src/components/v2/icon.tsx
n1ru4l: I don't see why we should keep unused stuff within the code-base? What is the point of it?

B2o5T: I want only keep icons, as they are already dark-mode friendly, we’ll use them in the future

n1ru4l: When we use them you can just pick them from the git history? I don't see the point in having commented/dead code on the `main` branch..

B2o5T: Ok remove it

B2o5T (approved):

PR #132, opened 2022-06-06
Improvements for the Docker layer to work with VirtioFS

dotansimha:


dotansimha:
Closing for now, seems like latest Docker works find with VirtioFS and our existing config.

PR #134, opened 2022-06-07
Fix missing service url in supergraph

### LINKED ISSUE #133: Missing service url in supergraph
When publishing the subgraph (applies to GraphQL API as well) the service url in empty.

    $ hive schema:publish --service users --url http://localhost:6969 supergraph.graphql

kamilkisiela:
Fixes #133


n1ru4l (approved):
always keep it 69

PR #147, opened 2022-06-19
fix: only allow CDN access with the correct access key

n1ru4l:


dotansimha (approved):

n1ru4l:
@dotansimha Note: the tests are still failing.

PR #150, opened 2022-06-20
enable and fixes errors for `@typescript-eslint/no-floating-promises` rule

B2o5T:
fixes https://github.com/kamilkisiela/graphql-hive/issues/148


packages/libraries/client/src/apollo.ts
```diff
@@ -39,7 +39,7 @@ export function hiveApollo(clientOrOptions: HiveClient | HivePluginOptions): Apo
         },
       });
 
-  hive.info();
+  void hive.info();
```
B2o5T, line 42: I think in this line we don't need to await, in cases like this `void` operator can be used [more info on the docs](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md#ignorevoid)

packages/services/api/src/modules/shared/providers/message-bus.ts
```diff
@@ -7,7 +7,7 @@ import Emittery from 'emittery';
 export class MessageBus {
   private emitter = new Emittery();
 
-  async on<TPayload>(event: string, listener: (payload: TPayload) => Promise<void>) {
+  on<TPayload>(event: string, listener: (payload: TPayload) => Promise<void>): void {
```
B2o5T, line 10: this method was never `async`, removed `async` keyword

packages/services/server/src/graphql-handler.ts
```diff
@@ -163,15 +163,14 @@ export const graphqlHandler = (options: GraphQLHandlerOptions): RouteHandlerMeth
         });
 
         response.headers.forEach((value, key) => {
-          reply.header(key, value);
+          reply.header(key, value); // eslint-disable-line @typescript-eslint/no-floating-promises -- false positive, FastifyReply.then returns void
```
B2o5T, line 166: this rule has some issues with `FastifyReply.then` type, more info https://github.com/typescript-eslint/typescript-eslint/issues/4961

packages/services/server/src/graphql-handler.ts
```diff
@@ -163,15 +163,14 @@ export const graphqlHandler = (options: GraphQLHandlerOptions): RouteHandlerMeth
...
         });
 
         if (!reply.hasHeader('x-request-id')) {
-          reply.header('x-request-id', requestId || '');
+          reply.header('x-request-id', requestId || ''); // eslint-disable-line @typescript-eslint/no-floating-promises -- false positive, FastifyReply.then returns void
         }
 
-        reply.status(response.status);
-        reply.send(response.body);
+        reply.status(response.status).send(response.body); // eslint-disable-line @typescript-eslint/no-floating-promises -- false positive, FastifyReply.then returns void
```
B2o5T, line 173: can be chained

packages/services/storage/migrations/migrate.ts
```diff
@@ -29,17 +29,15 @@ const isDownCommand = cmd === 'down';
...
-      process.exit(1);
-    });
+
+  try {
+    await migrator.up();
+    await migrateClickHouse();
+  } catch (error) {
+    console.error(error);
+    process.exit(1);
+  }
```
B2o5T, line 39: `Async/Await` is better, and `top-level await` should also be available, also `process.exit(0)` is redundant and can be removed

packages/services/usage/src/buffer.ts
```diff
@@ -254,7 +254,7 @@ export function createKVBuffer<T>(config: {
       const estimatedBufferSize = currentBufferSize + estimatedReportSize;
 
       if (currentBufferSize >= config.limitInBytes || estimatedBufferSize >= config.limitInBytes) {
-        send(true);
+        void send(true);
```
B2o5T, line 257: I think in this case we also don't need to `await`, so I added `void`

packages/services/usage/src/buffer.ts
```diff
@@ -269,7 +269,7 @@ export function createKVBuffer<T>(config: {
     } else {
       buffer.push(report);
       if (sumOfOperationsSizeInBuffer() >= config.size) {
-        send(true);
+        void send(true);
```
B2o5T, line 272: same

packages/services/schema/src/index.ts
```diff
@@ -21,15 +21,16 @@ async function main() {
...
     name: 'schema',
     tracing: false,
   });
 
   registerShutdown({
     logger: server.log,
     async onShutdown() {
-      await Promise.all([server.close(), redis.disconnect(false)]);
+      await server.close();
+      redis.disconnect(false);
```
B2o5T, line 33: this function do not return `Promise`

packages/services/storage/tools/register.cjs
```diff
@@ -1,3 +0,0 @@
-const { register } = require('esbuild-register/dist/node');
-
-register({ extensions: ['.mjs', '.ts'], format: 'cjs' });
```
B2o5T, line 3: I removed this file for the following reasons
1. removing `esbuild-register`
2. removing `strorage/migrations/package.json`
3. removing `storage/tools/register.cjs`
4. exec `db:migrator` script with `node --experimental-specifier-resolution=node --loader ts-node/esm migrations/migrate.ts` as `ts-node` already installed in package.json

saihaj (approved):

packages/libraries/client/src/client.ts
```diff
@@ -128,8 +128,8 @@ export function createHive(options: HivePluginOptions): HiveClient {
       } else {
         logger.error(`[hive][info] Error ${response.status}: ${response.statusText}`);
       }
-    } catch (error: any) {
-      logger.error(`[hive][info] Error ${error.message}`);
+    } catch (error) {
+      logger.error(`[hive][info] Error ${(error as Error).message}`);
```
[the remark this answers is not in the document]
B2o5T: the most of errors are instances of Error

n1ru4l, line 132:
```suggestion
      logger.error(`[hive][info] Error ${(error as Error)?.message ?? error}`);
```

packages/services/storage/tools/register.cjs
kamilkisiela: the thing is, it has to run in docker and as cli, it it works, I'm fine with the changes.

packages/libraries/client/src/client.ts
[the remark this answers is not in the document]
B2o5T: my newly changes don't change previous behaviour of code

B2o5T:
> Can you confirm we can run `migration:create`, `db:generate` and `db` scripts in `@hive/storage`?

@kamilkisiela yes, everything works like before but now code is simpler

kamilkisiela (approved):

PR #163, opened 2022-06-23
Show warnings in Client when commit, author or token are missing

kamilkisiela:
This should give better feedback to people, because based on issues like SENTRY-2632938996, GraphQL's validation is limiting us.

### DIFF
--- a/packages/libraries/client/src/internal/reporting.ts
+++ b/packages/libraries/client/src/internal/reporting.ts
@@ -3,7 +3,7 @@ import { getDocumentNodeFromSchema } from '@graphql-tools/utils';
 import { createAgent } from './agent';
 import { version } from '../version';
 import type { HivePluginOptions } from './types';
-import { warnIf } from './utils';
+import { logIf } from './utils';
 
 export interface SchemaReporter {
   report(args: { schema: GraphQLSchema }): void;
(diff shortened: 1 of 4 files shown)


packages/libraries/client/src/internal/reporting.ts
```diff
@@ -19,8 +20,20 @@ export function createReporting(pluginOptions: HivePluginOptions): SchemaReporte
 
   const token = pluginOptions.token;
   const reportingOptions = pluginOptions.reporting;
-
   const logger = pluginOptions.agent?.logger ?? console;
+
+  warnIf(
```
charlypoly, line 25: Since it's required by the `publishSchema()` mutation, shouldn't these be errors?

(I've put errors on the Ruby client for this scenario)

kamilkisiela: It could be an error, yes but not an exception.

packages/libraries/client/src/internal/reporting.ts
```diff
@@ -3,6 +3,7 @@ import { getDocumentNodeFromSchema } from '@graphql-tools/utils';
 import { createAgent } from './agent';
 import { version } from '../version';
 import type { HivePluginOptions } from './types';
+import { warnIf } from './utils';
```
kamilkisiela, line 6:
```suggestion
import { logIf } from './utils';
```

packages/libraries/client/src/internal/reporting.ts
```diff
@@ -19,8 +20,20 @@ export function createReporting(pluginOptions: HivePluginOptions): SchemaReporte
...
+    '[hive][reporting] author is missing',
+    logger
+  );
+  warnIf(
+    typeof reportingOptions.commit !== 'string' || reportingOptions.commit.length === 0,
+    '[hive][reporting] commit is missing',
+    logger
+  );
+  warnIf(typeof token !== 'string' || token.length === 0, '[hive][reporting] token is missing', logger);
+
```
kamilkisiela, line 36:
```suggestion
  logIf(
    typeof reportingOptions.author !== 'string' || reportingOptions.author.length === 0,
    '[hive][reporting] author is missing',
    logger.error
  );
  logIf(
    typeof reportingOptions.commit !== 'string' || reportingOptions.commit.length === 0,
    '[hive][reporting] commit is missing',
    logger.error
  );
  logIf(typeof token !== 'string' || token.length === 0, '[hive][reporting] token is missing', logger.error);

```

packages/libraries/client/src/internal/types.ts
```diff
@@ -22,6 +22,7 @@ export interface ClientInfo {
 export interface Logger {
   info(msg: string): void;
   error(error: any, ...data: any[]): void;
+  warn(msg: string): void;
```
kamilkisiela, line 25:
```suggestion
```

packages/libraries/client/src/internal/reporting.ts
kamilkisiela: @charlypoly 93ea93e4635b501031c80b034564f22085f7449f

PR #168, opened 2022-06-23
Use graphql as direct dependency of CLI

kamilkisiela:
Implements #162


saihaj (commented):
so issue this would cause is if someone is using `v15` in codebase and also uses Hive CLI they can get weird errors then?

saihaj (approved):

PR #215, opened 2022-06-30
feat: always log unexpected errors

n1ru4l:
In persuit of figuring out what causes https://github.com/kamilkisiela/graphql-hive/issues/188 I realized that the actual issue [for my test here](https://github.com/kamilkisiela/graphql-hive/pull/201#issuecomment-1169651610) is that the `GraphQLError` thrown from `parse` within our business logic causes the `unexpected error.`. This PR resolves that by wrapping the logic and rethrowing the error.

Note that this is probably still not a fix for the error mentioned in https://github.com/kamilkisiela/graphql-hive/issues/187, as the CLI would have failed before doing any API requests, because it is also using `parse` and `print`.

- Always log unexpected errors to the console, even in dev mode
- Catch `GraphQLError` raised by `parse` and format it into a more obvious error message for both the CLI and API

### DIFF
--- a/.changeset/little-dingos-exercise.md
+++ b/.changeset/little-dingos-exercise.md
@@ -1,0 +1,5 @@
+---
+'@hive/api': patch
+---
+
+Stop masking GraphQL syntax errors within `Mutation.schemaPublish`.
--- a/integration-tests/dockest.ts
+++ b/integration-tests/dockest.ts
@@ -9,9 +9,7 @@ async function main() {
     logLevel: logLevel.DEBUG,
     jestOpts: {
       runInBand: true,
-      testMatch: process.env.TEST_FILTER
-        ? [`**/${process.env.TEST_FILTER}?(*.)+(spec|test).[jt]s?(x)`]
-        : undefined,
+      testMatch: process.env.TEST_FILTER ? [`**/${process.env.TEST_FILTER}?(*.)+(spec|test).[jt]s?(x)`] : undefined,
       config: JSON.stringify({
         roots: ['<rootDir>/tests'],
         transform: {
--- a/integration-tests/dockest.ts
+++ b/integration-tests/dockest.ts
@@ -9,6 +9,9 @@ async function main() {
     logLevel: logLevel.DEBUG,
     jestOpts: {
       runInBand: true,
+      testMatch: process.env.TEST_FILTER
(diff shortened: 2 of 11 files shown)


packages/services/api/src/shared/entities.ts
```diff
@@ -54,9 +54,27 @@ export interface PersistedOperation {
...
+}
+
 export function createSchemaObject(schema: Schema): SchemaObject {
+  let document: DocumentNode;
+  
+  try {
+    document = parse(schema.source)
+  }  catch (err) {
+    if (err instanceof GraphQLError) {
+      throw new GraphQLDocumentStringInvalidError(err.message,  err.locations?.[0])
```
n1ru4l, line 71: To be honest we could also use `HiveError` here. But I did not check all other usages of `createSchemaObject` and whether those are safe to be exposed to the outside world.

.changeset/little-dingos-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
```
dotansimha, line 1: you don't have to add changesets to the internal services (i think? @kamilkisiela ?)

integration-tests/dockest.ts
```diff
@@ -9,6 +9,9 @@ async function main() {
     logLevel: logLevel.DEBUG,
     jestOpts: {
       runInBand: true,
+      testMatch: process.env.TEST_FILTER
```
dotansimha, line 12: ❤️

.changeset/little-dingos-exercise.md
kamilkisiela: yes, no need

integration-tests/tests/cli/schema.spec.ts
```diff
@@ -70,6 +70,69 @@ test('can publish and check a schema with target:registry:read access', async ()
...
+      '--commit',
+      'abc123',
+      'fixtures/init-invalid-schema.graphql',
+    ]);
+    throw allocatedError;
+  } catch (err) {
+    if (err === allocatedError) {
+      throw err;
+    }
+    expect(String(err)).toContain(`The SDL is not valid at line 1, column 1:\nSyntax Error: Unexpected Name "iliketurtles"`);
```
kamilkisiela, line 132:
```suggestion
    expect(String(err)).toMatch(`The SDL is not valid at line 1, column 1:`);
    expect(String(err)).toMatch(`Syntax Error: Unexpected Name "iliketurtles"`);
```

PR #261, opened 2022-07-22
Initial Emails service

### LINKED ISSUE #181: Emails service with support for enqueue jobs
related: https://github.com/kamilkisiela/graphql-hive/issues/179

kamilkisiela:
Closes #181


packages/services/emails/src/context.ts
```diff
@@ -0,0 +1,9 @@
+import type { FastifyLoggerInstance } from '@hive/service-common';
+import type { Job } from 'bullmq';
+import type { RenderedTemplate } from './templates';
+
+export interface Context {
+  logger: FastifyLoggerInstance;
+  errorHandler(message: string, error: Error, logger?: FastifyLoggerInstance | undefined): void;
+  schedule(input: RenderedTemplate): Promise<Job<any, any, string>>;
```
dotansimha, line 8: I think the input should also contain a `key`, and we can use that as `jobId` for BullMQ. 
This way the rest of the services can trigger a call as many as they want, and the "debounce" process will be done by the BullMQ because it can't create the same Job with the same ID. 

This way, services can control the key creation and the debounce process on it's own, something like `rate-limit-{ORG_ID}-{WEEK_NUMBER}` and this will make sure to send only 1 message a week about rate-limit 🤔

kamilkisiela: it's here https://github.com/kamilkisiela/graphql-hive/blob/90dede92cdd4e43fa1c2831c3298decb9e68e848/packages/services/emails/src/templates.ts#L66-L87

kamilkisiela: I did it this way because the email templates live in the emails service.

dotansimha: Oh I see, so emails service is aware of the templates, got it, thanks!

PR #304, opened 2022-08-22
Implement new ClickHouse DB structure

### LINKED ISSUE #193: Add TTL to materialized views in ClickHouse
Use `toDate(expires_at)` in the SummingMergeTree + add it to `ORDER BY`.
Do it with `MODIFY`. 

Think about:
-  running `OPTIMIZE` to delete expired rows.
- old rows without `expires_at` column

### LINKED ISSUE #302: Improve ClickHouse performance
**What do we show currently?**

- total number of requests
- requests per minute
- number of unique operations
- success and failure rates
- p90, p95, p99 of latency
- top 5 clients names (with number of requests)
- top 5 client versions (with number of requests)
- operations over time (total and failures)
- RPM over time
- latency over time
- latency histogram (super heavy)
- list of unique operations (with p90, p95, p99, number of requests, failure rate)

**What filters do we have?**

- date range
- operations

**What filters do we want to have?**

- client names
- date range
- operations (if the number of selected operations is greater than half, let's use `NOT IN (not-selected-list)`

**What else do we want to show?**

- hide histogram
- dedicated page for a single operation

kamilkisiela:
Closes #302 
Closes #193

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -126,6 +126,83 @@ jobs:
         if: always()
         run: cat integration-tests/*.log
 
+  integration-tests-ff:
+    name: Integration Tests (with Feature Flags enabled)
+    runs-on: ubuntu-latest
+    needs: setup
+    timeout-minutes: 15
+
+    env:
+      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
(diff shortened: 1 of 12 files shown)


kamilkisiela:
Reduce the size of storage by 55%

PR #327, opened 2022-09-01
Fix invoking router with shell

rcrowe:
Allow arguments to be easily passed to the docker image


rcrowe:
Improved by https://github.com/kamilkisiela/graphql-hive/pull/329

PR #328, opened 2022-09-01
No public env

n1ru4l:


n1ru4l:
See https://github.com/kamilkisiela/graphql-hive/pull/382

PR #336, opened 2022-09-06
fix: set config.api.externalResolver to true in order to avoid weird …

n1ru4l:
…behavior (https://github.com/getsentry/sentry-javascript/issues/3852#issuecomment-918820923)

### DIFF
--- a/packages/web/app/src/lib/api/extract-access-token-from-request.ts
+++ b/packages/web/app/src/lib/api/extract-access-token-from-request.ts
@@ -16,7 +16,7 @@ export async function extractAccessTokenFromRequest(req: NextApiRequest, res: Ne
     req,
     res
   );
-  // Session can be undefined in case access token was sent.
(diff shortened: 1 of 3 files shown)


packages/web/app/src/lib/api/extract-access-token-from-request.ts
```diff
@@ -8,8 +8,15 @@ import { backendConfig } from '@/config/backend-config';
...
-  const accessToken = (req as any).session.getAccessToken();
+  await superTokensNextWrapper(
+    async next =>
+      await verifySession({
+        sessionRequired: false,
+      })(req as any, res as any, next),
+    req,
+    res
+  );
+  // Session can be undefined in case access token was sent.
```
n1ru4l, line 19:
```suggestion
  // Session can be undefined in case no access token was sent.
```

PR #338, opened 2022-09-06
Sentry on staging

### LINKED ISSUE #337: enable sentry on staging
we had downtime due to unexpected issues with sentry on production. We did not spot this issue in staging/local dev as we did not enable sentry there.

kamilkisiela:
Closes #337


n1ru4l (commented):
Couldn't SENTRY_ENABLED be computed from whether a SENTRY_DSN is set?

n1ru4l (approved):
Convinced

PR #349, opened 2022-09-09
feat: import auth0 accounts into supertokens when doing a password reset

n1ru4l:
Currently, the password of accounts that have not been migrated from Auth0 to SuperTokens (via a login) cannot be recovered.

This PR solves this by applying the following flow upon a password reset:

- Check if user exists in our database but not within supertokens
- If he does not exist in supertokens, create an supertokens user with a random password
- Continue the flow as usual

### DIFF
--- a/packages/web/app/src/config/backend-config.ts
+++ b/packages/web/app/src/config/backend-config.ts
@@ -6,6 +6,7 @@ import type { TypeInput as ThirdPartEmailPasswordTypeInput } from 'supertokens-n
 import { fetch } from 'cross-undici-fetch';
 import { appInfo } from './app-info';
 import zod from 'zod';
+import * as crypto from 'crypto';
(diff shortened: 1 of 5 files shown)


packages/web/app/src/config/backend-config.ts
```diff
@@ -381,3 +456,15 @@ const generateAuth0AccessToken = async (config: LegacyAuth0ConfigEnabled): Promi
...
+  return await new Promise<string>((resolve, reject) =>
+    crypto.randomBytes(20, (err, buf) => {
+      if (err) {
+        reject(err);
+        return;
+      }
+      resolve(buf.toString('hex'));
+    })
+  );
+}
```
n1ru4l, line 470: Not sure whether this is the best method of generating a password on the fly 😄

kamilkisiela (approved):

PR #354, opened 2022-09-12
Update Pino and Fastify

kamilkisiela:


dotansimha:
we'll switch to use whatwg-server

PR #355, opened 2022-09-13
chore: imported dependencies should be in package json

rperryng:
Was syncing my fork and my merge commit failed the js linting

```
/Users/rperrynguyen/code/wealthsimple/graphql-hive/packages/web/app/pages/api/auth/[[...path]].ts
  4:1  error  'express' should be listed in the project's dependencies. Run 'npm i -S express' to add it  import/no-extraneous-dependencies
```

All I've done was `yarn add express` for the api package and then `yarn-deduplicate yarn.lock`


rperryng:
Hmm I think this was just an issue with my fork, I can't seem to reproduce on `main`, even though the dependency _is_ missing from `app`'s `package.json` 🤔

PR #363, opened 2022-09-14
chore: set supertokens log level to info

n1ru4l:


kamilkisiela:
The v3.16 (#366) has it by default

PR #365, opened 2022-09-14
Update timestamp description and value to be withing used Plan

cookieMr:


cookieMr:
This doc update is here, since it took me more than a week to figure this out. I hope this will save some time to other people testing this Hive solution :)

PR #379, opened 2022-09-18
collect input fields from variables

dimatill:
initial PR for https://github.com/kamilkisiela/graphql-hive/issues/369


kamilkisiela (changes requested):
Can we make it opt-in?

dimatill:
@kamilkisiela yes, I've added the `processVariables` option, please let me know if the naming seems ok to you

PR #382, opened 2022-09-19
feat: remove app next.js built-time environment variable requirement

n1ru4l:
Running your own hive requires setting some environment variables that are currently set during built time dynamically.

Since we do not want people to build hive from scratch, this PR replaces all `NEXT_PUBLIC_` environment variables with custom handling, where, the environment variables are passed to the frontend via a `script` tag within the `_document.tsx` file.

This allows dynamically setting all environment variables used by the frontend without having to rebuild the application from scratch.

**Checklist before deploying**:
- [x] Add `STRIPE_PUBLIC_KEY` to the `app` config
- [x] Add `MIXPANEL_TOKEN` to the `app` config
- [x] Add `GA_TRACKING_ID` to the `app` config
- [x] Add `CRISP_WEBSITE_ID` to the `app` config


.github/workflows/ci.yaml
```diff
@@ -404,3 +402,59 @@ jobs:
 
       - name: Lint Prettier
         run: yarn lint:prettier
+
+  publish_app:
```
kamilkisiela, line 406: @n1ru4l should I merge it with the docker things?

n1ru4l: its okay for now

PR #386, opened 2022-09-20
feat: docker images

n1ru4l:
## ~~TODO~~

- [x] handle storage container
- [x] use new images within integration tests
- [x] apply labels conditionally
- [ ] ~~use images for deployment (in a separate pull request)~~

## published docker images

https://github.com/kamilkisiela?tab=packages&repo_name=graphql-hive

image


docker-compose.build.yml
```diff
@@ -0,0 +1,63 @@
+version: '3.8'
+
+services:
+  emails:
+    build:
+      dockerfile: '${PWD:-.}/services.dockerfile'
```
n1ru4l, line 6: docker lol https://github.com/docker/compose/issues/4926#issuecomment-370500372

.github/workflows/ci.yaml
```diff
@@ -441,20 +396,146 @@ jobs:
...
+        run: |
+          docker-compose -f integration-tests/docker-compose.yml pull \
+            db \
+            clickhouse \
+            zookeeper \
+            broker \
+            redis \
+            supertokens \
+            local_cdn \
+            external_composition
```
kamilkisiela, line 412: @n1ru4l Why not just `pull`? Is there a reason?

docker.hcl
```diff
@@ -0,0 +1,246 @@
...
+  ]
+}
+
+target "webhooks" {
+  inherits = ["service-base", get_target()]
+  context = "${PWD}/packages/services/webhooks/dist"
+  args = {
+    IMAGE_TITLE = "graphql-hive/webhooks"
+    IMAGE_DESCRIPTION = "The webhooks ingestor service of the GraphQL Hive project."
+    HEALTHCHECK_CMD = "wget --spider -q http://127.0.0.1:3005/_readiness"
```
kamilkisiela, line 208: @n1ru4l Most of the containers accepts the `PORT` env var. Maybe we should inline the healthcheck with the `PORT` variable?

.github/workflows/ci.yaml
n1ru4l: because it fail due to the other service images not being available (they need to be built locally).

integration-tests/docker-compose.yml
```diff
@@ -98,11 +98,72 @@ services:
...
+    volumes:
+      - '../packages/libraries/external-composition/dist/example.mjs:/example.mjs'
+      - './run-external-composition.sh:/run-external-composition.sh'
+    environment:
+      NODE_ENV: production
+      PORT: 3012
+      SECRET: secretsecret
+
+  storage:
+    image: '${DOCKER_REGISTRY}storage${DOCKER_TAG}'
```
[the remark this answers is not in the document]
n1ru4l: You run `yarn build:local && yarn dockest`. Building the docker images locally is super cheap and fast so no big impact on performance compared to the mount fs into container approach.

docker.hcl
n1ru4l: storage does not have one, so it would already fail there

n1ru4l: instead we would have to declare an additional dockerfile

n1ru4l: But i agree that we should favor the local PORT env variable

integration-tests/docker-compose.yml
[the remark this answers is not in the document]
kamilkisiela: I see, so `DOCKET_REGISTRY` and `DOCKER_TAG` are predefined?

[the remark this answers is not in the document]
n1ru4l: no need to define them locally, they are only needed on CI for pushing or pulling the image from/to the correct destination

docker.hcl
n1ru4l: done in https://github.com/kamilkisiela/graphql-hive/pull/386/commits/2819a6542badf97676e644021c3b641bc13809ba

PR #397, opened 2022-09-22
Improve #379 and add more tests

kamilkisiela:


packages/libraries/client/src/internal/usage.ts
```diff
@@ -386,12 +401,7 @@ export function createCollector({
     };
   }
 
-  return cache(
-    (doc: DocumentNode, variables: ExecutionArgs['variableValues']) =>
-      collect(doc, processVariables ? variables : undefined),
-    cacheDocumentKey,
-    LRU<CacheResult>(max, ttl)
-  );
+  return cache(collect, cacheDocumentKey, LRU<CacheResult>(max, ttl));
```
dimatill, line 404: @kamilkisiela it looks to me that with this change the cache always considers variables even if `processVariables` is false

kamilkisiela: oh, you're write

kamilkisiela: https://github.com/kamilkisiela/graphql-hive/pull/399

PR #398, opened 2022-09-22
app environment validation

n1ru4l:


packages/web/app/src/env/frontend.ts
```diff
@@ -0,0 +1,23 @@
...
+      if (prop === Symbol.toPrimitive) {
+        return () => undefined;
+      }
+      if (prop in String.prototype) {
+        return obj[prop as any];
+      }
+      return noop();
+    },
+  });
+}
```
n1ru4l, line 23: My masterpiece...

packages/web/app/src/config/frontend-env.ts
```diff
@@ -0,0 +1,36 @@
+import { env as backendEnv } from '@/env/backend';
+
+/**
+ * The configuration for the frontend environment.
+ * Don't add any sensitive values here!
+ * Everything here must be serializable.
```
n1ru4l, line 6:
```suggestion
 * Everything here must be JSON serializable.
```

packages/web/app/environment.ts
```diff
@@ -0,0 +1,164 @@
...
+  authGithub: AuthGitHubConfigSchema.safeParse(process.env),
+  authGoogle: AuthGoogleConfigSchema.safeParse(process.env),
+  authLegacyAuth0: LegacyAuth0Config.safeParse(process.env),
+};
+
+const environmentErrors: Array<string> = [];
+
+for (const config of Object.values(configs)) {
+  if (config.success === false) {
+    environmentErrors.push(JSON.stringify(config.error.format(), null, 4));
```
n1ru4l, line 80: We should figure out better formatting here :)

eric-burel:
Just linking relevant docs because there has been muuuch discussion about that in Next:
- next.config.ts https://github.com/vercel/next.js/issues/4381
- disable SSR: https://github.com/vercel/next.js/issues/4381

For env variable checking, do you need that in next.config.ts specifically? What I've done so far is runtime check, a config files loads all env variables in a centralized and typed manner, so validation runs at this stage.
Example in Devographics code: https://github.com/Devographics/Monorepo/tree/d419395b4905168da92a4b55615cc46a9702df43/surveyform/src/config
(in another project I've used Joi and all with the same approach).
You can add an Eslint rule to forbid "process.env" throughout the code, it's a bad habit imo, and disable it only in this file.

Otherwise, you can check our config also in Devographics surveyform to build scripts: https://github.com/Devographics/Monorepo/blob/d419395b4905168da92a4b55615cc46a9702df43/surveyform/tsup.config.ts
Those scripts can reuse the app code (eg the logic to connect Mongo is very useful) and be written in TS. You can choose to bundle node_modules in the script (they become extremely big though eg multiple MOs) or not (then you just need to be sure that node_modules are there when the script is run, which is usually the case if those scripts are tied to the app anyway).
We have a generic "onStartup" script for instance: https://github.com/Devographics/Monorepo/blob/d419395b4905168da92a4b55615cc46a9702df43/surveyform/scripts/onServerStart.ts
You can run it when you need to depending on the situation: postbuild, predev, prestart...

n1ru4l:
For env variable checking, do you need that in next.config.ts specifically?

Unless there is another way to ensure that the environment variables are checked when running `next dev`, that is the kind of developer experience I would like to have.

packages/web/app/environment.ts
n1ru4l: It should be okay for now - but in the future, we definitely should make this more obvious.

PR #401, opened 2022-09-22
Batching in the usage-ingestor service

kamilkisiela:
- [x] Adds a configurable batching of writes to ClickHouse (the usage-ingestor service)
- [x] Migrates from `CSV` to `CSVWithNames` input format
- [x] Adds more tests to make sure we ingest data correctly (same order of fields and values)
- [x] gzip compression is now done in the ClickHouse Client
- [x] Uses `@clickhouse/client`, official NodeJS client
- [x] Removes the legacy ClickHouse DB structure (the migration process finished)
- [x] add a safety mechanism to make sure we won't lose any data when ClickHouse is down (think about it in the context of more than 1 replica and pod's memory)
- [x] adds a cron job to insert data from S3 to ClickHouse (it's done within the usage-ingestor service, safe to be used in many replicas)
- [x] retry on failures
- [ ] run load test and check the impact on memory


packages/services/usage-ingestor/src/fallback.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  return {
+    async write(buffer: Buffer, table: string) {
+      try {
+        const Body = await compress(buffer);
+        const ChecksumSHA256 = createHash('sha256').update(Body).digest('base64');
+        await retry(() =>
+          client.send(
+            new PutObjectCommand({
+              Bucket: s3.bucket, // `graphql-hive-usage-ingestor`,
```
kamilkisiela, line 146:
```suggestion
    region: s3.region,
    apiVersion: s3.apiVersion,
    credentials: {
      accessKeyId: s3.accessKeyId,
      secretAccessKey: s3.secretAccessKey,
    },
  });

  return {
    async write(buffer: Buffer, table: string) {
      try {
        const Body = await compress(buffer);
        const ChecksumSHA256 = createHash('sha256').update(Body).digest('base64');
        await retry(() =>
          client.send(
            new PutObjectCommand({
              Bucket: s3.bucket,
```

PR #416, opened 2022-09-29
Use Apollo Router v1

kamilkisiela:


ksvital:
Hi @kamilkisiela, looks like the publish to docker step keeps timing out close to completion, would be great if you could get the docker image published so that we could tryout router v1 with graphql-hive. Thanks 🙏

kamilkisiela:
Yeah, that's my task for today :)

kamilkisiela:
https://github.com/kamilkisiela/graphql-hive/pkgs/container/graphql-hive%2Fapollo-router

PR #427, opened 2022-09-29
Use ETag and If-None-Match in CDN and clients

### LINKED ISSUE #414: Add quick checksum comparison for CDN
Allow Gateways to send the checksum and compare it against the latest state.

Use either [304](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) status code as for Not Modified or a `HEAD` request.

kamilkisiela:
1. A GET request to CDN -> http response with data and ETag header (OK 200)
1. A GET request to CDN with If-None-Match: <etag> -> http response with no data (Not Modified 304)
1. A GET request to CDN with If-None-Match: <etag> -> http response with no data (Not Modified 304)
1. CDN state changes and previously fetched data becomes stale
1. A GET request to CDN with If-None-Match: <etag> -> http response with new data and ETag header (OK 200)
1. A GET request to CDN with If-None-Match: <new-etag> -> http response with no data (Not Modified 304)

Closes #414

### DIFF
--- a/packages/libraries/client/src/apollo.ts
+++ b/packages/libraries/client/src/apollo.ts
@@ -7,24 +7,53 @@ import { createHive } from './client';
 import { isHiveClient } from './internal/utils';
 
 export function createSupergraphSDLFetcher({ endpoint, key }: SupergraphSDLFetcherOptions) {
+  let cacheETag: string | null = null;
+  let cached: {
+    id: string;
+    supergraphSdl: string;
+  } | null = null;
+
   return function supergraphSDLFetcher() {
+    const headers: {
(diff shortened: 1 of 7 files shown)


n1ru4l (commented):
will you add docs for this?

kamilkisiela:
Not sure we should, it's too technical for the user. I think it makes sense to say that you don't poll the data over and over again, just when it changes.

The technical details would make more sense in a specification for CDN (if we would have one).

WDYT?

n1ru4l:
@kamilkisiela agreed!

n1ru4l (approved):

PR #456, opened 2022-10-05
(processVariables: true) do not collect input if variable is missing

dimatill:
Fix for the case when an optional variable is not provided

### DIFF
--- a/packages/libraries/client/tests/usage-collector.spec.ts
+++ b/packages/libraries/client/tests/usage-collector.spec.ts
@@ -336,6 +336,31 @@ test('(processVariables: true) collect used-only input fields', async () => {
   expect(info.fields).not.toContain(`PaginationInput.offset`);
 });
 
+test('(processVariables: true) should not collect input when corresponding variable is not provided', async () => {
+  const collect = createCollector({
+    schema,
+    max: 1,
+    processVariables: true,
+  });
+  const info = collect(
+    parse(/* GraphQL */ `
+      query getProjects($pagination: PaginationInput, $type: ProjectType!) {
+        projects(filter: { pagination: $pagination, type: $type }) {
+          id
+        }
+      }
+    `),
+    {
+      type: 'STITCHING',
+    }
+  ).value;
+
+  expect(info.fields).toContain(`FilterInput.type`);
+  expect(info.fields).toContain(`FilterInput.pagination`);
+  expect(info.fields).not.toContain(`PaginationInput.limit`);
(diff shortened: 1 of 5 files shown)


packages/libraries/client/tests/usage-collector.spec.ts
```diff
@@ -336,6 +336,31 @@ test('(processVariables: true) collect used-only input fields', async () => {
...
+    {
+      type: 'STITCHING',
+    }
+  ).value;
+
+  expect(info.fields).toContain(`FilterInput.type`);
+  expect(info.fields).toContain(`FilterInput.pagination`);
+  expect(info.fields).not.toContain(`PaginationInput.limit`);
+  expect(info.fields).not.toContain(`PaginationInput.offset`);
+});
```
kamilkisiela, line 362: hmmm, if it's not collecting `PaginationInput` then when `PaginationInput` is removed (because it's safe from the usage reports perspective) this query will fail, no? Because it contains `PaginationInput` and it's not defined.

dimatill: oh, yes, you're right. I need to think about it and find a way to handle this case

dimatill:
@kamilkisiela , I just pushed new changes. I didn't test anything yet, but it looks to me like it SHOULD be ok this way both from UI and GraphQL Inspector perspectives. Just want to hear your opinion before going forward.

dimatill:
@kamilkisiela I've tested it manually with a local instance of Hive and it looks good to me.
The test was done with the same schema as in `usage-collector.spec.ts`.
The test scenario was the same as in `(processVariables: true) should not collect input when corresponding variable is not provided` test case.

Please check attached screenshots.
Here is a description of screenshots:

In the UI I can see that `PaginationInput` was used 1 time while all fields were used zero times. This could be a little bit confusing, but this is what actually happened: fields weren't used but the input itself was used in the query definition.

`hive schema:check` also works as expected: removing `FilterInput.order` and `ProjectOrderByInput` is considered a non-breaking change because they weren't used. Removing `PaginationInput` and `FilterInput.pagination` is considered a breaking change because they were used in the query.

What do you think? Can we merge this PR?

Screenshot 2022-10-09 at 17 50 25
Screenshot 2022-10-09 at 18 07 37

kamilkisiela (approved):

kamilkisiela:
I will deploy it to production tomorrow

dimatill:
woohoo, thank you @kamilkisiela !

PR #496, opened 2022-10-17
Ensure no parsing errors in usage reports

kamilkisiela:


packages/services/usage/src/usage.ts
```diff
@@ -243,11 +241,23 @@ export function createUsage(config: {
...
         const validationResult = validateOperationMapRecord(record);
 
         if (validationResult.valid) {
+          // Make sure
+          const key = createHash('md5')
+            .update(outgoing.target)
+            .update(record.operation)
+            .update(record.operationName ?? '')
+            .update(JSON.stringify(record.fields))
+            .digest('hex');
```
kamilkisiela, line 257:
```suggestion
          // The key is used for lru cache (usage-ingestor) so we need to make sure, the record is unique per target, operation body, name and the list of fields
          const key = createHash('md5')
            .update(outgoing.target)
            .update(record.operation)
            .update(record.operationName ?? '')
            .update(JSON.stringify(record.fields.sort()))
            .digest('hex');
```

PR #499, opened 2022-10-17
Add self-hosting options to GraphQL Hive client

kamilkisiela:
```diff
 {
   usage: {
-    endpoint: 'https://api.com/usage'
   },
   reporting: {
-    endpoint: 'https://api.com/graphql'
   },
+  selfHosting: {
+    graphqlEndpoint: 'https://api.com/graphql',
+    applicationUrl: 'https://app.com',
+    usageEndpoint: 'https://api.com/usage',
+  }
}
```

This way everything that's related to GraphQL Hive Cloud can be replaced, the links, the endpoints, token info, everything.

### DIFF
--- a/packages/libraries/client/README.md
+++ b/packages/libraries/client/README.md
@@ -304,3 +304,30 @@ const server = new ApolloServer({
   ]
 })
 ```
+
+## Self-Hosting
+
+To align the client with your own instance of GraphQL Hive, you should use `selfHosting` options in the client configuration.
+
+The example is based on GraphQL Yoga, but the same configuration applies to Apollo Server and others.
+
+```ts
+import { createYoga } from '@graphql-yoga/node'
+import { useHive } from '@graphql-hive/client'
+
+const server = createYoga({
+  plugins: [
+    useHive({
+      enabled: true,
+      token: 'YOUR-TOKEN',
+      selfHosting: {
+        graphqlEndpoint: 'https://your-own-graphql-hive.com/graphql',
+        applicationUrl: 'https://your-own-graphql-hive.com',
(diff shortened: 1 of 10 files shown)


packages/libraries/client/README.md
```diff
@@ -304,3 +304,30 @@ const server = new ApolloServer({
...
+import { useHive } from '@graphql-hive/client'
+
+const server = createYoga({
+  plugins: [
+    useHive({
+      enabled: true,
+      token: 'YOUR-TOKEN',
+      selfHosting: {
+        graphqlEndpoint: 'https://your-own-graphql-hive.com/graphql',
+        applicationUrl: 'https://your-own-graphql-hive.com',
```
danielfigueiredo, line 325: Nice, this `applicationUrl` I think was the only missing parameter so we can configure everything for `selfHosting` 🎉

packages/libraries/client/README.md
```diff
@@ -304,3 +304,30 @@ const server = new ApolloServer({
   ]
 })
 ```
+
+## Self-Hosting
+
+To align the client with your own instance of GraphQL Hive, you should use `selfHosting` options in the client configuration.
```
danielfigueiredo, line 310: Would it be valuable to explain that the `selfHosting` options take precedence over the legacy ones `options.hosting.endpoint` and `options.usage.endpoint`
I believe we might want to remove the legacy ones at some point later via breaking change? Unless there is some other reason for keeping them I'm not aware of, but in either case might be beneficial to explain it so newcomers are not confused

danielfigueiredo (approved):
LGTM

PR #516, opened 2022-10-20
dont break the redis connection when we get a non breaking error Cont…

rickbijkerk:


rickbijkerk:
verified the change locally and after this change i am able to connect webhooks and emails to google MemoryStore

PR #525, opened 2022-10-24
Architecture Diagram

Enrico2:
Adds a design diagram.

I verified the link works on the fork, it should work after it merges:

Design Diagram

### DIFF
--- a/docs/architecture.puml
+++ b/docs/architecture.puml
@@ -1,0 +1,63 @@
+@startuml
+
+actor "UI User" as uiuser
+actor "CLI User" as cliuser
+actor "Running Server" as gqlserver
+
+component zookeeper
+component supertokens
+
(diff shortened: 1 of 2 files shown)


docs/architecture.puml
```diff
@@ -0,0 +1,63 @@
+@startuml
```
dotansimha, line 1: @B2o5T any idea if we can add some magic to display this kind of file (similar to what we have with mermaid)? 

In the same context, @Enrico2 any plan to move this to be in mermaid format?

Enrico2:
No, I'm not familiar with the mermaid format, this will render nicely on
the readme once it merges to main because the URL I set as the image URL
assumes the main branch..

The example in the PR description actually uses the PR branch url to show
it'll render.

PR #563, opened 2022-10-30
Fix createServicesFetcher handle null service url

PabloSzx:
We found that when a service doesn't set an explicit service URL, which seems to be valid state, the fetcher fails when it tries to create the hash since you can't give `null` into a Node.js crypto hash

screenshot of reference:

cc @saihaj @kamilkisiela

### DIFF
--- a/.changeset/famous-readers-cheat.md
+++ b/.changeset/famous-readers-cheat.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/client': patch
+---
+
+Fix createServicesFetcher handling null service url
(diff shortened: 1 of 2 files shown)


.changeset/famous-readers-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-hive/client': patch
```
PabloSzx, line 2:
```suggestion
'@graphql-hive/client': minor
```

packages/libraries/client/src/gateways.ts
```diff
@@ -4,7 +4,7 @@ import type { SchemaFetcherOptions, ServicesFetcherOptions } from './internal/ty
 
 interface Schema {
   sdl: string;
-  url: string;
+  url: string | null;
```
PabloSzx, line 7: this change is actually breaking since it's also used as return type

dotansimha (approved):

kamilkisiela:
`0.21.0`

PR #606, opened 2022-11-07
feat: revoke all sessions after a password reset

n1ru4l:


n1ru4l:
@dotansimha This solves the concern of the security audit.

kamilkisiela (approved):

PR #612, opened 2022-11-08
Support log level and request logs

### LINKED ISSUE #593: Request Logging
Could the `graphql-hive` server instances log their requests?  status code, runtime, information about the caller, e.g. adding something like https://github.com/mgcrea/fastify-request-logger ?  Currently I'm just using raw timestamps to try and correlate logs from client requests to the instance that I'm hosting

rperryng:
Adds support for:
1. configuring the logger level via the `LOG_LEVEL` env var
2. request logging (Fixes #593).
  a. ~Can be disabled with `LOG_DISABLE_REQUEST_LOGGING`~
  b. really simple and lightweight format.  Personally, i found the built-in request logging too awkward and [this plugin](https://github.com/mgcrea/fastify-request-logger) to be a bit more than what I'm looking for.

example logs from running (dockerized) locally:
```
server_1          | {"level":30,"time":1668522424590,"pid":1,"hostname":"5b5d04c96b55","msg":"[200] (172.19.0.11) POST /graphql 'organizationActivities' (+6CDHcG1SB+PXYOPpJiD1A/0000000004)"}
```


rperryng:
@kamilkisiela any thoughts :)?

n1ru4l:
@kamilkisiela Thoughts on the flag for disabling request logging? I think request logging should be determined via the log level (`info`) and not an additional option.

kamilkisiela:
I think we could show operation's name as well and the value of the `request_id` header and I guess making it use `info` level is fine.

I know I said it should be opt-in but on the other hand it does not bring much noise to the logs, so maybe having it enabled by default (with no option to opt-out) is actually the best option. If you all agree, then I will create a PR to your PR @rperryng to remove the flag `LOG_DISABLE_REQUEST_LOGGING` flag.

rperryng:
I removed the flag to disable request logging :).

With latest changes, the log message now looks like so:

```
server_1          | {"level":30,"time":1668522424590,"pid":1,"hostname":"5b5d04c96b55","msg":"[200] (172.19.0.11) POST /graphql 'organizationActivities' (+6CDHcG1SB+PXYOPpJiD1A/0000000004)"}
```

Is this close to what you had in mind @kamilkisiela ?

packages/services/emails/src/environment.ts
```diff
@@ -156,6 +173,9 @@ export const env = {
   },
   heartbeat: base.HEARTBEAT_ENDPOINT ? { endpoint: base.HEARTBEAT_ENDPOINT } : null,
   sentry: sentry.SENTRY === '1' ? { dsn: sentry.SENTRY_DSN } : null,
+  log: {
+    level: log.LOG_LEVEL ?? 'debug',
```
n1ru4l, line 177: I think the default fallback log level should be info

packages/services/emails/src/environment.ts
```diff
@@ -71,6 +71,20 @@ const PrometheusModel = zod.object({
...
+    .union([
+      zod.literal('trace'),
+      zod.literal('debug'),
+      zod.literal('info'),
+      zod.literal('warn'),
+      zod.literal('error'),
+      zod.literal('fatal'),
+      zod.literal('silent'),
+    ])
+    .optional(),
```
n1ru4l, line 85: these should be wrapped in `emptyString()` so that `""` is also interpreted as an empty value rather than raise a runtime error.

n1ru4l (changes requested):
I added two small comments that apply to all the files where the logging is parsed from the env variables

rperryng:
@n1ru4l thanks for your feedback, it should be addressed now - let me know if I misunderstood anything

kamilkisiela:
@rperryng https://github.com/rperryng/graphql-hive/pull/1

rperryng:
@kamilkisiela thanks!  merged your changes 🙂 

I figured there must have been a way to avoid re-parsing since all the envelop servers are already doing it, but yeah, wasn't sure how to slide it in the current config 😅.  Maybe I can give it a look when graphql-hive has moved off of fastify

PR #614, opened 2022-11-08
envoy - bump timeoutPolicy to 60s and retry twice on reset

kamilkisiela:
I updated the helm chart `7.8.0 -> 10.0.0`
Whenever a request is dropped because of connection reset, we do a retry (max 2) now.
The timeout (response) is higher `60s -> 15s` for all upstream routes.

### DIFF
--- a/deployment/utils/reverse-proxy.ts
+++ b/deployment/utils/reverse-proxy.ts
@@ -102,7 +102,7 @@ export class Proxy {
 
     const proxyController = new k8s.helm.v3.Chart('contour-proxy', {
       chart: 'contour',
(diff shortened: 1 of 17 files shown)


n1ru4l (approved):

deployment/utils/reverse-proxy.ts
```diff
@@ -130,6 +147,7 @@ export class Proxy {
             'user_agent',
             'x_forwarded_for',
           ],
+          debug: true,
```
kamilkisiela, line 150:
```suggestion
```

PR #620, opened 2022-11-09
fix: fastify wants a string or buffer not object

n1ru4l:


kamilkisiela:
Let's give it what it wants

kamilkisiela (approved):

PR #626, opened 2022-11-10
enable `@typescript-eslint/require-await`

B2o5T:


packages/services/server/src/index.ts
```diff
@@ -307,16 +307,15 @@ export async function main() {
...
         },
       });
       server.route({
         method: 'POST',
         url: '/__legacy/check_auth0_email_user_without_associated_supertoken_id_exists',
         async handler(req, reply) {
           if (req.headers['x-authorization'] !== auth0Config.apiKey) {
-            reply.status(401).send({ error: 'Invalid update user id mapping key.', code: 'ERR_INVALID_KEY' }); // eslint-disable-line @typescript-eslint/no-floating-promises -- false positive, FastifyReply.then returns void
-            return;
+            void reply.status(401).send({ error: 'Invalid update user id mapping key.', code: 'ERR_INVALID_KEY' });
```
B2o5T, line 318:
```suggestion
            void reply.status(401).send({ error: 'Invalid update user id mapping key.', code: 'ERR_INVALID_KEY' });
            return;
```

packages/web/app/src/lib/supertokens/third-party-email-password-react-oidc-provider.ts
```diff
@@ -30,7 +30,7 @@ export const getOIDCOverrides = (): UserInput['override'] => ({
           ? {
               ...input,
               options: {
-                preAPIHook: async options => {
+                preAPIHook: options => {
```
kamilkisiela, line 33: now it's not a Promise :( and it should be.
The same applies to all the functions you changed from `async () => {}` to `() => {}`.

kamilkisiela (changes requested):
All the `async` function are supposed to be `async`, we expect a Promise there even though some of them don't do any `await`.

packages/web/app/src/lib/supertokens/third-party-email-password-react-oidc-provider.ts
B2o5T: yeah this PR needs more attention that I expected 😬, I'll close it for now

PR #655, opened 2022-11-15
feat: add user-agent header to fetcher functions

n1ru4l:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent

### DIFF
--- a/packages/libraries/cli/src/base-command.ts
+++ b/packages/libraries/cli/src/base-command.ts
@@ -124,7 +124,7 @@ export default abstract class extends Command {
       new GraphQLClient(registry, {
         headers: {
           Accept: 'application/json',
-          'User-Agent': `HiveCLI@${this.config.version}`,
+          'User-Agent': `hive-cli/${this.config.version}`,
           Authorization: `Bearer ${token}`,
           'graphql-client-name': 'Hive CLI',
           'graphql-client-version': this.config.version,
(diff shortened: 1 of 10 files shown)


saihaj (approved):

packages/libraries/cli/src/base-command.ts
```diff
@@ -124,7 +124,7 @@ export default abstract class extends Command {
       new GraphQLClient(registry, {
         headers: {
           Accept: 'application/json',
-          'User-Agent': `HiveCLI@${this.config.version}`,
+          'User-Agent': `hive-cli/${this.config.version}`,
```
n1ru4l, line 127: @kamilkisiela I changed this so it is following the mdn recommendation.

packages/libraries/router/src/registry.rs
```diff
@@ -74,6 +76,7 @@ impl HiveRegistry {
         let client = reqwest::blocking::Client::new();
         let mut headers = reqwest::header::HeaderMap::new();
 
+        headers.insert(reqwest::header::USER_AGENT, format!("graphql-hive-router@{}", COMMIT.unwrap_or_else(|| "local")));
```
n1ru4l, line 79:
```suggestion
        headers.insert(reqwest::header::USER_AGENT, format!("graphql-hive-router/{}", COMMIT.unwrap_or_else(|| "local")));
```

PR #668, opened 2022-11-22
Fix ESM/CJS

kamilkisiela:


.github/workflows/ci.yaml
```diff
@@ -213,6 +213,9 @@ jobs:
       - name: Build
         run: pnpm build
 
+      - name: Test ESM & CJS exports integrity
+        run: pnpm bob check
```
enisdenjo, line 217: This does not check each package recursively. 

If you run `pnpm bob check` inside of `packages/libraries/client` you'll get more errors.

PR #702, opened 2022-11-28
update graphiql to v2

B2o5T:
- [x] update graphiql to v2
- [x] reexport all hooks from `hooks/index.ts`
- [x] `useToogle` for all modals
- [x] update `DEVELOPMENT.md`


packages/web/app/pages/[orgId]/members.tsx
```diff
@@ -242,9 +242,9 @@ const Page = ({ organization }: { organization: OrganizationFieldsFragment }) =>
   });
 
   const [checked, setChecked] = useState<string[]>([]);
-  const [selectedMemberId, setSelectedMemberId] = useState<string>('');
```
Comment (B2o5T) [inline packages/web/app/pages/[orgId]/members.tsx:245]: inherited from initial value

README.md
```diff
@@ -77,7 +77,7 @@ See you in Hive! 🐝
 - App: React, Next.js, Tailwind CSS, Radix Primitives
 - CLI: Oclif
 - Deployment: Pulumi, K8s, Nginx Proxy, Azure Cloud, CloudFlare Workers + KV Cache
-- Monitoring: Promthues, Grafana, Sentry
+- Monitoring: Prometheus, Grafana, Sentry
```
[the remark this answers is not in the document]
kamilkisiela: We do types from time to time to check if anyone reads documentation

packages/web/app/src/lib/hooks/use-toggle.ts
```diff
@@ -1,10 +1,10 @@
 import { useState, useCallback } from 'react';
 
-export function useToggle(defaultValue: boolean) {
+export function useToggle(defaultValue = false): [boolean, () => void] {
```
kamilkisiela, line 3:
```suggestion
export function useToggle(defaultValue = false) {
```

packages/web/app/src/lib/hooks/use-toggle.ts
```diff
@@ -1,10 +1,10 @@
...
   const [value, setValue] = useState(defaultValue);
   const toggle = useCallback(() => {
     setValue(value => !value);
   }, []);
 
-  return [value, toggle] as const;
+  return [value, toggle];
```
kamilkisiela, line 9:
```suggestion
  return [value, toggle] as const;
```

packages/web/app/src/lib/hooks/use-toggle.ts
B2o5T: does it really matter? In this case, I explicitly provided a return type for the hook to avoid mistakes in function body

kamilkisiela: just less manual typing

kamilkisiela: I usually don't define the return type and let the typescript do the work

PR #709, opened 2022-11-29
Fixed parse env variable EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED

SpeedyGonzaless:
Fixed parse from env, as env variable is always string and zod cannot parse string to boolean - it is needed to convert it to boolean

### DIFF
--- a/packages/services/emails/src/environment.ts
+++ b/packages/services/emails/src/environment.ts
@@ -56,7 +56,7 @@ const SMTPEmailModel = zod.object({
   EMAIL_PROVIDER_SMTP_PORT: NumberFromString,
   EMAIL_PROVIDER_SMTP_AUTH_USERNAME: zod.string(),
   EMAIL_PROVIDER_SMTP_AUTH_PASSWORD: zod.string(),
-  EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED: zod.boolean().optional(),
+  EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED: zod.string(),
 });
 
 const SendmailEmailModel = zod.object({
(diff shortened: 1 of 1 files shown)


packages/services/emails/src/environment.ts
```diff
@@ -56,7 +56,7 @@ const SMTPEmailModel = zod.object({
   EMAIL_PROVIDER_SMTP_PORT: NumberFromString,
   EMAIL_PROVIDER_SMTP_AUTH_USERNAME: zod.string(),
   EMAIL_PROVIDER_SMTP_AUTH_PASSWORD: zod.string(),
-  EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED: zod.boolean().optional(),
+  EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED: zod.string().optional(),
```
n1ru4l, line 59: Please use `emptyString(zod.union([zod.literal('0')], zod.literal('1')]).optional())`

n1ru4l (changes requested):
Thank you for this contribution! I added a comment.

packages/services/emails/src/environment.ts
SpeedyGonzaless: I got it, fixed to stay consistent

n1ru4l (approved):

n1ru4l:
@SpeedyGonzaless thank you

SpeedyGonzaless:
> @SpeedyGonzaless thank you

Can you merge pls? I want to have a final working build, to finish helm chart with self-hosted graphql-hive and publish it

packages/services/emails/src/environment.ts
```diff
@@ -156,7 +158,7 @@ const emailProviderConfig =
           pass: email.EMAIL_PROVIDER_SMTP_AUTH_PASSWORD,
         },
         tls: {
-          rejectUnauthorized: email.EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED,
+          rejectUnauthorized: !(email.EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED === '0'),
```
kamilkisiela, line 161: it's `true` by default, right?

SpeedyGonzaless: Yes, by default it's true
https://nodejs.org/api/tls.html#tlscreateserveroptions-secureconnectionlistener
`rejectUnauthorized - If not false the server will reject any connection which is not authorized with the list of supplied CAs. This option only has an effect if requestCert is true. Default: true.`

PR #713, opened 2022-11-29
Fix error message overflow

essaji:
When the error message is very long, the text is overflowing. Adding `whitespace-pre-wrap` class from tailwind to wrap the text.

Before:
image

After:
image


n1ru4l (approved):

saihaj (approved):

essaji:
Looks like one of the check is stuck, do we need to re-trigger it? @kamilkisiela @n1ru4l @saihaj

PR #735, opened 2022-12-04
[🔧 ESLint] fixes for `unicorn/numeric-separators-style` rule

B2o5T:


n1ru4l (approved):

kamilkisiela:
It improves readability sure, but can't we just let developer to decide whether it makes sense to use `2_000` or `2000`? I'm not a big fan of eslint rules for styling, even with autofix.

B2o5T:
@kamilkisiela we decided to add this rule to separate big numbers, unfortunately I guess it impossible to distinguish `2000` or `2_000`

PR #736, opened 2022-12-04
[🔧 ESLint] import sort

B2o5T:
I want to extend the guild eslint config, so will make PR with separated fixes

Current errors:

image


n1ru4l (approved):

kamilkisiela:
Does it really matter what import is first? The more eslint rules we have to slower `git commit` is.

B2o5T:
we discussed enabling this rule to keep imports organized, also it keeps git history clean when your remove/add new imports, now you will add them in right alphabetical place!

PR #738, opened 2022-12-04
[🔧 ESLint] fixes for yaml with `yml/plain-scalar` rule

B2o5T:
https://ota-meshi.github.io/eslint-plugin-yml/rules/plain-scalar.html


n1ru4l (approved):

kamilkisiela:
Why do we need it?

B2o5T:
@kamilkisiela it's a yaml eslint plugin, in this case it's removes unneeded quotes, but keeps quotes where it's required

PR #745, opened 2022-12-05
Add healthcheck endpoint to CDN Worker

kamilkisiela:


saihaj (approved):
hmm this will just instantiate the worker every minute then?

n1ru4l (approved):

PR #751, opened 2022-12-06
feat: include service name within the schema check

n1ru4l:
closes https://github.com/kamilkisiela/graphql-hive/issues/750

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-publisher.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-publisher.ts
@@ -225,7 +225,7 @@ export class SchemaPublisher {
         }
 
         await this.gitHubIntegrationManager.createCheckRun({
-          name: 'GraphQL Hive - schema:check',
+          name: buildGitHubActionCheckName(input.target, input.service ?? null),
           conclusion: validationResult.valid ? 'success' : 'failure',
           sha: input.github.commit,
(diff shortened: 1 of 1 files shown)


kamilkisiela:
It's a breaking change if somebody has the check marked as "required" on Github

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -167,7 +167,7 @@ export class SchemaPublisher {
         }
 
         await this.gitHubIntegrationManager.createCheckRun({
-          name: 'GraphQL Hive - schema:check',
+          name: 'GraphQL Hive - schema:check' + input.service ? `: ${input.service}` : '',
```
klippx, line 170: Would it be possible to have target as well?

n1ru4l:
> It's a breaking change if somebody has the check marked as "required" on Github
What strategy do you propose here? Unfortunately, we cannot check who marked it as required.

We could just tell users that the name will change. (low effort by both parties involved)
We could make it backward compatible (opt-in). (more effort on our side; no effort for existing users)

dotansimha:
@n1ru4l I rebased the PR to latest, CI should pass now. 
I think the concern about the breaking changes is valid, but the impact is actually low.
@kamilkisiela what should be the next step?

PR #756, opened 2022-12-06
poc redlock schema publish

n1ru4l:


packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -31,6 +31,13 @@ import { SCHEMA_MODULE_CONFIG } from './config';
 import { SchemaHelper } from './schema-helper';
 import { HiveError } from '../../../shared/errors';
 import { ArtifactStorageWriter } from './artifact-storage-writer';
+import { GraphQLError } from 'graphql';
+
+import Redlock from 'redlock';
+
+const redlock = new Redlock([null as any], {
```
n1ru4l, line 38: for the pseudo code like demonstration i wanted to avoid all the dependency injection overhead

PR #764, opened 2022-12-07
fix(api): Synchronise schema publishing per target

### LINKED ISSUE #728: Race condition in the schema publishing flow
## Problem 

Every new version inherits commits from a previous version (all when a new version introduces a new service, and all except one when an already existing service gets updated).

One of our customers reported an inconsistency. I checked and here's the timeline of versions:

```
version 1   c_1
version 2   c_1  c_2
version 3   c_1  c_2  c_3
version 4   c_1  c_2  c_3  c_4
version 5   c_1  c_2  c_3  c_4  c_5
version 6   c_1   ?   c_3  c_4       c_6
version 7   c_1  c_2  c_3  c_4        ?   c_7
version 8   c_1       c_3  c_4       c_6      c_8
version 9   c_1       c_3  c_4       c_6      c_8  c_9
```

To me, it looks like 2 versions published at the same time were competing with each other and persisted an incorrect state.

## Proposed solution

The flow should be something like this.
1. An attempt to publish a new schema
1. Validate the token, access etc.
1. If another attempt to publish the schema is already in progress, wait for it to complete (regardless the status - failure or success)
1. Continue with the regular flow.

We need a FIFO queue but grouped by target's id.

Why? Because we don’t want it to block the entire queue. For every "publish" action, only actions related to the same target should be executed one by one, first in, first out.

Context: https://tortilla-hq.slack.com/archives/C01E8ATBQGN/p1669909867184569?thread_ts=1669908661.254419&cid=C01E8ATBQGN

enisdenjo:
Closes #728
Closes #756


packages/services/storage/src/index.ts
```diff
@@ -2511,6 +2548,37 @@ export async function createStorage(connection: string, maximumPoolSize: number)
...
+      async lock(id) {
+        const lockConn = await getLockConn();
+
+        const idInt = hashFnv32a(id);
+
+        // wait while there's a lock
+        let lock;
+        while ((lock = locks.get(id))) {
+          // first db then process
+          await lockConn.query(sql`select pg_advisory_lock(${idInt})`);
```
n1ru4l, line 2562: What will this statement do, will it be blocked until the lock is released?
Will it raise an exception?

enisdenjo: I moved it below the while loop (and added more explanations), it must run only once per lock and should unlock the in-process lock if it fails - good catch!

https://github.com/kamilkisiela/graphql-hive/blob/b59aa864a9662dfab70375a43c2b0392cc27d488/packages/services/storage/src/index.ts#L2560-L2582

packages/services/storage/src/index.ts
```diff
@@ -2511,6 +2548,104 @@ export async function createStorage(connection: string, maximumPoolSize: number)
...
+            locks.set(id, {
+              p,
+              release,
+            });
+
+            // wait and acquire lock on the database (intra-process sync)
+            let advisoryLock = false;
+            try {
+              while (!advisoryLock) {
+                ({ advisoryLock } = await lockConn.one<{ advisoryLock: boolean }>(
```
n1ru4l, line 2593: There is also oneFirst which would avoid this destructuring assignment

packages/services/storage/src/index.ts
```diff
@@ -2511,6 +2548,104 @@ export async function createStorage(connection: string, maximumPoolSize: number)
...
+                }
+              }
+            } catch (err) {
+              // lock not acquired in database, unlock and bubble error
+              locks.delete(id);
+              release();
+              if (advisoryLock) {
+                lockConn.query(sql`select pg_advisory_unlock(${idInt})`).catch(err => {
+                  // TODO: what to do instead?
+                  console.error(
```
n1ru4l, line 2611: I think we can use logger.warn for now - if we notice it keeps happening a lot we can think more about it.

n1ru4l (approved):
I wonder whether we should also add a test for publishing two targets in parallel.

Aside from that the logic makes sense to me!

enisdenjo:
@n1ru4l right, that test would make sense. I'll see what I can do. 👌

kamilkisiela (approved):

enisdenjo:
@kamilkisiela fixed, the issue was mainly caused by https://github.com/nodejs/node/issues/38924.

kamilkisiela (approved):

PR #768, opened 2022-12-08
Revert "[🔧 ESLint] fixes for yaml with `yml/plain-scalar` rule"

kamilkisiela:
Reverts kamilkisiela/graphql-hive#738

It breaks docker compose files:

```
services.s3_provision_buckets.restart must be a string
```

@B2o5T


B2o5T:
@kamilkisiela do you know which change breaks? so I'll submit an issue

PR #776, opened 2022-12-09
fix: correctly set enabled value to "true"

n1ru4l:


dotansimha (approved):

packages/libraries/client/src/client.ts
```diff
@@ -9,25 +9,34 @@ import { version } from './version.js';
...
+
+  const usage = createUsage(mergedOptions);
+  const schemaReporter = createReporting(mergedOptions);
+  const operationsStore = createOperationsStore(mergedOptions);
 
   function reportSchema({ schema }: { schema: GraphQLSchema }) {
-    void schemaReporter.report({ schema });
+    if (enabled === false) {
+      return;
+    }
```
kamilkisiela, line 32:
```suggestion
```

The `createReporting` itself reacts to `enabled`.
The reason why it's there and not here is that the `HiveClient` expects a specific return type from these functions and it's a bit easier to have everything next to each other, types, real logic and noop.

packages/libraries/client/src/client.ts
```diff
@@ -9,25 +9,34 @@ import { version } from './version.js';
...
+    schemaReporter.report({ schema });
   }
 
   function collectUsage(args: ExecutionArgs) {
+    if (enabled === false) {
+      return noop;
+    }
```
kamilkisiela, line 39:
```suggestion
```

packages/libraries/client/src/client.ts
```diff
@@ -173,6 +182,8 @@ export function createHive(options: HivePluginOptions): HiveClient {
   };
 }
 
+function noop() {}
+
```
kamilkisiela, line 186:
```suggestion
```

PR #789, opened 2022-12-12
Registry v2

### LINKED ISSUE #279: Can't remove a service
No easy way to remove a subgraph

### LINKED ISSUE #310: Improve schema publishing flow and db

kamilkisiela:
Closes #310 
Closes #279 

**Breaking changes (when using non-legacy models)**
- Build errors (like these returned by graphql-js `validateSDL`) are now fatal errors meaning they block schema publishing
- Composition errors are now all errors that are not "build errors" (described above). For example, some errors returned by Federation's validation are unique for Federation, but some are just "build errors". In the new models, when publishing a schema and only composition errors are detected (only these related to Federation), the publish action will be accepted but the status will be `non-composable`.
- In order to follow this new logic (build and composition errors) the External Composition library and the docker image (for federation v2) needs to be updated on customer's side (as we added `source` property to errors).
- Breaking changes are accepted when publishing a schema, they won't affect the composition status.
- Service URL is now required in Stitching projects.
- To rename a service, user has to remove the service and publish a new one, with the new name.
- The `CUSTOM` projects are no longer available

**Other changes related to non-legacy models**
- `--force` and `--experimental_acceptBreakingChanges` have no effect on the logic of the new models
- New projects use the new models by default (with an option to opt-out).

TODO
- [x] integration tests for all non-legacy models
- [x] describe changes
- [x] migrate integration tests to use legacy and non-legacy models
- [x] integrate with CLI
- [x] `schema:delete`
- [x] no `updateSchemaVersionStatus` for modern models
- [x] no `schemaSyncCDN`
- [x] no `updateSchemaServiceName` (but only when `schema:delete` is available)
- [x] no `CUSTOM` type projects
- [x] Document the migration from the legacy models
- [x] Use `base_schema` in models
- [x] Disable `Mark as Valid` button for non-legacy projects
- [x] Add UI component in the settings for migrating a project to the modern models
- [x] Add more tests for `schema:delete`
- [x] Mark `--force` and `--experimental_acceptBreakingChanges` as deprecated
- [x] clean up the code, simplify it
- [ ] fix dropdown menu of Radix UI (major bump broke it)


kamilkisiela:
First step, make it work (write shit tons of integration tests).
Second step, reduce the complexity of the code.
Thirds step, merge.

packages/services/storage/src/db/types.ts
```diff
@@ -11,6 +11,7 @@ export type alert_channel_type = "SLACK" | "WEBHOOK";
 export type alert_type = "SCHEMA_CHANGE_NOTIFICATIONS";
 export type operation_kind = "mutation" | "query" | "subscription";
 export type organization_type = "PERSONAL" | "REGULAR";
+export type schema_registry_action = "DELETE" | "PUSH";
```
dotansimha, line 14: I can see the changes here but i don't see a migration for those?

packages/web/docs/src/pages/specs/schema-reports.md
```diff
@@ -34,7 +34,8 @@ input SchemaPublishInput {
   """
   GraphQL Hive prevents from publishing breaking changes or broken schemas by default, use this flag to override this behavior.
   """
-  force: Boolean
+  force: Boolean @deprecated(reason: "No longer needed for new projects")
+  """
```
n1ru4l, line 38: why is the field commented out?

packages/services/api/src/shared/schema.ts
```diff
@@ -15,10 +15,14 @@ import {
...
-import { emptySource, Schema, SchemaObject } from './entities';
-
-export function hashSchema(schema: Schema): string {
-  return createHash('md5').update(schema.source, 'utf-8').digest('hex');
+import type { SchemaObject } from './entities';
+
+export function hashSchema(schema: SchemaObject): string {
+  return createHash('md5')
+    .update(schema.source, 'utf-8')
+    .update(`service_name: ${schema.source}`)
```
n1ru4l, line 23: why are we including service name and url in the hash?

packages/services/api/src/modules/schema/resolvers.ts
```diff
@@ -458,16 +435,47 @@ export const resolvers: SchemaModule.Resolvers = {
...
         project: version.project,
         target: version.target,
       });
+
+      if (log.kind === 'single') {
+        return {
+          __typename: 'PushedSchemaLog',
+          author: log.author,
+          commit: log.commit,
+          date: log.date as any,
```
n1ru4l, line 451: You can select the date as a string from the database. with `to_json("fieldname") as "date"`. This is actually the better way of loading the date because if it is loaded as a number it will be inaccurate. E.g. if you load the date as number and then try to query for the record with that number you would get zero results.  See https://tortilla-hq.slack.com/archives/CAY2119MX/p1674036685685139

packages/migrations/src/actions/2023.01.18T11.03.41.registry-v2.sql
```diff
@@ -0,0 +1,117 @@
+-- 
+
+-- Describes an action that was performed on a registry.
+-- PUSH - a new schema was pushed to the registry, an existing schema was updated
+-- DELETE - a schema was deleted from the registry
+CREATE TYPE schema_registry_action AS ENUM ('PUSH', 'DELETE');
```
n1ru4l, line 6: note that enum can not be altered - no new members can be easilz added - so it might be better to make this a text column with insert check or keep the enum validation on the application side

packages/migrations/src/actions/2023.01.18T11.03.41.registry-v2.sql
```diff
@@ -0,0 +1,117 @@
...
+-- Describes the relationship between a schema version and the actions that were performed previously but are related to the matching schema version.
+CREATE TABLE public.schema_version_to_log (
+  version_id uuid NOT NULL REFERENCES public.schema_versions(id) ON DELETE CASCADE,
+  action_id uuid NOT NULL REFERENCES public.schema_log(id) ON DELETE CASCADE,
+  PRIMARY KEY(version_id, action_id)
+);
+
+-- This is used to determine if the project is using the new registry model or the old one.
+ALTER TABLE public.projects
+  ADD COLUMN legacy_registry_model boolean NOT NULL DEFAULT FALSE;
```
n1ru4l, line 47: adding a column with a default value to an existing table will cause every row in the table to be rewritten, which will immediatly double the tables used space and also increase the migration runtime.

It might be better to add the column without a default value and do the default value mapping / interpret null as "false" within the business logic of the application.

packages/services/api/src/shared/schema.ts
kamilkisiela: https://github.com/kamilkisiela/graphql-hive/blob/8d10cb0d96fb15aa2b6c5680aa9b149527f9c940/packages/services/api/src/modules/schema/providers/schema-validator.ts#L76

packages/services/api/src/modules/schema/resolvers.ts
kamilkisiela: Add it as a task to change it everywhere

packages/migrations/src/actions/2023.01.18T11.03.41.registry-v2.sql
kamilkisiela: I don't think it will cause any perf issues

packages/migrations/src/actions/2023.01.18T11.03.41.registry-v2.sql
dotansimha: Yeah, we used it in other places as well. In other projects (lance) we had this problem multiple times :/

packages/services/api/src/shared/schema.ts
n1ru4l: Because a service name and url can change?

dotansimha (approved):

packages/web/docs/src/pages/specs/schema-reports.md
kamilkisiela: Not sure what you mean

PR #858, opened 2022-12-22
1️⃣ refactor(services: rate-limit): migrate from fastify to @whatwg-node/server

YassinEldeeb:
related to https://github.com/kamilkisiela/graphql-hive/issues/657 but doesn't close it entirely

### DIFF
--- a/packages/services/storage/src/index.ts
+++ b/packages/services/storage/src/index.ts
@@ -1813,7 +1813,7 @@ export async function createStorage(connection: string, maximumPoolSize: number)
               SELECT svi.created_at FROM public.schema_versions as svi WHERE svi.id = ${version}
             ) AND sv.target_id = ${target} ORDER BY sv.created_at DESC LIMIT 1
           ) AND sl.action = 'PUSH'
-          ORDER BY c.created_at DESC
+          ORDER BY sl.created_at DESC
         `,
       );
 


packages/services/rate-limit/src/index.ts
```diff
@@ -24,67 +23,28 @@ async function main() {
...
-      rateLimitConfig: {
-        interval: env.limitCacheUpdateIntervalMs,
-      },
-      rateEstimator: env.hiveServices.usageEstimator,
-      emails: env.hiveServices.emails ?? undefined,
-      storage: {
-        connectionString: createConnectionString(env.postgres),
-      },
-    });
+  const logger = createLogger() as FastifyLoggerInstance;
```
dotansimha, line 26: If you are dropping the `fastify` dependency, I think you should also get rid of using the `FastifyXYZ` types?

packages/services/rate-limit/src/index.ts
```diff
@@ -24,67 +23,28 @@ async function main() {
...
+  try {
     registerShutdown({
-      logger: server.log,
+      logger,
       async onShutdown() {
-        await Promise.all([ctx.stop(), server.close()]);
-      },
-    });
-
-    server.route({
```
dotansimha, line 63: Why did you remove these? they are used by the K8s and Docker for health and readiness

packages/services/service-common/src/index.ts
```diff
@@ -5,3 +5,4 @@ export * from './metrics';
 export * from './heartbeats';
```
dotansimha, line 5: I assume we should remove `FastifyLoggerInstance` ?

packages/services/rate-limit/package.json
```diff
@@ -15,10 +15,13 @@
     "@sentry/tracing": "7.21.1",
     "@trpc/client": "10.4.3",
     "@trpc/server": "10.4.3",
+    "@valu/trpc-fetch-api-adapter": "^0.0.3",
     "@whatwg-node/fetch": "0.5.3",
+    "@whatwg-node/server": "^0.4.17",
     "date-fns": "2.29.3",
     "dotenv": "10.0.0",
     "got": "12.5.3",
+    "itty-router": "2.6.6",
```
dotansimha, line 24: @ardatan you mentioned that `itty-router` have some issues recently? should we use it here?

dotansimha (changes requested):
Please refer to the comments

packages/services/rate-limit/src/index.ts
YassinEldeeb: yes, the only concern would be to ensure that all services have been migrated to `whatwg` because I think there maybe is 1 or 2 that are still using fastify which would cause a type error when we change the arg type of `logger` for `registerShutdown` in `services/service-common`

packages/services/rate-limit/src/index.ts
YassinEldeeb: they haven't been removed but extracted to an external router for simplifying the testing process in the future by using the router without the need to launch the server.

https://github.com/kamilkisiela/graphql-hive/pull/858/files#diff-783f53b46dea4ebd59badd5a393cefdb1b6bca12a68e67324e6ffdcbbf89473eR34-R55

packages/services/service-common/src/index.ts
YassinEldeeb: this's the type we're using for the `registerShutdown` `logger` arg to avoid erroring for type conflicts, as you've mentioned above, there's going to be no need for it when we ensure all services have been migrated so that we can safely change the type for it knowing every service that's using it is passing the right arg type

packages/services/rate-limit/package.json
YassinEldeeb: curious about that 👀

packages/services/rate-limit/src/index.ts
dotansimha: I see. I'm also fine with grouping all these PRs into one if they are all similar and does the same thing - and then we can just drop `fastify` types in the same PR?

packages/services/rate-limit/src/index.ts
dotansimha: ok cool!

packages/services/service-common/src/index.ts
dotansimha: Got it, see my comment above.

packages/services/rate-limit/src/index.ts
YassinEldeeb: I was planning to open a single PR with everything but @n1ru4l advised to open a PR per service.

image

YassinEldeeb:
:one:

YassinEldeeb:
I accidentally, discarded the changes in this branch while I was syncing my fork with the latest of the original repo, I'm gonna open another PR with the same content.

PR #930, opened 2022-12-29
Improvements in change detection and the strictness of the registry

kamilkisiela:
- Make `Query.latestVersion` nullable
- Make `Query.latestValidVersion` nullable
- Enforce service name in schema:check
- Remove `Built a new rate-limit map` log
- Pass SDL when available but orchestrator collected composition errors - to still detect changes between versions

### Background

Migrates some of the improvements from #789

### DIFF
--- a/integration-tests/tests/models/stitching.spec.ts
+++ b/integration-tests/tests/models/stitching.spec.ts
@@ -1,0 +1,456 @@
+import { ProjectType } from '@app/gql/graphql';
+import { prepareProject } from '../../testkit/registry-models';
+import { createCLI } from '../../testkit/cli';
+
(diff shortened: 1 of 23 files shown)


integration-tests/tests/models/stitching.spec.ts
```diff
@@ -0,0 +1,456 @@
...
+        type Query {
+          topProduct: String
+        }
+      `,
+      serviceName: 'products',
+      expect: 'approved',
+    });
+  });
+
+  // TODO: this should be rejected, add a check to `main`
```
kamilkisiela, line 331:
```suggestion
```

PR #971, opened 2023-01-09
feat: use non-presigned url for S3 HEAD requests

n1ru4l:
This replaces the AWS s3 SDK for the isomorphic code (Node.js + Cloudflare Workers) with an inlined version of [`aws4fetch`](https://github.com/mhart/aws4fetch).

The reason for this is that we used to generate pre-signed URLs for HEAD requests instead of doing direct requests with the AWS S3 SDK.
We could not do the requests through the AWS S3 SDK as it provided no way of providing a custom `fetch` implementation that is able to successfully execute requests on Cloudflare Workers (https://github.com/aws/aws-sdk-js-v3/issues/3104).

There is another issue with the Cloudflare R2 S3 compatible API. We noticed 500 (https://community.cloudflare.com/t/sometimes-r2-response-code-500/430205/14), sometimes 503, and recently also 499 status codes issued for the pre-signed HEAD requests.

Thus, we want to try non-signed HEAD requests instead, to see if that might reduce the number of failures.

Because of https://github.com/mhart/aws4fetch/issues/22 I had to inline the `aws4fetch` library instead of adding it as a dependency.

### DIFF
--- a/packages/services/api/src/shared/aws.ts
+++ b/packages/services/api/src/shared/aws.ts
@@ -1,0 +1,463 @@
+/**
+ * This is a copy of https://github.com/mhart/aws4fetch which is licensed MIT
+ * See https://github.com/mhart/aws4fetch/issues/22
+ */
+import { crypto, fetch, Headers, Request, TextEncoder } from '@whatwg-node/fetch';
+
+const encoder = new TextEncoder();
+
+const HOST_SERVICES: Record<string, string | void> = {
+  appstream2: 'appstream',
+  cloudhsmv2: 'cloudhsm',
+  email: 'ses',
+  marketplace: 'aws-marketplace',
+  mobile: 'AWSMobileHubService',
+  pinpoint: 'mobiletargeting',
+  queue: 'sqs',
+  'git-codecommit': 'codecommit',
+  'mturk-requester-sandbox': 'mturk-requester',
+  'personalize-runtime': 'personalize',
+};
+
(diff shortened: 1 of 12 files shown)


packages/services/api/src/shared/aws.ts
```diff
@@ -0,0 +1,458 @@
...
+      if (res.status < 500 && res.status !== 429) {
+        return res;
+      }
+      await new Promise(resolve =>
+        setTimeout(resolve, Math.random() * this.initRetryMs * Math.pow(2, i)),
+      );
+    }
+    throw new Error('An unknown error occurred, ensure retries is not negative');
+  }
+}
```
n1ru4l, line 134: @kamilkisiela This thing has 5XX reties included

kamilkisiela: 500 retries is a lot (joke)

Maybe we could use https://github.com/elbywan/wretch or https://github.com/developit/redaxios to not do this the manual way? Not saying we should, just wanted to share the libs with you, you decide.

dotansimha (approved):
LGTM. Let's test it and see how it behaves :) 
@kamilkisiela please review

n1ru4l:
@dotansimha @kamilkisiela  I also added a simple 10x retry for the client SDK.

packages/services/api/src/shared/aws.ts
n1ru4l: @kamilkisiela I am a fan of being explicit over abstractions so I won't wadd this myself

kamilkisiela: why Yoga then? :))

PR #996, opened 2023-01-12
Fix type in Using Registry Doc Page

cookieMr:
### Background

Type in doc

### Description

Type in doc


dotansimha (approved):
Oops, great catch! Thanks @cookieMr !

PR #1003, opened 2023-01-13
feat: store legacy cdn keys on s3

n1ru4l:
### Background

We want to allow creating/revoking of multiple CDN access tokens.
This is the first PR in a series for allowing the handling of both the existing and new CDN access tokens in a backward-compatible way.

### Description

This PR introduces a new database table `cdn_access_tokens`, that is empty by default.
Currently, all issues tokens per target are identical.
In order to prepare for the new dynamic tokens, we want to import the existing already-used target tokens into the SQL database and the S3 bucket.

Every time a user issues a (legacy) token, it will now be written to the S3 bucket, and if the corresponding SQL database record will be created if it does not yet exist.

**Note:** After deploying this to production, we also need to run a manual script that will loop through all the already existing targets and seed the database records for those.

In the next step, we can then roll out the new CDN token management UI + artifatcs API token verification strategy described in https://github.com/the-guild-org/graphql-hive-deployment/issues/1334

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/cdn/providers/cdn.provider.ts
```diff
@@ -44,7 +53,45 @@ export class CdnProvider {
...
+      method: 'PUT',
+      body: tokenHash,
+    });
+
+    if (response.status !== 200) {
+      throw new Error(`Unexpected Status for storing key. (status=${response.status})`);
+    }
+
+    await this.storage.createCDNAccessToken({
+      id: crypto.randomUUID(),
```
n1ru4l, line 81: We don't let the DBMS decide the UUID, as we want to use the same one as within our new tokens that are introduced in a later PR. Conflicts are very unlikely to happen.

n1ru4l:
@dotansimha @kamilkisiela What do you think is the best way to run a script post-deployment for seeding the s3 bucket with the keys of the already existing targets in the database?

packages/migrations/src/actions/down/2023.01.12T17.00.23.cdn-tokens.sql
```diff
@@ -0,0 +1 @@
+raise 'down migration not implemented'
```
dotansimha, line 1: I think it worth implementing it?

packages/services/api/src/modules/cdn/providers/cdn.provider.ts
```diff
@@ -1,17 +1,22 @@
...
-import { CDN_CONFIG } from './tokens';
+import { type S3Config, S3_CONFIG } from '../../shared/providers/s3-config';
+import { Storage } from '../../shared/providers/storage';
+import { type CDNConfig, CDN_CONFIG } from './tokens';
 
 type CdnResourceType = 'schema' | 'supergraph' | 'metadata';
 
 @Injectable({
-  scope: Scope.Singleton,
+  scope: Scope.Operation,
```
dotansimha, line 19: Why this was changed? this will recreate the provider on every GraphQL operation, is that intended?

packages/services/api/src/modules/schema/providers/artifact-storage-writer.ts
```diff
@@ -24,7 +25,7 @@ const artifactMeta = {
  * Write an Artifact to an S3 bucket.
  */
 export class ArtifactStorageWriter {
-  constructor(private s3Client: S3Client, private bucketName: string) {}
+  constructor(@Inject(S3_CONFIG) private s3: S3Config) {}
```
dotansimha, line 28: @n1ru4l you can create a wrapping Provider (as class) if you want to avoid all the manually typing + Inject + using the branded token

dotansimha (approved):

packages/services/api/src/modules/cdn/providers/cdn.provider.ts
n1ru4l: Yes it is because we inject the auth provider for doing access checks

packages/migrations/src/actions/down/2023.01.12T17.00.23.cdn-tokens.sql
n1ru4l: Down migrations are never worth implementing imho

dotansimha (approved):

kamilkisiela (approved):

PR #1027, opened 2023-01-16
prettier should be responsible for import sorting instead of eslint

enisdenjo:
https://github.com/the-guild-org/shared-config/pull/154


dotansimha:
are we good to go with this one? @B2o5T @enisdenjo

B2o5T:
I merged prettier config but proposed to do 2 small enhancements

enisdenjo:
Rebased on main (I've completely deleted the previous history, started afresh). This PR installs https://github.com/the-guild-org/shared-config/pull/162.

B2o5T (approved):

PR #1030, opened 2023-01-16
Use webp in Hive SSO guide to reduce the size of images

kamilkisiela:
90% reduction


saihaj (approved):

saihaj:
maybe we should consider moving all these assets to CF R2 bucket so we don't have too many images in the repo.

PR #1043, opened 2023-01-17
feat: store/import legacy s3 key on s3 [2/4]

n1ru4l:
Follow up PR for https://github.com/kamilkisiela/graphql-hive/pull/1003

This runs a database migration for importing the legacy CDN access keys of targets into Postgres/S3.

The migration is skipped when environment variables for running it are not set.


packages/migrations/src/actions/2023.01.17T10.46.28.import-legacy-s3-keys-to-database.mts
```diff
@@ -0,0 +1,613 @@
...
+    `;
+
+    const items = await connection.query(query);
+    return TargetsModel.parse(items.rows);
+  }
+
+  let lastCursor: null | Cursor = null;
+
+  do {
+    const items: zod.TypeOf<typeof TargetsModel> = await getPaginationTargets(lastCursor);
```
dotansimha, line 102: @n1ru4l any real reason to do a paginated query here? I mean, why not just fetch it all? 😆

packages/migrations/src/actions/2023.01.17T10.46.28.import-legacy-s3-keys-to-database.mts
```diff
@@ -0,0 +1,613 @@
...
+      // After creation on database
+      const response = await s3Client.fetch(
+        [env.S3_ENDPOINT, env.S3_BUCKET_NAME, s3Key].join('/'),
+        {
+          method: 'PUT',
+          body: accessKeyHash,
+        },
+      );
+
+      // Should probably do a retry here
```
dotansimha, line 119: what's the scenario? is it possible that this will happen frequently?

packages/migrations/src/actions/2023.01.17T10.46.28.import-legacy-s3-keys-to-database.mts
```diff
@@ -0,0 +1,613 @@
...
+      if (response.status !== 200) {
+        throw new Error(`Unexpected Status for storing key. (status=${response.status})`);
```
dotansimha, line 121: seems like the rest of the migration script is kinda graceful if it runs twice, any reason to throw an error error and just stop the migration from running?

packages/migrations/src/actions/2023.01.17T10.46.28.import-legacy-s3-keys-to-database.mts
```diff
@@ -0,0 +1,613 @@
...
+      lastCursor = {
+        lastId: items[items.length - 1].id,
+        lastCreatedAt: items[items.length - 1].created_at_cursor,
+      };
+    }
+  } while (lastCursor !== null);
+};
+
+export const down: slonik.Migration = async ({ context: { connection, sql } }) => {
+  await connection.query(sql`raise 'down migration not implemented'`);
```
dotansimha, line 159: you can just `throw` here

packages/migrations/src/actions/2023.01.17T10.46.28.import-legacy-s3-keys-to-database.mts
```diff
@@ -0,0 +1,613 @@
...
+    cache?: Map<string, ArrayBuffer>;
+    datetime?: string;
+    signQuery?: boolean;
+    appendSessionToken?: boolean;
+    allHeaders?: boolean;
+    singleEncode?: boolean;
+  };
+};
+
+class AwsClient {
```
dotansimha, line 206: any reason to copy the code here? we should be able to use deps, no?

n1ru4l: Not sure how it would work, to be honest. 😓 

We also have this code within `@hive/api`.

I would prefer to avoid `@hive/api` being a dependency of this migration.
E.g. `@hive/api` could change and then we would have to adjust the DB migrations accordingly.
I would rather prefer the migration to be a frozen"" standalone piece of that never needs to change.

But maybe I am being overcautious here? 😁

packages/migrations/src/actions/2023.01.17T10.46.28.import-legacy-s3-keys-to-database.mts
n1ru4l: this is what the generator generated 🤣

packages/migrations/src/actions/2023.01.17T10.46.28.import-legacy-s3-keys-to-database.mts
n1ru4l: The AwsClient already does a retry internall

packages/migrations/src/actions/2023.01.17T10.46.28.import-legacy-s3-keys-to-database.mts
n1ru4l: Best practices? 😁 I want to avoid selecting everything from the table 😁

packages/services/api/src/shared/aws.ts
```diff
@@ -122,7 +122,7 @@ export class AwsClient {
         return fetched; // No need to await if we're returning anyway
       }
       const res = await fetched;
-      if (res.status < 500 && res.status !== 429) {
+      if (res.status < 500 && res.status !== 429 && res.status !== 499) {
```
kamilkisiela, line 125: nginx <3

kamilkisiela (approved):

packages/services/api/src/shared/aws.ts
n1ru4l: cloudflare our friend

PR #1055, opened 2023-01-18
Use the same image of Redis in all docker compose files

kamilkisiela:


dotansimha (approved):
@kamilkisiela heads up that this will also affect production environment, because of https://github.com/kamilkisiela/graphql-hive/blob/main/deployment/utils/redis.ts#L17

dotansimha:
Seems like `redis` image is breaking things, let's wait with this one

PR #1072, opened 2023-01-19
Change strict char limit for URLs from 200 to 500, revert Redis image for integration tests 

dotansimha:


packages/services/api/src/modules/alerts/resolvers.ts
```diff
@@ -13,7 +13,7 @@ export const resolvers: AlertsModule.Resolvers = {
     async addAlertChannel(_, { input }, { injector }) {
       const AddAlertChannelModel = z.object({
         slack: MaybeModel(z.object({ channel: SlackChannelNameModel })),
-        webhook: MaybeModel(z.object({ endpoint: z.string().url().max(200) })),
+        webhook: MaybeModel(z.object({ endpoint: z.string().url() })),
```
n1ru4l, line 16: pls still have a upper limit

dotansimha: did 500

n1ru4l (approved):

PR #1107, opened 2023-01-23
Increase max replicas of usage and usage-ingestor

kamilkisiela:


dotansimha (approved):

deployment/services/usage.ts
```diff
@@ -35,7 +35,7 @@ export function deployUsage({
 }) {
   const replicas = 1; /*isProduction(deploymentEnv) ? 2 : 1*/
```
dotansimha, line 36: @kamilkisiela should we do 2 by default? or 3?

kamilkisiela, line 36:
```suggestion
  const replicas = isProduction(deploymentEnv) ? 2 : 1;
```

kamilkisiela: @dotansimha

PR #1132, opened 2023-01-25
Pass errors coming from the external composition to the end-user

### LINKED ISSUE #1128: Error during external composition are not exposed to the end-user
In our logs:

```
External composition failure: 500 {"error":{"code":"500","message":"A server error has occurred"}}
```

Users see in the CLI:

```
:heavy_multiplication_x: Failed to publish schema
 »   Error: Unexpected error.
 »   Reference: 5ea78e05-bf5d-45f3-8cd0-858c42b60165
```

kamilkisiela:
Closes #1128

### DIFF
--- a/integration-tests/tests/api/schema/external-composition.spec.ts
+++ b/integration-tests/tests/api/schema/external-composition.spec.ts
@@ -81,3 +81,93 @@ test.concurrent('call an external service to compose and validate services', asy
   // expect `products` service to be composed externally
   await expect(history()).resolves.toContainEqual(productsServiceName);
 });
+
+test.concurrent(
+  'the error coming from the external composition service should be visible to the user',
+  async () => {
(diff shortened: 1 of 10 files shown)


integration-tests/tests/api/schema/external-composition.spec.ts
```diff
@@ -81,3 +81,93 @@ test.concurrent('call an external service to compose and validate services', asy
   // expect `products` service to be composed externally
   await expect(history()).resolves.toContainEqual(productsServiceName);
 });
+
+test.concurrent(
```
dotansimha, line 85: @kamilkisiela can we please also have a test that tried to call a service that does not exists?  just to make sure the `fetch` network error is also visible to the user

kamilkisiela: that's exactly what I'm doing now

PR #1134, opened 2023-01-25
Make tokens service replicable

kamilkisiela:
Slow `set`, fast `get` which is totally fine.
Tested it locally and it has the same performance as the current `tokens` service (over 12k reqs/s on my machine).

The idea here is to have `token:{hashed_token}` available directly ~~but tagged with `organization:{id}`, `project:{id}` and `target:{id}` so we could invalidate it quickly~~. We could get rid of tags and invalidate the tokens by fetching them from PG (when org, project or target is deleted). I will consider the second approach as it's easier.

The reason why we want to have Redis here is to allow for multiple replicas of the `tokens` service. We could reuse what we have today (the in-memory cache) and enhance it with a PubSub (over Redis) to invalidate the cache in all replicas, but it's too weird and has a similar performance.
Why multiple replicas if the `tokens` service can handle 12k reqs/s? Just to have a fallback service in case something happens (and yet we still end up with a single point of failure, this time Redis :))

Every request for a token is cached for 60 seconds in memory and for 5 minutes in Redis (to force clean the tokens, in case we missed something). We read from the in-memory cache first, to offload Redis.

In case when we got a request for a non-existing token (in the DB), we keep it in the in-memory cache for 60 seconds.
In case there was an exception when reading the token from the DB, we keep the error in the in-memory cache for 60 seconds.

Every token is read on demand, we don't fill up the cache, only if requested.


kamilkisiela:
@dotansimha I enabled at least 2 replicas for all services on dev and staging, just to match it with production. If we would have use a single replica of the tokens service, we might miss some issues.

PR #1253, opened 2023-02-06
chore: group supertokens?

n1ru4l:
### Background

For some reason renovate creates two PRs instead of a single one...

- https://github.com/kamilkisiela/graphql-hive/pull/1246
- https://github.com/kamilkisiela/graphql-hive/pull/1193

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -53,7 +53,12 @@
     },
     {
       "groupName": "supertokens",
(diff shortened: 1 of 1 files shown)


kamilkisiela:
One is major bump, one is not. I think it's normal @dotansimha

PR #1256, opened 2023-02-06
Remove twin.macro, use pure tailwindcss3

B2o5T:
Plz run the app to see that nothing was broken


packages/web/app/src/components/common/Navigation.tsx
```diff
@@ -212,7 +211,6 @@ export function Navigation() {
             </div>
           </div>
           <ThemeButton />
-          <div className="hive-release-notes" />
```
B2o5T, line 215: does this class is somewhere used?

packages/web/app/src/components/project/migration-toast.tsx
```diff
@@ -19,7 +19,7 @@ export function ProjectMigrationToast({ projectId, orgId }: { projectId: string;
         href={`/${orgId}/${projectId}/view/settings`}
         variant="link"
         size="small"
-        className="shrink-1 mr-2 self-center text-sm font-medium text-gray-300"
```
B2o5T, line 22: `shrink-1` not exist, seems typo

enisdenjo (approved):

packages/web/app/src/components/common/Navigation.tsx
dotansimha: I think it's unused?

dotansimha:
@B2o5T my impression so far:

### `Create Project` does not work

This button on the main org page:
image

is no longer functional; I'm not able to click on it at all. 

### Modals position

image

image

image

seems a bit off, no longer centered 

 ### CDN `Connect` button is not working 

image

can't click at all 

> I tested all other screens and it looks the same as before ✅

dotansimha (changes requested):
Please refer to my comment above.

PR #1275, opened 2023-02-07
Alter instead of creating a new table

kamilkisiela:
To speed up the migration and move on :)


docker/docker-compose.dev.yml
```diff
@@ -53,6 +53,8 @@ services:
     environment:
       MINIO_ROOT_USER: minioadmin
       MINIO_ROOT_PASSWORD: minioadmin
+    volumes:
```
dotansimha, line 56: intended?

dotansimha (approved):

PR #1407, opened 2023-02-17
Fix an issue in Federation when a type definition is not found for a type extension

kamilkisiela:
[Failing test](https://github.com/kamilkisiela/graphql-hive/actions/runs/4205307339/jobs/7297201194)
Apollo Federation struggles with `extend type EntityWithoutDefinition`, this PR fixes it


scripts/patch-manifests.js
```diff
@@ -1,6 +1,5 @@
 /*
   https://github.com/octokit/webhooks-methods.js/issues/45
-
```
dotansimha, line 3: any reason for all of those extra spaces? 😆

patches/@octokit__webhooks-methods@3.0.1.patch
```diff
@@ -0,0 +1,12 @@
...
+index 7c6e9a1af624be81938679edba36a046fd78f5f9..3dd36bc7573673a082dd8cc4910e0e3aca51c4a9 100644
+--- a/package.json
++++ b/package.json
+@@ -42,6 +42,5 @@
+   },
+   "source": "dist-src/index.js",
+   "types": "dist-types/index.d.ts",
+-  "main": "dist-node/index.js",
+-  "module": "dist-web/index.js"
++  "main": "dist-node/index.js"
```
dotansimha, line 11: this one is something else? or it's here on purpose?

scripts/patch-manifests.js
kamilkisiela: `prettier --write`

patches/@octokit__webhooks-methods@3.0.1.patch
kamilkisiela: We have two versions

kamilkisiela: indirect dependencies

PR #1451, opened 2023-02-22
remove redundant `globalThis.window` in favour just `window`

B2o5T:


packages/web/app/pages/_app.tsx
```diff
@@ -124,7 +124,7 @@ export default function App({ Component, pageProps }: AppProps): ReactElement {
     </>
   );
 }
-if (globalThis.window) {
+if (typeof window !== 'undefined') {
```
B2o5T, line 127:
```suggestion
if (globalThis.window) {
```

packages/web/app/pages/auth/[[...path]].tsx
```diff
@@ -46,7 +46,7 @@ export default function Auth(props: { oidcProviderId: string | null }): ReactEle
     // In case we are directed here from the Okta dashboard we automatically start the login flow.
     const isOkta =
       env.auth.okta !== null &&
-      new URLSearchParams(globalThis.window?.location.search ?? '').get('provider') === 'okta';
+      new URLSearchParams(window.location.search ?? '').get('provider') === 'okta';
```
Comment (n1ru4l) [inline packages/web/app/pages/auth/[[...path]].tsx:49]: We can also just use `globalThis?.location/search`

I prefer `globalThis` as it does not need a `typof window === ""undefined` check.

Comment (B2o5T) [reply packages/web/app/pages/auth/[[...path]].tsx:49]: `globalThis?.location/search` make it confusing since we can't figure out which environment it's

> I prefer globalThis as it does not need a typof window === ""undefined check.

in the above code `windows` is always defined

Comment (n1ru4l) [reply packages/web/app/pages/auth/[[...path]].tsx:49]: In that case `globalThis.window?.location` seems like the best trade-off to me. No need to `typeof window === "undefined"` checks + it is always clear that we access `window`.

> in the above code window is always defined
> For me it is more about consistency instead of doing it different based on in which function the code is written

Comment (dotansimha) [reply packages/web/app/pages/auth/[[...path]].tsx:49]: I agree with @n1ru4l on this one. 
@B2o5T what should we do with this PR?

PR #1452, opened 2023-02-22
2️⃣ Drop Personal organizations 

dotansimha:
Step 2 of https://github.com/kamilkisiela/graphql-hive/pull/1450 

Closes https://github.com/kamilkisiela/graphql-hive/issues/98

## TODO

- [x] Add a migration to delete all empty Personal organizations
- [x] Delete column `organizations.type` and delete PG type `organization_type` 
- [x] Delete `OrganizationType` from GraphQL API
- [x] Adjust all GraphQL queries and mutations 
- [x] Update admin panel ("Manage Instance")
- [x] Update web app
- [x] Make sure migration creates the correct permissions for organizations that previously marked as Personal
- [x] Adjust integration tests if needed 
- [x] Add setup for migrations testing + CI

### Screenshots

image

image

### DIFF
--- a/codegen.cjs
+++ b/codegen.cjs
@@ -11,7 +11,7 @@ const config = {
       plugins: [
         {
           add: {
-            content: "import { StripeTypes } from '@hive/stripe-billing';",
+            content: "import type { StripeTypes } from '@hive/stripe-billing';",
           },
         },
         'typescript',
@@ -26,7 +26,6 @@ const config = {
         immutableTypes: true,
         contextType: 'GraphQLModules.ModuleContext',
         enumValues: {
-          OrganizationType: '../shared/entities#OrganizationType',
           ProjectType: '../shared/entities#ProjectType',
           TargetAccessScope: '../modules/auth/providers/target-access#TargetAccessScope',
           ProjectAccessScope: '../modules/auth/providers/project-access#ProjectAccessScope',
--- a/packages/migrations/src/actions/2023.02.22T09.27.02.delete-personal-org.sql
+++ b/packages/migrations/src/actions/2023.02.22T09.27.02.delete-personal-org.sql
@@ -1,0 +1,13 @@
+-- Find and delete all organizations of type PERSONAL that have no projects
+DELETE FROM public.organizations as o
+WHERE
+  o.type = 'PERSONAL'
+  AND NOT EXISTS (
+    SELECT id from public.projects as p WHERE p.org_id = o.id LIMIT 1
+  );
+
+-- Delete the "type" column from organizations
+ALTER TABLE public.organizations DROP COLUMN type;
+
+-- Delete the "organization_type" enum, as it's unused now
+DROP TYPE organization_type;
--- a/.github/workflows/migrations-test.yaml
+++ b/.github/workflows/migrations-test.yaml
@@ -1,0 +1,38 @@
+on:
+  workflow_call:
+
+jobs:
+  test:
+    runs-on: ubuntu-22.04
+
+    services:
+      postgres:
+        image: postgres:14.6-alpine
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_USER: postgres
+          POSTGRES_DB: registry
+        options: >-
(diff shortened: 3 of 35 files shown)


codegen.cjs
```diff
@@ -11,7 +11,7 @@ const config = {
       plugins: [
         {
           add: {
-            content: "import { StripeTypes } from '@hive/stripe-billing';",
+            content: "import type { StripeTypes } from '@hive/stripe-billing';",
```
dotansimha, line 14: just a tiny fix

packages/migrations/src/actions/2023.02.22T09.27.02.delete-personal-org.sql
```diff
@@ -0,0 +1,13 @@
+-- Find and delete all organizations of type PERSONAL that have no projects
+DELETE FROM public.organizations as o
+WHERE
+  o.type = 'PERSONAL'
+  AND (
+    SELECT COUNT(*) from public.projects as p WHERE p.org_id = o.id
+  ) = 0;
```
kamilkisiela, line 7: We need to double check that and check how long it will take

turbo.json
```diff
@@ -7,7 +7,8 @@
     "tsconfig.json",
     "scripts/patch-manifests.js",
     "scripts/banner.js",
-    "patches/*"
+    "patches/*",
+    "packages/services/api/src/shared/*"
```
kamilkisiela, line 11: Why? I guess what you tried to fix is caused by having relative imports, we should stick to `import { ... } from '@hive/xyz'` instead of `import { ... } from '.././packages/xyz'` and we will follow what Turbo expects from us.

packages/migrations/src/actions/2023.02.22T09.27.02.delete-personal-org.sql
dotansimha: I checked and it takes 4s to run this as `SELECT`. Since all of the found records (844 in total) are not references to any other object, i don't expect the `DELETE` to take much longer?

n1ru4l: couldnt you just instead fo

`SELECT id from public.projects where project.org_id = o.id LIMIT 1`?

If there is one row that should be sufficient, no need to get all

dotansimha: for sure!

dotansimha: i did:
```
AND NOT EXISTS (
    SELECT id from public.projects as p WHERE p.org_id = o.id LIMIT 1
  );
```

even faster 🚀

n1ru4l (approved):
Okay but quoting myself from slack:

> There will be a time-frame where the old app container might try to fetch `organization.type` `, but the column got already removed by the migration. The better approach here would be to first remove the non-null constraint via migration, then deploy server changes to no longer select/insert the type column; then do another deployment with a database migration that drops the column for sure.
>
> Regarding the migrations, they are atm running in a transaction and blocking all other operations on the table. I think the migration that deletes the organizations could instead probably run outside a transaction and use single statements (per organization), instead of one big for all.
>
> Not saying that we need to do it like this now. I just think at some point we need to do it like this, otherwise we cause downtime/degraded performance.

dotansimha:
> > There will be a time-frame where the old app container might try to fetch `organization.type` `, but the column got already removed by the migration. The better approach here would be to first remove the non-null constraint via migration, then deploy server changes to no longer select/insert the type column; then do another deployment with a database migration that drops the column for sure.

Fully agree. And I think we are getting pretty close to this point.  

> > Regarding the migrations, they are atm running in a transaction and blocking all other operations on the table. I think the migration that deletes the organizations could instead probably run outside a transaction and use single statements (per organization), instead of one big for all.
> > Not saying that we need to do it like this now. I just think at some point we need to do it like this, otherwise we cause downtime/degraded performance.

Yeah, that's why I tried to optimize the SQL statement we had there. It seems better and faster now. We can use the stats from staging and see if it makes sense, or if an additional optimization (or, break to 2 parts) is needed.

PR #1454, opened 2023-02-22
Fix missing list of changes and add a list of messages to notifications

kamilkisiela:


integration-tests/tests/models/single.spec.ts
```diff
@@ -68,6 +68,49 @@ describe('publish', () => {
       expect: 'ignored',
     });
   });
+
+  test.concurrent('CLI output', async () => {
```
B2o5T, line 72:
```suggestion
  test.concurrent('CLI output', async ({ expect }) => {
```

I guess when using `test.concurrent` you need to destructure `expect`

kamilkisiela: none of our `test.concurrent` tests gets `expect` like this

B2o5T: oops, you need to destructure only if you use `toMatchInlineSnapshot` inside `test.concurrent` in vitest

kamilkisiela: :)

PR #1513, opened 2023-02-27
Upgrade Postgres to v14

dotansimha:
Partially closes https://github.com/kamilkisiela/graphql-hive/issues/902 

### TODO 

- [x] Upgrade `slonik/migrator` and other Slonik utils to latest 
- [x] Upgrade `slonik` to latest version that doesn't have breaking changes 
- [x] Upgrade CI `services` to match PG version we are using 
- [x] Upgrade Docker compose versions  
- [x] Testing


patches/slonik@30.1.2.patch
```diff
@@ -1,30 +0,0 @@
-diff --git a/dist/src/routines/executeQuery.js b/dist/src/routines/executeQuery.js
```
dotansimha, line 1: @kamilkisiela I think it's fine?

kamilkisiela: `get-stack-trace` was causing some damage but I don't remember why. I add comments now to remember but this patch was done a year ago or something.

kamilkisiela: I think we should keep using this patch, just in case and once we figure out if we should use Slonik (upgrade to the latest) or migrate to some other library, we will decide if we want to keep it or not.

dotansimha: Yeah I think it seems fine now, I also remember an issue we had but I wasn't sure how to reproduce?

kamilkisiela: "it seems fine now" maybe :) We'll see when we deploy it to production

PR #1514, opened 2023-02-27
Make PG connection check as part of readiness check

kamilkisiela:
This should restart the pod if it's unhealthy


dotansimha (approved):
LGTM 🚀

PR #1519, opened 2023-02-27
Replace tRPC with FETS in rate-limit

ardatan:


packages/services/rate-limit/src/api.ts
```diff
@@ -1,50 +1,166 @@
...
+    },
+  },
+  additionalProperties: false,
+  required: ['id', 'entityType', 'type', 'token'],
+} as const;
 
-export type RateLimitInput = z.infer<typeof VALIDATION>;
+export type RateLimitInput = FromSchema<typeof VALIDATION>;
 
-const VALIDATION = z
```
Urigo, line 19: to make the diff smaller, we can actually use the same zod schema and use https://github.com/StefanTerdell/zod-to-json-schema ?
just to show the actual difference between tRPC and this instead of also json-schema<->zod change?

ardatan: But it should be a constant type definition. I don't think that library provides that.

packages/services/api/src/modules/rate-limit/providers/rate-limit.provider.ts
```diff
@@ -1,20 +1,21 @@
...
+type CheckRateLimitInput = RateLimitApiInput['/check-rate-limit']['post']['json'];
+
 @Injectable({
   global: true,
   scope: Scope.Singleton,
 })
 export class RateLimitProvider {
   private logger: Logger;
-  private rateLimit;
+  private rateLimit: RateLimitApi['__client'] | null;
```
dotansimha, line 18: Can we consider improving the API of this? the `[__client]` feels like a hack?

packages/services/rate-limit/package.json
```diff
@@ -11,13 +11,15 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@graphql-yoga/logger": "0.0.1-alpha-20230303125454-5d621255",
```
dotansimha, line 14: let's avoid using alpha if it's already released

packages/services/rate-limit/src/api.ts
```diff
@@ -1,50 +1,149 @@
...
 
 export interface Context {
-  req: FastifyRequest;
   limiter: Limiter;
 }
 
-const t = initTRPC.context<Context>().create();
-const errorMiddleware = t.middleware(handleTRPCError);
-const procedure = t.procedure.use(errorMiddleware);
+const VALIDATION = {
```
dotansimha, line 18: can we have some type checks on this object? like `RateLimitValidation: JsonSchema`?

packages/services/rate-limit/src/api.ts
```diff
@@ -1,50 +1,149 @@
...
 
 export type RateLimitApi = typeof rateLimitApiRouter;
-export type RateLimitApiInput = inferRouterInputs<RateLimitApi>;
-export type RateLimitApiOutput = inferRouterOutputs<RateLimitApi>;
+export type RateLimitApiInput = RouterInput<RateLimitApi>;
+export type RateLimitApiOutput = RouterOutput<RateLimitApi>;
+
+export const rateLimitApiRouter = createRouter<Context>({
+  title: 'Rate Limit API',
+  version: '0.0.0',
```
dotansimha, line 52: can we avoid the need for these by default? (title, version)

packages/services/rate-limit/src/api.ts
ardatan: No because it needs to be a `const` type to infer types from JSON Schema correctly.

packages/services/rate-limit/src/api.ts
ardatan: Yes of course, I just want to show them :)

packages/services/api/src/modules/rate-limit/providers/rate-limit.provider.ts
```diff
@@ -23,18 +24,13 @@ export class RateLimitProvider {
...
     this.rateLimit = rateLimitServiceConfig.endpoint
-      ? createTRPCProxyClient<RateLimitApi>({
-          links: [
-            httpLink({
-              url: `${rateLimitServiceConfig.endpoint}/trpc`,
-              fetch,
-            }),
-          ],
+      ? createClient({
+          endpoint: rateLimitServiceConfig.endpoint,
```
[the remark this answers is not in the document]
ardatan: Currently `fetch` can be replaced but we can add a support for middlewares on the client side.

packages/services/api/src/modules/rate-limit/providers/rate-limit.provider.ts
ardatan: I forgot to change it, we have `RouterClient` type.

packages/services/rate-limit/package.json
ardatan: It hasn't been released yet. I'll update it after we release.

packages/services/api/src/modules/rate-limit/providers/rate-limit.provider.ts
ardatan: Done! 👍

packages/services/rate-limit/package.json
ardatan: Done!

Parables:
Great work @ardatan 🥂

Please look into the following:
https://github.com/kamilkisiela/graphql-hive/actions/runs/4343125662/jobs/7584781721#step:4:172
https://github.com/kamilkisiela/graphql-hive/actions/runs/4343125662/jobs/7584781721#step:4:172

Thanks very much

packages/services/api/src/modules/rate-limit/providers/rate-limit.provider.ts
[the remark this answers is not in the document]
ardatan: I just added a plugin system in the client as well 
https://github.com/ardatan/fets/blob/master/packages/fets/src/client/types.ts#L135

packages/services/rate-limit/src/api.ts
ardatan: We can also implement a helper lib like Zod.

packages/services/rate-limit/src/api.ts
ardatan: Changed to Zod 👍

packages/services/rate-limit/src/api.ts
ardatan: Changed to Zod

dotansimha:
Closing. This PR is outdated and we'll reconsider it later.

PR #1572, opened 2023-03-02
docs: add callout to have TLS1.2 or higher to install CLI

saihaj:
### Background

For security compliance we use `TSL v1.2` if users are running `1.1` then using `cUrl` they will not be able to install the CLI.

### Description

| Before | After | 
| ------|-------|
|CleanShot 2023-03-02 at 15 41 48 | CleanShot 2023-03-02 at 15 41 23|

### DIFF
--- a/packages/web/docs/src/pages/features/publish-schema.mdx
+++ b/packages/web/docs/src/pages/features/publish-schema.mdx
@@ -21,6 +21,8 @@ Install [`@graphql-hive/cli`](https://www.npmjs.com/package/@graphql-hive/cli) p
 
 Install latest version:
 
+<Callout type="info">You need TLS v1.2 or higher to install this script.</Callout>
+
 ```bash
 curl -sSL https://graphql-hive.com/install.sh | sh
 ```
--- a/packages/web/docs/src/pages/features/publish-schema.mdx
+++ b/packages/web/docs/src/pages/features/publish-schema.mdx
@@ -21,7 +21,7 @@ Install [`@graphql-hive/cli`](https://www.npmjs.com/package/@graphql-hive/cli) p
 
 Install latest version:
 
-<Callout type="info">You need TLS v1.2 or higher to install this script.</Callout>
+<Callout type="warning">You need TLS v1.2 or higher to install this script.</Callout>
 
 ```bash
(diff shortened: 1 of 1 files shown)


B2o5T (approved):

packages/web/docs/src/pages/features/publish-schema.mdx
```diff
@@ -21,6 +21,8 @@ Install [`@graphql-hive/cli`](https://www.npmjs.com/package/@graphql-hive/cli) p
 
 Install latest version:
 
+<Callout type="info">You need TLS v1.2 or higher to install this script.</Callout>
```
B2o5T, line 24:
```suggestion
<Callout type="warning">You need TLS v1.2 or higher to install this script.</Callout>
```
make change type to give more attention?

packages/web/docs/src/pages/features/publish-schema.mdx
```diff
@@ -21,6 +21,8 @@ Install [`@graphql-hive/cli`](https://www.npmjs.com/package/@graphql-hive/cli) p
 
 Install latest version:
 
+<Callout type="warning">You need TLS v1.2 or higher to install this script.</Callout>
```
dotansimha, line 24: I feel like it might be clear on "how do I check I have TLS 1.2" or "what do I do if I don't have it"?

saihaj: It is kinda tricky because if you just do `curl --version` on Mac it may not tell you what TLS protocols are supported directly but it uses SecureTransport which supports it. 
CleanShot 2023-03-03 at 09 44 33

Parables: I think it would be convenient to include this in the docs.

I believe that was what @dotansimha  was suggesting... Instead of just displaying a callout warning, 
have this information included in a Prerequisite/ Requirement section of the docs and have the callout warning link to that section.

Love your work @saihaj  💖

dotansimha:
Closing, it's referring to outdated docs. We fixed it since than, and I think we also changed the TLS of the domain, it should work also with 1.1

PR #1652, opened 2023-03-09
Timout errors should be visible by the user

kamilkisiela:


integration-tests/tests/api/schema/external-composition.spec.ts
```diff
@@ -272,3 +272,95 @@ test.concurrent(
...
+        type Product @key(fields: "id") {
+          id: ID!
+          name: String
+        }
+      `,
+      service: productsServiceName,
+    })
+    .then(r => r.expectNoGraphQLErrors());
+
+  // Schema publish should be unsuccessful and the error coming from the external composition service should be visible
```
kamilkisiela, line 348:
```suggestion
  // Schema publish should be unsuccessful and the timeout error should be visible
```

dotansimha (approved):

integration-tests/tests/api/schema/external-composition.spec.ts
```diff
@@ -272,3 +272,95 @@ test.concurrent(
...
+      __typename: 'SchemaPublishError',
+      changes: {
+        total: 0,
+        nodes: [],
+      },
+      errors: {
+        total: 1,
+        nodes: [
+          {
+            message: expect.stringContaining('Timeout: took longer than'),
```
kamilkisiela, line 360:
```suggestion
            message: expect.stringMatching('Timeout: took longer than'),
```

PR #1716, opened 2023-03-14
Remove `graphql-request` and use raw `fetch` in Hive CLI 

dimaMachina:
Closes https://github.com/kamilkisiela/graphql-hive/issues/1703 https://github.com/kamilkisiela/graphql-hive/pull/1563

### DIFF
--- a/packages/libraries/cli/package.json
+++ b/packages/libraries/cli/package.json
@@ -60,7 +60,6 @@
     "colors": "1.4.0",
     "env-ci": "7.3.0",
     "graphql": "^16.8.1",
-    "graphql-request": "5.1.0",
     "graphql-tag": "2.12.6",
     "log-symbols": "4.1.0",
     "mkdirp": "2.1.6",
--- a/packages/libraries/cli/src/base-command.ts
+++ b/packages/libraries/cli/src/base-command.ts
@@ -1,11 +1,11 @@
 import colors from 'colors';
 import { print, type GraphQLError } from 'graphql';
+import type { ExecutionResult } from 'graphql';
 import symbols from 'log-symbols';
+import type { TypedDocumentNode } from '@graphql-typed-document-node/core';
 import { Command, Errors, Config as OclifConfig } from '@oclif/core';
-import { Config, GetConfigurationValueType, ValidConfigurationKeys } from './helpers/config';
 import { fetch } from '@whatwg-node/fetch';
-import type { TypedDocumentNode } from "@graphql-typed-document-node/core";
-import type { ExecutionResult } from 'graphql'
+import { Config, GetConfigurationValueType, ValidConfigurationKeys } from './helpers/config';
 
 type OmitNever<T> = { [K in keyof T as T[K] extends never ? never : K]: T[K] };
 
@@ -146,6 +146,7 @@ export default abstract class extends Command {
 
   registryApi(registry: string, token: string) {
     const requestHeaders = {
+      'Content-Type': 'application/json',
       Accept: 'application/json',
       'User-Agent': `hive-cli/${this.config.version}`,
       Authorization: `Bearer ${token}`,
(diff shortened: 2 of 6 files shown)


packages/libraries/cli/package.json
```diff
@@ -59,16 +59,14 @@
     "env-ci": "7.3.0",
     "git-parse": "3.0.1",
     "graphql": "^16.0.0",
-    "graphql-request": "5.1.0",
+    "@whatwg-node/fetch": "0.8.2",
     "graphql-tag": "2.12.6",
     "log-symbols": "4.1.0",
     "mkdirp": "2.1.5",
     "rimraf": "4.4.0",
-    "ts-node": "10.9.1",
```
B2o5T, line 67: nowhere used for this package

dotansimha: should we also remove `graphql-tag`? I think we don't really need it?

packages/libraries/cli/src/base-command.ts
```diff
@@ -173,6 +174,13 @@ export default abstract class extends Command {
   }
 }
 
+type ClientError = Error & {
```
dotansimha, line 177: I'm not sure that we need that now?

dotansimha: becuase the structure is different, we'll got `{ data: {}, errors: [] }`

packages/libraries/cli/src/base-command.ts
```diff
@@ -119,18 +120,18 @@ export default abstract class extends Command {
...
+      headers: {
+        Accept: 'application/json',
+        'User-Agent': `hive-cli/${this.config.version}`,
+        Authorization: `Bearer ${token}`,
+        'graphql-client-name': 'Hive CLI',
+        'graphql-client-version': this.config.version,
+      },
+    });
+
+    return response.json();
```
kamilkisiela, line 134: What if we get 5XX or 4XX error? :)

dotansimha:
I rebased and made adjustments to match the usage of client-preset in the CLI. 
Also, I added error handling similar to what GraphQL-Request was running before.

packages/libraries/cli/src/base-command.ts
dotansimha: fixed, now it's the same as graphql-request (throwing an error). Also added handling of `response.errors` with status 200

packages/libraries/cli/src/base-command.ts
dotansimha: fixed

dotansimha (approved):

kamilkisiela:
Let’s think if it’s a breaking change.
Does a user need to upgrade to newer version of node?

dotansimha:
> Let’s think if it’s a breaking change. Does a user need to upgrade to newer version of node?

It will require >18 as far as I understand maybe? (because of native `fetch` deps)

packages/libraries/cli/src/base-command.ts
```diff
@@ -142,17 +145,41 @@ export default abstract class extends Command {
...
+    const requestHeaders = {
+      Accept: 'application/json',
+      'User-Agent': `hive-cli/${this.config.version}`,
+      Authorization: `Bearer ${token}`,
+      'graphql-client-name': 'Hive CLI',
+      'graphql-client-version': this.config.version,
+    };
+
+    return {
+      async request<TData = any, TVariables = Record<string, any>>(operation: TypedDocumentNode<TData, TVariables>, variables: TVariables) {
```
n1ru4l, line 157: Here we can use the function signature as mentioned in https://github.com/kamilkisiela/graphql-hive/issues/1703#issuecomment-1588634237

https://github.com/dotansimha/graphql-code-generator/blob/2139f1db5704aa34a0cd89342736bf67ec556da3/examples/yoga-tests/src/yoga.spec.ts#L6-L23

The benefit of this one is that `TVariables` will be an optional parameter, based on whether variables are defined or not.

packages/libraries/cli/src/commands/operations/check.ts
```diff
@@ -115,7 +115,7 @@ export default class OperationsCheck extends Command {
         return;
       }
 
-      const result = await this.registryApi(endpoint, accessToken).request(fetchLatestVersionQuery);
+      const result = await this.registryApi(endpoint, accessToken).request(fetchLatestVersionQuery, {});
```
n1ru4l, line 118: This change can be removed with the updated function signature.

packages/libraries/cli/src/commands/whoami.ts
```diff
@@ -80,7 +80,7 @@ export default class WhoAmI extends Command {
     });
 
     const result = await this.registryApi(registry, token)
-      .request(myTokenInfoQuery)
+      .request(myTokenInfoQuery, {})
```
n1ru4l, line 83: This change can be removed with the updated function signature.

n1ru4l (changes requested):

n1ru4l:
@dotansimha @kamilkisiela The code uses [`@whatwg-node/fetch`](https://www.npmjs.com/package/@whatwg-node/fetch). This will not be a breaking change

packages/libraries/cli/src/base-command.ts
```diff
@@ -142,17 +145,41 @@ export default abstract class extends Command {
...
-        'graphql-client-version': this.config.version,
-      },
-    });
+    const requestHeaders = {
+      Accept: 'application/json',
+      'User-Agent': `hive-cli/${this.config.version}`,
+      Authorization: `Bearer ${token}`,
+      'graphql-client-name': 'Hive CLI',
+      'graphql-client-version': this.config.version,
+    };
```
n1ru4l, line 154: Integration tests are failing with status code 415, because `content-type: application/json` header is missing here.

packages/libraries/cli/src/base-command.ts
dotansimha: much better. thanks

dotansimha:
> @dotansimha @kamilkisiela The code uses [`@whatwg-node/fetch`](https://www.npmjs.com/package/@whatwg-node/fetch). This will not be a breaking change

Hmm, I think we are still using the global `Headers` (which has `.get` function). So this might work only in Node 18?

n1ru4l (approved):

PR #1723, opened 2023-03-14
Fetsival of joy

kamilkisiela:


packages/services/rate-limit/src/api.ts
```diff
@@ -101,10 +101,12 @@ export const rateLimitApiRouter = createRouter<Context>({
         200: {
           type: 'number',
         },
+        //^ in tRPC, every response is 200 by default. Only if it throws, it's 500 and connection errors or parsing errors are handled by the library through tRPCError class
       },
     } as const,
     async handler(request, ctx) {
       const jsonBody = await request.json();
+      //                     ^ does it use runtime validation?
```
ardatan, line 109: Yes 👍 it throws 400 with a specific validation error so client throws.

PR #1746, opened 2023-03-16
Use FETS in cdn worker instead of itty-router

ardatan:
image


packages/services/cdn-worker/src/artifact-handler.ts
```diff
@@ -1,27 +1,24 @@
...
     { type: 'notModified' } | { type: 'notFound' } | { type: 'redirect'; location: string }
   >;
   isKeyValid: KeyValidator;
   analytics?: Analytics;
   fallback?: (
-    request: Request,
+    request: Request | RouterRequest,
     params: { targetId: string; artifactType: string },
-  ) => Promise<Response | undefined>;
+  ) => Promise<any>;
```
n1ru4l, line 21: why any?

packages/services/cdn-worker/src/artifact-handler.ts
```diff
@@ -1,27 +1,24 @@
...
 import type { KeyValidator } from './key-validation';
 
-const { Response } = createFetch({ useNodeFetch: true });
-
 type ArtifactRequestHandler = {
   getArtifactAction: (
     targetId: string,
     artifactType: ArtifactsType,
-    eTag: string | null,
+    eTag?: string | null,
```
n1ru4l, line 12:
```suggestion
    eTag: string | null,
```

Please don't alter my code for no reason 😁

ardatan: It was an undefined, null conflict. Fixing it in FETS.

packages/services/cdn-worker/src/artifact-handler.ts
ardatan: Somehow default `Response` and `TypedResponse` were conflicting. Adding a workaround to make it pass for now.

n1ru4l:
closing this for now because of staleness and will re-consider it later on.

PR #1884, opened 2023-03-29
FeatureFlag: use latest composable version for comparisons

kamilkisiela:
The core difference between the default behavior and when then flag is enabled is the `diff` check.
By default, Hive fetches the schemas of the `latest` version.
When the flag is enabled, it uses the schemas of the `latest-and-composable` version.
The `composition` check stays untouched, it still uses `latest`.

When the flag is enabled, Hive accepts the published schema, even if it has GraphQL errors (unknown type etc).
The reason for it is to unblock users when they try to publish incomplete schemas (Federation can give really weird errors when some composition requirements are not met).

The outcome of that change is that the `diff` check runs only for composable versions, meaning Hive is able to detect breaking changes only when the version is `composable`.


integration-tests/testkit/seed.ts
```diff
@@ -59,6 +72,16 @@ export function initSeed() {
 
           return {
             organization,
+            async setFeatureFlag(name: string, enabled: boolean) {
```
kamilkisiela, line 75: I could do it over GraphQL API but we don't need to set features flags over a GraphQL request or any other form of remote calls. Once we have a need for a more complex admin panel, we could implement it, but until then, the feature flag might be no longer available.

packages/services/api/src/modules/schema/providers/models/composite-legacy.ts
```diff
@@ -367,108 +358,4 @@ export class CompositeLegacyModel {
       reasons,
     };
   }
-
-  async delete({
```
kamilkisiela, line 371: The `delete` feature was available for the legacy models for some reason, and it shouldn't, so I removed it.

PR #1948, opened 2023-04-05
Display service url changes in UI when comparing versions

kamilkisiela:
### Background

The History page receives no changes and no composition errors when comparing versions and it displays "it's your first composable versions" message. It shouldn't do that when a service url changes.

### DIFF
--- a/integration-tests/tests/cli/schema.spec.ts
+++ b/integration-tests/tests/cli/schema.spec.ts
@@ -170,7 +170,7 @@ describe.each`
     },
   );
 
-  test.only
+  test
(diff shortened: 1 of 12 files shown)


integration-tests/tests/cli/schema.spec.ts
```diff
@@ -167,4 +169,74 @@ describe.each`
       await expect(output).rejects.toThrowError('Invalid token provided');
     },
   );
+
+  test.only
```
kamilkisiela, line 173:
```suggestion
  test
```

PR #2019, opened 2023-04-13
fix: logging issues with router plugin

YassinEldeeb:


packages/libraries/router/src/logger.rs
```diff
@@ -0,0 +1,57 @@
...
+type LogLevel = &'static str;
+
+impl Logger {
+    fn find_log_level_score(log_level: &str) -> u8 {
+        LOG_LEVEL_SCORES
+            .iter()
+            .find(|(l, _)| *l == log_level)
+            .map(|(_, score)| *score)
+            .expect("Invalid log level")
+    }
```
YassinEldeeb, line 26: I guess this could also be as you've mentioned before, just an Array with those items and we depend on the index instead of the tuples with scores

n1ru4l:
@YassinEldeeb @kamilkisiela Is this PR still relevant? Can I close it?

YassinEldeeb:
@n1ru4l you can close it, Kamil shipped a fix.

PR #2051, opened 2023-04-18
Hive Cloud: drop trial for Pro plan

dotansimha:
Since Hive Free plan provides access to all features, there is no real need for a trial period on the Pro plan. 
Users who wish to experiment and test Hive, can use the Free plan, and upgrade to Pro if they need more data to be stored.


dotansimha:
Decided to keep it for now. No need to drop.

PR #2056, opened 2023-04-18
minor ui changes

beerose:
Couldn't help myself and made two small changes:

1. Added a background for the current active organization — when I was using it it was confusing and I never knew which one I'm currently on (I know there's this info on the left side, but it's small and kinda makes you look at two sides of the screen).

The color is up to discussion — didn't want it to stand out too much.
CleanShot 2023-04-18 at 17 39 05@2x

2. Removed left margin from the avatar because the arrow was quite far from the avatar and I kept assuming that these are two separate buttons.

CleanShot 2023-04-18 at 17 40 18@2x


beerose:
Does it need a changeset?

beerose:
I suppose the CI fails because of missing keys in the fork. Will close this PR and open it in the original repo

PR #2078, opened 2023-04-20
Rework mutex using redlock for redis

enisdenjo:


kamilkisiela (approved):

kamilkisiela:
Maybe let's add an integration test where:
1. Two `schema:publish` requests are called at similar time (with small delay to make sure one always runs first) to the same target, with different schemas.
2. Two `schema:publish` requests are called at similar time (with small delay, as described above) but this time to two different targets.

(1) will tell us that the lock does not allow to two parallel inserts.
(2) will tell us that one lock is not blocking other locks (across targets).

packages/services/api/src/modules/shared/providers/mutex.ts
```diff
@@ -0,0 +1,108 @@
...
+          retryCount: retries,
+          retryDelay,
+        });
+        if (signal.aborted) {
+          lock.release().catch(err => {
+            // it is safe to not throw the error, as the lock will
+            // automatically expire after its duration is exceeded
+            // TODO: should this be logged as an error? a release may fail if there
+            //       is no lock to release, like when the duration gets exceeded
+            this.logger.warn('Lock release problem after aborted (id=%s, err=%s)', id, err);
```
n1ru4l, line 77: I think we should lock in case a lock times out

packages/services/api/src/modules/shared/providers/mutex.ts
```diff
@@ -0,0 +1,108 @@
...
+          throw new Error('Locking aborted');
+        }
+        this.logger.debug('Lock acquired (id=%s)', id);
+        return async () => {
+          this.logger.debug('Releasing lock (id=%s)', id);
+          await lock.release().catch(err => {
+            // it is safe to not throw the error, as the lock will
+            // automatically expire after its duration is exceeded
+            // TODO: should this be logged as an error? a release may fail if there
+            //       is no lock to release, like when the duration gets exceeded
```
n1ru4l, line 88: we can still remove the logging afterward. For now I think we should keep this

kamilkisiela (approved):

PR #2087, opened 2023-04-21
Use simple redis cache instead of tasks over redis

kamilkisiela:
Because of the changes done in #2078 (redlock) we can simplify the caching logic and remove IdempotentRunner. Instead of caching a task on top of the lock, we move the lock to be on top of the cache. This way we can use a simple read and write cache instead of a dealing with the status of the task. There's a chance it has a performance impact but it's a bit hard to tell without doing a proper stress test.

It affects `schema:delete` action. We no longer cache the task (I don't think there's a reason to cache it, the redlock is enough to make sure there is only one `schema:delete` action happening).


enisdenjo (approved):
I think this is just fine by the looks of it, reasoning makes sense. Lets try it out.

PR #2129, opened 2023-04-26
ci: migrate-to-pnpm-v8

YassinEldeeb:


dotansimha:
I noticed that with pnpm v8, we have some warnings now during installation: 
```
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/client/dist
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/core/dist
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/core/dist
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/core/dist
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/core/dist
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/external-composition/dist
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/external-composition/dist
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/client/dist
 WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/core/dist
```

Also, please note that in some workflows we are not using the shared-pipelines (but a similar one, that lives in this repo). So it needs to adjustments as well.

YassinEldeeb:
> I noticed that with pnpm v8, we have some warnings now during installation:
> ```
>  WARN  Local dependency not found at /home/runner/work/graphql-hive/graphql-hive/packages/libraries/client/dist
> ```

@dotansimha hmm, but that looks ok, the local dependency's files were not found as those packages are specifying their `main` entry to be `dist/index.js` and `dist` doesn't exist unless after doing a successful build.

I have noticed that in the root `tsconfig.json`, there is a configured `paths` like this `"@graphql-hive/client": ["./packages/libraries/client/src/index.ts"],` which means that in development we reference the local dependencies as TS files to avoid having to re-build multiple projects continuously, so I think this warning doesn't have a lot of meaning in development, but just a warning that if we didn't have that tsconfig `paths` that we would have trouble using the local packages without them being built.

YassinEldeeb:
>Our pnpm setup here also have some configuration overrides (see .npmrc) - so we need to make sure pnpm v8 does not change any defaults that might effect this.

I have verified that both `.npmrc` configs work locally and are still supported in pnpm@v8 from the [release notes](https://github.com/pnpm/pnpm/releases/tag/v8.0.0), nothing is listed regarding deprecating any of the listed below:
```
enable-pre-post-scripts=true
shamefully-hoist=true
```

dotansimha (approved):

PR #2173, opened 2023-05-09
Better prettier plugin for sorting imports

enisdenjo:
https://github.com/the-guild-org/shared-config/pull/228


packages/services/api/src/modules/alerts/providers/alerts-manager.ts
```diff
@@ -11,7 +12,6 @@ import { ProjectManager } from '../../project/providers/project-manager';
 import { Logger } from '../../shared/providers/logger';
 import type { ProjectSelector } from '../../shared/providers/storage';
 import { Storage } from '../../shared/providers/storage';
-import type { AlertsModule } from '../__generated__/types';
```
B2o5T, line 14: now `import type` always on top? not sure why it was moved

packages/services/api/src/modules/rate-limit/resolvers.ts
```diff
@@ -1,5 +1,5 @@
-import { Logger } from './../shared/providers/logger';
 import { RateLimitModule } from './__generated__/types';
+import { Logger } from './../shared/providers/logger';
 import { RateLimitProvider } from './providers/rate-limit.provider';
```
B2o5T, line 3:
```suggestion
import { RateLimitModule } from './__generated__/types';
import { Logger } from './../shared/providers/logger';
import { RateLimitProvider } from './providers/rate-limit.provider';
```

I guess here we need apply ESLint rule to suggest do not use `./` in path :)

packages/services/api/src/modules/alerts/providers/alerts-manager.ts
enisdenjo: They're not always on top ([see `ProjectSelector` in same file](https://github.com/kamilkisiela/graphql-hive/pull/2173/files/acca56a7ecb3243291dba08b6624c8711c93a0c8#diff-230a1e435104ca30e66815972d61fb01469757b836f03d3cdb7495c09836c5d6L12))

I think the sort is just better now haha.

packages/services/api/src/modules/rate-limit/resolvers.ts
enisdenjo: I agree!

enisdenjo: Lovely!!

PR #2235, opened 2023-05-15
feat: supergraph field information in explorer

n1ru4l:
### Background

On the Explorer page, we want to show users from which subgraph a specific schema coordinate (field/type/etc.) originates.
We can extract this data from the annotated supergraph SDL.

https://www.notion.so/theguildoss/Display-the-subgraph-from-which-the-type-is-originated-0734755cb4cf4a7fb298f923da6140ea?pvs=4

### Description

- Refactors the insert/decoding logic for schema versions
- Store supergraph SDL in the schema_versions database table
- Attempt to construct meta information based on the supergraph SDL and expose it via the GraphQL API if available
image

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/storage/src/index.ts
+++ b/packages/services/storage/src/index.ts
@@ -1597,6 +1597,7 @@ export async function createStorage(connection: string, maximumPoolSize: number)
             sv.has_persisted_schema_changes,
             sv.previous_schema_version_id,
             sv.composite_schema_sdl,
+            sv.supergraph_sdl,
             sv.schema_composition_errors
           FROM public.schema_versions as sv
           WHERE sv.target_id = ${target} AND sv.is_composable IS TRUE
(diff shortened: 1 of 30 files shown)


packages/services/storage/src/index.ts
```diff
@@ -1849,9 +1837,8 @@ export async function createStorage(connection: string, maximumPoolSize: number)
           sv.has_persisted_schema_changes,
           sv.previous_schema_version_id,
           sv.composite_schema_sdl,
-          sv.schema_composition_errors,
-          sl.author,
-          lower(sl.service_name) as "service_name"
```
kamilkisiela, line 1854: @n1ru4l are you sure we don't need to expose `service_name`?

kamilkisiela (approved):

packages/services/storage/src/index.ts
n1ru4l: yeah otherwise we would have a typescript/runtime error

dotansimha (approved):

PR #2270, opened 2023-05-19
Add hash to sort key of clients_daily view

kamilkisiela:
Adds a `hash` column to the `ORDER BY` of `clients_daily`. Fixes an issue with partial data in client exclusion logic.

### DIFF
--- a/deployment/services/db-migrations.ts
+++ b/deployment/services/db-migrations.ts
@@ -60,6 +60,7 @@ export function deployDbMigrations({
         CLICKHOUSE_USERNAME: clickhouse.config.username,
         CLICKHOUSE_PASSWORD: clickhouse.config.password,
         CLICKHOUSE_PROTOCOL: clickhouse.config.protocol,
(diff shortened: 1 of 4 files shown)


deployment/services/db-migrations.ts
```diff
@@ -60,6 +60,7 @@ export function deployDbMigrations({
         CLICKHOUSE_USERNAME: clickhouse.config.username,
         CLICKHOUSE_PASSWORD: clickhouse.config.password,
         CLICKHOUSE_PROTOCOL: clickhouse.config.protocol,
+        CLICKHOUSE_MIGRATOR_GRAPHQL_HIVE_CLOUD: '1',
```
dotansimha, line 63: this is something that self-hosted users will need to have as well? do we want to have some strategy to support self-hosted as well?

kamilkisiela: What do you mean? The `CLICKHOUSE_MIGRATOR_GRAPHQL_HIVE_CLOUD` is to tell the migration script that it runs within GraphQL Hive. Self-hosters have it `false` by default.

PR #2299, opened 2023-05-23
feat: make sure unexpected serialization errors are reported

n1ru4l:
### Background

See https://github.com/n1ru4l/envelop/issues/1808

### Description

This maps GraphQLErrors raised by scalars `serialize` function/method to normal Errors in order to get them masked and reported via Sentry.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/server/src/use-unexpected-serialization-error.ts
+++ b/packages/services/server/src/use-unexpected-serialization-error.ts
@@ -1,4 +1,3 @@
-import { GraphQLScalarType } from 'graphql';
 import { type Plugin } from 'graphql-yoga';
 import { isGraphQLError } from '@envelop/core';
 import { MapperKind, mapSchema } from '@graphql-tools/utils';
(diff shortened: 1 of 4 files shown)


kamilkisiela (approved):

dotansimha (approved):

packages/services/server/src/use-unexpected-serialization-error.ts
```diff
@@ -0,0 +1,44 @@
...
+          const config = type.toConfig();
+          return new GraphQLScalarType({
+            ...config,
+            serialize(value) {
+              try {
+                return config.serialize(value);
+              } catch (err) {
+                if (isGraphQLError(err)) {
+                  throw new Error(err.message);
+                }
```
enisdenjo, line 30: This makes all scalars plain Errors, are they then reported to the user? Like, what if its the users fault having provided a malformed value?

enisdenjo (commented):
.

packages/services/server/src/use-unexpected-serialization-error.ts
n1ru4l: The `serialize` function is only used for resolver values. These do not really matter to the user and could even be a security risk. The malformed values are handled by the `parse` function,

enisdenjo (approved):

PR #2409, opened 2023-06-07
fix: avoid leaking internal (unexpected) errors to the outside world

n1ru4l:
### Background

Related https://github.com/kamilkisiela/graphql-hive/issues/333

### Description

After talking with @dotansimha, we came to the conclusion that these checks can potentially expose sensitive messages. It makes more sense to drop the try-catch statement and let those kind of errors escalate.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/schema/providers/registry-checks.ts
+++ b/packages/services/api/src/modules/schema/providers/registry-checks.ts
@@ -20,7 +20,7 @@ import type {
 import { Logger } from './../../shared/providers/logger';
 import { Inspector } from './inspector';
 import { SchemaCheckWarning } from './models/shared';
-import { ensureSDL, extendWithBase, isCompositeSchema, SchemaHelper } from './schema-helper';
+import { extendWithBase, isCompositeSchema, SchemaHelper } from './schema-helper';
 
 // The reason why I'm using `result` and `reason` instead of just `data` for both:
 // https://bit.ly/hive-check-result-data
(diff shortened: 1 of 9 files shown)


packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -139,7 +133,6 @@ export class RegistryChecks {
           errorsBySource: {
             graphql: validationErrors.filter(isGraphQLValidationError),
             composition: validationErrors.filter(isCompositionValidationError),
-            policy: validationErrors.filter(isPolicyValidationError),
```
n1ru4l, line 142: Compose and validate can never yield policy errors, it does not make sense to add them here.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -184,46 +177,31 @@ export class RegistryChecks {
...
+        result: {
+          warnings,
         },
       } satisfies CheckResult;
     }
+
+    return {
+      status: 'failed',
+      reason: {
+        errors: policyResult.errors.map(toSchemaCheckWarning),
```
n1ru4l, line 201: We now transform both the warnings and errors in order to be more consistent. This is helpful for doing https://github.com/kamilkisiela/graphql-hive/issues/333

dotansimha (approved):

kamilkisiela (approved):

PR #2430, opened 2023-06-13
Use lighter solution to get commit sha and author from git

### LINKED ISSUE #2429: CLI: schema:check and schema:publish slow when git repo is huge
**Problem**

Hi, I have a repo with 331488 commits and it took more than 60 seconds for hive cli to parse those commits.

**Solution ?**

It would be great to either skip reading git repo commits at all if author and commit are provided or as another solution, read only latest commit from git repo (but AFAIN https://github.com/wayfair/git-parse does not support reading only latest commit)

kamilkisiela:
Closes #2429

### DIFF
--- a/packages/libraries/cli/src/helpers/git.ts
+++ b/packages/libraries/cli/src/helpers/git.ts
@@ -1,36 +1,48 @@
-import { existsSync, readFileSync } from 'fs';
-import { join } from 'path';
+import { exec } from 'child_process';
+import { readFileSync } from 'fs';
 import ci from 'env-ci';
-import { gitToJs } from 'git-parse';
-
-function splitPath(path: string) {
-  const parts = path.split(/(\/|\\)/);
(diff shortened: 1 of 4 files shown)


packages/libraries/cli/src/helpers/git.ts
```diff
@@ -1,36 +1,48 @@
...
+function getLatestCommitFromGit() {
+  return new Promise<{
+    hash: string;
+    author: string;
+  } | null>(resolve => {
+    exec(latestCommitCommand, { cwd: process.cwd() }, (_, stdout) => {
+      if (stdout.includes(splitBy)) {
+        const [hash, authorName, authorEmail] = stdout.split(splitBy);
+        console.log({ hash, authorName, authorEmail });
+
```
kamilkisiela, line 27:
```suggestion

```

PR #2439, opened 2023-06-15
feat: useYogaHive

n1ru4l:
### Background

When using the yoga response cache plugin, hive metrics wont be reported with the `useHive` envelop plugin, as the execution phase is skipped.

### Description

This Yoga plugin that acts in the `onParams`  and `onResultProcess` hooks, which are always invoked. This leads to the usage data being always reported.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/libraries/client/src/yoga.ts
+++ b/packages/libraries/client/src/yoga.ts
@@ -1,0 +1,61 @@
+import { DocumentNode, GraphQLSchema, parse } from 'graphql';
+import type { Plugin } from 'graphql-yoga';
+import LRU from 'tiny-lru';
+import { createHive } from './client.js';
+import type { CollectUsageCallback, HiveClient, HivePluginOptions } from './internal/types.js';
+import { isHiveClient } from './internal/utils.js';
+
+export function useHive(clientOrOptions: HiveClient): Plugin;
+export function useHive(clientOrOptions: HivePluginOptions): Plugin;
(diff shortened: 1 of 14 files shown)


packages/libraries/client/src/yoga.ts
```diff
@@ -0,0 +1,61 @@
...
+        ...clientOrOptions,
+        agent: {
+          name: 'hive-client-yoga',
+          ...clientOrOptions.agent,
+        },
+      });
+
+  void hive.info();
+
+  const lru = LRU<DocumentNode>(10_000);
```
n1ru4l, line 23: Not a fan of hardcoding this here. If we could somehow share the parser cache from yoga that would be cool. Any idea on whether that is possible @ardatan?

packages/libraries/client/src/yoga.ts
```diff
@@ -0,0 +1,61 @@
...
+          // ignore
+        }
+      }
+    },
+    onResultProcess(context) {
+      const callback = cache.get(context.request);
+      if (callback) {
+        // we don't support batching :)
+        if (Array.isArray(context.result) === false) {
+          callback(context.result as any);
```
n1ru4l, line 56: catsing to any as the Yoga type is too complicated :D

PR #2546, opened 2023-06-28
Memory Leak Fix

nopor:
Fixing a memory leak.

### Background

The close event gets set every time a request is send via the same socket and fixing the following warning:

```
(node:1) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [Socket]. Use emitter.setMaxListeners() to increase limit
```

### Description

The event will only be set once, if the abortController doesn't exist yet.

### Checklist

- [x] Memory management


n1ru4l:
Hey @nopor, can you share a reproduction that causes the following warning?

```
(node:1) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [Socket]. Use emitter.setMaxListeners() to increase limit
```

nopor:
Hi @n1ru4l,

This warning pops up, when a connection/socket is being reused and hasn't been closed before.
We are using this in our graphql gateway.
If you close the socket/end the connection after every request, this is obviously no issue.

n1ru4l (approved):

n1ru4l:
@nopor can you please format the code with prettier?

n1ru4l:
Also just out of curiosity, can you share why your gateway sends requests to the Hive GraphQL API?

packages/services/server/src/graphql-handler.ts
```diff
@@ -247,15 +249,22 @@ export const graphqlHandler = (options: GraphQLHandlerOptions): RouteHandlerMeth
...
+    let controller = abortControllerCache.get(req.socket);
 
     // we use the socket.close over req.close because req.close is emitted
     // when the request gets processed (not canceled)
     // see more: https://github.com/nodejs/node/issues/38924
     // TODO: socket.once might break for http/2 because
-    req.raw.socket.once('close', () => {
-      controller.abort();
-    });
+    if(!controller) {
```
kamilkisiela, line 258:
```suggestion
    if (!controller) {
```

packages/services/server/src/graphql-handler.ts
```diff
@@ -247,15 +249,22 @@ export const graphqlHandler = (options: GraphQLHandlerOptions): RouteHandlerMeth
...
+      controller = new AbortController();
+      abortControllerCache.set(req.socket, controller);
+
+      req.raw.socket.once('close', () => {
+        controller.abort();
+        abortControllerCache.delete(req.socket);
+      });
+    }
+
```
kamilkisiela, line 267:
```suggestion
```

nopor:
Will do later - applied the fix on the web, sorry.

The gateway is re-using that function, it's not sending the data to/through hive.

PR #2555, opened 2023-06-30
Add ca certificates to base docker images

andriihrachov:
### Background

When deploying hive with using Hashicorp Vault init container, the container fails with error

```
time="2022-10-19T17:58:45Z" level=error msg="failed to request new Vault token" app=vault-env err="Put \"********/login\": x509: certificate signed by unknown authority"
```

### Description

Add CA-certificates to base images

### DIFF
--- a/docker/app.dockerfile
+++ b/docker/app.dockerfile
@@ -1,5 +1,7 @@
 FROM node:18.16.1-slim
 
+RUN apt-get update && apt-get install -y ca-certificates
+
 ENV NODE_ENV production
 
 WORKDIR /usr/src/app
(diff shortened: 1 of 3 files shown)


n1ru4l:
@andriihrachov, please explain why we need to add this within the Dockerfile in this repository. We want to keep the images simple and small. You could extend the published Docker Images for your use case instead, like the following.

```dockerfile
FROM ghcr.io/kamilkisiela/graphql-hive/server:latest

RUN apt-get update && apt-get install -y ca-certificates
```

We are happy to consider adding this to the images we provide given it makes sense. However, it seems like this is very specific to the use-case and set up you are using.

PR #2559, opened 2023-07-03
Display a helpful error message when commit not found (gh schema:check)

kamilkisiela:


packages/services/api/src/modules/integrations/providers/github-integration-manager.ts
```diff
@@ -166,7 +168,7 @@ export class GitHubIntegrationManager {
     });
 
     if (!installationId) {
-      throw new Error(
+      throw new HiveError(
```
kamilkisiela, line 171: Makes the error visible to the user

packages/services/api/src/modules/integrations/providers/github-integration-manager.ts
```diff
@@ -177,18 +179,42 @@ export class GitHubIntegrationManager {
...
+          'GitHub error details (message=%s, status=%s)',
+          error.message,
+          error.status,
+        );
+
+        if (error.message.includes('No commit found for SHA:')) {
+          throw new HiveError(
+            `Commit ${input.sha} not found in repository ${input.repositoryOwner}/${input.repositoryName}`,
+          );
+        }
```
kamilkisiela, line 210: Instead of throwing `No commit found for SHA: 123456`, we add more details (useful to detect a wrong repo connected to a project)

n1ru4l (approved):

PR #2593, opened 2023-07-12
Allow to leave an organization

kamilkisiela:
Allow user to leave an organization only if:
- not an owner
- not last member

### DIFF
--- a/packages/services/api/src/modules/organization/module.graphql.ts
+++ b/packages/services/api/src/modules/organization/module.graphql.ts
@@ -15,6 +15,7 @@ export default gql`
     deleteOrganization(selector: OrganizationSelectorInput!): OrganizationPayload!
     deleteOrganizationMembers(selector: OrganizationMembersSelectorInput!): OrganizationPayload!
     joinOrganization(code: String!): JoinOrganizationPayload!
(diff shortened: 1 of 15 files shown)


packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -57,6 +58,19 @@ export default gql`
     inputErrors: CreateOrganizationInputErrors!
   }
 
+  type LeaveOrganizationResult {
```
dotansimha, line 61: Annotate with `@oneOf` (just as a comment, like in other places?)

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -57,6 +58,19 @@ export default gql`
...
+    ok: LeaveOrganizationOk
+    error: LeaveOrganizationError
+  }
+
+  type LeaveOrganizationOk {
+    id: ID!
```
dotansimha, line 67: should be `organizationId`? calling it `id` might imply that this is an entity? :D

dotansimha (approved):

n1ru4l (approved):

PR #2663, opened 2023-07-25
IPv4 only

### LINKED ISSUE #2572: Issue with ipv6 notation for server.listen (EAFNOSUPPORT: address family not supported :::3069)
Having this issue in multiple of the docker images:
image

There seems to be an issue with the ipv6 way the server is started: https://github.com/kamilkisiela/graphql-hive/blob/main/packages/services/external-composition/federation-2/src/index.ts#L10

I was able to reproduce the issue with a really simple code sandbox: https://codesandbox.io/s/hopeful-robinson-4w6f5k?file=/src/index.ts 

I guess this is a general node issue. But it breaks our deployment. Any help is welcome :-) Will post any findings here.

kamilkisiela:
Drop compatibility with both IPv4 and IPv6, let's stick to v4 only.
Fixes #2572
Closes #2573


integration-tests/.env
```diff
@@ -7,7 +7,7 @@ POSTGRES_PORT=5432
 POSTGRES_DB=registry
 HIVE_ENCRYPTION_SECRET=wowverysecuremuchsecret
 HIVE_EMAIL_FROM=no-reply@graphql-hive.com
-HIVE_APP_BASE_URL=http://localhost:8080
+HIVE_APP_BASE_URL=http://0.0.0.0:8080
```
n1ru4l, line 10: This is should either be localhost or 127.0.0.1.

kamilkisiela: https://github.com/kamilkisiela/graphql-hive/pull/1317
it was `0.0.0.0` before

platzhersh:
@kamilkisiela thanks for tackling this. As a reference, here are the changes i made locally to build my own images: https://github.com/platzhersh/graphql-hive/pull/1

PR #2671, opened 2023-07-25
add optional s3_session_token variable and remove setting region to auto

tapaderster:
### Background
This fixes an issue when using graphql-hive with aws s3 as the artifact storage provider.

### Description

Adding a way to pass the optional S3_SESSION_TOKEN variable. This should not change any existing behavior.

Also removing setting the region to 'auto'  while creating the AWS client. The region should be guessed correctly via:
https://github.com/kamilkisiela/graphql-hive/blob/d6c417837d9bcd80412d1b2283244d0dbf9e3649/packages/services/cdn-worker/src/aws.ts#L413

nit: added some indentation fixes to the Testing readme file.

By the way, this is a great project!

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -18,7 +18,7 @@ Integration tests are based pre-built Docker images, so you can run it in 2 mode
 
 #### Running from Source Code
 
-**TL;DR**: Use `pnpm integration:prepare` command to setup the complete environment from locally
+**TL;DR**: Use `pnpm integration:prepare` command to set up the complete environment from locally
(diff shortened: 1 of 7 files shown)


n1ru4l (commented):
Can you please also document the environment variable in the `README.md` file?

kamilkisiela:
Thank you :)
Until we fix https://github.com/kamilkisiela/graphql-hive/issues/1436#issuecomment-1465111551, I need to merge your PRs to a local branch and run CI from there.

PR #2676, opened 2023-07-26
Improvements to error handling in Yoga

kamilkisiela:
+ extract operation name from DocumentNode if not available on the request object
+ show only the first error message from `result.errors[]` instead of using `CombinedError` class from URQL (it combines error messages and usually they are more than one and we end up with duplicates)

### DIFF
--- a/packages/services/server/src/graphql-handler.ts
+++ b/packages/services/server/src/graphql-handler.ts
@@ -5,7 +5,16 @@ import type {
   FastifyRequest,
   RouteHandlerMethod,
 } from 'fastify';
-import { DocumentNode, GraphQLError, print, ValidationContext, ValidationRule } from 'graphql';
+import {
+  DocumentNode,
+  GraphQLError,
+  Kind,
+  print,
+  ValidationContext,
+  ValidationRule,
+  type DefinitionNode,
+  type OperationDefinitionNode,
+} from 'graphql';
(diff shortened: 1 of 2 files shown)


packages/services/server/src/graphql-handler.ts
```diff
@@ -165,15 +179,21 @@ export const graphqlHandler = (options: GraphQLHandlerOptions): RouteHandlerMeth
       }),
       useSentryUser(),
       useErrorHandler(({ errors, context }): void => {
+        // Not sure what changed, but the `context` is now an object with a contextValue property.
+        // We previously relied on the `context` being the `contextValue` itself.
+        const ctx = ('contextValue' in context ? context.contextValue : context) as Context;
+
```
kamilkisiela, line 185: @ardatan any idea why? I thought `context` === `contextValue` but apparently it's not.

packages/services/server/src/graphql-handler.ts
```diff
@@ -165,15 +179,21 @@ export const graphqlHandler = (options: GraphQLHandlerOptions): RouteHandlerMeth
...
+            console.error(error.originalError);
             continue;
+          } else {
+            console.error(error);
           }
 
-          if (hasFastifyRequest(context)) {
-            context.req.log.error(error);
+          if (hasFastifyRequest(ctx)) {
+            ctx.req.log.error(error);
```
kamilkisiela, line 196: we gain here `reqId` label

PR #2679, opened 2023-07-28
Use Yargs and compile binaries with Deno

kamilkisiela:
This is a proof of concept for using Yargs with Deno. Binaries are pushed to a branch tagged with `cli-{version}`, e.g. cli-0.0.1. 
This way we can use github to host the binaries and let users to specify the version they want to use.

It should speed up docker image building (we no longer need to `npm install` stuff).

No tarball, no packing, no oclif, no S3, no weird stuff...
The only "weird" thing here is Deno but there's no need to install Deno for development, it's required only in CI/CD.


packages/libraries/cli/package.json
```diff
@@ -18,32 +19,49 @@
...
   "scripts": {
-    "build": "tsc",
-    "oclif:pack": "npm pack && pnpm oclif pack tarballs --no-xz && pnpm oclif pack win",
-    "oclif:upload": "pnpm oclif upload tarballs --no-xz && pnpm oclif upload win",
-    "postpack": "rm -f oclif.manifest.json",
+    "build": "pnpm tsx ./scripts/version.ts && bob build --single && tsx ../../../scripts/runify.ts",
+    "compile": "pnpm tsx ./scripts/node-to-deno.ts && pnpm compile:lin && pnpm compile:mac && pnpm compile:win",
+    "compile:lin": "deno compile -A deno/index.ts --no-check --output binaries/hive-linux   --target x86_64-unknown-linux-gnu",
+    "compile:mac": "deno compile -A deno/index.ts --no-check --output binaries/hive-macos   --target x86_64-apple-darwin",
+    "compile:win": "deno compile -A deno/index.ts --no-check --output binaries/hive-windows --target x86_64-pc-windows-msvc",
```
n1ru4l, line 55: what about arm?

kamilkisiela: only available arm target is for macos

n1ru4l: hmm, we already use arm linux everywhere possible, we need to make sure others that do the same can still use the cli

PR #2689, opened 2023-08-01
Fix issues with CSP errors in Lab page

dotansimha:


deployment/services/cloudflare-security.ts
```diff
@@ -51,14 +51,14 @@ export function deployCloudFlareSecurityTransform(options: {
 
   const contentSecurityPolicy = `
   default-src 'self';
-  frame-src ${stripeHost} https://game.crisp.chat https://giscus.app https://www.youtube.com;
```
n1ru4l, line 54: we use giscus on the docs 👀

dotansimha: yeah but now it's no longer on the Hive app, because the landing page and docs are part of the main Guild website

PR #2697, opened 2023-08-03
Adds introspect command

kamilkisiela:
```bash
hive introspect https://example.com/graphql
hive introspect https://example.com/graphql --header "Key:Value"
hive introspect https://example.com/graphql --write schema.graphql
hive introspect "./src/**/*.graphql"
```

### DIFF
--- a/.changeset/spotty-panthers-accept.md
+++ b/.changeset/spotty-panthers-accept.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/cli': minor
+---
+
+Adds introspect command
--- a/packages/libraries/cli/src/commands/introspect.ts
+++ b/packages/libraries/cli/src/commands/introspect.ts
@@ -1,0 +1,96 @@
+import { writeFileSync } from 'node:fs';
+import { extname, resolve } from 'node:path';
+import { buildSchema, GraphQLError, introspectionFromSchema } from 'graphql';
+import { Args, Flags } from '@oclif/core';
+import Command from '../base-command';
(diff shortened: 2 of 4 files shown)


packages/libraries/cli/src/commands/introspect.ts
```diff
@@ -0,0 +1,96 @@
...
+import { Args, Flags } from '@oclif/core';
+import Command from '../base-command';
+import { loadSchema } from '../helpers/schema';
+
+export default class Introspect extends Command {
+  static description = 'introspects a GraphQL Schema';
+  static flags = {
+    write: Flags.string({
+      aliases: ['W'],
+      description: 'Write to a file',
```
n1ru4l, line 13: This could also list the possible file extensions

n1ru4l (approved):
Approved with optional suggestion

.changeset/spotty-panthers-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-hive/cli': minor
+---
+
+Adds introspect command
```
n1ru4l, line 5: This could include an example usage 😇

kamilkisiela: it's in readme!

n1ru4l: Then link to the readme from here 😇
My point is that the changelog is a great place to give an quick example, without people having to dig docs. As I said it is optional.

Issue #2603, opened 2023-07-14
Optimize schema check storage of composite sdl and supergraph


n1ru4l:
Right now we store the composite SDL and supergraph on every record within the `schema_checks` table.

When a schema check runs that has no active changes within the schema, we store a lot of redundant data within our database. E.g. bots like renovate cause a lot of these runs. Some GraphQL schemas are quite huge.

A lot of `schema_checks` are linked to the `schema_versions` table, which also stores the composite SDL and supergraph.
In case a schema check has no active changes (read as the same checksum), we could omit to store the composite SDL and supergraph within the `schema_checks` table and instead retrieve them from the `schema_versions` table when consuming the field via the GraphQL API.

A further optimization could be to even store schemas and their checksums in a completely new table, e.g. `checksum_schemas`, that is a `checksum` to `SDL` mapping. That way we could furthermore reduce the redundancy of large SDL strings.

**Note:** If we do the latter, we also need to make sure the cleanup happens when the checksum entry is no longer referenced by any `schema_checks` entry, which would furthermore complicate the database schema.

andriihrachov:
would like to take it

n1ru4l:
Hey @andriihrachov, feel free to send a pull request! I would be happy to review it ☺️

kamilkisiela:
@andriihrachov before touching code, do you mind sharing with us when do you plan to work on it and how you want to implement it?

andriihrachov:
@kamilkisiela sorry I missed your message, but opened a PR with first steps (migration, recording schema SDL and checksum in separate table, etc,) - it's quite small to kill if I took wrong way, pls feedback :)

PR #2711, opened 2023-08-05
[WIP] Schema checksums

andriihrachov:
### Current status:

Draft.

### Background

#2603 implementation

### Description

- Do not store schema_sdls as strings in versions and schema_checks, store checksum, which acts as a primary key for normalization.

- Store schema SDLs and schema checksums in separate table
- Schema checksums writes are idempotent based on checksum (e.g.  subsequent writes will not create a new schema SDL entry)

### Checklist

- [ ] Testing


packages/services/api/src/shared/entities.ts
```diff
@@ -1,5 +1,5 @@
 import { DocumentNode, GraphQLError, parse, SourceLocation } from 'graphql';
-import { z } from 'zod';
+import { number, z } from 'zod';
```
andriihrachov, line 2: no need

andriihrachov:
example  future retrieval of normalized data

```
SELECT s.schema_checksum,s.composite_schema_checksum, s.supergraph_schema_checksum, c.schema_sdl, c2.schema_sdl, c3.schema_sdl 
FROM schema_checks s
JOIN schema_sdl_checksums c ON  s.target_id = c.target_id AND c.checksum = s.schema_checksum
LEFT OUTER JOIN schema_sdl_checksums c2 ON  s.target_id = c2.target_id AND c2.checksum = s.composite_schema_checksum
LEFT OUTER JOIN schema_sdl_checksums c3 ON  s.target_id = c3.target_id AND c3.checksum = s.supergraph_schema_checksum;

```

packages/migrations/src/actions/2023.07.24T10.10.04.schema_sdl_checksums.ts
```diff
@@ -0,0 +1,31 @@
...
+export default {
+  name: '2023.07.24T10.10.04.schema_sdl_checksums.sql',
+  run: ({ sql }) => sql`
+CREATE TABLE IF NOT EXISTS "public"."schema_sdl_checksums" (
+    "checksum" text NOT NULL,
+    "schema_sdl" text NOT NULL,
+    "target_id" uuid NOT NULL REFERENCES "targets" ("id") ON DELETE CASCADE,
+    "created_at" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    "updated_at" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    PRIMARY KEY(target_id, checksum));
```
n1ru4l, line 12: if we go for the `"schema_sdl_checksums"` path, we also need to make sure that we delete `"schema_sdl_checksums"` automatically, once they are no longer referenced by either a `schema_versions` or `schema_checks` table entry.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -300,6 +300,21 @@ export class SchemaPublisher {
...
+        schemaSDL: sdl,
+      }),
+      this.storage.createSchemaSDLChecksum({
+        targetId: target.id,
+        schemaSDL: checkResult.state?.composition.compositeSchemaSDL,
+      }),
+      this.storage.createSchemaSDLChecksum({
+        targetId: target.id,
+        schemaSDL: checkResult.state?.composition.supergraphSDL,
+      }),
```
n1ru4l, line 315: It feels like this could be a bulk insert operation instead.

```ts
const [
  schemaChecksum,
  compositeSchemaChecksum,
  supergraphSchemachecksum
] = await this.storage.createSchemaSDLChecksums([
  { },
  { },
  { },
])
```

packages/services/storage/package.json
```diff
@@ -24,13 +24,15 @@
     "slonik": "30.4.4",
     "slonik-interceptor-query-logging": "1.4.7",
     "slonik-utilities": "1.9.4",
+    "ssri": "10.0.4",
```
n1ru4l, line 27: Do we really need this additional dependency? Couldn't we just us the Node.js `crypto` module similar to this code? https://github.com/kamilkisiela/graphql-hive/blob/fc441974d3a08d31e65fb63fc1aff8a8011a85f8/packages/libraries/client/src/apollo.ts#L54

packages/services/storage/src/index.ts
```diff
@@ -2146,6 +2167,21 @@ export async function createStorage(connection: string, maximumPoolSize: number)
...
+      if (args.schemaSDL === null || args.schemaSDL === undefined) {
+        return null;
+      }
+
+      const checksum = getIntegrityHashOf(args.schemaSDL);
+      const result = await pool.one<schema_sdl_checksums>(sql`
+          INSERT INTO schema_sdl_checksums ("schema_sdl", "checksum", "target_id")
+          VALUES (${args.schemaSDL}, ${checksum}, ${args.targetId}) ON CONFLICT(checksum, target_id)
+          DO UPDATE SET updated_at = NOW()
+          RETURNING *;
```
n1ru4l, line 2180: We prefer listing the fields for these return statements instead of returning `*`.

You can do something like

```SQL
RETURNING
  "checksum",
  to_json("created_at") as "createdAt",
  to_json("updated_at") as "updatedAt",
  "target_id" as "targetId",
  "schema_sdl" as "schemaSdl"
```

n1ru4l (changes requested):
Thank you for this PR, I added a few comments!

packages/services/storage/package.json
andriihrachov: yeah, no prob with that :)

n1ru4l:
I just had another chat with @kamilkisiela and we came to the conclusion that we should probably avoid having that extra table for now.

for 99,99% of schema checks (that do not alter the graphql schema), we would already solve the storage issue if we do not store the SDL on the schema checks in case it hasn't changed (aka is still the same as in the referenced `schema_version`) record.

And for the other 0,01% of schema checks that actually alter the schema, we will delete them once the expired date is reached.

So the temporary increase and storage of redundant data is fine.

@andriihrachov Would you be fine with doing that instead? It is what I already described within https://github.com/kamilkisiela/graphql-hive/issues/2603

andriihrachov:
tx, will go over them. I will proceed also with adding schemaversion -> schemachecksum relation

andriihrachov:
> I just had another chat with @kamilkisiela and we came to the conclusion that we should probably avoid having that extra table for now.
> 
> for 99,99% of schema checks (that do not alter the graphql schema), we would already solve the storage issue if we do not store the SDL on the schema checks in case it hasn't changed (aka is still the same as in the referenced `schema_version`) record.
> 
> And for the other 0,01% of schema checks that actually alter the schema, we will delete them once the expired date is reached.
> 
> So the temporary increase and storage of redundant data is fine.
> 
> @andriihrachov Would you be fine with doing that instead? It is what I already described within #2603

yeah, it will be easier

n1ru4l:
@andriihrachov the relation is not necessary, if we avoid this optimization.

andriihrachov:
closing in favor of #2731

PR #2712, opened 2023-08-07
ClickHouse storage size and performance improvements

### LINKED ISSUE #1129: Possible improvement of ClickHouse Cloud performance
ClickHouse Cloud stores data in AWS S3. In order to do the computation, it needs to fetch the objects.
I'm not sure about this, but my guess is that a partition equals an object. If that's true or partially true it means ClickHouse has to fetch multiple objects (the greater the number of partitions involved in the query, the more objects needs to be fetched).
When these objects are fetched, they are cached and the computation part happens.

What does it mean?

If the objects are in memory, the ClickHouse instance skips the fetching part and does the computation. This translates to a much faster response time.

A possible improvement would be to reduce the number of partitions, but on the other side, it will have to be invalidated more often (as the new data is inserted).

kamilkisiela:
Some big and exciting changes :)

`operation_collection` table:
- use `CODEC(T64, ZSTD(1))` on `total` column
- use `CODEC(DoubleDelta, LZ4)` on `timestamp` and `expires_at` columns
- partition monthly instead of daily
- add TTL of 3 hours

`operations` table:
- use `CODEC(DoubleDelta, LZ4)` on `timestamp` and `expires_at` columns
- use `CODEC(T64, ZSTD(1))` on `errors` and `duration` columns
- add TTL of 3 hours

`operations_daily` view:
- use `CODEC(DoubleDelta, LZ4)` on `timestamp` and `expires_at` columns
- use `CODEC(T64, ZSTD(1))` on `total` and `total_ok` columns
- partition monthly, instead of daily
- add `client_version` column
- add `client_version` to `ORDER BY` (we will now able to show stats per client_version)

`operations_hourly` view:
- use `CODEC(DoubleDelta, LZ4)` on `timestamp` and `expires_at` columns
- use `CODEC(T64, ZSTD(1))` on `total` and `total_ok` columns
- do not partition
- add TTL of 30 days (we can't show only 30 data points on the operations view, meaning we need to fetch data from hourly aggregation view to serve more than 30 data points, instead of doing it from daily)
- add `client_version` column
- add `client_version` to `ORDER BY` (we will now able to show stats per client_version)

**NEW** `operations_minutely` view:
- same as `hourly` aggregation view but with TTL set to 24 hours
- by introducing this table, we no longer need to read data from `operations`
- it should bring significant read performance improvements but also for storage size as we store only 1 data point (kind of) per every 60 seconds.

`coordinates_daily` view:
- use `CODEC(DoubleDelta, LZ4)` on `timestamp` and `expires_at` columns
- use `CODEC(T64, ZSTD(1))` on `total` column
- add TTL of `expires_at` (full data retention period)
- partition by month, not by day

`clients_daily` view:
- use `CODEC(DoubleDelta, LZ4)` on `timestamp` and `expires_at` columns
- use `CODEC(T64, ZSTD(1))` on `total` column
- partition by month, not by day

**NEW** `operation_collection_body` view:
- stores `body` per each `hash` and `target` combination
- thanks to this view, we no longer need to read data from `operation_collection`
- because it's `ReplacingMergeTree` and partitioned by month, the read performance should improve by a lot
- partitioned by month
- TTL of `expires_at` (full data retention period)
- should significantly reduce the storage size as rows are deduplicated

**NEW** `operation_collection_details` view:
- stores `name`, `hash`, `operation_kind` per `hash` and `target` combination
- thanks to this view, we no longer need to read data from `operation_collection`
- because it's `ReplacingMergeTree` and partitioned by month, the read performance should improve by a lot
- partitioned by month
- TTL of `expires_at` (full data retention period)
- should significantly reduce the storage size as rows are deduplicated

The effect of those changes is quite significant. The storage size should drop to around 3% of what we had before.

## Migration

In short, both `operation_collection` and `operations` tables require a complete reinsert of rows.
The reason for it is that the new db schema adds TTL to every table, modifies the sort key and adds three new materialized views.

### Gotchas!

**`CREATE TABLE` statement can’t include TTL definition**

The `operation_collection` and `operations` tables have a Time-To-Live applied to all rows. The TTL is only 3 hours.
The materialized views have shorter or longer (depending on the view) TTLs than their source tables.

When data older than 3 hours (TTL) is inserted into the source tables, the row is dropped and not propagated to associated materialized views.

In order to apply TTL, we need to alter the table after full data migration is completed.

**The aggregation state remains unchanged when rows are deleted from a materialized view.**

Once inserted, there’s no way back.

### Migration steps

https://github.com/kamilkisiela/graphql-hive/pull/2712#issuecomment-1693101109

Closes #1129


kamilkisiela:
TODOS
- check if Buffer can be read read in parallel by two functions
- double check if `expires_at` has always the latest value when it's not part of `ORDER BY` (SummingMergeTree and ReplacingMergeTree)

kamilkisiela:
# Migration

There are two scenarios.

## a) My `operations` table has LESS than 50M rows

The migration script will do all the migration steps automatically. The only downside is a tiny time window (a few seconds - depending on the amount of rows in `operations` table) when some usage reports might be lost.

## b) My `operations` table has MORE than 50M rows

Migration happens in 3 phases.

### 1. Deploy **[hash-here]** 

Deploy it with `MIGRATION_V2_INGEST_AFTER_UTC` environment variable applied to the `usage-ingestor` service. 
The variable accepts a future date in `YYYY-MM-DD` format (**it cannot be today's date**).
The `usage-ingestor` service will selectively write only those operations that happened after `YYYY-MM-DD 00:00:00 UTC`.
For example, if you deploy on 2023-08-10 08:50 UTC, set the date to `2023-08-11` (or any future date). This is important as the migration needs a clear point in time when data started flowing in.

### 2. Wait for the first inserted rows

Once at least an hour passed since `YYYY-MM-DD 00:00:00 UTC` (a date defined by `MIGRATION_V2_INGEST_AFTER_UTC` during deployment), you're ready to move to the next step.

For example, if `MIGRATION_V2_INGEST_AFTER_UTC` points to `2023-08-11`, wait at least until `2023-08-11 01:00:00 URC`.

### 3. Run **[link-to-migration-script]**

1. Git clone this repository
2. Run `pnpm install`
3. Go to `cd packages/migrations`
4. Create `.env` file and pass these variables with correct values:

```
CLICKHOUSE_HOST="<value>"
CLICKHOUSE_PORT="<value>"
CLICKHOUSE_USERNAME="<value>"
CLICKHOUSE_PASSWORD="<value>"
MIGRATION_V2_INGEST_AFTER_UTC="yyyy-mm-dd"
```

You don't have to create `.env` file as long as you pass those environment variables correctly to the command from the next step.

5. Run `pnpm tsx src/scripts/004-cloud.ts`.
6. And wait... It may take an hour or even a few hours. It depends on the amount of data and ingestion speed of your ClickHouse instance.
7. The migration script prints a list of commands to be executed manually. It's a manual process to avoid accidental deletion of old data and gives you time to apply time-to-live when you feel confident and ready.

This migration should reduce the storage size by 92%.

PR #2721, opened 2023-08-08
Retry calls to external composition service on 404s

kamilkisiela:


packages/libraries/external-composition/package.json
```diff
@@ -59,7 +59,7 @@
     "graphql": "^0.13.0 || ^14.0.0 || ^15.0.0 || ^16.0.0"
   },
   "devDependencies": {
-    "@apollo/composition": "2.5.1",
+    "@apollo/composition": "2.4.12",
```
n1ru4l, line 62: Why are we reverting?

PR #2731, opened 2023-08-10
Do not write schema to schema_check table

andriihrachov:
### Background

#2603 implementation

### Description

 Do not write schema/composite schema/ supergraph schema to schema_check db table if there were no changes

### Checklist

- [x] Testing


packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -305,6 +305,8 @@ export class SchemaPublisher {
     const retention = await this.rateLimit.getRetention({ targetId: target.id });
     const expiresAt = retention ? new Date(Date.now() + retention * millisecondsPerDay) : null;
 
+    const hasChemaChanges = !!checkResult?.state?.schemaChanges;
```
andriihrachov, line 308: ouch

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -384,8 +386,8 @@ export class SchemaPublisher {
         schemaPolicyWarnings: checkResult.state?.schemaPolicyWarnings ?? null,
         schemaPolicyErrors: null,
         schemaCompositionErrors: null,
-        compositeSchemaSDL: composition.compositeSchemaSDL,
-        supergraphSDL: composition.supergraphSDL,
+        compositeSchemaSDL: hasSchemaChanges ? composition.compositeSchemaSDL : '',
+        supergraphSDL: hasSchemaChanges ? composition.supergraphSDL : null,
```
n1ru4l, line 390: We want to insert `null` instead.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -372,8 +372,10 @@ export class SchemaPublisher {
         }
       }
 
+      const hasSchemaChanges = !!checkResult?.state?.schemaChanges;
+
       schemaCheck = await this.storage.createSchemaCheck({
-        schemaSDL: sdl,
+        schemaSDL: hasSchemaChanges ? sdl : '',
```
n1ru4l, line 378: We want to insert `null` instead.

n1ru4l:
What is missing here is the business logic to instead load the schema from the linked `schema_version` reocrd when serving it via the API.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
andriihrachov: just to double check. 

When I publish some version now for "single" project - I have empty "base_schema" field all the time in "schema_versions", and the field that is populated is "composite_sdl". should i use composite_sdl from "schema_versions" as a "fallback" or

packages/services/api/src/modules/schema/providers/schema-publisher.ts
andriihrachov: done

packages/services/api/src/modules/schema/providers/schema-publisher.ts
andriihrachov: done

packages/services/storage/src/index.ts
```diff
@@ -4054,6 +4058,28 @@ const schemaCheckSQLFields = sql`
   , "manual_approval_user_id" as "manualApprovalUserId"
 `;
 
+const schemaCheckSQLFieldsFallback = sql`
+  "sc"."id"
+  , to_json("sc"."created_at") as "createdAt"
+  , to_json("sc"."updated_at") as "updatedAt"
+  , coalesce("sc"."schema_sdl", "sv"."base_schema") as "schemaSDL"
```
andriihrachov, line 4065: I am not sure if this is a proper fallback TBH

n1ru4l: The fallback would be `sv.composite_schema_sdl`

n1ru4l: But it needs to be done in the business logic because a schema check has no SDL if the composition failed.

andriihrachov: added to zod object.transform  on SuccessfulSchemaCheckPartialModel

andriihrachov:
Hi @n1ru4l  can u pls look at this version?

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -384,8 +386,8 @@ export class SchemaPublisher {
         schemaPolicyWarnings: checkResult.state?.schemaPolicyWarnings ?? null,
         schemaPolicyErrors: null,
         schemaCompositionErrors: null,
-        compositeSchemaSDL: composition.compositeSchemaSDL,
-        supergraphSDL: composition.supergraphSDL,
+        compositeSchemaSDL: hasSchemaChanges ? composition.compositeSchemaSDL : null,
+        supergraphSDL: hasSchemaChanges ? composition.supergraphSDL : null,
```
n1ru4l, line 390: in case there has been no initial schema published the SDL and supergraph need to be stored as well. (schemaVersionId` is `null`).

packages/services/storage/src/index.ts
```diff
@@ -4054,6 +4058,28 @@ const schemaCheckSQLFields = sql`
...
+  , "sc"."target_id" as "targetId"
+  , "schema_version_id" as "schemaVersionId"
+  , "is_success" as "isSuccess"
+  , "sc"."schema_composition_errors" as "schemaCompositionErrors"
+  , "breaking_schema_changes" as "breakingSchemaChanges"
+  , "safe_schema_changes" as "safeSchemaChanges"
+  , "schema_policy_warnings" as "schemaPolicyWarnings"
+  , "schema_policy_errors" as "schemaPolicyErrors"
+  , coalesce("sc"."composite_schema_sdl", "sv"."composite_schema_sdl") as "compositeSchemaSDL"
+  , coalesce("sc"."supergraph_sdl", "sv"."supergraph_sdl") as "supergraphSDL"
```
n1ru4l, line 4077: In case the `"sc"."composite_schema_sdl"` and `"sc"."supergraph_sdl"` is `null`, because `sc"."schema_composition_errors"` is non-null, `"compositeSchemaSDL"` and "supergraphSDL" need to be `null`.

n1ru4l:
@andriihrachov I added two comments.

packages/services/storage/src/index.ts
andriihrachov: I think this thing is handled by zod (supergraphSchemaSDL is declared as null there on partialFailedSchemaCheckModel)

packages/services/api/src/modules/schema/providers/schema-publisher.ts
andriihrachov: can do, but does it makes sense?

n1ru4l: Yes, it makes sense because otherwise there is no compositeSchemaSDL/supergraphSDL that can be displayed. It is an edge-case that should be handled correctly.

packages/services/storage/src/index.ts
n1ru4l: I don't see why this would be handled by zod. If zod only validates, it does not transform the data.

packages/services/storage/src/schema-change-model.ts
```diff
@@ -824,39 +825,44 @@ const FailedSchemaCheckPartialModel = z.intersection(
...
+      }),
+      z.object({
+        isManuallyApproved: z.literal(false),
+        manualApprovalUserId: z.null(),
+      }),
+    ]),
+  )
+  .transform(value => ({
+    ...value,
+    schemaSDL: value.schemaSDL ?? value.compositeSchemaSDL,
```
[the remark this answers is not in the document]
andriihrachov: composite should be null in case of 

"single service", right? 

so this will be null ?? null

packages/services/api/src/modules/schema/providers/schema-publisher.ts
andriihrachov: I tested this case - empty federation project, it gives me this and no schema checks are created. 

Screenshot 2023-08-21 at 17 54 38

packages/services/storage/src/index.ts
andriihrachov: added transformation

PR #2757, opened 2023-08-21
add optional ssl options for kafka client

tapaderster:
### Background

This provides a way to configure graphql-hive with additional SSL options for local Kafka setup.

### Description

Adding a way to pass the optional SSL options for Kafka client configurations. 
This should not change any existing behavior.

PR #2758, opened 2023-08-22
add optional ssl options for kafka client (#2757)

kamilkisiela:
#2757


packages/services/usage-ingestor/src/environment.ts
```diff
@@ -1,4 +1,5 @@
-import zod from 'zod';
+import * as fs from 'fs';
+import zod, { optional } from 'zod';
```
kamilkisiela, line 2:
```suggestion
import zod from 'zod';
```

PR #2948, opened 2023-09-18
Use latest debian to match Apollo Router's glibc requirement

kamilkisiela:


docker/router.dockerfile
```diff
@@ -31,7 +31,7 @@ RUN touch ./src/main.rs
 RUN cargo build --release
 
 # Runtime
-FROM debian:bullseye-slim as runtime
+FROM debian:latest as runtime
```
dotansimha, line 34: @kamilkisiela can we use the explicit version of `latest`?

n1ru4l: yes

kamilkisiela, line 34:
```suggestion
FROM debian:12-slim as runtime
```

dotansimha (approved):

PR #2957, opened 2023-09-21
Improve strawberry export-schema snippet

patrick91:
This is the recommended approach for exporting a schema to sdl when using strawberry 😊


saihaj (approved):
Thank you!

PR #2980, opened 2023-09-29
Sort fields in Explorer by usage DESC, name ASC

### LINKED ISSUE #583: Feature Request: Sorting of Fields
As a user of GraphQL HIVE, 
that is having a **public** GraphQL API with mostly anonymous operations
I want to be able to sort the fields in the **Explorer** Tab by query count.

With a public GraphQL API, the operations tile is not really helpfull:
image

What we are interessted in is the following informations based on a time selection:
- What is the most used _Root Query / Root Mutation_?
- What is the most used Field of my Queries? Displayed like _"Me.id" 100 queries_
   - Optional splitted by the path, as Type A could be fetched over multiple different paths.
   - Example: _"Me.id" 90 queries_ & _"Organization.users.id" 10 queries_
- Which fields were not used in the selected timeframe?
- ... 

Probably this list can grow with the input of other Public API users. It's just a first idea.

---
So the easiest solution for us would be that you add sorting on the **Explorer Page**
For the future it could be interessting to have predefined lists like you have for Operations at the moment.

---

Another idea for visualization of the field usage could a combination of GraphQL Voyager and GraphQL Hive.
Bold Lines --> High Usage
Thin Lines --> Small Usage
Dotted Lines --> Zero Usage

Hovering a line could show more details.

kamilkisiela:
Closes #583

### DIFF
--- a/packages/web/app/src/components/target/explorer/common.tsx
+++ b/packages/web/app/src/components/target/explorer/common.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement, ReactNode } from 'react';
+import React, { ReactElement, ReactNode, useMemo } from 'react';
 import { clsx } from 'clsx';
 import { Tooltip } from '@/components/v2';
 import { PulseIcon, UsersIcon } from '@/components/v2/icon';
@@ -209,6 +209,7 @@ const GraphQLFields_FieldFragment = graphql(`
     isDeprecated
     deprecationReason
     usage {
+      total
       ...SchemaExplorerUsageStats_UsageFragment
     }
     args {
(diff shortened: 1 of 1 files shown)


packages/web/app/src/components/target/explorer/common.tsx
```diff
@@ -401,11 +404,16 @@ export function GraphQLFields(props: {
...
   const { totalRequests } = props;
-  const [fields, collapsed, expand] = useCollapsibleList(
-    useFragment(GraphQLFields_FieldFragment, props.fields),
-    5,
-    props.collapsed ?? false,
+  const fieldsFromFragment = useFragment(GraphQLFields_FieldFragment, props.fields);
+  const sortedFields = useMemo(
+    () =>
+      [...fieldsFromFragment].sort(
+        // Sort by usage DESC, name DESC
```
kamilkisiela, line 411:
```suggestion
        // Sort by usage DESC, name ASC
```

PR #2992, opened 2023-10-03
Use KV storage to deduplicate SDLs from schema_checks

kamilkisiela:
Closes https://github.com/kamilkisiela/graphql-hive/issues/2603

Adds `sdl_store` table with `id` (hash of sdl) and `sdl` columns.
Replaces `schema_sdl`, `supergraph_sdl`, `composite_schema_sdl` with `schema_sdl_store_id`, `supergraph_sdl_store_id` and `composite_schema_sdl_store_id`.

### DIFF
--- a/packages/migrations/package.json
+++ b/packages/migrations/package.json
@@ -14,7 +14,7 @@
     "db:init": "pnpm db:create && pnpm migration:run",
     "db:migrator": "node --experimental-specifier-resolution=node --loader ts-node/esm src/index.ts",
     "migration:run": "pnpm db:migrator up",
-    "test": "node --experimental-specifier-resolution=node --loader ts-node/esm ./test/*.test.ts"
+    "test": "TS_NODE_TRANSPILE_ONLY=true node --experimental-specifier-resolution=node --loader ts-node/esm ./test/index.ts"
   },
   "devDependencies": {
     "@graphql-hive/core": "workspace:*",
--- a/packages/services/server/src/index.ts
+++ b/packages/services/server/src/index.ts
@@ -90,26 +90,25 @@ export async function main() {
 
   const storage = await createPostgreSQLStorage(createConnectionString(env.postgres), 10);
 
-  let expiredSchemaCheckPurgeTaskRunner: null | ReturnType<typeof createTaskRunner> = null;
+  let dbPureTaskRunner: null | ReturnType<typeof createTaskRunner> = null;
 
   if (!env.hiveServices.usageEstimator) {
-    server.log.debug(
-      'Usage estimation is disabled. Skip scheduling purge tasks for expired schema checks.',
-    );
+    server.log.debug('Usage estimation is disabled. Skip scheduling purge tasks.');
   } else {
     server.log.debug(
-      `Usage estimation is enabled. Start scheduling purge tasks for expired schema checks every ${env.hiveServices.usageEstimator.dateRetentionPurgeIntervalMinutes} minutes.`,
+      `Usage estimation is enabled. Start scheduling purge tasks every ${env.hiveServices.usageEstimator.dateRetentionPurgeIntervalMinutes} minutes.`,
     );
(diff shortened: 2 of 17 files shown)


n1ru4l:
I was wondering whether the schema_sdl (if did not change), could instead be loaded from the linked `schema_version` (`schema_version_id` column)? I think that might cover 90% of duplications, and might avoid us having to run one more cron job.

kamilkisiela:
> I was wondering whether the schema_sdl (if did not change), could instead be loaded from the linked `schema_version` (`schema_version_id` column)? I think that might cover 90% of duplications, and might avoid us having to run one more cron job. Eventually, schema checks will be deleted after some time anyways.

Some supergraph SDLs can be 1MB big so If somebody is rebasing all the time or pushing many commits we still have a lot of duplicated schemas.

packages/migrations/package.json
```diff
@@ -14,7 +14,7 @@
     "db:init": "pnpm db:create && pnpm migration:run",
     "db:migrator": "node --experimental-specifier-resolution=node --loader ts-node/esm src/index.ts",
     "migration:run": "pnpm db:migrator up",
-    "test": "node --experimental-specifier-resolution=node --loader ts-node/esm ./test/*.test.ts"
+    "test": "TS_NODE_TRANSPILE_ONLY=true node --experimental-specifier-resolution=node --loader ts-node/esm ./test/index.ts"
```
kamilkisiela, line 17: I did that because we use `Storage` in tests and otherwise I would have to include it and a few more paths in `tsconfig`.
It's type-checked anyway by `pnpm typecheck`.

packages/migrations/src/actions/2023.10.25T14.41.41.schema-checks-dedup.ts
```diff
@@ -0,0 +1,27 @@
...
+      ADD COLUMN "composite_schema_sdl_store_id" text REFERENCES "public"."sdl_store" ("id");
+
+    ALTER TABLE "public"."schema_checks"
+      ALTER COLUMN "schema_sdl" DROP NOT NULL,
+      ALTER COLUMN "supergraph_sdl" DROP NOT NULL,
+      ALTER COLUMN "composite_schema_sdl" DROP NOT NULL;
+
+    CREATE INDEX "schema_check_by_schema_sdl_store_id" ON "public"."schema_checks" ("schema_sdl_store_id" ASC);
+    CREATE INDEX "schema_check_by_supergraph_sdl_store_id" ON "public"."schema_checks" ("supergraph_sdl_store_id" ASC);
+    CREATE INDEX "schema_check_by_composite_schema_sdl_store_id" ON "public"."schema_checks" ("composite_schema_sdl_store_id" ASC);
```
n1ru4l, line 25: As discussed, we should verify whether we want to create these indices `CONCURRENTLY` or not.

packages/services/server/src/index.ts
```diff
@@ -90,26 +90,25 @@ export async function main() {
...
-      `Usage estimation is enabled. Start scheduling purge tasks for expired schema checks every ${env.hiveServices.usageEstimator.dateRetentionPurgeIntervalMinutes} minutes.`,
+      `Usage estimation is enabled. Start scheduling purge tasks every ${env.hiveServices.usageEstimator.dateRetentionPurgeIntervalMinutes} minutes.`,
     );
-    expiredSchemaCheckPurgeTaskRunner = createTaskRunner({
+    dbPureTaskRunner = createTaskRunner({
       async run() {
         await storage.purgeExpiredSchemaChecks({
           expiresAt: new Date(),
         });
+        await storage.purgeUnusedSchemasInStore();
```
kamilkisiela, line 106:
```suggestion
        // TODO: activate await storage.purgeUnusedSchemasInStore();
```

n1ru4l (approved):

n1ru4l (approved):

PR #3052, opened 2023-10-18
Check graphql server response for validity before updating supergraph file on regular poll.

andriihrachov:
### Background

If hive registry is reverse-proxied and becomes unreachable, there can be non-graphql response returned with different statuscode (e.g. 404 due to gateway misconfiguration, 503/502 if upstream is unreachable).

In this case plain text response is returned and supergraph file is updated with this plain text response, which cause apollo router to crash.

### Description

Regardless of response code returned, check response for validity. log error and do not update the supergraph file.

### Checklist

- [x] Testing


andriihrachov:
@kamilkisiela can you pls take a look ?

dotansimha (approved):

PR #3186, opened 2023-10-25
add status filters to the schema checks view

beerose:
Fixes https://github.com/kamilkisiela/graphql-hive/issues/2540

So, it may not be the best UX, but I couldn't figure out a better one without introducing another way of doing filters (we may need to unify it in the future). I was also thinking about explicit checkboxes, but I'm not sure what will be the most common usages (I guess we can change that later if needed?).

CleanShot 2023-10-25 at 18 49 43@2x

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/web/app/pages/[organizationId]/[projectId]/[targetId]/checks.tsx
+++ b/packages/web/app/pages/[organizationId]/[projectId]/[targetId]/checks.tsx
@@ -6,6 +6,7 @@ import { authenticated } from '@/components/authenticated-container';
 import { Page, TargetLayout } from '@/components/layouts/target';
 import { SchemaEditor } from '@/components/schema-editor';
 import { ChangesBlock, labelize } from '@/components/target/history/errors-and-changes';
+import { Label } from '@/components/ui/label';
 import { Subtitle, Title } from '@/components/ui/page';
 import { QueryError } from '@/components/ui/query-error';
 import {
@@ -17,6 +18,7 @@ import {
   Heading,
   MetaTitle,
   Modal,
+  Switch,
   TimeAgo,
   Tooltip,
 } from '@/components/v2';
@@ -49,6 +51,12 @@ const SchemaChecks_NavigationQuery = graphql(`
               commit
               author
             }
+            breakingSchemaChanges {
+              total
+            }
+            safeSchemaChanges {
+              total
+            }
           }
         }
         pageInfo {
@@ -61,11 +69,17 @@ const SchemaChecks_NavigationQuery = graphql(`
   }
 `);
 
+interface SchemaCheckFilters {
+  showOnlyFailed?: boolean;
+  showOnlyChanged?: boolean;
+}
+
 const Navigation = (props: {
   after: string | null;
   isLastPage: boolean;
   onLoadMore: (cursor: string) => void;
-}): React.ReactElement => {
+  filters?: SchemaCheckFilters;
+}) => {
   const router = useRouteSelector();
   const [query] = useQuery({
     query: SchemaChecks_NavigationQuery,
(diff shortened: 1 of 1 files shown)


dotansimha:
@beerose if I understand correctly, this only applies to the frontend? I think it might create weird UX because of pagination?

n1ru4l:
As @dotansimha said, we should rather do this pagination on the backend.

In order to be performant we should also set corresponding indices.
E.g. a schema change with changes would be a schema change with the SQL condition AFAIK

```sql
WHERE
  jsonb_typeof("safe_schema_changes") = 'array'
  OR jsonb_typeof("breaking_schema_changes") = 'array'
```

e.g.

```sql
CREATE INDEX "schema_checks_connection_pagination_with_changes" ON "schema_checks" (
  "target_id" ASC
  , "created_at" DESC
  , "id" DESC
)
WHERE
  jsonb_typeof("safe_schema_changes") = 'array'
  OR jsonb_typeof("breaking_schema_changes") = 'array'
```

https://github.com/kamilkisiela/graphql-hive/blob/519dec1bb12d6442967f9f8a22db6cfd9cd5382c/packages/migrations/src/actions/2023.06.06T11.26.04.schema-checks.ts#L31-L35

So in order to query all combinations of the filters, we would need to set multiple index permutations, which should be fine. 🤔 

This is the backend code where we would need to adjust the SQL query: https://github.com/kamilkisiela/graphql-hive/blob/519dec1bb12d6442967f9f8a22db6cfd9cd5382c/packages/services/storage/src/index.ts#L3545-L3615

The filter should then be passed through as a GraphQL argument: https://github.com/kamilkisiela/graphql-hive/blob/519dec1bb12d6442967f9f8a22db6cfd9cd5382c/packages/services/api/src/modules/schema/module.graphql.ts#L175-L178

packages/services/api/src/modules/shared/providers/storage.ts
```diff
@@ -722,6 +723,10 @@ export interface Storage {
      * Optional mapper for transforming the raw schema check loaded from the database.
      */
     transformNode?: (check: SchemaCheck) => TransformedSchemaCheck;
+    /**
+     * Optional filters config for filtering failed and/or changed the schema checks.
```
beerose, line 727:
```suggestion
     * Optional filters config for filtering failed and/or changed schema checks.
```

n1ru4l:
image

When there is nothing shown in the list, we could have a better UI experience 🤔

packages/services/storage/src/index.ts
```diff
@@ -3574,6 +3576,25 @@ export async function createStorage(connection: string, maximumPoolSize: number)
...
               : sql``
           }
+          ${
+            failed
+              ? sql`
+                AND (
+                  "is_success" = false
+                )
+              `
+              : sql``
```
n1ru4l, line 3586:
```sql
explain SELECT
  "id"
  , to_json("created_at") as "createdAt"
  , to_json("updated_at") as "updatedAt"
  , "schema_sdl" as "schemaSDL"
  , "service_name" as "serviceName"
  , "meta"
  , "target_id" as "targetId"
  , "schema_version_id" as "schemaVersionId"
  , "is_success" as "isSuccess"
  , "schema_composition_errors" as "schemaCompositionErrors"
  , "breaking_schema_changes" as "breakingSchemaChanges"
  , "safe_schema_changes" as "safeSchemaChanges"
  , "schema_policy_warnings" as "schemaPolicyWarnings"
  , "schema_policy_errors" as "schemaPolicyErrors"
  , "composite_schema_sdl" as "compositeSchemaSDL"
  , "supergraph_sdl" as "supergraphSDL"
  , "github_check_run_id" as "githubCheckRunId"
  , "github_repository" as "githubRepository"
  , "github_sha" as "githubSha"
  , coalesce("is_manually_approved", false) as "isManuallyApproved"
  , "manual_approval_user_id" as "manualApprovalUserId"
FROM
  "public"."schema_checks"
WHERE
  "target_id" = '97e7ccae-0e08-4fb7-84e0-c589f528816d'
  AND "is_success" = true
ORDER BY
  "target_id" ASC
  , "created_at" DESC
  , "id" DESC
LIMIT 21
```

```
Limit  (cost=0.14..8.17 rows=1 width=530)
  ->  Index Scan using schema_checks_connection_pagination on schema_checks  (cost=0.14..8.17 rows=1 width=530)
        Index Cond: (target_id = '97e7ccae-0e08-4fb7-84e0-c589f528816d'::uuid)
        Filter: is_success
```

We can add another index, to improve the query:

```sql
CREATE INDEX "schema_checks_connection_pagination_with_no_success" ON "schema_checks" (
  "target_id" ASC
  , "created_at" DESC
  , "id" DESC
)
WHERE
  "is_success" = false
;
```

```
Limit  (cost=0.12..8.15 rows=1 width=530)
  ->  Index Scan using schema_checks_connection_pagination_with_no_success on schema_checks  (cost=0.12..8.15 rows=1 width=530)
        Index Cond: (target_id = '97e7ccae-0e08-4fb7-84e0-c589f528816d'::uuid)
```

At the same time when both filters are enabled, we can also make the query use the index by combining the conditions:

```
CREATE INDEX "schema_checks_connection_pagination_with_no_success_and_changes" ON "schema_checks" (
  "target_id" ASC
  , "created_at" DESC
  , "id" DESC
)
WHERE
  "is_success" = false
  AND  (
    jsonb_typeof("safe_schema_changes") = 'array'
    OR jsonb_typeof("breaking_schema_changes") = 'array'
  )
;
```

beerose:
@n1ru4l One case I'm not sure about is the "empty state with filters" but with an active check. Should we redirect? Leave it as is? Add it to the list anyway?
CleanShot 2023-10-27 at 14 53 29@2x

n1ru4l:
Regarding https://github.com/kamilkisiela/graphql-hive/pull/3186#issuecomment-1782871448, I think it is an okay trade-off for now. We can think about how to improve the user experience in a follow-up issue/pr.

n1ru4l:
@beerose Can you please fix the prettier formatting issue?

dotansimha (approved):

n1ru4l:
@beerose One last thing 😓 We recently started a product update page on Hive: https://the-guild.dev/graphql/hive/product-updates

Can you please also add a short update at https://github.com/kamilkisiela/graphql-hive/tree/main/packages/web/docs/src/pages/product-updates for these changes?

You also need to add yourself to the authors here: https://github.com/kamilkisiela/graphql-hive/blob/17ea9228b7383fdd7aa6cc42955fa2948f0e313b/packages/web/docs/src/authors.ts#L8-L19

n1ru4l (approved):

PR #3187, opened 2023-10-25
add link to a github commit for schema checks

beerose:
fixes https://github.com/kamilkisiela/graphql-hive/issues/3030

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -755,6 +755,10 @@ export default gql`
     The URL of the schema check on the Hive Web App.
     """
     webUrl: String
+    """
```
beerose, line 758: So the reason why I did was that I noticed that it was possible to:
- Have a schema check **with** associated github repository (`github_repository` set in the `shema_checks` table in the database) — when a `schema:check` was done with a `--github` flag,
- But its corresponding schema version has a `github_repository` empty (i.e. when `schema: publish` was never run with `--github`). 
 
Currently, a way to get github info is through the `schemaVersion` field, which means that in the case mentioned above, we wouldn't be able to show the commit link for schema checks. 

Unless there was a reason for that?

n1ru4l: A schema check or schema version can be done from multiple GitHub repositories as federated/distributed schemas can reside in different Git repositories.

So a schema version can be triggered by one git repository and a schema check by another repository.
We should only link to the GitHub repository that is linked to the check or schema version.

n1ru4l (approved):

PR #3230, opened 2023-10-31
feat: run migrations outside transaction

n1ru4l:
### Background

Adding indices to the `schema_checks` table is slow and would block pending operations on the table (writes, updates) for at least 30 seconds, according to @kamilkisiela estimations), which would block our customers from using any features related to schema_checks for that timeframe...

Instead, we can create indices concurrently using the [`CONCURRENTLY` keyword](https://www.postgresql.org/docs/current/sql-createindex.html).
However, this requires us to run database migrations outside a transaction, which is fine but requires adjustments to our migrations-runner, which is running everything inside a single transaction. 
I also added some time logging, which makes it easier for us to see the actual runtimes of the single migrations.

Example output:

image

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/migrations/src/actions/2023.10.26T12.44.36.schema-checks-filters-index.ts
+++ b/packages/migrations/src/actions/2023.10.26T12.44.36.schema-checks-filters-index.ts
@@ -3,37 +3,50 @@ import { type MigrationExecutor } from '../pg-migrator';
 export default {
   name: '2023.10.26T12.44.36.schema-checks-filters-index.ts',
   noTransaction: true,
-  run: ({ sql }) => sql`
-    CREATE INDEX CONCURRENTLY "schema_checks_connection_pagination_with_changes" ON "schema_checks" (
-      "target_id" ASC
-      , "created_at" DESC
-      , "id" DESC
-    )
-    WHERE
-      jsonb_typeof("safe_schema_changes") = 'array'
-      OR jsonb_typeof("breaking_schema_changes") = 'array'
-    ;
-
-    CREATE INDEX CONCURRENTLY "schema_checks_connection_pagination_with_no_success" ON "schema_checks" (
-      "target_id" ASC
-      , "created_at" DESC
-      , "id" DESC
-    )
-    WHERE
-      "is_success" = false
-    ;
-
-    CREATE INDEX CONCURRENTLY "schema_checks_connection_pagination_with_no_success_and_changes" ON "schema_checks" (
-      "target_id" ASC
-      , "created_at" DESC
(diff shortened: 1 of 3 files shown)


packages/migrations/src/actions/2023.10.26T12.44.36.schema-checks-filters-index.ts
```diff
@@ -2,37 +2,51 @@ import { type MigrationExecutor } from '../pg-migrator';
...
+        )
+        WHERE
+          "is_success" = false
+          AND  (
+            jsonb_typeof("safe_schema_changes") = 'array'
+            OR jsonb_typeof("breaking_schema_changes") = 'array'
+          )
+        ;
+      `,
+    },
```
n1ru4l, line 50: Note here: If we have multiple index creations within a single tagged `sql` block, it will automatically run it inside a transaction block. `CONCURRENTLY` cannot be run within a transaction block. That is why I introduced these steps and as a bonus we will also log the execution times of each of these steps.

packages/migrations/src/index.ts
```diff
@@ -1,17 +1,14 @@
...
 import { runPGMigrations } from './run-pg-migrations';
 
-const slonik = await createPool(createConnectionString(env.postgres));
-
-const __dirname = path.dirname(url.fileURLToPath(import.meta.url));
-const actionsDirectory = __dirname + path.sep + 'actions';
-console.log('Actions in:', actionsDirectory);
+const slonik = await createPool(createConnectionString(env.postgres), {
+  // 10 minute timeout per statement
+  statementTimeout: 10 * 60 * 1000,
```
n1ru4l, line 10: I increased the timeout time as the CONCURRENTLY statements will run for a longer time.

n1ru4l:
Before running db migration indices sizes.

image


image


```
Running migration: 2023.10.26T12.44.36.schema-checks-filters-index.ts
  Starting step schema_checks_connection_pagination_with_changes
  Finished in 52.130s
  Starting step schema_checks_connection_pagination_with_no_success
  Finished in 309ms
  Starting step schema_checks_connection_pagination_with_no_success_and_changes
  Finished in 376ms
Finished in 52.931s
Running migration: 2023.10.30T00-00-00.drop-persisted-operations.ts
Finished in 301ms
```


Post-migration:

image

PR #3239, opened 2023-11-01
UI fixes for large screens 

dotansimha:
- [x] Allow small/medium screens to use 100% of the space
- [x] Limit huge screens to use some space
- [x] Fix for Lab page height limit


n1ru4l (approved):
Checked it on the following resolutions and it is looking fine!

```
3440×1440
1920×1080
1366×768
1440×900
1280×720
1280×1024
```

PR #3279, opened 2023-11-05
prettier v3 and apply fixes to existing files

dotansimha:
Closes https://github.com/kamilkisiela/graphql-hive/pull/2901 https://github.com/kamilkisiela/graphql-hive/pull/2173 https://github.com/kamilkisiela/graphql-hive/pull/2888


n1ru4l:
doesn't this break the vscode plugin?

dotansimha:
> doesn't this break the vscode plugin?

seems to work after restarting VSCode. Config loading works fine, I guess because we are still using CJS.

PR #3296, opened 2023-11-06
Fix legacy SDL, introspection and metadata (+ publishing metadata) endpoints

kamilkisiela:
`/sdl` and `/introspection` - both expected an object with `"sdl"` property in it, but we passed SDL to them directly.

`/metadata` - the endpoint was fine but apparently the schema publisher uploaded an array of metadata instead of a single object. In order to not break existing Mesh projects, I had to detect if a value returned from R2 was a stringified array of objects and if that object was in fact a Mesh artifact. If so, I had to turn it into an object.


kamilkisiela:
Applied in #3298

PR #3298, opened 2023-11-06
feat: no longer publish to the legacy CDN

n1ru4l:
- [x] /schema should return [{sdl: string}]
- [x] /supergraph should return a string
- [x] /sdl should return a string
- [x] /metadata should return what's in the R2 but {object} instead of [{object}] when it's Mesh artifact
- [x] /introspection should return introspection result

### DIFF
--- a/packages/services/cdn-worker/src/handler.ts
+++ b/packages/services/cdn-worker/src/handler.ts
@@ -1,5 +1,7 @@
 import { buildSchema, introspectionFromSchema } from 'graphql';
 import { Analytics, createAnalytics } from './analytics';
+import { GetArtifactActionFn } from './artifact-handler';
+import { ArtifactsType as ModernArtifactsType } from './artifact-storage-reader';
 import {
   CDNArtifactNotFound,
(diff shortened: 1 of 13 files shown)


packages/services/cdn-worker/src/handler.ts
```diff
@@ -167,6 +177,17 @@ function createArtifactTypesHandlers(analytics: Analytics) {
 const VALID_ARTIFACT_TYPES = artifactTypes;
 const AUTH_HEADER_NAME = 'x-hive-cdn-key';
 
+const legacyToModernArtifactTypeMap: Record<
+  ArtifactType,
+  Extract<ModernArtifactsType, 'sdl' | 'services' | 'metadata' | 'supergraph'>
+> = {
+  schema: 'services',
```
kamilkisiela, line 184: Artifacts published to KV as `schema` were either `[{sdl: ""}, {sdl: ""}]` or `{sdl: ""}` but I checked and everyone who requested `/schema` got an array anyway, so it's not a breaking change.

PR #3315, opened 2023-11-07
Limit metadata to SINGLE project and fix being published as array

kamilkisiela:
- [ ] double check if metadata is used outside `single` project.


kamilkisiela:
#3326

PR #3326, opened 2023-11-08
Metadata as array for COMPOSITE projects, as object for SINGLE project

kamilkisiela:
1. fixes metadata being published as an array in SINGLE projects by rewriting Mesh artifacts (only mesh users are fetching metadata from CDN on SINGLE projects) to objects (on CDN)
2. adds tests for metadata
3. makes metadata in FEDERATION projects officially supported 

Once all R2 objects storing metadata as an array are migrated / re-published, we will be able to reduce the "rewriting of Mesh artifacts" part. We need to see if all targets were updated after this PR was deployed to production.

### DIFF
--- a/integration-tests/tests/models/single-legacy.spec.ts
+++ b/integration-tests/tests/models/single-legacy.spec.ts
@@ -443,7 +443,7 @@ describe('other', () => {
 
       let cdnMetadataResult = await fetchMetadataFromCDN();
       expect(cdnMetadataResult.status).toEqual(200);
-      expect(cdnMetadataResult.body).toEqual([{ c0: 1 }]);
+      expect(cdnMetadataResult.body).toEqual({ c0: 1 });
 
       const versions = await fetchVersions(3);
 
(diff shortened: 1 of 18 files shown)


packages/services/cdn-worker/src/artifact-handler.ts
```diff
@@ -125,14 +133,66 @@ export const createArtifactRequestHandler = (deps: ArtifactRequestHandler) => {
...
     if (result.type === 'redirect') {
+      if (params.artifactType === 'metadata') {
+        // To not change a lot of logic and still reuse the etag bits, we
+        // fetch metadata using the redirect location.
+        // Once we convert all the legacy metadata (SINGLE project passes an array instead of an object),
+        // we can remove this and continue serving a redirect.
+        // In case of metadata, we need to fetch the artifact and transform it.
+        // We're using here a private location, because the public S3 endpoint may differ from the internal S3 endpoint. E.g. within a docker network,
+        // and we're fetching the artifact from within the private network.
+        // If they are the same, private and public locations will be the same.
```
kamilkisiela, line 150: @n1ru4l

integration-tests/tests/models/single-legacy.spec.ts
```diff
@@ -443,7 +443,7 @@ describe('other', () => {
 
       let cdnMetadataResult = await fetchMetadataFromCDN();
       expect(cdnMetadataResult.status).toEqual(200);
-      expect(cdnMetadataResult.body).toEqual([{ c0: 1 }]);
+      expect(cdnMetadataResult.body).toEqual({ c0: 1 });
```
kamilkisiela, line 446: Every past call we registered was a Mesh artifact and it's expected to be an object, not an array, so it's not a breaking change.

n1ru4l (approved):

PR #3329, opened 2023-11-08
Use version.compositeSchemaSDL instead of keep recomposing the schema

kamilkisiela:


packages/services/api/src/modules/lab/resolvers.ts
```diff
@@ -59,6 +68,8 @@ export const resolvers: LabModule.Resolvers = {
       const orchestrator = schemaManager.matchOrchestrator(type);
       const helper = injector.get(SchemaHelper);
 
+      // TODO: should we use compositeSchemaSDL here, instead of calling orchestrator.composeAndValidate?
```
kamilkisiela, line 71:
```suggestion
```

n1ru4l (approved):

PR #3356, opened 2023-11-09
add option to pass validateOptions

jmvtrinidad:
Closes https://github.com/kamilkisiela/graphql-hive/issues/3379 

### Background

By default apollo is false, so we are proposing to enable to pass some options for graphql inspector validate.
This will be used if the operation that is being check has `@client` directive
https://github.com/kamilkisiela/graphql-inspector/blob/master/packages/core/src/validate/index.ts#L47

### DIFF
--- a/packages/libraries/cli/README.md
+++ b/packages/libraries/cli/README.md
@@ -222,6 +222,9 @@ FLAGS
   --require=<value>...
       [default: ] Loads specific require.extensions before running the command
 
+  --validateOptions=<value>...
+      [default: '{}'] validations options to be passed to graphql inspector validate library, as JSON
+
   --token=<value>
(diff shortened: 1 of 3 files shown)


dotansimha:
Thanks @jmvtrinidad .
Few notes: 
- Documentation for this change missing.
- Tests for this change are missing. 
- There is no issue linked to this PR, so I'm not sure what's the background for that?

Also, `validateOptions` is an object so I'm not sure how this will be passed from the shell? can you please share an example?

jmvtrinidad:
Hello @dotansimha 
Please see the created issue
https://github.com/kamilkisiela/graphql-hive/issues/3379
Thanks

PR #3357, opened 2023-11-09
Overwrite a query's request timeout using CLICKHOUSE_REQUEST_TIMEOUT env var

kamilkisiela:
If not set, use query-level values.


packages/services/api/src/modules/operations/providers/clickhouse-client.ts
```diff
@@ -107,7 +107,8 @@ export class ClickHouse {
           lookup: 1000,
           connect: 1000,
           secureConnect: 1000,
-          request: timeout,
+          // override the provided timeout of a query with the globally configured timeout
+          request: this.config.requestTimeout ?? timeout,
```
tapaderster, line 111: does this apply to the harcoded timeouts like this one:
https://github.com/kamilkisiela/graphql-hive/blob/main/packages/services/api/src/modules/operations/providers/operations-reader.ts#L694

PR #3359, opened 2023-11-09
feat: remember breaking change approvals in context of a pull request

n1ru4l:
### Background

See https://github.com/kamilkisiela/graphql-hive/issues/2698
See https://github.com/kamilkisiela/graphql-hive/issues/3216

### Description

image

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/migrations/src/actions/2023.11.09T00.00.00.schema-check-approval.ts
```diff
@@ -0,0 +1,19 @@
+import type { MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2023.11.09T00.00.00.schema-check-approval.ts',
+  run: ({ sql }) => sql`
+    CREATE TABLE "public"."schema_change_approvals" (
```
kamilkisiela, line 6: should we do TTL here?

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -310,6 +310,16 @@ export class SchemaPublisher {
       }
     }
 
+    let contextId: string | null = input.contextId ?? null;
+
+    if (
+      input.contextId === undefined &&
```
kamilkisiela, line 316:
```suggestion
      !contextId &&
```

maybe like this to prevent a situation where we pass an empty string or `null` accidentally (typescript is typescript)?

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -331,6 +341,18 @@ export class SchemaPublisher {
 
     let checkResult: SchemaCheckResult;
 
+    const approvedSchemaChanges = new Map<string, SchemaChangeType>();
+
+    if (contextId !== null) {
```
kamilkisiela, line 346: do we check if `contextId` is an empty string (or string with whitespaces only)?

packages/services/server/src/index.ts
```diff
@@ -116,6 +116,12 @@ export async function main() {
           );
           transaction.setMeasurement('deletedSchemaCheckCount', result.deletedSchemaCheckCount, '');
           transaction.setMeasurement('deletedSdlStoreCount', result.deletedSdlStoreCount, '');
+          transaction.setMeasurement(
```
kamilkisiela, line 119: oh so we do TTL, nvm

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: Good catch in general we should do some `contextId` validation to avoid someone using a novel as the context id

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: I did an `undefined` check so that in case the c`ontextId` value sent from the client is null, no context fallback from github input is used. Not sure if really needed though

n1ru4l: https://github.com/kamilkisiela/graphql-hive/pull/3359/commits/66ca884fe98c59867d15fd26ca165c7ceddd0d26

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: https://github.com/kamilkisiela/graphql-hive/pull/3359/commits/66ca884fe98c59867d15fd26ca165c7ceddd0d26

PR #3362, opened 2023-11-10
laboratory: persist and sync local and remote changes

beerose:
Fixes https://github.com/kamilkisiela/graphql-hive/issues/2737

1. Breadcrumbs + a yellow dot next to the save icon to subtly prompt a user to save changes: 
CleanShot 2023-11-20 at 09 11 28@2x
  If there's not current operation, the breadcrumb displays `New operation`.

2. Added a `Discard Changes` button:
CleanShot 2023-11-22 at 07 05 15@2x

---
One of the ideas was also to add a diff modal (to sync local and remote versions). We decided to go with a simpler solution, but I'm mentioning it here for reference.

CleanShot 2023-11-21 at 18 02 49@2x

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/web/app/pages/[organizationId]/[projectId]/[targetId]/laboratory.tsx
```diff
@@ -869,7 +864,7 @@ function LaboratoryPageContent() {
               </>
             ),
           }}
-          showPersistHeadersSettings={false}
+          showPersistHeadersSettings
```
Comment (beerose) [inline packages/web/app/pages/[organizationId]/[projectId]/[targetId]/laboratory.tsx:867]: I figured that always persisting headers can be a security concern, but we can allow turning in on in the settings.

packages/web/docs/src/pages/product-updates/2023-11-10-laboratory-persist-state.mdx
```diff
@@ -0,0 +1,10 @@
+---
```
beerose, line 1: Not sure if that isn't too minor to include in the updates

packages/web/app/pages/[organizationId]/[projectId]/[targetId]/laboratory.tsx
Comment (dotansimha) [reply packages/web/app/pages/[organizationId]/[projectId]/[targetId]/laboratory.tsx:867]: Yes, I think this is why we originally disabled it by default. Using the settings to control that is a good decision 👍

packages/web/docs/src/pages/product-updates/2023-11-10-laboratory-persist-state.mdx
dotansimha: I think it can be added, especially if this changes the default behavior and might effect things like security.
@beerose maybe we should mention that this new behavior can be disabled from the editor's settings?

packages/web/docs/src/pages/product-updates/2023-11-10-laboratory-persist-state.mdx
```diff
@@ -0,0 +1,10 @@
...
+title: Persist query and variables state for the laboratory GraphiQL playground
+description:
+  Query and variables state is now saved in the browser for the laboratory GraphiQL playground.
+date: 2023-11-10
+authors: [aleksandra]
+---
+
+The laboratory GraphiQL playground now saves the query and variables state in the browser. This
+means that you can now refresh the page and the query and variables will be preserved.
```
beerose, line 10: note to self: update this once there's a final version of this pr

packages/web/app/src/components/v2/diff-editor.tsx
```diff
@@ -49,9 +52,9 @@ export const DiffEditor = ({
   }
 
   return (
-    <div className="h-full w-full">
+    <div className={cx('h-full w-full', className)}>
       <div className="flex items-center justify-between p-5">
-        <div className="font-semibold">{title}</div>
+        {title && <div className="font-semibold">{title}</div>}
```
beerose, line 57: I can revert it now that I don't need `DiffEditor` after all, but I think it might be a nice-to-have API change for this component.

n1ru4l (approved):

PR #3364, opened 2023-11-10
Member access roles

### LINKED ISSUE #1157: Group of access scopes
Allow creating groups of access scopes, that could be assigned to users.
We could have a recommended set of groups, but not something that is built-in, but rather created on demand.

This will result in a modification of the user invite logic. When inviting a user to an organization, this user will receive a selected access group when accepting the invitation.

### LINKED ISSUE #2718: Default permissions for newly added users
It would be nice to be able to setup Roles and be able to have default role for newly added users to be able to give specific permissions, now we have almost no permissions assigned and need to manually adjust each new user.

kamilkisiela:
# Member Roles.

Closes #2718
Closes #1157

## Description

- Introduces two built-in roles: Admin (full access) and Viewer (read-only access).
- The built-in roles are locked, meaning they cannot be modified or deleted.
- Users with `organization:members` access scope can create/modify/assign/delete the roles.
- Allows to assign a role when inviting a new member.

## Creating a role

To create a role a user is required to:
- have access to `organization:members` scope
- have access to all selected scopes
- define a unique (per organization) name of the role

## Deleting a role

To delete a role
- a user needs an access to `organization:members` scope
- a user is required to have access to all scopes of the role
- the role can't be locked (Admin or Viewer)
- can't have members
- when a role is deleted, but some of the pending invitations are assigned to that role, the invitations are modified and a Viewer role is used instead.

Screenshot 2023-11-20 at 09 51 43

## Assigning a role

- a user needs an access to `organization:members` scope
- selected role needs to be part of the organization
- user needs to have access to all the scopes of the **new** role
- user needs to have access to all the scopes of the **old** role
- new role has to have at least the same access scopes as the old role (to avoid downgrading users) - Admin role is an exception here

Screenshot 2023-11-16 at 21 58 25

## Modifying a role

- a user needs an access to `organization:members` scope
- it can't be a locked role (Admin or Viewer)
- a user needs to have access to all the selected scopes
- role's name has to be unique
- a modified role has to have at least the same access scopes as the old role (if it has members), to avoid downgrading members (Admin role is an exception)

Screenshot 2023-11-17 at 14 34 16

## Inviting a new member

- a user with `organization:members` access scope can invite new members.
- a role must be assigned to the invited member (roles are available based on user's permissions).

Screenshot 2023-11-16 at 21 58 47
Screenshot 2023-11-16 at 21 58 29

## Migration story

- existing members won't be assigned to any role.
- existing members will keep their access.
- modifying member's access in the legacy way (granular selection of scopes) is still possible for users without a role
- a migration wizard will be available to help admins create and assign roles to members during the migration period (doing it manually would be painful...)
- we give Hive users X days to perform the migration, otherwise all users without a role will be assigned to custom-created role that aligns with their permissions.

Screenshot 2023-11-20 at 09 50 21
Screenshot 2023-11-20 at 13 35 49
Screenshot 2023-11-21 at 16 21 45
Screenshot 2023-11-21 at 16 21 30


packages/services/api/src/modules/organization/resolvers.ts
```diff
@@ -1,13 +1,43 @@
...
+      required_error: 'Please enter role name',
+    })
+    .min(2, 'Role name must be at least 2 characters long')
+    .pipe(capitalLetter)
+    .pipe(builtInRole),
+  description: z
+    .string({
+      required_error: 'Please enter role description',
+    })
+    .min(2, 'Role description must be at least 2 characters long'),
```
n1ru4l, line 38: Also, a max length so nobody tries to store Harry Potter and the Order of Phoenix as the description.

packages/services/api/src/modules/organization/resolvers.ts
```diff
@@ -1,13 +1,43 @@
...
+const capitalLetter = z.custom<string>(val => {
+  return typeof val === 'string' && val.length > 0 && val[0] === val[0].toUpperCase();
+}, 'Role name must start with a capital letter');
+
+const createOrUpdateMemberRoleInputSchema = z.object({
+  name: z
+    .string({
+      required_error: 'Please enter role name',
+    })
+    .min(2, 'Role name must be at least 2 characters long')
```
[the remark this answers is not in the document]
kamilkisiela: [a0941ac](https://github.com/kamilkisiela/graphql-hive/pull/3364/commits/a0941acff52342b33fc07bb5f65c43a87ac6480a)

packages/services/api/src/modules/organization/resolvers.ts
kamilkisiela: [a0941ac](https://github.com/kamilkisiela/graphql-hive/pull/3364/commits/a0941acff52342b33fc07bb5f65c43a87ac6480a)

PR #3366, opened 2023-11-10
Persist the schema filter checks as query parameters on page refresh

### LINKED ISSUE #3261: persist schema check filter state in url parameters
Right now, when you change a schema filter and then refresh the page the filter state is reset. Instead we should persist the state in the URL.

image

```
https://app.graphql-hive.com/the-guild/graph-ql-hive/staging/checks?filter_changes=1
https://app.graphql-hive.com/the-guild/graph-ql-hive/staging/checks?filter_changes=1&filter_errors=1
```

deep-maheshwari:
### Background

This fixes #3261 . The state is initialized to empty and on re-render we need to fetch the state from router's query params so that the checks are persisted.

### Description

This affects the frontend module for the checks page under target. The state for filter checks would be fetched from router's query params and would be updated when filters are changed.

### Checklist

- [x] Testing

### DIFF
--- a/packages/web/app/pages/[organizationId]/[projectId]/[targetId]/checks.tsx
+++ b/packages/web/app/pages/[organizationId]/[projectId]/[targetId]/checks.tsx
@@ -841,7 +841,7 @@ function ChecksPageContent() {
       showOnlyFailed: !filters.showOnlyFailed,
     }));
   };
-  
+
   return (
     <>
       <TargetLayout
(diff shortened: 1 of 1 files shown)


deep-maheshwari:
Hi @dotansimha thanks for adding the reviewers, I see some checks failing at step run_containers, is there anything that needs to be done for it?

deep-maheshwari:
@n1ru4l @beerose could you please review this PR, Thanks!

n1ru4l (approved):

n1ru4l:
> Hi @dotansimha thanks for adding the reviewers, I see some checks failing at step run_containers, is there anything that needs to be done for it?

We currently, do not run integration tests for contributions. We will consider this in the future.

Thank you for your contribution, I just verified the functionality locally!

PR #3396, opened 2023-11-15
feat: revise breaking change presentation in web app

n1ru4l:
### Background

>We show “breaking changes” that are safe based on usage on the “Safe Changes” section.
I feel like it makes more sense to show them on the “Breaking Changes” section with an annotation instead (similar to screenshot attached). We actually already have the annotation “(safe based on usage)“, but having the change within Safe Changes still feels a bit misleading.

[Internal discussion](https://guild-oss.slack.com/archives/C01E8ATBQGN/p1699872922797379)

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/schema/module.graphql.ts
+++ b/packages/services/api/src/modules/schema/module.graphql.ts
@@ -296,13 +296,22 @@ export default gql`
   type SchemaChange {
     criticality: CriticalityLevel!
     criticalityReason: String
-    message: String!
+    message(
+      """
+      Whether to include a note about the safety of the change based on usage data within the message.
+      """
+      withSafeBasedOnUsageNote: Boolean = true
+    ): String!
     path: [String!]
     """
     Approval metadata for this schema change.
     This field is populated in case the breaking change was manually approved.
     """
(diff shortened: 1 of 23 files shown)


packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -296,13 +296,22 @@ export default gql`
   type SchemaChange {
     criticality: CriticalityLevel!
     criticalityReason: String
-    message: String!
+    message(
+      """
+      Whether to include a note about the safety of the change based on usage data within the message.
+      """
+      withSafeBasedOnUsageNote: Boolean = true
```
n1ru4l, line 303: We have legacy clients (CLI) that don't select the `isSafeBasedOnUsage` field.
Those should still receive the same message as before.

PR #3503, opened 2023-11-30
Fix usage-based exclusion of clients

comp615:
### Background

We have a query like the following done from multiple clients:
query getVisitorPropertiesQuery($marketId: String!, $userId: ID!) {
  clinicPodForUser(userId: $userId) {
    name
  }
  clinicsForMarket(marketId: $marketId) {
    id
    name
  }
  currentUser {
    createdAt
    permissions
  }
}

In a PR, I removed the `clinicPodForUser` field from top-level Query. However, the Hive check is flagging this as safe based on usage. We only exclude one client, and this query is made on multiple, so this should be flagged as broken

image

### Description

Instead of filtering hashes where the client names are present, we actually want to select hashes where a non-excluded client is present

### Checklist

TBD: Trying to discover tests run via CI

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/operations/providers/operations-reader.ts
+++ b/packages/services/api/src/modules/operations/providers/operations-reader.ts
@@ -251,9 +251,10 @@ export class OperationsReader {
         hash NOT IN (
           SELECT hash FROM clients_daily ${this.createFilter({
             target,
-            period,
-            extra: [sql`client_name IN (${sql.array(excludedClients, 'String')})`],
-          })} GROUP BY hash
+            period
+          })}
+          GROUP BY hash
+          HAVING COUNT(CASE WHEN `client` NOT IN (${sql.array(excludedClients, 'String')}) THEN 1 END) > 0
         )
       `);
     }


n1ru4l:
@comp615 running the integration tests is quite simple (if u have docker installed)

> **TL;DR**: Use `pnpm integration:prepare` command to set up the complete environment from locally
running integration tests. You can ignore the rest of the commands in this section, if this script
worked for you, and just run `pnpm test:integration` to run the actual tests.


https://github.com/bczerwinski-equinix/graphql-hive/blob/925ba5514b7430f31e63104e0abb342d73eee24e/docs/TESTING.md#L19-L23

I fixed the test case and will have another look at the actual fix next week!

PR #3504, opened 2023-11-30
show affected operations for a schema change

beerose:
### Background

Fixes https://github.com/kamilkisiela/graphql-hive/issues/2442

### Description

The flow:
1. On `schema:check` — for each breaking change, get relevant data from clickhouse (insights about usage). Note: we only get top 10 operations so that we don't overload the DB.
2. Store that data in `schema_checks` table (inside breaking_changes JSON)
3. In the hive app, on the Checks page, get check info along with the affected operations we got before.

UI:

image

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/schema/providers/inspector.ts
```diff
@@ -100,6 +100,18 @@ export class Inspector {
       .sort((a, b) => a.criticality.localeCompare(b.criticality));
   }
 
+  @sentry('Inspector.getUsageForCoordinate')
+  async getUsageForCoordinate(coordinate: string, selector: Types.TargetSelector) {
+    return this.operationsManager.getTopOperationForCoordinate({
+      coordinate,
+      period: createPeriod('7d'),
+      limit: 10,
```
beerose, line 108: Tempted to limit it to even less, e.g. 5

packages/web/app/src/components/target/history/errors-and-changes.tsx
```diff
@@ -32,20 +34,97 @@ export function ChangesBlock(props: {
...
+                                      projectId: router.projectId,
+                                      targetId: router.targetId,
+                                      operationName: `${hash.substring(0, 4)}_${name}`,
+                                      operationHash: hash,
+                                      period: '7d',
+                                    },
+                                  }}
+                                >
+                                  {hash.substring(0, 4)}_{name}
+                                </Link>
```
beerose, line 89: Question: Currently, if we open the insights page with a non-existing operation, it just shows empty stats components. Should we change it to display some information that _we don't have any stats on this operation/  operation does not exist_?

n1ru4l: @beerose That is a good idea!

packages/services/api/src/modules/schema/providers/inspector.ts
```diff
@@ -100,6 +100,18 @@ export class Inspector {
       .sort((a, b) => a.criticality.localeCompare(b.criticality));
   }
 
+  @sentry('Inspector.getUsageForCoordinate')
+  async getUsageForCoordinate(coordinate: string, selector: Types.TargetSelector) {
+    return this.operationsManager.getTopOperationForCoordinate({
+      coordinate,
+      period: createPeriod('7d'),
```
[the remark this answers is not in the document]
n1ru4l: We should use the accounts data retention period here .🤔 It is the `limit_retention_days` column on the `projects` table.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -329,6 +329,22 @@ export class RegistryChecks {
...
+          selector,
+        );
+
+        if (affectedOperations) {
+          change.affectedOperations = affectedOperations;
+        }
+        return change;
+      }),
+    );
+
```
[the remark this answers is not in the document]
n1ru4l: @kamilkisiela What would you recommend here?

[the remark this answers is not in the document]
kamilkisiela: hmm, it depends what we wish to show here.
Do we want to show a list of client names as well, or not?

WillBoudnik:
What would be nice for our situation is to see the affected clients too.  We have cloud functions and multiple projects that call the same graphql with different client names.  This change would help us identify pieces of code we didn't check.

packages/web/app/src/components/target/history/errors-and-changes.tsx
beerose: I figured that can be a separate PR as it will require more changes

beerose:
> What would be nice for our situation is to see the affected clients too. We have cloud functions and multiple projects that call the same graphql with different client names. This change would help us identify pieces of code we didn't check.

Hmm, we could add it next to the operations usage data. The "go-to insights" button will take you to live data (which includes all clients using a field), though — is that sufficient?

kamilkisiela:
@n1ru4l @beerose @dotansimha 
We need to answer two main questions.

Screenshot 2023-12-14 at 14 09 56

What about the option above? Should we follow this client name exclusion when we fetch a list of affected operations?

Screenshot 2023-12-14 at 14 09 58

Should we show affected operations of selected targets?

beerose:
> @n1ru4l @beerose @dotansimha We need to answer two main questions.

It depends on whether we want to think about it as "just" stats. If it's saved based on usage (= we show "save based on usage" info), then the additional affected stuff is more of a statistics detail. We can also make the messaging less aggressive (different colour + "used by" instead of "affected").

However, if we decide to do the filtering out, we should also filter out operations that were used less than the user set in the breaking changes settings. Currently, it only takes the period from the settings.

n1ru4l:
I agree with @kamilkisiela that we should leverage the breaking change target configuration for the displayed data. Users want to see why a schema check failed depending on their specific configuration.

n1ru4l:
Todos:

- [ ] Load operations only where the change `isSafeBasedOnUsage` is `false`
- [ ] Load operations based on breaking change configuration. This means that we need to store the target within the changes `affectedOperations` value, we also need to make this information obvious on the schema check UI 🤔  Maybe show for the schema checks against which target we checked!

klippx:
> On schema:check — for each breaking change, get relevant data from clickhouse (insights about usage). Note: we only get top 10 operations so that we don't overload the DB.

Is this still the case that you only fetch the top 10 operations, or is the PR desription outdated? I am worried that we might not see which operation is affected, if you are breaking "small" (in terms of usage) but critical operations.

n1ru4l:
@klippx Storing all the operations for each schema change for each schema check would significantly increase the used storage size as some fields could be referenced by hundreds of operations. We decided that for the schema check, we will store the top 10 used operations, but then also later on provide a new view/link/table for seeing all the affected operations queried from the live usage data.

n1ru4l (approved):

n1ru4l (approved):

PR #3505, opened 2023-12-01
SQL statements should not have PostgreSQL schema names hardcoded

bczerwinski-equinix:
### Background

This PR fixes a bug which I found when I have used PostgreSQL user with a not default schema set via `search_path` - migration scripts were failing.

### Description

Fixes: https://github.com/kamilkisiela/graphql-hive/issues/3502

There should be no schema hardcoded (as a prefix) in SQL statements. `public` schema is a default one anyway.

### Checklist

- [x] pnpm test
- [ ] pnpm test:integration

### DIFF
--- a/packages/migrations/src/actions/2021-03-05T19-06-23.initial.ts
+++ b/packages/migrations/src/actions/2021-03-05T19-06-23.initial.ts
@@ -28,7 +28,7 @@ COMMENT
 
 --- Tables
 CREATE TABLE
-  public.users (
+  users (
     id UUID PRIMARY KEY DEFAULT uuid_generate_v4 (),
     email VARCHAR(320) NOT NULL UNIQUE,
     external_auth_user_id VARCHAR(50) NOT NULL UNIQUE,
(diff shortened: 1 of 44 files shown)


n1ru4l (changes requested):
@bczerwinski-equinix For me the integration tests are passing locally. 🤔  Could you please share more information on the exact errors that make these tests fail? Did you alter the `docker-compose.yml` files?

Could you also update the `createConnectionString` function to always append the [`public` `search_path`](https://www.postgresql.org/docs/8.1/runtime-config-client.html) by default?  https://github.com/bczerwinski-equinix/graphql-hive/blob/925ba5514b7430f31e63104e0abb342d73eee24e/packages/services/storage/src/db/utils.ts#L3-L13

bczerwinski-equinix:
> @bczerwinski-equinix For me the integration tests are passing locally. 🤔 Could you please share more information on the exact errors that make these tests fail? Did you alter the `docker-compose.yml` files?
> 
> Could you also update the `createConnectionString` function to always append the [`public` `search_path`](https://www.postgresql.org/docs/8.1/runtime-config-client.html) by default? https://github.com/bczerwinski-equinix/graphql-hive/blob/925ba5514b7430f31e63104e0abb342d73eee24e/packages/services/storage/src/db/utils.ts#L3-L13

Possibly I have some issues with my local env as running those tests takes a lot of time and they are failing randomly - let's not waste time on that, provided they will pass in CI and in your env.

Why do you want to append `search_path` to the connection string? (by the way - you have sent a link to an outdated Postgresql version)
The idea of this change is to **not** restrict schema in any type of SQL statements by hardcoding it either in statements or connection strings. `public` schema is a default one - in `search_path` docs you can find: `The default value for this parameter is "$user", public.`.
Actually, what could be done next is to implement support for `search_path` value provided via configuration, as I have described here: https://github.com/kamilkisiela/graphql-hive/issues/3482

kamilkisiela (approved):

klys-equinix (commented):
@n1ru4l Could you take a look since the PR was approved by @kamilkisiela ?

bczerwinski-equinix:
@n1ru4l , you have requested changes - could you review, as it is blocking this PR?

n1ru4l:
@bczerwinski-equinix I was out sick. Can you send the follow-up PR for defining the pg schema?

PR #3555, opened 2023-12-08
Update `exclude` arg to accept RegExp

kongMina:
### Background
https://github.com/kamilkisiela/graphql-hive/issues/3493

There are certain operations that are prefixed or suffixed in our system that we use internally, and we have no benefit from posting this to Hive.

In order to not make this a flaky list that always needs updating, it would be nice if we can match on regexp values along with perfect string matches.

### Description

Updating the `exclude` argument type to accept strings and RegExps


kongMina:
@kamilkisiela 👋 Would you be able to give this one a review? Thank you!

n1ru4l (approved):

dotansimha:
@kongMina can you please also push a changeset file? (run `pnpm changeset` in the root and it will guide you). I think we should do a `minor` bump for `client` package in this case.

PR #3607, opened 2023-12-13
filter deprecated fields in the explorer view

beerose:
### Background

Fixes https://github.com/kamilkisiela/graphql-hive/issues/1580

With https://github.com/kamilkisiela/graphql-hive/pull/3606 the filters section gets a bit crowded. I currently split it into two rows, but it may need a redesign for a long-term solution (especially since I suppose we may have more filters).

CleanShot 2023-12-13 at 22 55 53@2x

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


kamilkisiela:
In my opinion, we should present deprecated fields and types in a similar fashion to how the unused schema parts are display.
If we do a client-side filter to hide non-deprecated bits, then the page may be empty in some cases. One major issue with this approach is a situation where all root-level fields are non-deprecated and user won't be able to even navigate to deprecated parts of schema.

PR #3711, opened 2023-12-31
fix: error on delete organization

### LINKED ISSUE #3706: Deletion of organization results in error
In case an organization is deleted in the web app, the user is navigated to the same page and thus an error page is shown. Instead, you should be navigated back to main page.

PlayAnyData:
### Background

Fixes #3706 

### Description

Navigates to main page

### Checklist

- [x] Testing


jdolle:
It looks like this change was done https://github.com/graphql-hive/console/blob/hive-metadata/packages/web/app/src/pages/organization-settings.tsx#L591
I believe this can be closed.

PR #3751, opened 2024-01-04
feat: Make schema file path configurable in router registry plugin

thrawny:
### Background

Make it easier to as non root by allowing the user to specify where to download the supergraph schema file from the hive cdn. Typically, /tmp is writable and is a suitable place to download the file to. Alternatively, if using kubernetes, one could create an emptyDir volume and download the schema file to there.

### Description

The change affects the router rust package, specifically the registry module.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/libraries/router/src/registry.rs
+++ b/packages/libraries/router/src/registry.rs
@@ -21,6 +21,7 @@ pub struct HiveRegistryConfig {
     key: Option<String>,
     poll_interval: Option<u64>,
     accept_invalid_certs: Option<bool>,
+    schema_file_path: Option<String>,
 }
 
 static COMMIT: Option<&'static str> = option_env!("GITHUB_SHA");
@@ -32,6 +33,7 @@ impl HiveRegistry {
             key: None,
             poll_interval: None,
             accept_invalid_certs: Some(true),
+            schema_file_path: None
         };
 
         // Pass values from user's config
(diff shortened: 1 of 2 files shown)


packages/libraries/router/src/registry.rs
```diff
@@ -76,6 +79,12 @@ impl HiveRegistry {
             }
         }
 
+        if config.schema_file_path.is_none() {
+            if let Ok(schema_file_path) = env::var("HIVE_CDN_SCHEMA_FILE_PATH") {
```
klippx, line 83: I think you should update the [docs](https://the-guild.dev/graphql/hive/docs/integrations/apollo-router) too!

- [ ] Document new env var `HIVE_CDN_SCHEMA_FILE_PATH`
- [ ] Document new config option `schema_file_path` 
- [ ] Add `schema_file_path` to the commented `router.yaml` - not sure if this is a docs only thing or actually part of the code bootstrapping the config too?

thrawny: schema_file_path cant be configured in router.yaml, since this is not a router plugin

thrawny: did update the docs regarding the new env var

kamilkisiela (approved):

PR #3931, opened 2024-02-07
fix: use latest composable version when FF compareToPreviousComposableVersion is on

kamilkisiela:
It only affects a schema check with no detected changes

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-publisher.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-publisher.ts
@@ -171,6 +171,42 @@ export class SchemaPublisher {
       scope: TargetAccessScope.REGISTRY_READ,
     });
 
+    const target$ = this.targetManager.getTarget({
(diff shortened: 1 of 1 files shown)


packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -180,39 +216,13 @@ export class SchemaPublisher {
...
-        organization: input.organization,
-        project: input.project,
-        target: input.target,
-      }),
-      this.schemaManager.getMaybeLatestValidVersion({
-        organization: input.organization,
-        project: input.project,
-        target: input.target,
-      }),
+      target$,
```
kamilkisiela, line 219: @n1ru4l not sure you like it, but with this amount of function calls (and lines of code) it's hard connect a variable to a promise.

n1ru4l:
closed via https://github.com/kamilkisiela/graphql-hive/pull/3932

PR #3932, opened 2024-02-07
fix: correctly handle schema check composition skip

n1ru4l:
### Background

An error popped up within our error monitoring system regarding this code

### Description

We made the wrong assumption that a skipped schema check is always successful. This resolves it by determining the schema check result based on the schema version we compared against.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-publisher.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-publisher.ts
@@ -606,22 +606,29 @@ export class SchemaPublisher {
         githubSha: githubCheckRun?.commit ?? null,
         expiresAt,
         contextId,
-        contracts:
-          contractVersions?.edges.map(edge => ({
-            contractId: edge.node.contractId,
-            contractName: edge.node.contractName,
-            // TODO: this is not correct, we should compare with the latest valid version...
-            comparedContractVersionId:
-              edge.node.schemaCompositionErrors === null
-                ? edge.node.id
-                : contractVersionIdByContractName.get(edge.node.contractName) ?? null,
-            isSuccess: !!edge.node.schemaCompositionErrors,
-            compositeSchemaSdl: edge.node.compositeSchemaSdl,
-            supergraphSchemaSdl: edge.node.supergraphSdl,
-            schemaCompositionErrors: edge.node.schemaCompositionErrors,
-            breakingSchemaChanges: null,
-            safeSchemaChanges: null,
-          })) ?? null,
+        contracts: contractVersions
+          ? await Promise.all(
+              contractVersions?.edges.map(async edge => ({
+                contractId: edge.node.contractId,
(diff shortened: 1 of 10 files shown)


packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -518,51 +525,7 @@ export class SchemaPublisher {
...
-        }
-
-        if (latestSchemaVersion.compositeSchemaSDL) {
-          composition = {
-            compositeSchemaSDL: latestSchemaVersion.compositeSchemaSDL,
-            supergraphSDL: latestSchemaVersion.supergraphSDL,
-          };
-        } else {
-          // LEGACY CASE if the schema version record has no sdl
-          // -> we need to do manual composition
```
n1ru4l, line 540: this case did not make sense as schema versions always have SDL if composable. If not composable we should not attempt to compose...

I mixed up schema versions with schema checks when I implemented this as legacy schema check records have no SDL saved.

In addition to that I made the mistake of assuming that `SchemaCheckConclusion.Success` that was skipped is always composable...

n1ru4l: actually... we have legacy schema version records without SDL, the above still needs to happen in some cases... 😮‍💨

n1ru4l: handled in https://github.com/kamilkisiela/graphql-hive/pull/3932/commits/396f348fea1aee9cf62eabdb56f626f946db85fd

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -600,13 +565,87 @@ export class SchemaPublisher {
             safeSchemaChanges: contract.schemaChanges?.safe ?? null,
           })) ?? null,
       });
+    } else if (checkResult.conclusion === SchemaCheckConclusion.Skip) {
+      if (!comparedVersion || !comparedSchemaVersion) {
+        throw new Error('This cannot happen :)');
```
kamilkisiela, line 570: "Hey Siri, remind me about this comment in 2 months from now"

n1ru4l: Either this or refactor everything and pull the checksum check out of the orchestrator in order to please TypeScript 😇

dotansimha (approved):
LGTM, but let's wait for @kamilkisiela approval

kamilkisiela (approved):

PR #4009, opened 2024-02-20
added idleTimout=30 for pg connections

dotansimha:
Might fix https://github.com/kamilkisiela/graphql-hive/issues/4008

Suggested in https://github.com/brianc/node-postgres/issues/2512

### DIFF
--- a/packages/services/storage/src/db/pool.ts
+++ b/packages/services/storage/src/db/pool.ts
@@ -16,6 +16,7 @@ export async function getPool(connection: string, maximumPoolSize: number) {
     interceptors: dbInterceptors,
     captureStackTrace: false,
     maximumPoolSize,
+    idleTimeout: 30000,
   });
 
   function interceptError<K extends Exclude<keyof CommonQueryMethods, 'transaction'>>(


packages/services/storage/src/db/pool.ts
```diff
@@ -16,6 +16,7 @@ export async function getPool(connection: string, maximumPoolSize: number) {
     interceptors: dbInterceptors,
     captureStackTrace: false,
     maximumPoolSize,
+    idleTimeout: 30000,
```
kamilkisiela, line 19: Looks like [slonik has `5_000`](https://github.com/gajus/slonik/blob/1228dba2fafcb002a7a78b76864011c08a388453/src/factories/createClientConfiguration.ts#L19) as default value, and [node-postgres has `30_000`](https://github.com/brianc/node-postgres/blob/2a8efbee09a284be12748ed3962bc9b816965e36/packages/pg/lib/defaults.js#L39).

dotansimha: Yeah I'm not sure why `slonik` uses another default 🤔

PR #4017, opened 2024-02-21
Rate limit Mutation.inviteToOrganizationByEmail

kamilkisiela:


packages/services/api/src/modules/organization/resolvers.ts
```diff
@@ -255,6 +256,13 @@ export const resolvers: OrganizationModule.Resolvers = {
       };
     },
     async inviteToOrganizationByEmail(_, { input }, { injector }) {
+      await injector.get(InMemoryRateLimiter).check(
```
n1ru4l, line 259: Instead of raising HiveError we can also use the union type and return an error response

kamilkisiela: I don't think we should, it's not a piece of logic that user in his/her daily workflow

dotansimha (approved):

packages/services/api/src/modules/organization/resolvers.ts
n1ru4l: In that case - should we disclose that the request failed due to rate limit?

n1ru4l: just asking I am fine with merging either way

kamilkisiela: User will see `Exceeded rate limit for inviting to organization by email.`

kamilkisiela: I think it's not different from `x-rate-limit-retry-at` etc so yeah, we should show it

PR #4025, opened 2024-02-22
Auto-fix persisted composite schema sdl

kamilkisiela:
There's a possibility that the composite schema SDL contains parts of the spergraph spec. This is a problem because we want to show the public schema to the user, and the supergraph spec is not part of that. This may happen when composite schema was produced with an old version of transformSupergraphToPublicSchema or when supergraph sdl contained something new.

the-guild-org/federation#44

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-version-helper.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-version-helper.ts
@@ -1,12 +1,18 @@
+import { print } from 'graphql';
 import { Injectable, Scope } from 'graphql-modules';
 import { CriticalityLevel } from '@graphql-inspector/core';
 import type { SchemaChangeType } from '@hive/storage';
+import {
+  containsSupergraphSpec,
+  transformSupergraphToPublicSchema,
+} from '@theguild/federation-composition';
 import { ProjectType } from '../../../shared/entities';
 import { cache } from '../../../shared/helpers';
 import type { SchemaVersion } from '../../../shared/mappers';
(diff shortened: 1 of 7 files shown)


packages/services/api/src/modules/schema/providers/schema-version-helper.ts
```diff
@@ -97,7 +106,9 @@ export class SchemaVersionHelper {
 
   async getSupergraphSdl(schemaVersion: SchemaVersion) {
     if (schemaVersion.hasPersistedSchemaChanges) {
-      return schemaVersion.supergraphSDL;
+      return schemaVersion.supergraphSDL
+        ? this.autoFixCompositeSchemaSdl(schemaVersion.supergraphSDL, schemaVersion.id)
+        : null;
```
kamilkisiela, line 111:
```suggestion
      return schemaVersion.supergraphSDL;
```

packages/services/api/src/modules/schema/providers/schema-version-helper.ts
```diff
@@ -302,4 +311,34 @@ export class SchemaVersionHelper {
...
+      this.logger.debug(
+        transformedSdl === sdl
+          ? 'Transformation did not change the original SDL'
+          : 'Transformation changed the original SDL',
+      );
+
+      return transformedSdl;
+    }
+
+    return sdl;
```
n1ru4l, line 342: this should be outside the if block? We can also flip the condition for early return and avoid indent

PR #4040, opened 2024-02-23
Ignore descriptions when checking native federation compatibility

kamilkisiela:


dotansimha (approved):

n1ru4l:
why does this allow more people to use native federation?

kamilkisiela:
> why does this allow more people to use native federation?

We do `print(sortSDL(sdl))` on two supergraph sdls (native and existing) and compare.
If they are identical, we say 👍🏻, if not do we show 👎🏻 .
If we ignore differences in descriptions, it could show 👍🏻 to some projects that have slightly different descriptions (if you have a description on an object type, the order of services matters - A -> Z).

n1ru4l (approved):

PR #4109, opened 2024-02-28
Dispose hive client on signal termination for yoga and envelop

EmrysMyrddin:
### Background

Today, the client should be disposed the cleanly exit the Node process. Otherwise, the client keep a Timmer hanging in the background, which prevent the process from naturally exit once all other handles are closed.

### Description

This PR aims to workaround this in the Envelop and Yoga plugin to be on par with Apollo Server plugin features. The Apollo Server plugin is actually automatically disposing the client when the Apollo Server is about to stop.

Since Yoga and Envelop can't have this kind of hooks, we rely on node's `process.on(signal, handler)` function.

I'm registering this automatic dispose only when the client is created by the plugin, because we can't be really sure about the use of the client outside of Envelop/Yoga, so it's can unsafe to dispose automatically.

### Alternative

An other solution we can implement is that the client itself registers the signal handler. This way, it will always automatically dispose itself on process exit, whatever integration is used. We just have to make the `dispose` function safe and idempotent so that it's safe to call it twice. This way, if the user already disposed the client manually, it will not be a problem.


kamilkisiela (approved):

EmrysMyrddin:
I've updated for a more universal approach, I think it's better than duplicating the dispose logic everywhere, and the dispose is already pretty much idempotent.

I can also make sure that the we can't run 2 dispose in parallel, which from what I see can probably lead to a race condition if 2 dispose are ran during the same tick.

packages/libraries/client/src/internal/types.ts
```diff
@@ -192,6 +192,12 @@ export type HivePluginOptions = OptionalWhenFalse<
      * Disabled by default
      */
     reporting?: HiveReportingPluginOptions | false;
+    /**
+     * Automatically dispose the client when the process is terminated
+     *
+     * Enabled by default for SIGINT and SIGTERM signals
+     */
+    autoDispose?: Boolean | NodeJS.Signals[];
```
n1ru4l, line 200:
```suggestion
    autoDispose?: boolean | NodeJS.Signals[];
```

packages/libraries/client/src/apollo.ts
```diff
@@ -255,25 +255,10 @@ export function hiveApollo(clientOrOptions: HiveClient | HivePluginOptions): Apo
...
-          },
-        } as any;
-      }
-
       // Works on v3 and v4
-
       return Promise.resolve({
-        async serverWillStop() {
-          await hive.dispose();
-        },
```
n1ru4l, line 276: Removing this seems to be a breaking change?

packages/libraries/client/src/internal/types.ts
```diff
@@ -192,6 +192,12 @@ export type HivePluginOptions = OptionalWhenFalse<
      * Disabled by default
      */
     reporting?: HiveReportingPluginOptions | false;
+    /**
+     * Automatically dispose the client when the process is terminated
+     *
+     * Enabled by default for SIGINT and SIGTERM signals
+     */
```
n1ru4l, line 199: If it is enabled by default, how will this affect non-Node.js runtimes?

packages/libraries/client/src/apollo.ts
kamilkisiela: It is a breaking change

packages/libraries/client/src/client.ts
```diff
@@ -36,6 +36,18 @@ export function createHive(options: HivePluginOptions): HiveClient {
...
+  if (options.autoDispose) {
+    const signals = Array.isArray(options.autoDispose)
+      ? options.autoDispose
+      : ['SIGINT', 'SIGTERM'];
+    for (const signal of signals) {
+      process.once(signal, async () => {
+        logger.info(`[hive] Received ${signal}, disposing.`);
+        await dispose();
+      });
+    }
```
kamilkisiela, line 48: Let's check first if `process` is even available (non-node envs), if it's not available, let's print a warning maybe?

kamilkisiela (changes requested):
Needs few adjustments

packages/libraries/client/src/apollo.ts
kamilkisiela: we shouldn't remove it...

EmrysMyrddin: Is it actually a breaking change ? Since the dispose is now automatically done on process exit ?

kamilkisiela: Could be. It's `serverWillClose` of Apollo Server that controls when the server is about to be closed. It may have extra steps before shutting down the server, like waiting inflight requests to be completed.
That's why I prefer to keep Apollo plugin as close to Apollo as possible in this case, but have autoDispose for Yoga and the rest.

EmrysMyrddin: Ok, let me rollback to the privous implementation then

packages/libraries/client/src/client.ts
EmrysMyrddin: Hummm didn't though about that :/
I'm not very comfortable with the idea of having env specific behaviours like this.

Is it actually a warning ? Because we can't actually detect if the user has made a proper dispose setup or not. Would be terrible to always show the warning even if the user has actually handled this properly ^^'

Perhaps we should not do this and simply make it clear in the documentation that `dispose` *should* be called.

kamilkisiela: This code runs when autodispose === true, so the intent is to auto dispose the client, but if a user tries to do something that is not possible, let's tell him/her about it.

EmrysMyrddin: Yes, but `autodispose` is true by default for now.
Should I make it disabled by default ?
Or enabled by default only of `process` exists ?

kamilkisiela: Maybe we should add "to disable this warning, disable autodispo" in the warning messge?

kamilkisiela: Maybe something like:

> It seems that GraphQL Hive is not being executed in Node.js, which is why the `autodispose` feature is not working as intended. Please attempt manual client disposal and use `autodispose: false`.

kamilkisiela: this way we tell user that `autodispose` is ineffective and there's a chance we notified him/her about an issue, but if the user is handling it manually anyway, it explains how to get rid of the warning.

EmrysMyrddin: Should it be a warning then ? It seems it probably should be an exception at this point ?

kamilkisiela: exception = user's app is broken and it'd be our fault

packages/libraries/client/src/apollo.ts
```diff
@@ -260,7 +260,7 @@ export function hiveApollo(clientOrOptions: HiveClient | HivePluginOptions): Apo
 
       hive.reportSchema({ schema: ctx.schema });
 
-      if (isLegacyV0) {
+      if (isLegacyV0 && hive[autoDisposeSymbol]) {
         return {
           async serverWillStop() {
             await hive.dispose();
```
kamilkisiela, line 266:
```suggestion
      if (isLegacyV0) {
        return {
          async serverWillStop() {
            if(hive[autoDisposeSymbol]) {
              await hive.dispose();
            }
```

kamilkisiela: Just in case, could be that v0.X expects an object back

kamilkisiela (approved):

Issue #3013, opened 2023-10-06
Changelog Feed within hive app


n1ru4l:
On GitHub, there is a list of the latest changes within the UI.
<img width="795" alt="image" src="https://github.com/kamilkisiela/graphql-hive/assets/14338007/f84c4d41-9382-408b-99ce-450115ede817">

It would be great to have something similar within the Hive app. We already expose a list of all changes via the changelog.json file.

kamilkisiela:
@TuvalSimha it would be nice to display it in a format similar to https://headwayapp.co, not GitHub

PR #4149, opened 2024-03-04
feat: Changelog page from GitHub repo

TuvalSimha:
### Background

#3013 

### Description

image

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/web/app/src/components/v2/changelog/changelog.tsx
```diff
@@ -0,0 +1,150 @@
...
+  const [changelogItems, setChangelogItems] = useLocalStorage<Changelog[]>('changelogItems', []);
+  const [changeItemDisplay, setChangeItemDisplay] = useState<Changelog | null>(null);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+        if (changelogItems?.length) {
+          return;
+        }
+        const data = await fetchChangelogsFromGitHub();
```
n1ru4l, line 38: can we instead embed a json file within the app container from the docs and use that? I don't think GitHub is a good source, as the paths could change and GitHub changes could be newer than the self-hosted version you are running.

packages/web/app/src/components/v2/changelog/changelog.tsx
```diff
@@ -0,0 +1,150 @@
+import React, { ReactElement, useEffect, useState } from 'react';
+import { Button } from '@/components/v2';
```
n1ru4l, line 2: We don't use these components anymore, please don't use any component from `@/components/v2`, use `@/components/ui` instead. Also, please don't place any new components within the  `v2` folder. Our final goal is to remove that folder.

packages/web/app/src/components/v2/changelog/changelog.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { useToggle } from '@/lib/hooks';
+import { fetchChangelogsFromGitHub } from './fetch-changelogs-from-github';
+import {
+  Popover,
+  PopoverContent,
+  PopoverFooter,
+  PopoverHeader,
+  PopoverTrigger,
+} from './pop-over-changelog';
+import { useLocalStorage } from './useLocalStorage';
```
n1ru4l, line 12: We already have this hook in `@/lib/hooks`, you can re-use that one

packages/web/app/src/components/v2/changelog/pop-over-changelog.tsx
```diff
@@ -0,0 +1,89 @@
+import * as React from 'react';
+import { cn } from '@/lib/utils';
+import * as PopoverPrimitive from '@radix-ui/react-popover';
```
n1ru4l, line 3: Please use https://ui.shadcn.com/docs/components/popover (`@/components/ui/popover`) instead.

n1ru4l (changes requested):
see my comments

packages/web/app/package.json
```diff
@@ -8,7 +8,10 @@
     "build-storybook": "storybook build",
     "build:config": "tsup-node --no-splitting --out-dir . --loader \".mts=ts\" --format esm --target node21 next.config.mts",
     "dev": "pnpm build:config && next dev | pino-pretty",
+    "generate-changelogs": "pnpm prettier --write ../../../scripts/generate-changelogs.js",
     "postbuild": "pnpm --filter @hive/app deploy --prod --no-optional deploy-tmp/web && rimraf dist/node_modules && mv deploy-tmp/web/node_modules dist && rimraf deploy-tmp/web",
+    "pre:generate-changelogs": "node ../../../scripts/generate-changelogs.js",
+    "prepare": "pnpm run generate-changelogs",
```
n1ru4l, line 14: Instead of using prepare you can add this to the build script

packages/web/app/package.json
```diff
@@ -8,7 +8,10 @@
     "build-storybook": "storybook build",
     "build:config": "tsup-node --no-splitting --out-dir . --loader \".mts=ts\" --format esm --target node21 next.config.mts",
     "dev": "pnpm build:config && next dev | pino-pretty",
+    "generate-changelogs": "pnpm prettier --write ../../../scripts/generate-changelogs.js",
     "postbuild": "pnpm --filter @hive/app deploy --prod --no-optional deploy-tmp/web && rimraf dist/node_modules && mv deploy-tmp/web/node_modules dist && rimraf deploy-tmp/web",
+    "pre:generate-changelogs": "node ../../../scripts/generate-changelogs.js",
```
n1ru4l, line 13: I would group generate-changelogs and pre:generate-changelogs into a single command

packages/web/app/src/components/ui/changelog/changelog.tsx
```diff
@@ -0,0 +1,162 @@
...
+export function Changelog(): ReactElement {
+  const [isOpen, toggle] = useToggle();
+  return <ChangelogPopover isOpen={isOpen} onClose={toggle} />;
+}
+
+function ChangelogPopover({ isOpen, onClose }: { isOpen: boolean; onClose: () => void }) {
+  const [changelogItems, setChangelogItems] = useLocalStorage<Changelog[]>(
+    'changelogItems',
+    latestChangelogs,
+  );
```
n1ru4l, line 27: i manually edited the generated changelog file

image

After refreshing the page the changes did not show up:

image


I suspect this is because the items are always read from local storage.

____

I don't think we need to store the full list in local storage, we can always use the changelog items from the generated file.

Then we only store the d/url of the first item in the changelog items in the local storage.

If the user opens the menu, we set the local storage value to the first item in the list

if the local storage value is not equal to the first item in the generated list, we show a red notification dot

Screenshot 2024-03-14 at 13 25 31

kamilkisiela:
I added a storybook story (?) so we could have a live preview of the component

packages/web/app/.gitignore
```diff
@@ -35,3 +35,6 @@ npm-debug.log*
 
 # Graphql
 lib/graphql-operations.ts
+
+# Changelog
+/packages/web/app/src/components/ui/changelog/generated-changelogs.ts
```
kamilkisiela, line 40:
```suggestion
src/components/ui/changelog/generated-changelogs.ts
```

PR #4206, opened 2024-03-12
feat: use pg_bouncer

saihaj:
Introducing `pg_bouncer` secret in cluster which can be used by services. For now enabled it for
- GraphQL API
- Tokens Service
- Billing Service
- Rate Limit Service

This way if we want to scale these services horizontally we can easily scale without DB pool being a constraint.


deployment/services/postgres.ts
```diff
@@ -33,7 +33,20 @@ export function deployPostgres() {
     ssl: connectionString.apply(connection => (connection.ssl ? '1' : '0')),
   });
 
-  return { secret };
+  const pgBouncerSecret = new PostgresConnectionSecret('postgres-bouncer', {
+    connectionString: rawConnectionString,
```
dotansimha, line 37: Remove this or change the port in this one as well.

deployment/services/postgres.ts
```diff
@@ -33,7 +33,20 @@ export function deployPostgres() {
...
+    connectionStringPostgresql: rawConnectionString.apply(str =>
```
dotansimha, line 38: Remove this or change the port in this one as well.

dotansimha (approved):

PR #4228, opened 2024-03-14
feat: process subscription usage reporting

n1ru4l:
### Background

Since subscriptions work differently than query and mutation operations, we decided to store these in a new clickhouse table and start by only collecting the necessary information for the breaking change detection feature.

Later on, we can also start more subscription-specific data.

### Description

- Seed new clickhouse table `subscription_operations` and materialized view `subscription_operations_daily`
- adjust usage ingestor to insert subscription operations
- adjust breaking change detection clickhouse queries to also use the subscription usage data

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/usage-ingestor/src/processor.ts
```diff
@@ -136,10 +156,29 @@ export function createProcessor(config: { logger: ServiceLogger }) {
...
+          if (operationType === 'subscription') {
+            serializedSubscriptionOperations.push(
+              ...processedOperations.map(record => stringifySubscriptionOperation(record)),
+            );
+          } else {
+            serializedOperations.push(
+              ...processedOperations.map(record => stringifyQueryOrMutationOperation(record)),
+            );
+          }
+        }
```
n1ru4l, line 176: I delayed the stringification until we determined the operation type. We could also from the client send the operation type.

E.g. if we want also to support live queries later on/immediately, these would also be stored within `subscription_operations`.

So, I am tempted to instead send a type with every operation:
- `subscription`: graphql live query, graphql subscription
- `request`: graphql mutation, graphql query

As a legacy fallback, we can always use `request`, as today we don't send any live query/graphql subscription stuff to hive.

packages/services/api/src/modules/operations/providers/operations-reader.ts
```diff
@@ -1127,34 +1143,47 @@ export class OperationsReader {
         queryId: '_getTopOperationsForSchemaCoordinates',
         query: sql`
           WITH "top_operations_by_coordinates" AS (
-            SELECT
+            SELECT DISTINCT
               "coordinates_daily"."coordinate" AS "coordinate",
               "coordinates_daily"."hash" AS "hash",
-              SUM("coordinates_daily"."total") AS "total"
+              "operations_join"."total" AS "total"
             FROM
```
n1ru4l, line 1150: We selected the wrong total count here... This caused bugs where the percentage of traffic is calculated above 100% if two different clients send the same query (identical hash), but one client is excluded from the breaking change detection.

packages/services/usage-ingestor/src/processor.ts
n1ru4l: done via [`2d8fdd0` (#4228)](https://github.com/kamilkisiela/graphql-hive/pull/4228/commits/2d8fdd03a35c01119d27997c3e364cc214e4089d)

n1ru4l:
continued in https://github.com/kamilkisiela/graphql-hive/pull/4228

PR #4241, opened 2024-03-15
Target setting new design 

TuvalSimha:
### Background
This pull request come to change Target design for setting page.

### Video demo
https://github.com/kamilkisiela/graphql-hive/assets/37614975/1675a816-3b48-49eb-8a4a-0ebdda90701f


dotansimha:
I think we need to rethink the sidebar title and icons. Maybe something like that? 

⚙️ General
🔑 Tokens
🧱 Schema Composition

n1ru4l:
If we use icons please icons without color and not built-in emojis that will look completly different based on the OS being used.

As for the seperations, I would even separate CDN access tokens and registry tokens to their own pages, as well as conditional breaking changes and base schemas.

n1ru4l:
I think the yellow underline might be a bit too agressive, we already have a similar menu for the user role management settings on the organisation level, we could mostly re-use these styles.

kamilkisiela:
> **Settings**
> Manage your XYZ
is no longer needed

kamilkisiela:
And let's not use icons

kamilkisiela:
@TuvalSimha you can check the sub-menu I did for members, because we already have such thing in Hive, now we will have 2 differently looking menus that serve the same purpose.

n1ru4l:
image

Let's also separate CDN access tokens and registry tokens to their own pages

https://github.com/kamilkisiela/graphql-hive/pull/4241#issuecomment-2067958416

n1ru4l (changes requested):
see my two comments!

TuvalSimha:
@n1ru4l  @dotansimha @kamilkisiela 
Little bit confused 🤣

Lets make it clear:

(tab section | content page components)
1. General: Target Name Edit + GraphQL Endpoint URL + Delete Target
2. CDN: CDN Access Token
3. Tokens: Registry Access Tokens
4. Conditional Breaking Changes: Conditional Breaking Changes
5. Base Schema: Extend Base Schema

right?

n1ru4l:
@TuvalSimha Let's also drop the "shape"/frame around the single sections:

image

Instead, we can make it look more flat similar to GitHub and separate different sections through headings and spacings like GitHub settings:

image

TuvalSimha:
So i worked on it, created new components for that to make sure all be with the same design.
You can watch the video below

https://github.com/kamilkisiela/graphql-hive/assets/37614975/6d0d2750-3203-45c0-98fd-e0f79dc4d484

Thoughts?

TuvalSimha:
> @dotansimha @kamilkisiela @n1ru4l
> 
> So i worked on it, created new components for that to make sure all be with the same design. You can watch the video below
> 
>  Screen.Recording.2024-05-05.at.17.53.48.mov 
> Thoughts?

Oh, and for this moment every sub page for Members without any main title (see the main title in Target setting) i can add if we want. Also, what about the organization settings design?

packages/web/app/src/components/v2/modals/connect-schema.tsx
```diff
@@ -166,6 +166,9 @@ export const ConnectSchemaModal = (props: {
           <span className="text-sm text-gray-500">
             To authenticate,{' '}
             <Link
+              search={{
+                page: 'general',
```
kamilkisiela, line 170: shouldn't this be `cdn`?

PR #4288, opened 2024-03-19
Move SuperTokens-node to GraphQL server

kamilkisiela:
- [ ] docker
- [ ] integration tests
- [ ] document breaking changes (self-hosting)


packages/services/server/src/environment.ts
```diff
@@ -165,35 +207,28 @@ const LogModel = zod.object({
   ),
 });
 
+// eslint-disable-next-line no-process-env
+const processEnv = process.env;
```
n1ru4l, line 211: genius

PR #4357, opened 2024-03-26
feat: oidc debugging for subscriptions

n1ru4l:
### Background

- We want to use GraphQL subscriptions within Hive
- We want to make OIDC integration debugging easier

### Description

- This adds a basic GraphQL Subscription setup using the GraphQL over SSE protocol

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/context.ts
```diff
@@ -5,7 +5,7 @@ export interface RegistryContext {
   requestId: string;
   user: any;
   headers: Record<string, string | string[] | undefined>;
-  abortSignal: AbortSignal;
+  request: Request;
```
kamilkisiela, line 8: We now have `req` and `request`. Do you think we should merge these two?

packages/services/server/src/graphql-handler.ts
```diff
@@ -268,37 +268,19 @@ export const graphqlHandler = (options: GraphQLHandlerOptions): RouteHandlerMeth
   return async (req, reply) => {
     const requestIdHeader = req.headers['x-request-id'] ?? reqIdGenerate();
     const requestId = cleanRequestId(requestIdHeader);
-    let controller = abortControllerCache.get(req.socket);
```
kamilkisiela, line 271: no abort?

packages/services/api/src/context.ts
n1ru4l: They were in the context before, I only added type for `request: Request;`, so I can re-use it for the abort signal

n1ru4l: I think we should replace all usages of `req` with `request` where possible

kamilkisiela (approved):

packages/services/server/src/index.ts
```diff
@@ -154,6 +158,19 @@ export async function main() {
...
+  const redis = createRedisClient('Redis', env.redis, server.log.child({ source: 'Redis' }));
+
+  const pubSub = createPubSub({
+    eventTarget: createRedisEventTarget({
+      publishClient: redis,
+      subscribeClient: createRedisClient(
+        'subscriber',
+        env.redis,
+        server.log.child({ source: 'RedisSubscribe' }),
+      ),
```
n1ru4l, line 170: Subscriptions need a dedicated client as it blocks the Redis client.

packages/services/server/src/graphql-handler.ts
n1ru4l: we dont need to add our own abortController, we  already have the abortcontroller from whatwg-node/server on the context request object.

n1ru4l: I changed the code for the mutex to use that one instead

PR #4383, opened 2024-03-28
Allow to compose local subgraphs with subgraphs from the registry

### LINKED ISSUE #2750: local composition with CLI
Sometimes when you want to develop locally it makes sense that you want to use the latest valid schema version on CDN, but replace some of the services locally because you develop locally.

The command should allow you to generate the supergraph and/or SDL locally (aka artifacts you would pass to a graphql gateway you run locally).

E.g. the following example would watch/poll `htttp://localhost:3000/graphql`, use the Hive GraphQL API for doing composition then write it to the `supergraph.sdl` file. If there was already a `users` service within the registry on GraphQL Hive it would be completely replaced by the local service.

```bash
hive dev --watch --service users --url http://localhost:3000/graphql  --out supergraph.sdl
```

If you want to replace multiple services locally you could just pass more parameters:

```bash
hive dev --watch \
  --service users --url http://localhost:3000/graphql \
  --service posts --url http://localhost:3001/graphql \
  --out supergraph.sdl
```

After each successful composition we could send a system push notification and/or display the changes on the CLI. 🤔 

**Note**: We cannot do the composition locally as users could configure a wide range of custom external composition services. because of that we always need to do the composition through the API. In order to reduce the amount of network calls we can always check whether our local schemas have changed before doing the call.

**Note**: We cannot serve a gateway ourselves, as we don't know what composition type is used. We can only generate the artifact and write it to a file. Users are then advised to use their gateway of choice for consuming/watching that file.

kamilkisiela:
- [x] GraphQL API
- [x] decide whether we should use latest composable or just latest by default (+ allow user to change the default)
- [x] CLI
- [x] docs

Closes #2750

### DIFF
--- a/packages/web/docs/src/pages/docs/api-reference/cli.mdx
+++ b/packages/web/docs/src/pages/docs/api-reference/cli.mdx
@@ -232,6 +232,10 @@ the gateway.
 ```bash
 # Introspect the SDL of the local service
 hive dev --service reviews --url http://localhost:3001/graphql
+
+# Watch mode
+hive dev --watch --service reviews --url http://localhost:3001/graphql
+
 # Provide the SDL of the local service
 hive dev --service reviews --url http://localhost:3001/graphql --schema reviews.graphql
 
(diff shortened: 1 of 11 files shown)


packages/web/docs/src/pages/docs/api-reference/cli.mdx
```diff
@@ -187,6 +187,63 @@ In case you want to confirm deletion of the service without typing anything in t
 
 </Callout>
 
+### Develop schema locally
+
+<Callout type="info" emoji="🔑">
+  This CLI command requires an active registry token with **Read & Write** permissions to the target
```
kamilkisiela, line 193:
```suggestion
  This CLI command requires an active registry token with **Read** permissions to the target
```

Issue #2492, opened 2023-06-20
GraphQL Hive Client for workers


ilijaNL:
Looking at the current code of @graphql-hive/client, it seems that the client won't work inside an edge (cf) worker? Am I correct, and if so, is it on the roadmap?

The use-case is to use vercel/cf workers for usage reporting.

kamilkisiela:
It would require some adjustments (dispose the client with `context.waitUntil`) but in a matter of a few lines of code you could make it work. Here's [the core logic](https://github.com/kamilkisiela/graphql-hive/blob/ae98aa978536f7df541db7a975b877e3814d3cfe/packages/libraries/client/src/client.ts#L10) that you could use (both envelop and apollo plugins are using it under the hood).

ilijaNL:
I see, my concern was more about using axios as an agent client instead of whatwg fetch for example

ilijaNL:
> I see, yeah it needs to be replaced by something platform agnostic like Fetch API.
> 
> Instead of envelop, you could use our (of [The Guild](https://the-guild.dev/)) [GraphQL Yoga](https://the-guild.dev/graphql/yoga-server/docs/integrations/integration-with-cloudflare-workers).

I don't think this will fix using hive client in workers env since under the hood it still uses the @graphql-hive/client package. 

Anyways I will try to write my own agnostic hive client for the time being

kamilkisiela:
> I see, yeah it needs to be replaced by something platform agnostic like Fetch API.

But that's what I meant, to replace axios with something like Fetch API.

ilijaNL:
I created simple edge client (only usage reporting for now) here: https://www.npmjs.com/package/graphql-hive-edge-client

My use case is to use edge functions as a (security) proxy which does analytics with graphql hive

kamilkisiela:
CF Worker support coming soon #4494

kamilkisiela:
https://the-guild.dev/graphql/hive/product-updates/2024-04-23-graphql-server-integrations-workers-support

PR #4494, opened 2024-04-12
Node agnostic client

### LINKED ISSUE #4336: hive client usage reporting doesn't work with cloudflare workers
Hello, while trying to use hive usage reporting with coudflare, I get the following error : 

```
 (error) [hive][info] Error There is no suitable adapter to dispatch the request since :
- adapter xhr is not supported by the environment
- adapter http is not available in the build
✘ [ERROR]   ReferenceError: global is not defined
```

This is what the handler looks like : 

```ts
const handler = {
  async fetch(request: Request, env, context) {
    const cloudflare = { env, context };
    const yoga =   createYoga({
    schema,
    plugins: [
      useHive({
        token: env.HIVE_TOKEN,
        usage: {
          clientInfo() {
            return {
              name: `${env.ENVIRONMENT}-api`,
              version: '0.0.1'
            };
          }
        }
      })
    ],
    logging: true
  });
    const response = await yoga(request, [{ request, cloudflare }]);
    return response;
  }
} satisfies ExportedHandler<Env>
```

From what I can see it's because the plugin use Axios instead of fetch. 
Is there an available workaround ? Or would that requires a fetch based rewrite ?

kamilkisiela:
- [x] document Node 16+ requirement
- [x] document Yoga.dispose()
- [x] document new `printTokenInfo` option
- [x] document that token info is only available in debug mode or when printTokenInfo is enabled

Fixes #4336 #2492

### DIFF
--- a/packages/libraries/client/README.md
+++ b/packages/libraries/client/README.md
@@ -48,6 +48,35 @@ const server = createYoga({
 server.start()
 ```
 
+### GraphQL Yoga and Cloudflare Workers
+
+If you're using Cloudflare Workers, you can use the following code:
+
+```typescript
+import { useYogaHive, createYogaHive, useYogaHive } from '@graphql-hive/client'
+import { createYoga } from 'graphql-yoga'
+
+export default {
+  async fetch(request, env, ctx) {
+    const hive = createYogaHive({
+      enabled: true, // Enable/Disable Hive Client
+      token: 'YOUR-TOKEN',
+      usage: true // Collects schema usage based on operations
+    });
+
+    const yoga = createYoga({
+      plugins: [
+        useYogaHive(hive)
+      ]
+    });
+
+    const response = await yoga.fetch(request, env, ctx);
+    ctx.waitUntil(hive.dispose());
+    return response;
+  }
+}
+```
+
 ### With Envelop
 
 If you're not familiar with Envelop - in "short" it's a lightweight JavaScript library for wrapping
@@ -150,6 +179,11 @@ app.post(
     }
   })
 )
+
+// When server is shutting down
+async function onShutdown() {
+  await hive.close()
+}
 ```
 
 ### Using the registry when Stitching
--- a/packages/libraries/client/src/internal/http-client.ts
+++ b/packages/libraries/client/src/internal/http-client.ts
@@ -1,0 +1,82 @@
+import { fetch } from '@whatwg-node/fetch';
+
+export function get(
+  endpoint: string,
+  config: {
+    headers: Record<string, string>;
+    timeout?: number;
+    fetchImplementation?: typeof fetch;
+  },
+) {
+  return makeFetchCall(endpoint, {
+    method: 'GET',
+    headers: config.headers,
+    timeout: config.timeout,
+    fetchImplementation: config.fetchImplementation,
+  });
+}
+
+export async function post(
+  endpoint: string,
+  data: string | Buffer,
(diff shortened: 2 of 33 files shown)


packages/libraries/client/src/internal/http-client.ts
```diff
@@ -0,0 +1,77 @@
...
+    timeout?: number;
+  },
+) {
+  const controller = new AbortController();
+  let timeoutId: ReturnType<typeof setTimeout> | undefined = undefined;
+  const responsePromise = fetch(endpoint, {
+    method: config.method,
+    body: config.body,
+    headers: config.headers,
+    signal: controller.signal,
```
ardatan, line 48: You can use `AbortSignal.timeout` instead;
https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static

packages/libraries/client/src/internal/utils.ts
```diff
@@ -1,7 +1,48 @@
...
 import { hiveClientSymbol } from '../client.js';
 import type { HiveClient, HivePluginOptions } from './types.js';
 
+export const isCloudflareWorker = Boolean(
+  typeof caches !== 'undefined' && 'default' in caches && caches.default,
+);
+
+async function digest(algo: 'SHA-256' | 'SHA-1', output: 'hex' | 'base64', data: string) {
+  if (typeof crypto !== 'undefined' && typeof TextEncoder !== 'undefined') {
+    const buffer = await crypto.subtle.digest(algo, new TextEncoder().encode(data));
```
ardatan, line 10: We have this in `@whatwg-node/fetch`

kamilkisiela: super weird to use WebCrypto API from `whatwg-node/fetch` :))

packages/libraries/client/src/internal/http-client.ts
kamilkisiela: It was added in v16.14.0 so I cannot really use it if I want to support Node v16

packages/libraries/client/src/internal/utils.ts
ardatan: But we already have a dependency here :)
Also some bundlers can complain about the dynamic import below;
https://github.com/kamilkisiela/graphql-hive/pull/4494/files/840154d858b4708d6efb9d7bb1c67ae0ecf6eaa9#diff-1920db0a7d3df2907cd2cc1ed61b516ded3145e21a7487384f60b5d17854d10fR18

packages/libraries/client/README.md
```diff
@@ -48,6 +48,35 @@ const server = createYoga({
...
+
+If you're using Cloudflare Workers, you can use the following code:
+
+```typescript
+import { useYogaHive, createYogaHive, useYogaHive } from '@graphql-hive/client'
+import { createYoga } from 'graphql-yoga'
+
+const hive = createYogaHive({
+  enabled: true, // Enable/Disable Hive Client
+  token: 'YOUR-TOKEN',
```
Hebilicious, line 61: For cloudflare workers this needs to happen within the fetch handler as the env variables are not available globally. Could you consider an alternative API to pass the token down ? Maybe try to read it automatically from the cloudflare env assuming a specific variable name ?

kamilkisiela: You can run `createYogaHive` within fetch, it has no overhead really

Issue #4229, opened 2024-03-14
date ranger filter: suggest custom number interval


n1ru4l:
we have predefined (fixed) presets within the date range picker.

![image](https://github.com/kamilkisiela/graphql-hive/assets/14338007/ee597f73-e33d-4f74-ac4f-ace7771539bd)

In case we put in a number, we should also generate on-the-fly possible options for the given time range.

E.g if you enter `last 4`, we could automatically display:
- last 4 minutes
- last 4 hours
- last 4 days
- last 4 months

Sentry does this for example:
![image](https://github.com/kamilkisiela/graphql-hive/assets/14338007/5c561475-4cc8-471b-a96c-02e2dc3b46f6)

From a technical perspective we can just check whether the input string starts with `last <number>`, and then automatically add the ranges (minutes, hours, days, weeks, month) based on the user retention range.

dotansimha:
done in https://github.com/kamilkisiela/graphql-hive/pull/4594

PR #4594, opened 2024-04-24
new feature: date ranger filter suggest custom number interval

TuvalSimha:
### Background
Task: #4229  

### Description
Enhance DateRangePicker to dynamically suggest time presets while respecting valid time units specified by the user.

### Live Video
https://github.com/kamilkisiela/graphql-hive/assets/37614975/faee8145-23dd-45a4-9b55-5311cd560a2d

### Note
The list of suggestions based on the `dynamicPresets` prop.

### DIFF
--- a/packages/web/app/src/components/ui/date-range-picker.tsx
+++ b/packages/web/app/src/components/ui/date-range-picker.tsx
@@ -270,7 +270,6 @@ export function DateRangePicker(props: DateRangePickerProps): JSX.Element {
     } else {
       setDynamicPresets([]);
     }
-
   }, [quickRangeFilter, validUnits]);
 
   presets = [...presets, ...dynamicPresets];
--- a/packages/web/app/src/components/ui/date-range-picker.tsx
+++ b/packages/web/app/src/components/ui/date-range-picker.tsx
@@ -169,6 +169,7 @@ export function DateRangePicker(props: DateRangePickerProps): JSX.Element {
   const [toValue, setToValue] = useState(activePreset?.range.to ?? '');
   const [range, setRange] = useState<DateRange | undefined>(undefined);
   const [quickRangeFilter, setQuickRangeFilter] = useState('');
+  console.log('quickRangeFilter', quickRangeFilter);
 
   const fromParsed = parse(fromValue);
   const toParsed = parse(toValue);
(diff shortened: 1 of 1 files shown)


packages/web/app/src/components/ui/date-range-picker.tsx
```diff
@@ -231,6 +231,49 @@ export function DateRangePicker(props: DateRangePickerProps): JSX.Element {
     );
   };
 
+  // Dynamic presets
```
dotansimha, line 234: useless comment? 😆

dotansimha: TIP: It can be something like this: 
// The following code is suggesting the users filter options based on the input, when the input starts with "last" keyword.

packages/web/app/src/components/ui/date-range-picker.tsx
```diff
@@ -231,6 +231,49 @@ export function DateRangePicker(props: DateRangePickerProps): JSX.Element {
...
+        },
+        {
+          name: `last${number}w`,
+          label: `Last ${number} weeks`,
+          range: { from: `now-${number}w`, to: 'now' },
+        },
+      ];
+
+      // Filter out dynamic presets that use invalid units
+      const validDynamicPresets = dynamicPresets.filter(preset => {
```
dotansimha, line 263: I wonder why we could end up with invalid values? 🤔

packages/web/app/src/components/ui/date-range-picker.tsx
```diff
@@ -231,6 +231,49 @@ export function DateRangePicker(props: DateRangePickerProps): JSX.Element {
     );
   };
 
+  // Dynamic presets
+  const [dynamicPresets, setDynamicPresets] = useState<Preset[]>([]);
+  useEffect(() => {
+    if (quickRangeFilter.trim().match(/^Last [0-9]+$/)) {
```
dotansimha, line 237: maybe also make it case-insensitive? so `last` also works?

kamilkisiela: we need to cover a, where somebody types `3` and gets `Last 3 hours / days / weeks` etc.

Basically a similar experience to the one in [Sentry](https://the-guild-z4.sentry.io/issues/?project=6716298&statsPeriod=30d&utc=true)

TuvalSimha:
Fixed and improved the filter, here a quick video from the behavior:

https://github.com/kamilkisiela/graphql-hive/assets/37614975/04822e5d-2801-47c9-94ba-badcee55aae7

What do you think?

kamilkisiela:
Amazing @TuvalSimha I <3 it!

I suggest to add sorting to fix this issue:
Screenshot 2024-04-30 at 10 47 54and when a user types `1` it needs to show `Last 1 hour/day/month/year` only, not `Last 14 days`

once we have that, let's merge it

dotansimha (approved):

n1ru4l (approved):

PR #4612, opened 2024-04-29
Replace Next with Vite + tanstack/router

kamilkisiela:
## Build

- `pnpm deploy` is no longer needed
- `vite` builds static files, `runify` builds a server code to serve static files and handle api routes.
- `turbo` can now cache `pnpm build` of `@hive/app`

## Development

- `pnpm dev` gives us hot reloading and in general similar experience to what we had with Next
- The HTTP server framework of choice is `fastify` as it has a good integration with `vite` (`@fastify/vite`, there was `vite-express` but we already use `fastify` in Hive, so it was an obvious choice). The `@fastify/vite` is only used in development node.

## Production

- files generated by `vite build` are served by `@fastify/static`
- api routes are served by `fastify`

## Docker

From now on, we serve `@hive/app` in the same fashion as other services, no dedicated docker file or configuration. The size of a docker image is now 4x lower than the previous one.

## Routing

We moved from file-based to code-base routing and all routes are in `router.tsx` file. I hope you like it :)

All pages are in `src/page` directory, stored as flat files. To be honest, we could have a hierarchy there driven by directory structure, but I prefer flat files as it's easier to navigate and we avoid doing `../../../../` shit.

Thanks to `@tanstack/router`, all pieces of routing are type-safe, even params, you will love it.

I also removed `useRouteSelector` hook that gave us `{organization|project|target}Id` props based on an URL. We simply pass these using props.

## Runtime Environment Variables

To still be able to build the app, ship it as a Docker image and pass env vars when running the docker container, we handle `GET /__env.js` in `fastify`, instead of generating a file.

---

- [x] fix <Helmet><title> at target level
- [x] links to docs are not really pointing to docs


dotansimha (approved):

docker/app.dockerfile
```diff
@@ -1,24 +0,0 @@
-FROM node:21.7.3-slim
```
dotansimha, line 1: you can also consider using a simple HTTP server to serve the artifact, right? 
something like `busybox` (https://lipanski.com/posts/smallest-docker-image-static-website)

kamilkisiela: we could move them to `graphql-api` and serve from there, but I did not want to do it in this PR

TuvalSimha (changes requested):
left one comment about router and how to handle delete organization when user have more then 1 organizations

packages/web/app/src/components/v2/modals/delete-organization.tsx
```diff
@@ -61,10 +58,12 @@ export const DeleteOrganizationModal = ({
           onClick={async () => {
             await mutate({
               selector: {
-                organization: router.organizationId,
+                organization: props.organizationId,
               },
             });
-            void (await replace('/'));
+            void router.navigate({
```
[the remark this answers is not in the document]
kamilkisiela: hmmm it's redirecting to `/`.
When you enter `/`, it will find an organization you visited previously or redirect you to one of them.

PR #4695, opened 2024-05-07
Date range component: fix sorting and storybook

### LINKED ISSUE #4696: Date range component: fix order in search input

TuvalSimha:
### Background

Date range component sorting was confusing, this PR come to sort the range based on weight.

closes #4696

### DIFF
--- a/packages/web/app/src/components/ui/date-range-picker.tsx
+++ b/packages/web/app/src/components/ui/date-range-picker.tsx
@@ -64,6 +64,16 @@ function resolveRange(rawFrom: string, rawTo: string): ResolvedDateRange | null 
   return null;
 }
 
+function calculateWeight(preset: Preset): number {
+  const from = parse(preset.range.from);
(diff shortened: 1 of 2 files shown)


packages/web/app/src/components/ui/date-range-picker.tsx
```diff
@@ -273,13 +273,32 @@ export function DateRangePicker(props: DateRangePickerProps): JSX.Element {
   }, [quickRangeFilter, validUnits]);
 
   presets = [...presets, ...dynamicPresets].sort((a, b) => {
-    if (a.label < b.label) {
-      return -1;
+    const aWeight = a.label.includes('minutes')
```
dotansimha, line 276: hmm, I guess you can move this logic of calculating weight into something else? 

Also, maybe the ideal way to understand the "weight" is to measure how many minutes you have between now and the result date?

For example:
"last 10 minutes" = 10 
"last 2 days" =  2880

PR #4740, opened 2024-05-12
[DO NOT MERGE] Proof Of Concept - Migrating to Server Preset

eddeee888:
### Background

This is a Proof Of Concept for gradual migration to Server Preset (`@eddeee888/gcg-typescript-resolver-files`) from `@graphql-codegen/graphql-modules-preset`


codegen.mts
```diff
@@ -1,10 +1,143 @@
...
+            '../modules/schema/providers/contracts#ContractCheck as ContractCheckMapper',
+          ContractVersion:
+            '../modules/schema/providers/contracts#ContractVersion as ContractVersionMapper',
+          BreakingChangeMetadataTarget:
+            '../shared/mappers#BreakingChangeMetadataTarget as BreakingChangeMetadataTargetMapper',
+        },
+      },
+    }),
     // API
     './packages/services/api/src/modules': {
```
eddeee888, line 142: We still keep this until we finish the migration to avoid blocking features to be developed

codegen.mts
```diff
@@ -1,10 +1,143 @@
...
+      resolverMainFileMode: 'modules',
+      resolverTypesPath: './__generated__/types.next.ts',
+      whitelistedModules: ['usage-estimation', 'admin'],
+      scalarsOverrides: {
+        DateTime: { type: 'string' },
+        Date: { type: 'string' },
+        SafeInt: { type: 'number' },
+        ID: { type: 'string' },
+      },
+      typesPluginsConfig: {
```
eddeee888, line 26: This is kept the same as modules preset to keep the two generated versions the same

codegen.mts
```diff
@@ -1,10 +1,143 @@
...
+    // Server preset
+    './packages/services/api/src': defineConfig({
+      add: {
+        './__generated__/types.next.ts': {
+          content: "import type { StripeTypes } from '@hive/stripe-billing';",
+        },
+      },
+      typeDefsFilePath: false,
+      resolverMainFileMode: 'modules',
+      resolverTypesPath: './__generated__/types.next.ts',
```
eddeee888, line 18: We are making the Server Preset files use `types.next.ts` until the migration is done.

codegen.mts
```diff
@@ -1,10 +1,143 @@
...
+      whitelistedModules: ['usage-estimation', 'admin'],
```
eddeee888, line 19: We use `whitelistedModules` in this POC but using `blacklistedModules` is a bit more explicit. Using `blacklistedModules` also allows new modules to use Server Preset so we don't have to migrate them later.

packages/services/api/src/modules/usage-estimation/resolvers/UsageEstimation.ts
```diff
@@ -0,0 +1,5 @@
+import type { UsageEstimationResolvers } from './../../../__generated__/types.next';
+
+export const UsageEstimation: UsageEstimationResolvers = {
+  /* Implement UsageEstimation resolver logic here */
+};
```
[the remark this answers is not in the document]
eddeee888: That's true! It's common feedback I've been getting.
I'm working on a mode to avoid generating empty object type files like this. Will release that as a feature first before starting the migration 🙂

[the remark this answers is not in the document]
eddeee888: I've made an alpha release to avoid these empty files [in this commit](https://github.com/kamilkisiela/graphql-hive/pull/4740/commits/48c60d5e72423c163011eab48d16f089d00096de).

Now, only a small number of files are created:
- query 
- mutation 
- subscriptions 
- object types with mappers

If an object type _without_ a mapper needs to be created to have field resolvers, it can be created at the expected location on filesystem and it'd be prefilled and automatically wired up to `resolvers.generated.ts`:

https://github.com/kamilkisiela/graphql-hive/assets/33769523/f62e3b9a-fc23-4d53-aed3-ef3688aa4e8c

Issue #3468, opened 2023-11-27
schema page with multiple services UX is confusing


beerose:
Maybe we can show a list of services on the left (similar UI as with schema checks)? It got me confused for a second as I kind of thought that my other service was removed when publishing a new one (I didn't expect it to be shown one after another 😅). Especially that code mirror takes a lot of space. 

![CleanShot 2023-11-27 at 17 39 15@2x](https://github.com/kamilkisiela/graphql-hive/assets/9019397/c07f83b5-b292-4db8-8c1d-bc765d9fb768)

capaj:
@beerose I am not familiar with the app yet-can you post a screenshot of the schema checks UI you're mentioning?

PR #4742, opened 2024-05-12
Search list for schema page with multiple services

TuvalSimha:
### Background
This pull request come from #3468, the search input in schema page is bit confusing. So, we tried to make it more clear and user friendly when user have multiple services.

### Description
- Search input was changed to input with list on value changed.
- List of result should be more user friendly when have multiple services.
- Added Storybook stories.

### Before
https://github.com/kamilkisiela/graphql-hive/assets/37614975/e25cf003-1d05-4d0e-a5e9-f8a997d8d3c0

### After
https://github.com/kamilkisiela/graphql-hive/assets/37614975/b6862e06-ef9e-497a-82b5-e0e068e9eaef


dotansimha:
The `Search and select schema` label seems a bit weird for me, because it makes me feel like something will happen when I select an option. 
I think it should be "filter schemas" or something like that.

TuvalSimha:
> filter schemas

@dotansimha 
Fixed :) i changed it to `Filter schemas`

kamilkisiela (changes requested):
When a user searches for a subgraph and clicks on one of the options, the schema element should not be collapsed.

Another question, how does the list look like when there are 3 subgraphs and when there are 10?

TuvalSimha:
@kamilkisiela 
So i did rebase, fix conflicts and fixed some issues you mentioned. You can watch the video attachment and see the behaviour.

**3 schemas :**
https://github.com/kamilkisiela/graphql-hive/assets/37614975/05ba5859-b3aa-4206-8076-fe58741d4c6c

**10 schemas**
https://github.com/kamilkisiela/graphql-hive/assets/37614975/6424eb0a-6282-4984-8a09-a5355090ea73

n1ru4l (approved):
Approved!

Only nitpick is that "Filter schema" should probably be "Find Schema" as filter implies I can select n schemas to display?

dotansimha (approved):

TuvalSimha:
> Find Schema

i will change it :) @n1ru4l

TuvalSimha:
> Approved!
> 
> Only nitpick is that "Filter schema" should probably be "Find Schema" as filter implies I can select n schemas to display?

i will change it :) @n1ru4l

kamilkisiela:
Let's make subgraph schemas shareable. We could move up the state of the schema picker to the url (?service=XYZ).
We can do it in a separate Pull Request.

TuvalSimha:
> Let's make subgraph schemas shareable. We could move up the state of the schema picker to the url (?service=XYZ).
> We can do it in a separate Pull Request.

https://github.com/kamilkisiela/graphql-hive/issues/4793

kamilkisiela (approved):

PR #4805, opened 2024-05-22
reduce complexity

kamilkisiela:
**Service not found**

Screenshot 2024-05-22 at 13 50 29

**Keeping the state only in the url**

https://github.com/kamilkisiela/graphql-hive/assets/8167190/453d8126-f268-4188-ac85-c66e52875fd7

**View of a single schema**

Screenshot 2024-05-22 at 13 46 22

**View of a Federation project with a single schema**

No combobox as there's nothing to pick from.

Screenshot 2024-05-22 at 13 46 12

### DIFF
--- a/packages/web/app/src/pages/target.tsx
+++ b/packages/web/app/src/pages/target.tsx
@@ -304,7 +304,6 @@ const TargetSchemaPageQuery = graphql(`
 `);
 
 function TargetSchemaPage(props: { organizationId: string; projectId: string; targetId: string }) {
-  const router = useRouter();
   const [query] = useQuery({
     query: TargetSchemaPageQuery,
     variables: {
--- a/packages/web/app/src/pages/target.tsx
+++ b/packages/web/app/src/pages/target.tsx
@@ -1,7 +1,6 @@
-import { ChangeEventHandler, ReactElement, useCallback, useEffect, useRef, useState } from 'react';
-import { Check, ChevronsUpDown } from 'lucide-react';
+import { ReactElement, useState } from 'react';
+import { ChevronsUpDown, XIcon } from 'lucide-react';
 import { useQuery } from 'urql';
-import { useDebouncedCallback } from 'use-debounce';
 import { Page, TargetLayout } from '@/components/layouts/target';
 import { MarkAsValid } from '@/components/target/history/MarkAsValid';
 import { Button } from '@/components/ui/button';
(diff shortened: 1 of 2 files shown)


packages/web/app/src/pages/target.tsx
```diff
@@ -209,41 +164,23 @@ function SchemaView(props: {
...
   project: FragmentType<typeof SchemaView_ProjectFragment>;
   target: FragmentType<typeof SchemaView_TargetFragment>;
-  highlightedService: string | null;
 }): ReactElement | null {
   const organization = useFragment(SchemaView_OrganizationFragment, props.organization);
   const project = useFragment(SchemaView_ProjectFragment, props.project);
   const target = useFragment(SchemaView_TargetFragment, props.target);
-  const [filterService, setFilterService] = useState(props.highlightedService ?? '');
-  const [term, setTerm] = useState(props.highlightedService ?? '');
-  const [openItems, setOpenItems] = useState<string[]>([]);
```
kamilkisiela, line 219: There's no point of keeping open items in a state. The accordion does it for us.
The `filterService` is not really used.
The `term` is also not needed as combobox is handling the search input for us.

packages/web/app/src/pages/target.tsx
```diff
@@ -209,41 +164,23 @@ function SchemaView(props: {
...
-      setTerm(value);
-      setOpen(false);
-      setOpenItems(prevItems => [...new Set([...prevItems, value])]);
-      void router.navigate({
-        to: '/$organizationId/$projectId/$targetId',
-        search: { service: value },
-      });
-    },
-    [debouncedFilter, setTerm],
-  );
```
kamilkisiela, line 238: Because of the state I deleted, we have much simpler component

packages/web/app/src/pages/target.tsx
```diff
@@ -209,41 +164,23 @@ function SchemaView(props: {
...
+      ? router.latestLocation.search.service
+      : null;
+
+  const [open, setOpen] = useState(false);
+  const reset = () => {
     void router.navigate({
-      to: '/$organizationId/$projectId/$targetId',
+      search: {},
     });
-  }, [setFilterService]);
```
kamilkisiela, line 246: What `reset` now does is just updating the url, so our component re-renders and since we have no state, then there's no need to reset anything else.

packages/web/app/src/pages/target.tsx
```diff
@@ -266,110 +203,86 @@ function SchemaView(props: {
...
+  const singleSchema = schemas?.filter(schema => !isCompositeSchema(schema))[0] as
+    | SingleSchema
+    | undefined;
+  const schemasToDisplay = selectedServiceName
+    ? compositeSchemas.filter(schema => schema.service === selectedServiceName)
+    : compositeSchemas;
 
   return (
     <>
-      {showExtra ? (
```
kamilkisiela, line 293: this condition broke the UI. If a user has no access to click "Mark as valid" button then he also cannot see the combobox.

PR #4831, opened 2024-05-28
Set up server preset and migrate `usage-estimation` module

eddeee888:
### Background

This PR sets up Server Preset and migrates `usage-estimation` module.
With this setup, new modules will be using Server Preset.


codegen.mts
```diff
@@ -6,6 +7,62 @@ const config: CodegenConfig = {
...
+          'organization',
+          'policy',
+          'project',
+          'rate-limit',
+          'schema',
+          'shared',
+          'support',
+          'target',
+          'token',
+        ],
```
eddeee888, line 41: I will go through this list one or a few modules at a time. 
This PR starts with `usage-estimation` to test it out.

packages/services/api/src/modules/usage-estimation/resolvers/Query/usageEstimation.ts
```diff
@@ -0,0 +1,37 @@
...
+    year: args.input.year,
+  });
+
+  if (!result && result !== 0) {
+    throw new GraphQLError(`Failed to estimate usage, please try again later.`);
+  }
+
+  return {
+    operations: result,
+  };
```
eddeee888, line 36: This is just copy/pasted from the old resolver. I've tested the logic in dev.

n1ru4l (approved):

PR #4937, opened 2024-06-07
Migrate `token` module to Server Preset (Part 2)

eddeee888:
### Background

This PR migrates `token` module to use Server Preset.


codegen.mts
```diff
@@ -57,6 +56,11 @@ const config: CodegenConfig = {
             SupportTicketPriority: '../shared/entities#SupportTicketPriority',
             SupportTicketStatus: '../shared/entities#SupportTicketStatus',
           },
+          resolversNonOptionalTypename: {
+            interfaceImplementingType: true,
+            unionMember: true,
+            excludeTypes: ['TokenInfoPayload'],
```
eddeee888, line 62: This config is required because we are using [__isTypeOf in TokenInfo](https://github.com/kamilkisiela/graphql-hive/pull/4937/files#diff-90d57ac659bf1029c0543323045fe5202f300c6cdfffd232ea43f6a43d4cfecfR7-R9). Server Preset's default is using `__typename` but since all 3 options work and this is a migration, I want to keep logic the same

packages/services/api/src/modules/token/resolvers.generated.ts
```diff
@@ -0,0 +1,20 @@
+/* This file was automatically generated. DO NOT UPDATE MANUALLY. */
```
dotansimha, line 1: I think maybe this one can be removed from source control?

eddeee888: Yes! This is done in https://github.com/kamilkisiela/graphql-hive/pull/4937/commits/051a7ffe8d145f9b1785096981132c67c38a6957

PR #4955, opened 2024-06-08
Migrate `shared` module to Server Preset (Part 4)

eddeee888:
### Background

This PR migrates `shared` module to Server Preset

TODO:

- [x] Merge previous PR: https://github.com/kamilkisiela/graphql-hive/pull/4938
- [x] Release new Server Preset version: https://github.com/eddeee888/graphql-code-generator-plugins/pull/291


packages/services/api/src/modules/shared/resolvers/JSON.ts
```diff
@@ -0,0 +1,7 @@
+import { JSONResolver } from 'graphql-scalars';
+
+// `scalar JSON` in `module.graphql.ts` does not have a description
+// and it messes up the static analysis
+JSONResolver.description = undefined;
```
eddeee888, line 5: This comment only seems to be applicable to `JSONResolver` so I only left it here, and not on other scalar resolvers e.g. `DateTime`, `Date`, etc.

I'm happy to repeat the comment in other places it required 🙂

packages/services/api/src/modules/rate-limit/resolvers.generated.ts
eddeee888, line 1: `resolvers.generated.ts` files are now ignored, these are remnants of PRs before they are ignored

n1ru4l (approved):

PR #4956, opened 2024-06-10
Migrate `support` module to Server Preset (Part 5)

eddeee888:
### Background

This PR migrates `support` module to Server Preset

TODO

- [x] Merge part 4: https://github.com/kamilkisiela/graphql-hive/pull/4955
- [x] Merge https://github.com/eddeee888/graphql-code-generator-plugins/pull/299 and release version that supports enum handling
- [x] Merge this PR


packages/services/api/src/modules/rate-limit/resolvers.generated.ts
```diff
@@ -1,7 +1,7 @@
 /* This file was automatically generated. DO NOT UPDATE MANUALLY. */
 import type { Resolvers } from './../../__generated__/types.next';
-import { Organization } from './resolvers/Organization';
+import { Organization as rate_limit_Organization } from './resolvers/Organization';
```
eddeee888, line 3: This rename is expected because other modules are extending `type Organization`, so server preset adds module name to distinguish them, even if they are not all in one file

codegen.mts
```diff
@@ -51,8 +50,6 @@ const config: CodegenConfig = {
             ProjectAccessScope: '../modules/auth/providers/project-access#ProjectAccessScope',
             OrganizationAccessScope:
               '../modules/auth/providers/organization-access#OrganizationAccessScope',
-            SupportTicketPriority: '../shared/entities#SupportTicketPriority',
-            SupportTicketStatus: '../shared/entities#SupportTicketStatus',
```
eddeee888, line 55: We are using `enumValues` here, however, I've converted it to use `mappers` since Server Preset already handles mapper logic well enough.

Here's how we add these enums as mappers: https://github.com/kamilkisiela/graphql-hive/pull/4956/files#diff-4b4509fdcb8380696d4a5f77ddebf8d2f6678de05cb598c51edc1f1eb928ff2cR3-R4

packages/services/api/src/modules/support/resolvers/SupportTicket.ts
```diff
@@ -0,0 +1,26 @@
...
+        cursor: String(comment.id),
+      })),
+      pageInfo: {
+        endCursor: String(response.nodes[response.nodes.length - 1]?.id ?? ''),
+        hasNextPage: response.meta.has_more,
+        hasPreviousPage: false,
+        startCursor: String(response.nodes[0]?.id ?? ''),
+      },
+    };
+  },
```
eddeee888, line 25: The schema for this type is this:

```graphql
comments: SupportTicketCommentConnection
```

This seems a bit different from other `*Connection` as they are usually non-nullable i.e. `*Connection!`. 
Looking at the implementation, it can never be null. So I think this could be a way to get around having to declare mapper for `SupportTicket`? 🤔 

Either way, this is part of the migration so we should keep the implementation as-is, however, I think we should change the type to `SupportTicketCommentConnection!` so it correctly reflects the logic. 

Unless I misunderstood it, then please let me know!

codegen.mts
eddeee888: Actually, this is still needed 😅

packages/services/api/src/modules/support/resolvers/SupportTicket.ts
n1ru4l: @kamilkisiela Did you think about anything specific when u implemented this?

n1ru4l (approved):

PR #4962, opened 2024-06-12
[eslint] disallow duplicate unions

dimaMachina:


.eslintrc.cjs
```diff
@@ -95,6 +96,7 @@ module.exports = {
       plugins: [...guildConfig.plugins, 'hive'],
       extends: guildConfig.extends,
       rules: {
+        '@typescript-eslint/no-duplicate-type-constituents': 'error',
```
kamilkisiela, line 99:
```suggestion
```

kamilkisiela:
I don't think we should eslint integration tests

PR #4972, opened 2024-06-13
remove more unused v2 components, fix `href="emailto:` with `href="mailto:`

dimaMachina:


packages/web/app/src/components/error.tsx
```diff
@@ -14,7 +14,7 @@ export function ErrorComponent(props: { error: any }) {
     void flush(2000);
   }, []);
 
-  const isLoggedIn = (session.loading === false && session?.doesSessionExist === true) || false;
+  const isLoggedIn = !session.loading && session.doesSessionExist;
```
dimaMachina, line 17: simplify condition, no need for optional chaining too

PR #4974, opened 2024-06-13
[eslint] disallow unused collections, enable `no-inverted-boolean-check` rule

dimaMachina:


kamilkisiela:
I'm not fine with adding more eslint rules to this codebase, we should aim to have the least amount possible and enable only those that have a real impact on users (like preventing bugs). This one is just about the "style" of code and I honestly don't care.

dimaMachina:
`sonarjs/no-unused-collection` preventing mistakes when a collection is populated but its contents are never used

kamilkisiela:
We can do a deal :D 
I accept `sonarjs/no-unused-collection` but you remove `sonarjs/no-inverted-boolean-check` :D

.eslintrc.cjs
```diff
@@ -122,6 +122,8 @@ module.exports = {
           },
         ],
         '@typescript-eslint/no-floating-promises': 'error',
+        'sonarjs/no-unused-collection': 'error',
+        'sonarjs/no-inverted-boolean-check': 'error',
```
kamilkisiela, line 126: Make it a warning and I'll accept it. Unused code should not be an error.

kamilkisiela (changes requested):

PR #4975, opened 2024-06-13
[eslint] prefer object dot notation

dimaMachina:


packages/web/app/src/env/read.ts
```diff
@@ -1,13 +1,10 @@
-function isBrowser() {
-  // eslint-disable-next-line no-restricted-syntax
-  return Boolean(
-    typeof window !== 'undefined' && '__ENV' in window && window['__ENV'] !== undefined,
-  );
+function isBrowser(): boolean {
+  return typeof window !== 'undefined' && '__ENV' in window && window.__ENV !== undefined;
```
dimaMachina, line 2: unneeded cast to `boolean`

kamilkisiela:
I will merge it if you give me two arguments how it affects Hive users :)

dimaMachina:
- is easier to read
- less verbose
- works better with aggressive JavaScript minimizers

kamilkisiela:
> is easier to read
> less verbose

just a personal preference

> works better with aggressive JavaScript minimizers
we don't really care about 20b improvement in bundle size

not enough to add new eslint rules, I'm sorry

PR #4978, opened 2024-06-14
Fix constantly growing \200b backslash escaping

kamilkisiela:
Every time prettier runs, the amount of \ doubles

```css
after:content-['\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\200b']
```


capaj:
Yeah I got that too, nicely fixed

PR #4990, opened 2024-06-18
Audit log proposal wip

capaj:
just for review purposes, don't merge

### DIFF
--- a/docs/proposals/proposal-audit-log.md
+++ b/docs/proposals/proposal-audit-log.md
@@ -35,21 +35,6 @@ schemas, and other resources.
 - `deleteProject`
 - `updateName`
 
-### ArtifactStorageWriter
-
-- `writeArtifact`
-- `deleteArtifact`
-
-### CompositeLegacyModel
-
-- `check`
-- `publish`
-
-### SingleLegacyModel
-
-- `check`
-- `publish`
-
 ### TargetManager
 
 - `createTarget`
@@ -62,9 +47,7 @@ schemas, and other resources.
 
 ### SchemaPublisher
 
-- `updateVersionStatus`
 - `delete`
-- `publishToCDN`
 
 ### SchemaManager
 
@@ -101,9 +84,6 @@ schemas, and other resources.
 
 - `register`
 - `unregister`
-- `createCheckRun`
-- `updateCheckRun`
-- `updateCheckRunToSuccess`
 - `enableProjectNameInGithubCheck`
 
 ### SlackIntegrationManager
@@ -117,16 +97,10 @@ schemas, and other resources.
 - `updateOIDCIntegration`
 - `deleteOIDCIntegration`
 
-### ContractsManager
-
-- `createContract`
-- `disableContract`
-
 ### Contracts
 
 - `createContract`
 - `disableContract`
-- `approveContractChecksForSchemaCheckId`
 
 ### AlertsManager
 
@@ -135,10 +109,46 @@ schemas, and other resources.
 - `addAlert`
 - `deleteAlerts`
 
+## Storage
+
+- updateUser
+- updateOrganizationName
+- updateOrganizationPlan
+- updateOrganizationRateLimits
+- createOrganizationInvitation
+- deleteOrganizationInvitationByEmail
+- createOrganizationTransferRequest
+- answerOrganizationTransferRequest
+- addOrganizationMemberViaInvitationCode
+- deleteOrganizationMember
+- updateOrganizationMemberAccess
+- assignOrganizationMemberRole
+- assignOrganizationMemberRoleToMany
+- deleteOrganizationMemberRole
+- updateProjectRegistryModel
+- createVersion
+- updateVersionStatus
+- createActivity
+- addSlackIntegration
+- deleteSlackIntegration
+- addGitHubIntegration
+- deleteGitHubIntegration
+- setSchemaPolicyForOrganization
+- setSchemaPolicyForProject
+- createDocumentCollection
+- deleteDocumentCollection
+- updateDocumentCollection
+- createDocumentCollectionDocument
+- deleteDocumentCollectionDocument
+- updateDocumentCollectionDocument
+- createSchemaCheck
+
+
 ## Implementation
 
-Proposed implementation is to use a single clickhouse table to store all events. The table will have
+Proposed implementation is to use a single Clickhouse table to store all events. The table will have
 the following columns:
+
 
 ```sql
 CREATE TABLE audit_log (
--- a/docs/proposals/proposal-audit-log.md
+++ b/docs/proposals/proposal-audit-log.md
@@ -1,0 +1,215 @@
+# Audit log
+
+### Purpose of audit log
+
+The audit log is a feature that allows users to track changes made to their organization, projects,
+schemas, and other resources.
+
+## Events to track
+
+### OrganizationManager
+
+- `leaveOrganization`
+- `createOrganization`
+- `deleteOrganization`
+- `updatePlan`
+- `updateRateLimits`
+- `updateName`
+- `deleteInvitation`
+- `inviteByEmail`
+- `joinOrganization`
+- `requestOwnershipTransfer`
+- `answerOwnershipTransferRequest`
+- `deleteMember`
+- `updateMemberAccess`
+- `createMemberRole`
+- `deleteMemberRole`
+- `assignMemberRole`
+- `updateMemberRole`
+- `createRoleWithMembersMigration`
+- `assignRoleToMembersMigration`
+
+  ### ProjectManager
+
+- `createProject`
+- `deleteProject`
+- `updateName`
+
+### ArtifactStorageWriter
+
+- `writeArtifact`
+- `deleteArtifact`
+
+### CompositeLegacyModel
+
+- `check`
+- `publish`
+
+### SingleLegacyModel
+
+- `check`
+- `publish`
+
+### TargetManager
+
+- `createTarget`
+- `deleteTarget`
+- `setTargetValidation`
+- `updateTargetValidationSettings`
+- `updateName`
+- `updateTargetGraphQLEndpointUrl`
+- `updateTargetSchemaComposition`
+
+### SchemaPublisher
+
+- `updateVersionStatus`
+- `delete`
+- `publishToCDN`
+
+### SchemaManager
+
+- `updateSchemaVersionStatus`
+- `createVersion`
+- `updateBaseSchema`
+- `disableExternalSchemaComposition`
+- `enableExternalSchemaComposition`
+- `updateNativeSchemaComposition`
+- `updateRegistryModel`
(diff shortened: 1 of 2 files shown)


docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,209 @@
...
+The audit log is a feature that allows users to track changes made to their organization, projects,
+schemas, and other resources.
+
+## Events to track
+
+### OrganizationManager
+
+- `leaveOrganization`
+- `createOrganization`
+- `deleteOrganization`
```
n1ru4l, line 14: about these, since the audit log lives on the organization level, I am unsure whether we can have these.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,209 @@
...
+  ### ProjectManager
+
+- `createProject`
+- `deleteProject`
+- `updateName`
+
+### ArtifactStorageWriter
+
+- `writeArtifact`
+- `deleteArtifact`
```
n1ru4l, line 41: no need to track these, this logic is called as part of schema publish

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,209 @@
...
+- `updateCollection`
+- `deleteOperation`
+
+### GitHubIntegrationManager
+
+- `register`
+- `unregister`
+- `createCheckRun`
+- `updateCheckRun`
+- `updateCheckRunToSuccess`
```
n1ru4l, line 106: this is called as part of schema checks, no need to audit this.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,209 @@
...
+### ContractsManager
+
+- `createContract`
+- `disableContract`
+
+### Contracts
+
+- `createContract`
+- `disableContract`
+- `approveContractChecksForSchemaCheckId`
```
n1ru4l, line 129: this is part of approving a schema check, no need to track this individually.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,209 @@
...
+- `deleteTarget`
+- `setTargetValidation`
+- `updateTargetValidationSettings`
+- `updateName`
+- `updateTargetGraphQLEndpointUrl`
+- `updateTargetSchemaComposition`
+
+### SchemaPublisher
+
+- `updateVersionStatus`
```
n1ru4l, line 65: Part of SchemaPublished.publish, no need to track this individually.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,209 @@
...
+- `delete`
+- `publishToCDN`
```
n1ru4l, line 67: Part of SchemaPublished.publish, no need to track this individually.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,209 @@
...
+
+### CompositeLegacyModel
+
+- `check`
+- `publish`
+
+### SingleLegacyModel
+
+- `check`
+- `publish`
```
n1ru4l, line 51: these are invoked via SchemaPublisher and other classes, no need to track them individually.

n1ru4l:
@capaj Compiling this list based on GraphQL mutation root fields might be more useful as they already describe the actions from a user point of view. Then from there we can pick the classes/methods that have the business logic for these

docs/proposals/proposal-audit-log.md
capaj: why not? what if I self host and someone deletes my organisation by mistake? 
I would love to see which user did that.

capaj:
@n1ru4l I did it based on the provider typescript classes as suggested here:

https://guild-oss.slack.com/archives/C01E8ATBQGN/p1718616626696269?thread_ts=1718613813.068609&cid=C01E8ATBQGN

Are there still too many events missing? I have added a lot of them from Storage interface now.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,219 @@
...
+  event_details JSON
+) ENGINE = MergeTree ()
+ORDER BY
+  event_time;
+```
+
+our log function will be a simple function that inserts a row into the table:
+
+```ts
+
```
dimaMachina, line 174:
```suggestion
```

scripts/get-services-outline.ts
```diff
@@ -0,0 +1,77 @@
...
+    file,
+    fs.readFileSync(file, 'utf8'),
+    ts.ScriptTarget.Latest,
+    true,
+  );
+  const classes = extractClassesAndMethods(file, sourceFile);
+  if (classes.length > 0) {
+    classesAndMethods.push({ file, classes });
+  }
+  if (classes.length > 0) {
```
dimaMachina, line 74:
```suggestion
```

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,219 @@
...
+import { ClickHouse } from 'clickhouse';
+
+const clickhouse = new ClickHouse(...)
```
dimaMachina, line 177:
```suggestion
const clickhouse = new ClickHouse(/* ... */)
```
and prettier could prettify this code block!

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,220 @@
...
+
+Our log function will be a simple function that inserts a row into the table:
+
+```ts
+import { ClickHouse } from 'clickhouse';
+
+const clickhouse = new ClickHouse(/* ... */)
+
+type AuditLogEvent = {
+  userId: string;
```
n1ru4l, line 181: not every event can be pointed to a user, access tokens are currently not linked to any user account AFAIK.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,220 @@
...
+logAuditEvent({
+  userId: '690ae6ae-30e7-4e6c-8114-97e50e41aee5',
+  organizationId: 'da2dbbf8-6c03-4abf-964d-8a2d949da5cb',
+  eventKind: 'OrganizationManager',
+  action: 'joinOrganization',
+})
+```
+
+We would call this `logAuditEvent` function in all the places in the codebase to log the events
+listed above.
```
n1ru4l, line 220: We don't use global functions, we use service injections within the business logic services.

n1ru4l:
@capaj I think there was a misunderstanding before, in which exact method the logging is added is not as important right now. We need a list of the event types, the data stored for that specific event, and a sample formatted sentence on how to display that data within the app for users. We can take the mutation resolvers as an entry point to locate the class method where the logging should be added for the actual implementation later on, and get a better idea of the data worth storing.

docs/proposals/proposal-audit-log.md
capaj: I was thinking of just importing it, but if it needs to an injectable let it be an injectable. Will refactor it into an injectable class

capaj: should be an injectable class now

capaj:
@n1ru4l the list of events I propose to log is here between lines 10 and 144:
https://github.com/kamilkisiela/graphql-hive/pull/4990/files#diff-8f8e57f328cf7f2227e74556e03115d1b53187395e84323c8bc3b9085916b6c9R10-R144

>  a sample formatted sentence on how to display that data within the app for users. 

why is FE a necessary part of this task? 
IMHO audit logging has TREMENDOUS value for the user of hive even if it is not displayed anywhere.
It was never mentioned anywhere before. I can add it if you deem it is necessary and we need to implement it from the get got.
I would personally split it in two separate issues with two separate PRs, but if you think I can pull it off in a single PR fine by me.
It will take me like additional day just for the proposal to add this for all 120ish events we have listed.

> the data stored for that specific event

this will take me like 2-3 days to add to the proposal. I need to go through each event, basically find where it is going to be called.
For me easiest way to gather this is by simply implementing it and then gathering it by parsing all the calls to `logAuditEvent`  so we're crossing from proposal to implementation here.

capaj:
> We can take the mutation resolvers as an entry point to locate the class method where the logging should be added for the actual implementation later on, and get a better idea of the data worth storing.

I was suggesting using an http middleware where mutation resolver names would be used as event names in our initial conversation on slack and you said we should rather use our bussinec logic classes here: https://guild-oss.slack.com/archives/C01E8ATBQGN/p1718614773106589?thread_ts=1718613813.068609&cid=C01E8ATBQGN

so that's why I went for the class methods rather than GQL mutations.

If you think we should rather use GQL mutation resolver names I am totally fine with that, In that case I would probably suggest we track almost all of them, maybe with exception of `experimental__updateTargetSchemaComposition`.
In details we would store the inputs for each resolver.

For example for event  `retireAppDeployment` we would store `RetireAppDeploymentInput` in `event_details`

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,230 @@
...
+  };
+}
+
+// sample usage - do not await to avoid blocking the the thread
+
+
+this.auditLog.logAuditEvent({
+  userId: '690ae6ae-30e7-4e6c-8114-97e50e41aee5',
+  organizationId: 'da2dbbf8-6c03-4abf-964d-8a2d949da5cb',
+  eventKind: 'OrganizationManager',
```
n1ru4l, line 228: Just having an action/event is fine, no need to have both eventKind and action, the eventKind information does not really add any new value.

docs/proposals/proposal-audit-log.md
n1ru4l: I was talking from a users point of perspective. As admins will look at this and if the organization no longer exist, they would have to contact us directly. Which is fine in that scenario. So let's keep it.

docs/proposals/proposal-audit-log.md
capaj: removed now

capaj:
From what it seems to me there is misalignment between me and @kamilkisiela @n1ru4l on what the objective is for this issue.

They see it as fullstack solution that we can deploy and it will have a nice UI for the user to be able to search, filter, download and explore the audit logs.

I have never considered the UI part for this proposal.

Since the UI part is so important to both @kamilkisiela and @n1ru4l it's not productive to try to work on the proposal for anyone without the knowledge of actual usecases.
IMHO we should have a UI design done first or at least wireframes to show what is needed on FE. 
This has to be done by someone who knows the requirements, ideally someone who speaks to users who asked for this to find out how they plan to use this feature.

I am a bit confused why anyone would think this task is ready for getting specced when @kamilkisiela  admitted himself we do not have a list of filters from the UI: https://guild-oss.slack.com/archives/CAY2119MX/p1718801461050759?thread_ts=1718790068.341029&cid=CAY2119MX

PR #4996, opened 2024-06-18
[laboratory] remove `Merge fragments into query`, `Copy query` toolbar buttons, move `Prettify query` to end

dimaMachina:
- [x] remove `Merge fragments into query`, `Copy query` toolbar buttons, move `Prettify query` to end
image

- [x] increase text

image

- [x] show 3 dots on hover

https://github.com/kamilkisiela/graphql-hive/assets/7361780/a42a25f8-014d-46d3-9bc6-2a1a3b9bb885


patches/graphiql@3.3.0.patch
```diff
@@ -0,0 +1,25 @@
+# move prettify button to the end, we can't use `<GraphiQL.Toolbar />` because `usePrettifyEditors` needs to access `EditorContextProvider`
```
dimaMachina, line 1: described why we need patch

n1ru4l (approved):

PR #4999, opened 2024-06-19
[hive/app] fixes error `<button>` cannot appear as a descendant of `<button>` for `TabsTrigger` inside `TooltipTrigger

dimaMachina:
https://github.com/kamilkisiela/graphql-hive/assets/7361780/1371f379-ca8e-4312-8a49-7a6b95d72385

### DIFF
--- a/packages/web/app/src/components/ui/tabs.tsx
+++ b/packages/web/app/src/components/ui/tabs.tsx
@@ -28,7 +28,8 @@ const TabsTrigger = React.forwardRef<
   <TabsPrimitive.Trigger
     ref={ref}
     className={cn(
-      'ring-offset-background focus-visible:ring-ring data-[state=active]:bg-background data-[state=active]:text-foreground inline-flex items-center justify-center whitespace-nowrap rounded-sm px-3 py-1.5 text-sm font-medium transition-all focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50 data-[state=active]:shadow-sm',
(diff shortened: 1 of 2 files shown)


packages/web/app/src/components/ui/tabs.tsx
```diff
@@ -28,7 +28,8 @@ const TabsTrigger = React.forwardRef<
   <TabsPrimitive.Trigger
     ref={ref}
     className={cn(
-      'ring-offset-background focus-visible:ring-ring data-[state=active]:bg-background data-[state=active]:text-foreground inline-flex items-center justify-center whitespace-nowrap rounded-sm px-3 py-1.5 text-sm font-medium transition-all focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 disabled:pointer-events-none disabled:opacity-50 data-[state=active]:shadow-sm',
+      'ring-offset-background focus-visible:ring-ring data-[state=active]:bg-background data-[state=active]:text-foreground inline-flex items-center justify-center whitespace-nowrap rounded-sm px-3 py-1.5 text-sm font-medium transition-all focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-offset-2 disabled:opacity-50 data-[state=active]:shadow-sm',
+      'disabled:cursor-not-allowed active:disabled:pointer-events-none', // `:active` allows `:hover` state for using with `TooltipTrigger` with `asChild` prop
```
kamilkisiela, line 32: Can you explain a bit?

dimaMachina: with the `asChild` prop, there will be only 1 button, but since we had `disabled:pointer-events-none` previously - a tooltip will not be shown as we disable pointer-events, even on `:hover`

If we remove completely  `disabled:pointer-events-none` disabled tabs buttons will be clickable, so instead we set `pointer-events-none` only on `disabled` state + `:active`, so with this tooltips will be shown even on disabled tabs

PR #5000, opened 2024-06-19
[laboratory] scroll to operation in collections' accordion on first load

dimaMachina:
- [x] scroll to operation in collections' accordion on the first load
- [x] There is a bug when sometimes, the right collection is not opened on the first load with searchParams `operation`, fixed by using `value` instead `defaultValue` in `<Accordion />` element

https://github.com/kamilkisiela/graphql-hive/assets/7361780/fa17f94d-1ebe-4637-8b65-b405d4394a85

### DIFF
--- a/packages/web/app/src/pages/target-laboratory.tsx
+++ b/packages/web/app/src/pages/target-laboratory.tsx
@@ -1,4 +1,4 @@
-import { ReactElement, useCallback, useEffect, useMemo, useState } from 'react';
+import { ReactElement, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { cx } from 'class-variance-authority';
 import clsx from 'clsx';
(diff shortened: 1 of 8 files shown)


packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -1100,6 +1110,7 @@ function LaboratoryPageContent(props: {
             visiblePlugin={operationCollectionsPlugin}
             schema={schema}
             forcedTheme="dark"
+            disableTabs
```
dimaMachina, line 1113: use new feature instead 
image

n1ru4l (approved):

kamilkisiela:
@dimaMachina can you resolve merge conflicts? After that, you can merge.

PR #5004, opened 2024-06-20
[hive/app] remove `use client` comments, remove `globalThis.window` checks

dimaMachina:
- [x] remove `'use client'` comments
- [x] remove `globalThis.window` checks
- [x] <s>replace Sentry's deprecated `configureScope` with `getCurrentScope`</s>


packages/web/app/src/router.tsx
```diff
@@ -80,9 +79,8 @@ const LazyTanStackRouterDevtools = lazy(() =>
 );
 
 function identifyOnSentry(userId: string, email: string): void {
-  configureScope(scope => {
-    scope.setUser({ id: userId, email });
-  });
+  const scope = getCurrentScope();
+  scope.setUser({ id: userId, email });
```
kamilkisiela, line 83: Let's not do any Sentry related changes in this PR. We need to migrate to newer Sentry anyway, but not here.

kamilkisiela (changes requested):
I added one comment

PR #5047, opened 2024-06-24
[hive/app] replace `'` and `"` for `old`/`new` description

dimaMachina:
fixes https://guild-oss.slack.com/archives/C01E8ATBQGN/p1719154268309169

image

### DIFF
--- a/packages/services/storage/src/schema-change-model.ts
+++ b/packages/services/storage/src/schema-change-model.ts
@@ -197,6 +197,10 @@ export function implement<Model = never>() {
   };
 }
 
+function replaceQuotes(value: string): string {
+  return value.replaceAll("'", '&apos;').replaceAll('"', '&quot;');
+}
+
 export const FieldArgumentDescriptionChangedModel =
   implement<FieldArgumentDescriptionChangedChange>().with({
     type: FieldArgumentDescriptionChangedLiteral,
@@ -204,8 +208,8 @@ export const FieldArgumentDescriptionChangedModel =
       typeName: z.string(),
       fieldName: z.string(),
       argumentName: z.string(),
-      oldDescription: z.union([z.string(), z.null()]),
-      newDescription: z.union([z.string(), z.null()]),
+      oldDescription: z.union([z.string().transform(replaceQuotes), z.null()]),
+      newDescription: z.union([z.string().transform(replaceQuotes), z.null()]),
     }),
   });
 
@@ -253,8 +257,8 @@ export const DirectiveDescriptionChangedModel = implement<DirectiveDescriptionCh
     type: DirectiveDescriptionChangedLiteral,
     meta: z.object({
       directiveName: z.string(),
-      oldDirectiveDescription: z.union([z.string(), z.null()]),
-      newDirectiveDescription: z.union([z.string(), z.null()]),
+      oldDirectiveDescription: z.union([z.string().transform(replaceQuotes), z.null()]),
+      newDirectiveDescription: z.union([z.string().transform(replaceQuotes), z.null()]),
     }),
   },
 );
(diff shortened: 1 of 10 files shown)


packages/services/storage/src/schema-change-model.ts
```diff
@@ -197,15 +197,19 @@ export function implement<Model = never>() {
...
   implement<FieldArgumentDescriptionChangedChange>().with({
     type: FieldArgumentDescriptionChangedLiteral,
     meta: z.object({
       typeName: z.string(),
       fieldName: z.string(),
       argumentName: z.string(),
-      oldDescription: z.union([z.string(), z.null()]),
-      newDescription: z.union([z.string(), z.null()]),
+      oldDescription: z.union([z.string().transform(replaceQuotes), z.null()]),
+      newDescription: z.union([z.string().transform(replaceQuotes), z.null()]),
```
dimaMachina, line 212: I only transform `old` and `new` descriptions, I think we don't need transform other descriptions, like https://github.com/kamilkisiela/graphql-hive/blob/3c53032966f2ea7e63f40d22f32e463a4e8fe082/packages/services/storage/src/schema-change-model.ts#L549

packages/web/app/src/components/target/history/labelize.tsx
```diff
@@ -0,0 +1,10 @@
+import { ReactNode } from 'react';
+import reactStringReplace from 'react-string-replace';
+import { Label } from '@/components/common';
+
+export function labelize(message: string): ReactNode[] {
+  // Replace '...' with <Label>...</Label>
+  return reactStringReplace(message, /'([^']+)'/gim, (match, i) => (
```
dimaMachina, line 7: I removed previously `message.replace(/"/g, "'")` I think we don't need to replace `"`

packages/web/app/src/components/target/history/labelize.tsx
```diff
@@ -0,0 +1,10 @@
...
+    <Label key={i} dangerouslySetInnerHTML={{ __html: match }} />
```
dimaMachina, line 8: I think better to use `dangerouslySetInnerHTML` instead of `.replaceAll('&apos;', "'").replaceAll('&quot;', '"')`

packages/services/storage/src/schema-change-model.ts
```diff
@@ -197,15 +197,19 @@ export function implement<Model = never>() {
   };
 }
 
+function replaceQuotes(value: string): string {
+  return value.replaceAll("'", '&apos;').replaceAll('"', '&quot;');
```
dotansimha, line 201: Good catch. Do we have more things that need to be escaped? 🤔 
@kamilkisiela @n1ru4l @TuvalSimha

packages/web/app/src/components/target/history/labelize.tsx
dotansimha: My concern is that using `dangerouslySetInnerHTML` will inject things we do not expect. 
Let's say someone is using some HTML or malicious syntax in the schema description, it might be injected all the way to the frontend?

dimaMachina: you are right, I totally forgot about this case, I will push a fix

packages/services/storage/src/schema-change-model.ts
n1ru4l: How does this affect the output of the CLI, Slack and GitHub actions? We also use this information from there, not only from the web client.

kamilkisiela (changes requested):
https://github.com/kamilkisiela/graphql-hive/pull/5047#discussion_r1654632219

PR #5062, opened 2024-06-24
Audit log proposal II

capaj:
just for review purposes.

If you prefer to read as markdown open here: https://github.com/capaj/graphql-hive/blob/audit-log-proposal-2/docs/proposals/proposal-audit-log.md


docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+  target_id UUID,
+  target_name STRING,
+  schema_version_id UUID,
+  event_action STRING NOT NULL,
+  event_details JSON,
+  INDEX idx_user_id user_id TYPE set(0) GRANULARITY 64,
+  INDEX idx_user_email user_email TYPE set(0) GRANULARITY 64,
+  INDEX idx_organization_id organization_id TYPE set(0) GRANULARITY 64,
+  INDEX idx_project_id project_id TYPE set(0) GRANULARITY 64,
+  INDEX idx_target_id target_id TYPE set(0) GRANULARITY 64,
```
kamilkisiela, line 32:
```suggestion
```
These are not needed if we do the sort key correctly

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+) ENGINE = MergeTree ()
+ORDER BY event_time
```
kamilkisiela, line 34: let's use better sort key, here's what I mean:
https://www.tinybird.co/clickhouse/knowledge-base/sorting-key-performance

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+TTL timestamp + INTERVAL 3 MONTH;
```
kamilkisiela, line 35:
```suggestion
TTL timestamp + INTERVAL 2 YEAR;
```

@dotansimha how long should we keep the audit logs?

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Implementation
+
+Proposed implementation is to use a single Clickhouse table to store all events. The table will have
+the following schema:
+
+
+```sql
+CREATE TABLE audit_log (
+  event_time DateTime DEFAULT now(),
```
kamilkisiela, line 17:
```suggestion
  timestamp DateTime('UTC'),
```

Let's follow the same conventions as in other tables

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+  user_id UUID,
+  user_email STRING,
+  organization_id UUID,
+  project_id UUID,
+  project_name STRING,
+  target_id UUID,
+  target_name STRING,
+  schema_version_id UUID,
+  event_action STRING NOT NULL,
```
kamilkisiela, line 26:
```suggestion
  event_action LowCardinality(String),
```

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+TTL timestamp + INTERVAL 3 MONTH;
+```
+Data in clickhouse would be append only. We would never manually delete them, we would rely on TTL removing old rows after X months
+Of course we could make the interval configurable if necessary.
+
+Audit log is simple class with one method that inserts a single row into the clickhouse table:
+
+```ts
+const { createClient } = require('@clickhouse/client');
+const clickhouse = createClient()
```
kamilkisiela, line 44: we do have our own clickhouse client (packages/services/api/src/modules/operations/providers/clickhouse-client.ts)

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+
+type AuditLogEvent = {
```
kamilkisiela, line 46: We need to be support all events in a backwards compatible manner, so I guess we will need something more strict than `string` for an event type and for the details, something to avoid "reading 'foo' of undefined" errors.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+}
+
+@Injectable({
+  scope: Scope.Singleton,
+  global: true,
+})
+export class AuditLog {
+  logAuditEvent (event: AuditLogEvent) {
+    const { userId, organizationId, projectId, targetId, schemaVersionId, eventAction, details } = event;
+    const userEmail = userId ? await getUserEmail(userId) : null; // get user email would be cached in redis for 10 minutes to avoid hitting the DB too often
```
kamilkisiela, line 63: Every action that requires a user, has the user info cached already, so it won't be a problem. I will explain more once you will start implementing these in the codebase.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+      targetId,
+      schemaVersionId,
+      eventKind,
+      eventAction,
+      JSON.stringify(details),
+    ])
+  };
+}
+
+// sample usage - always await unless performance is critical
```
kamilkisiela, line 83: Yes, I think we should always `await` but also introduce a prometheus metric to measure how long it took to insert an event and for what action type.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+| USER_JOINED                    | User **john@acme.com** joined **Acme**. Approved by **admin@acme.com**. Referrer: **jane@acme.com**.      |
+| EXPIRED_INVITE_HIT             | User **john@acme.com** tried to join **Acme** with an expired/invalid invite. Referrer: **jane@acme.com** |
+| PROJECT_CREATED                | User **john@acme.com** created a project named **Project Alpha**.                                         |
+| TARGET_CREATED                 | User **john@acme.com** created a target named **Project Alpha**.                                          |
+| ROLE_CREATED                   | Admin **admin@acme.com** created a new role **SAMPLE_ROLE**.                                              |
+| ROLE_ASSIGNED                  | Admin **admin@acme.com** assigned a new role to user **john@acme.com**.                                   |
+| USER_REMOVED                   | Admin **admin@acme.com** removed user **john@acme.com**.                                                  |
+| ORG_TRANSFERRED                | Admin **admin@acme.com** transferred ownership.                                                           |
```
kamilkisiela, line 108:
```suggestion
| ORGANIZATION_TRANSFERRED                | Admin **admin@acme.com** transferred ownership.                                                           |
```

let's use full words

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+| SCHEMA_CHECKED                 | CI made a schema check for **Project Alpha**.                                                             |
+| SCHEMA_PUBLISH                 | User **john@acme.com** published a new schema version for **Project Alpha**.                              |
```
kamilkisiela, line 110:
```suggestion
| SCHEMA_PUBLISHED                 | User **john@acme.com** published a new schema version for **Project Alpha**.                              |
```

@n1ru4l do you think we should use past tense everywhere? `SCHEMA_PUBLISHED` vs `PUBLISH_SCHEMA`?

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+    ORGANIZATION_SETTINGS_UPDATED
+    TARGET_SETTINGS_UPDATED
+    SCHEMA_POLICY_SETTINGS_UPDATED
+  }
+
+  type AuditLog {
+    event_time: Date!
+    user_id: String
+    user_email: String
+    organization_id: String
```
kamilkisiela, line 162:
```suggestion
    organization_id: String!
```

I think we will always have organization_id

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+
+
+## Implementation
+
+Proposed implementation is to use a single Clickhouse table to store all events. The table will have
+the following schema:
+
+
+```sql
+CREATE TABLE audit_log (
```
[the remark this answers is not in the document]
n1ru4l: hmmm, I am more in favor of computing the message from the data. The thought behind this is that we want to make parts of the message interactive (e.g. link to the organization or user etc. from the UI). But for export purposes also having a message within the table might make sense.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+
+We store metadata about the export in a new postgres table `audit_log_export`:
+
+
+```sql
+CREATE TABLE audit_log_export (
+  id UUID PRIMARY KEY,
+  url TEXT,
+  filters JSON,
+  valid_until TIMESTAMP WITH TIME ZONE,
```
kamilkisiela, line 252:
```suggestion
  expires_at TIMESTAMP WITH TIME ZONE,
```
maybe? But I like the idea of using S3+TTL and a column in PG to make sure we hide the export

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+## Audit log export
+
+Stored file is either a CSV or plain text file with a list of events. The file is stored in S3 and a link to download it is returned to the user. The file is stored for 30 days-TTL would be set on S3 object so that we don't have to clean up old files using a CRON job.
+CSV file would be generated using clickhouse client and streamed to S3 to avoid any possible memory issues:
+
+```ts
+  await clickhouse.query({
+    query,
+    format: 'CSV',
+  }).stream().pipe(writableStream);
```
kamilkisiela, line 233: If we want to include formatted messages in the CSV or TEXT output, we will have to go row by row and render it.
Maybe storing full messages in clickhouse is good option to avoid such scenario?

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+  }).stream().pipe(formatToPlainText).pipe(writableStream);
+```
+
+We store metadata about the export in a new postgres table `audit_log_export`:
+
+
+```sql
+CREATE TABLE audit_log_export (
+  id UUID PRIMARY KEY,
+  url TEXT,
```
kamilkisiela, line 250: I guess it would have to be a presigned url

docs/proposals/proposal-audit-log.md
capaj: changed to
```suggestion
ORDER BY (event_time, organization_id, project_id, target_id)
```

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,256 @@
...
+    pageInfo: PageInfo
+  }
+
+  type Query {
+    auditLogs(filter: AuditLogFilter, first: Int, after: String): AuditLogConnection
+    auditLogExports(first: Int, after: String ): ExportResultConnection
+  }
+
+  type Mutation {
+    exportAuditLogsToFile(filter: AuditLogFilter): AuditLogFileExport
```
kamilkisiela, line 220: I guess here, we will show a link to the file, right? And filters, to render the thing correctly?

docs/proposals/proposal-audit-log.md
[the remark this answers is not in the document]
capaj: same opinion as @n1ru4l.
Also storing the message would be useless if we ever do i18n in the future. Rendering the message is trivial and we can share the code for it between FE/BE.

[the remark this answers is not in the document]
kamilkisiela: We can have both. If we store the message in clickhouse, we can export data from it in CSV format and ship it to S3, with no IO reading/writing/transforms, and still produce a message dynamically in React.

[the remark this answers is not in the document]
kamilkisiela: Imo we should store messages and use them for file exports (text and csv)

docs/proposals/proposal-audit-log.md
capaj: removed

docs/proposals/proposal-audit-log.md
capaj: changed to 2 years. If it is too expensive we can always lower it

docs/proposals/proposal-audit-log.md
capaj: changed `eventAction` to be AuditLogEventAction.
That can be a generated enum from the GQL schema.

docs/proposals/proposal-audit-log.md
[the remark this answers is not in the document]
capaj: ok I will add the column for that

docs/proposals/proposal-audit-log.md
capaj: yes writableStream is not defined, but it would be something like in this SO answer: https://stackoverflow.com/a/50291380/671457

docs/proposals/proposal-audit-log.md
capaj: renamed

docs/proposals/proposal-audit-log.md
capaj: correct. We can always make it non nullable if we introduce event for something like user signup

docs/proposals/proposal-audit-log.md
capaj: Yes I think we want to give users ability to filter in exports too, not just in UI.
url is a field in the AuditLogFileExport, so FE can pick that when doing the mutation

docs/proposals/proposal-audit-log.md
capaj: I renamed this for consistency with all the other ones

capaj:
I have made some changes based on your feedback @kamilkisiela.
I have resolved comments which were adressed. 
Please ping me on anything that still seems unresolved from your POV.

capaj:
what do you think @n1ru4l? I have only seen one comment from you. Do you reckon I could start implementing this feature based on this proposal?

kamilkisiela:
> I have made some changes based on your feedback @kamilkisiela. I have clicked "resolve" for comments which were adressed. Please ping me on anything that still seems unresolved from your POV.

All good. Next time you can message me on Slack, because I missed this comment and it's been a week.

capaj:
Sorry, will do next time.
I was not idly waiting, I have a branch WIP where I have a few things already implemented.

So I will close this PR as it seems everything is addressed.
Will open a draft with implementation soonish-hopefully Thursday.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,265 @@
...
+    user_id: String
+    user_email: String
+    organization_id: String!
+    project_id: String
+    project_name: String
+    target_id: String
+    target_name: String
+    schema_version_id: String
+    event_action: AuditLogEventAction!
+    event_details: JSON
```
n1ru4l, line 178: lets stay consistent and use camel case

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  type ExportResultEdge {
+    cursor: String
+    node: AuditLogFileExport
+  }
+
+  type ExportResultConnection {
+    edges: [ExportResultEdge]
+    pageInfo: PageInfo
```
n1ru4l, line 220: lets make these non nullable

docs/proposals/proposal-audit-log.md
capaj: looking at migrations here:
https://github.com/kamilkisiela/graphql-hive/tree/d07cc6b1c2992227d13c79325927e7f6ebb857de/packages/migrations/src/clickhouse-actions

it seems we are using snake case for clickhouse tables.

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  input AuditLogFilter {
+    startDate: Date
+    endDate: Date
+    userId: String
+    userEmail: String
+    projectId: String
+    targetId: String
+    eventAction: String
```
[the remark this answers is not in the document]
capaj: will change it

docs/proposals/proposal-audit-log.md
```diff
@@ -0,0 +1,265 @@
...
+  user_id LowCardinality(String) CODEC(ZSTD(1)),
+  user_email STRING,
+  organization_id LowCardinality(String) CODEC(ZSTD(1)),
+  project_id LowCardinality(String) CODEC(ZSTD(1)),
+  project_name STRING,
+  target_id LowCardinality(String) CODEC(ZSTD(1)),
+  target_name STRING,
+  schema_version_id LowCardinality(String) CODEC(ZSTD(1)),
+  event_action LowCardinality(String) CODEC(ZSTD(1)),
+  event_details JSON,
```
capaj, line 27:
```
Set setting allow_experimental_object_type = 1 in order to allow it. (ILLEGAL_COLUMN) (version 24.4.3.25 (official build))"
}
```

actually if we want to use JSON, we would have to enable `allow_experimental_object_type`.
Do you want to enable that or change this to plain string @kamilkisiela ?

PR #5088, opened 2024-06-27
feat: app deployment UI

n1ru4l:
UI implementation for app deployments. 

See https://github.com/kamilkisiela/graphql-hive/pull/4746

image
image

### DIFF
--- a/packages/web/app/src/components/ui/page-content-layout.tsx
+++ b/packages/web/app/src/components/ui/page-content-layout.tsx
@@ -51,14 +51,14 @@ SubPageLayout.displayName = 'SubPageLayout';
 
 type SubPageLayoutHeaderProps = {
   children?: ReactNode;
-  title?: string;
+  subPageTitle?: ReactNode;
   description?: string | ReactNode;
 } & HTMLAttributes<HTMLDivElement>;
 
 const SubPageLayoutHeader = forwardRef<HTMLDivElement, SubPageLayoutHeaderProps>((props, ref) => (
   <div className="flex flex-row items-center justify-between" ref={ref}>
     <div className="space-y-1.5">
-      <CardTitle>{props.title}</CardTitle>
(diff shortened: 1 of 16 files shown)


packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -362,6 +362,20 @@ function useOperationCollectionsPlugin(props: {
...
+              ? searchObj.operationString
+              : null;
+
+          // We provide an operation string when navigating to the laboratory from persisted documents
+          // in that case we want to show that operation within this tab.
+          if (operationString) {
+            editorContext.queryEditor?.setValue(operationString);
+            editorContext.variableEditor?.setValue('');
+            editorContext.headerEditor?.setValue('');
+          }
```
n1ru4l, line 377: I wanted to have a way to open any operation within the lab, so I added this.

packages/web/app/src/lib/urql.ts
```diff
@@ -34,6 +35,14 @@ export const urqlClient = createClient({
...
         Mutation,
       },
+      resolvers: {
+        Target: {
+          appDeployments: relayPagination(),
+        },
+        AppDeployment: {
+          documents: relayPagination(),
+        },
+      },
```
n1ru4l, line 45: This method is way more chilled than the patterns we use in other locations in the code. The drawback is the global configuration...

packages/web/app/src/components/ui/page-content-layout.tsx
```diff
@@ -51,14 +51,14 @@ SubPageLayout.displayName = 'SubPageLayout';
 
 type SubPageLayoutHeaderProps = {
   children?: ReactNode;
-  title?: string;
+  subPageTitle?: ReactNode;
```
n1ru4l, line 54: had to rename it as `title` property conflicted with `HTMLDivElement["title"]`, which can only be a string.

PR #5111, opened 2024-07-01
Custom Auth UI, proper SSO login flow and slug modifications

kamilkisiela:
Closes https://github.com/kamilkisiela/graphql-hive/issues/3055
Closes https://github.com/kamilkisiela/graphql-hive/issues/4367

**Ability to change organization slug**
So people could modify it, if the generated one bothers them, or if they wish to have something simpler for SSO login.

Screenshot 2024-07-03 at 12 06 47

**Login**
I added the section on left side to not only make the view a bit more interesting, but to also give people an idea where they landed :D Somebody could invite them to Hive and what they would see is `app.graphql-hive.com` without any links to `the-guild.dev/graphql/hive`, no context.

Screenshot 2024-07-03 at 12 02 58

**SSO**

Uses organization's slug for convenience, otherwise we would have to add options to define a list of allowed email domains, verify none of them are overlapping, avoid a situation where somebody types somebody's else company email domain and blocks it... things like that.

Screenshot 2024-07-03 at 12 41 21

**Register**

Screenshot 2024-07-03 at 12 03 11

**Password reset**

When a user typed an email address in the login form, and clicked "Forgot your password?", the email is passed to the password reset form. For convenience.

Screenshot 2024-07-03 at 12 03 06

**Email verification**

Screenshot 2024-07-03 at 12 04 10

Screenshot 2024-07-03 at 12 05 41

Screenshot 2024-07-03 at 12 05 34

### DIFF
--- a/packages/services/server/src/supertokens.ts
+++ b/packages/services/server/src/supertokens.ts
@@ -232,7 +232,6 @@ const getEnsureUserOverrides = (
         throw Error('emailPasswordSignUpPOST is not available');
       }
 
-      console.log(input.formFields);
       const response = await originalImplementation.emailPasswordSignUpPOST(input);
 
       const firstName = input.formFields.find(field => field.id === 'firstName')?.value ?? null;
--- a/packages/services/server/src/supertokens.ts
+++ b/packages/services/server/src/supertokens.ts
@@ -119,6 +119,19 @@ export const backendConfig = (requirements: {
     recipeList: [
       ThirdPartyEmailPasswordNode.init({
         providers,
+        signUpFeature: {
+          formFields: [
+            {
+              id: 'firstName',
+              // optional because of OIDC integration
+              optional: true,
+            },
+            {
+              id: 'lastName',
+              optional: true,
+            },
+          ],
+        },
         emailDelivery: {
           override: originalImplementation => ({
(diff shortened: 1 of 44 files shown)


packages/services/server/src/supertokens.ts
```diff
@@ -219,13 +232,19 @@ const getEnsureUserOverrides = (
         throw Error('emailPasswordSignUpPOST is not available');
       }
 
+      console.log(input.formFields);
```
kamilkisiela, line 235:
```suggestion
```

packages/services/server/src/supertokens/oidc-provider.ts
```diff
@@ -264,6 +264,7 @@ const OIDCTokenSchema = zod.object({ access_token: zod.string() });
 const getOIDCIdFromInput = (input: { userContext: any }, logger: FastifyBaseLogger): string => {
   const fastifyRequest = input.userContext._default.request as FastifyRequest;
   const originalUrl = 'http://localhost' + fastifyRequest.getOriginalURL();
+  console.log(originalUrl);
```
kamilkisiela, line 267:
```suggestion
```

n1ru4l (approved):
LGTM

PR #5112, opened 2024-07-01
Update high-availability-cdn.mdx

kamilkisiela:
Fixes this code block

Screenshot 2024-07-01 at 09 31 47

Screenshot 2024-07-01 at 09 35 56


kamilkisiela:
@dimaMachina is it an issue in Nextra or guild-docs? If a codeblock has no language specified, it renders with no padding/margin

PR #5136, opened 2024-07-03
[hive/app] add tabs for laboratory

dimaMachina:
currently uses patch from https://github.com/dimaMachina/graphiql/pull/3/files


kamilkisiela (changes requested):
There's no way to create a new tab and once I close the last one, the tabs are gone.

https://github.com/kamilkisiela/graphql-hive/assets/8167190/58b6a8e2-2628-486d-a5fe-03e0ae2dc62a

---

Dragging tabs is visually broken

https://github.com/kamilkisiela/graphql-hive/assets/8167190/f4fbe550-ba2f-4568-b8b7-ea4415f28543

---

Let's not do animations, there's just no point to have an animation here

https://github.com/kamilkisiela/graphql-hive/assets/8167190/b67b946e-636e-4b5b-bffe-e522e70c6620

Another point for no animation, it looks weird

Screenshot 2024-07-10 at 11 30 56

dimaMachina:
> There's no way to create a new tab and once I close the last one, the tabs are gone.

the new tab is created from sidebar, or by clicking on the existing

image

kamilkisiela (changes requested):
How do I create a tab? There's no button for it.
Screenshot 2024-08-01 at 11 09 04

I think we should make tabs touching the editor, otherwise it's floating and has only the top rounded, for no reason.
Screenshot 2024-08-01 at 11 11 21

Like this maybe...
Screenshot 2024-08-01 at 11 14 35


Also, not sure `<untitled>` is correct. The operations is titled in the collections. And yes I know GraphiQL takes the name from the operations, but we do have collections thingy and saved operations are named.
Screenshot 2024-08-01 at 11 14 56

The "My other collection" is centered and breaks weirdly (on smaller screens, but you can just open dev tools on right side and you will see).
Screenshot 2024-08-01 at 11 18 49

dimaMachina:
> How do I create a tab? There's no button for it.

image

> I think we should make tabs touching the editor, otherwise it's floating and has only the top rounded, for no reason.

done

> Also, not sure <untitled> is correct. The operations is titled in the collections. And yes I know GraphiQL takes the name from the operations, but we do have collections thingy and saved operations are named.

done

> The "My other collection" is centered and breaks weirdly (on smaller screens, but you can just open dev tools on right side and you will see).

fixed

kamilkisiela (changes requested):
There's still no way to create a tab. Yes you can create a collection (that is persisted in postgres) and add an operation (that is persisted in postgres as well), but it misses the whole point of tabs.

Users should be able to create as many operations as they need and save some to a collection only if they wants to. These operations are shared between users.

Screenshot 2024-08-06 at 09 04 28

kamilkisiela (changes requested):
The plus button, alone is hard to spot.
Screenshot 2024-08-12 at 09 43 52

Can we just not hide a tab and do it like here? If we always have a tab open then the plus sign button makes sense.
Screenshot 2024-08-12 at 09 44 00

kamilkisiela:
Hmm, I get a scrollbar here (chrome). I did hard refresh etc. 
Screenshot 2024-08-14 at 09 27 24

kamilkisiela:
other than that we're good I think and once it's fixed, we can merge it.

kamilkisiela (approved):

PR #5152, opened 2024-07-05
Migrate `activity` module to Server Preset (Part 9)

eddeee888:
### Background

This PR migrates `activity` module to use Server Preset

### DIFF
--- a/codegen.mts
+++ b/codegen.mts
@@ -58,9 +58,6 @@ const config: CodegenConfig = {
             unionMember: true,
             excludeTypes: ['TokenInfoPayload'],
           },
-          mappers: {
-            Activity: '../modules/activity/module.graphql.mappers#ActivityMapper', // TODO: @eddeee888 to fix server preset not detecting interface mappers
-          },
         },
       },
       {
--- a/codegen.mts
+++ b/codegen.mts
@@ -18,7 +18,6 @@ const config: CodegenConfig = {
         resolverMainFileMode: 'modules',
         resolverTypesPath: './__generated__/types.next.ts',
         blacklistedModules: [
-          'activity',
           'alerts',
           'auth',
           'billing',
--- a/codegen.mts
+++ b/codegen.mts
@@ -58,6 +58,9 @@ const config: CodegenConfig = {
             unionMember: true,
             excludeTypes: ['TokenInfoPayload'],
           },
+          mappers: {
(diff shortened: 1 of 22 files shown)


codegen.mts
```diff
@@ -59,6 +58,9 @@ const config: CodegenConfig = {
             unionMember: true,
             excludeTypes: ['TokenInfoPayload'],
           },
+          mappers: {
+            Activity: '../modules/activity/module.graphql.mappers#ActivityMapper', // TODO: @eddeee888 to fix server preset not detecting interface mappers
```
eddeee888, line 62: Server Preset is not detecting interface mappers so I have to this manually here. This should be an easy fix so we can do this manually here for now.

packages/services/api/src/modules/activity/resolvers/MemberLeftActivity.ts
```diff
@@ -0,0 +1,11 @@
+import type { ActivityObject } from '../../../shared/entities';
+import type { MemberLeftActivityResolvers } from './../../../__generated__/types.next';
+
+export const MemberLeftActivity: MemberLeftActivityResolvers = {
+  __isTypeOf(activity) {
+    return activity.type === 'MEMBER_LEFT';
+  },
+  email(activity: any) {
```
n1ru4l, line 8: Do you think we could as part of this PR make this more type-safe?

eddeee888: Let me see if I can wrangle this type in reasonable time 👀

eddeee888: I've looked into this and we could fix the `any` type issue now, but there's a few awkward things we need to do. 

Here's what I've tried:
1. Add `ActivityObject` as one of the implementing type mappers e.g.
```ts
export type ProjectCreatedActivityMapper = ActivityObject;
```
2. Run generate
3. Static analysis detects fields to update:
Screenshot 2024-07-09 at 8 26 23 PM
Here's how we can fix these issues:
- For (a.), update codegen config scalar `DateTime` to be `Date | string` as this is the correct expected type for `DateTime`
- For (b.) and (c.), make `ProjectCreatedActivityMapper`'s `project` and `user` fields to be required i.e. have a separate type from `ActivityObject`. We would need to make a mapper for every Activity type.

Since this change both codegen config and is quite verbose, I've marked it as a cleanup task to do at the end. This helps us finish the migration quickly, and then we can make code improvements after

n1ru4l (approved):

PR #5154, opened 2024-07-06
Migrate `auth` module to Server Preset (Part 10)

eddeee888:
### Background

This PR migrates `auth` module to Server Preset

### DIFF
--- a/packages/services/api/src/modules/auth/resolvers/User.ts
+++ b/packages/services/api/src/modules/auth/resolvers/User.ts
@@ -1,0 +1,6 @@
+import type { UserResolvers } from './../../../__generated__/types.next';
+
+export const User: Pick<
+  UserResolvers,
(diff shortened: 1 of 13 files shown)


packages/services/api/src/modules/auth/resolvers/User.ts
```diff
@@ -0,0 +1,6 @@
+import type { UserResolvers } from './../../../__generated__/types.next';
+
+export const User: Pick<
+  UserResolvers,
+  'displayName' | 'email' | 'fullName' | 'id' | 'isAdmin' | 'provider'
+> = {};
```
eddeee888, line 6: This is created automatically because we have a `UserMapper`. 
Currently, server preset still requires an empty object like this to do static analysis if there's a mapper - because if there's a mapper, there's a high chance user needs to write field mapping logic.

n1ru4l (approved):

dotansimha (approved):

PR #5166, opened 2024-07-07
Migrate `collection` module to Server Preset (Part 11)

eddeee888:
### Background

This PR migrates `collection`


packages/services/api/src/modules/collection/resolvers/DocumentCollection.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const DocumentCollection: DocumentCollectionResolvers = {
+  id: root => root.id,
+  name: root => root.title,
+  description: root => root.description,
+  operations: (root, args, { injector }) =>
+    injector.get(CollectionProvider).getOperations(root.id, args.first, args.after),
+  createdBy: async (_parent, _arg, _ctx) => {
+    /* DocumentCollection.createdBy resolver is required because DocumentCollection.createdBy exists but DocumentCollectionMapper.createdBy does not */
+  },
```
eddeee888, line 12: This was never implemented in this module, and I couldn't find this being implemented somewhere else... But it's not erroring at runtime... 
So, the only reason I can think of is this field is not being queried in any client. 

I'll need advice on how to turn the returned `createdByUserId` to a `User` object here 🙏 
Otherwise, we can ignore it in this migration with TS override and come back later

packages/services/api/src/modules/collection/resolvers/DocumentCollectionConnection.ts
```diff
@@ -0,0 +1,3 @@
+import type { DocumentCollectionConnectionResolvers } from './../../../__generated__/types.next';
+
+export const DocumentCollectionConnection: DocumentCollectionConnectionResolvers = {};
```
eddeee888, line 3: This empty object is created because of the mapper usage

packages/services/api/src/modules/collection/resolvers/DocumentCollectionOperationsConnection.ts
```diff
@@ -0,0 +1,4 @@
+import type { DocumentCollectionOperationsConnectionResolvers } from './../../../__generated__/types.next';
+
+export const DocumentCollectionOperationsConnection: DocumentCollectionOperationsConnectionResolvers =
+  {};
```
eddeee888, line 4: This empty object is created because of the mapper usage

packages/services/api/src/modules/collection/resolvers/DocumentCollection.ts
n1ru4l: You could call `injector.get(Storage).getUserById` or remove it altogether since it is not used from the frontend.

dotansimha: Yeah, I think it can be removed also from the schema

eddeee888: This is now done

n1ru4l (approved):

PR #5167, opened 2024-07-07
Apply rate-limit config to `auth-api` routes, using Contour

dotansimha:
This change is using the Local config for rate-limit: https://projectcontour.io/docs/1.29/config/rate-limiting/#local-rate-limiting

We might want to use the Global config with a proper service in the future. 

This PR also enabled sticky sessions for Contour (requests for the same user session will be routed to the same container), we can safely use the Local option at the moment.

### DIFF
--- a/deployment/services/proxy.ts
+++ b/deployment/services/proxy.ts
@@ -90,7 +90,7 @@ export function deployProxy({
         requestTimeout: '60s',
         retriable: true,
         rateLimit: {
-          maxRequests: 5,
+          maxRequests: 7,
           unit: 'hour',
         },
       },
--- a/deployment/services/proxy.ts
+++ b/deployment/services/proxy.ts
@@ -92,9 +92,6 @@ export function deployProxy({
         rateLimit: {
           maxRequests: 10,
           unit: 'minute',
-          responseHeadersToAdd: {
-            'x-ratelimit-active': 'true',
-          },
         },
       },
       {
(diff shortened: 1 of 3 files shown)


deployment/services/proxy.ts
```diff
@@ -89,6 +89,10 @@ export function deployProxy({
         service: graphql.service,
         requestTimeout: '60s',
         retriable: true,
+        rateLimit: {
+          maxRequests: 7,
```
kamilkisiela, line 93: That’s super low, no? The session is validated on every page visit (the initial one) and there’s a chance we get a oauth callback with a redirect to a new page that increases the amount by a lot

kamilkisiela (commented):
Looks good, but we should increase the limit. Maybe to 60 requests per minute or something?

deployment/services/proxy.ts
dotansimha: The calls for `/auth-api` made only when you first load the page (and then it's SPA). It's 3 calls, so you are right, it needs adjustments. 
How about max 10 requests a minute?

dotansimha: @kamilkisiela we can also set `burst: 3` and it will ignore the specific burst config from the token validation

kamilkisiela: Let's test it on staging and see if none of the auth flows are affected

dotansimha: @kamilkisiela seems like `maxRequests: 10` with burst works well (even when I'm refreshing very quickly). Also tried to login/logout multiple times.
When trying to send requests from curl, it blocks me at some point.

n1ru4l (approved):

kamilkisiela (approved):

PR #5182, opened 2024-07-09
Migrate `target` module to Server Preset (Part 12)

eddeee888:
### Background

This PR migrates `target` module to Server Preset


packages/services/api/src/modules/target/resolvers/Query/target.ts
```diff
@@ -0,0 +1,22 @@
+import { IdTranslator } from '../../../shared/providers/id-translator';
+import { TargetManager } from '../../providers/target-manager';
+import type { QueryResolvers } from './../../../../__generated__/types.next';
+
+export const target: NonNullable<QueryResolvers['target']> = async (
```
dotansimha, line 5: It seems like we are using `NonNullable` wrapper many times, maybe codegen can generate some signature for the non-nullable? 🤔 or do you think it's out of scope?

eddeee888: That's a good thought 🤔 Generally we want root-level resolvers to be implemented. If there's a codegen setting to make these non-optional (There's one for something else IIRC), we can just apply that and remove the `NonNullable`. 

Let me capture this and look into it next.

PR #5189, opened 2024-07-10
configures the Access-Control-Allow-Credentials CORS header

rickbijkerk:
### Background

Configures the Access-Control-Allow-Credentials CORS header

### Description
allows credentials to be used in requests

We ran into an issue after updating Hive, our cors preflight request got redirected to our SSO page due to missing cookies. This should fix that and allow us to update

### DIFF
--- a/packages/web/app/src/server/index.ts
+++ b/packages/web/app/src/server/index.ts
@@ -59,7 +59,9 @@ async function main() {
     });
   }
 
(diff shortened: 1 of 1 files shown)


kamilkisiela:
Oppsie :D I referenced your PR in my other PR to test the change and it got closed

PR #5228, opened 2024-07-15
Drop modals from V2: Create project modal

TuvalSimha:
## In this PR:

- `CreateProjectModal` split to 2 components: `CreateProjectModal` that handle the graphql, functions, etc... and `CreateProjectModalContent` handle the UI.
- Added `CreateProjectModalContent` to Storybook
- `create-project.tsx` moved to `packages/web/app/src/components/ui/create-project.tsx`
- `create-project.tsx` no more components from `V2`


packages/web/app/src/components/ui/create-project.tsx
```diff
@@ -60,44 +60,67 @@ export const CreateProjectMutation = graphql(`
...
   organizationId: string;
-}) => {
-  const { isOpen, toggleModalOpen } = props;
+};
+
+export const CreateProjectModal = ({
+  isOpen,
+  toggleModalOpen,
+  organizationId,
+}: CreateProjectModalProps): ReactElement => {
```
n1ru4l, line 123: Let's please not use destructuring for the props. Accessing `props.foo` is fine!

TuvalSimha: Fixed :)

PR #5237, opened 2024-07-16
Perform local schema composition in the hive dev command

### LINKED ISSUE #5113: hive dev - option to exclude unspecified subgraphs
Currently, `hive dev` includes schemas from all subgraphs in the supergraph, even when not specified in the CLI.
We need to implement a flag/option, allowing users to use only the provided subgraphs in the supergraph generation process.

kamilkisiela:
Closes #5113

Changes the default behavior of `hive dev` command. Now schema composition is done locally, without substituting subgraphs available in the registry.

I added `--remote` flag to the `hive dev` command to mimic the previous behavior.

The `$ hive dev` command was marked as "work in progress" (stated in the command description).
That's why I'm are not considering this a breaking change, but a minor change.

**_Before:_**

The `hive dev` command would substitute subgraphs available in the registry with their local counterparts, performing schema composition over the network according to your project's configuration.

**_After:_**

The `hive dev` command will now perform schema composition locally, without substituting subgraphs available in the registry. This is the default behavior.

To mimic the previous behavior, a user can apply the `--remote` flag and continue using the command as before.

---

- [ ] test `dependsOn` oclif options

### DIFF
--- a/.changeset/bright-lobsters-smell.md
+++ b/.changeset/bright-lobsters-smell.md
@@ -2,7 +2,7 @@
 '@graphql-hive/cli': minor
 ---
 
-Changes the default behavior of `hive dev` command where schema composition is done locally, without substituting subgraphs available in the registry.
+Changes the default behavior of `hive dev` command. Now schema composition is done locally, without substituting subgraphs available in the registry.
 
 We added `--remote` flag to the `hive dev` command to mimic the previous behavior.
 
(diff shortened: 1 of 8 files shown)


.changeset/bright-lobsters-smell.md
```diff
@@ -0,0 +1,23 @@
+---
+'@graphql-hive/cli': minor
+---
+
+Changes the default behavior of `hive dev` command where schema composition is done locally, without substituting subgraphs available in the registry.
```
kamilkisiela, line 5:
```suggestion
Changes the default behavior of `hive dev` command. Now schema composition is done locally, without substituting subgraphs available in the registry.
```

PR #5242, opened 2024-07-17
Migrate `policy` module to Server Preset (Part 16)

eddeee888:
### Background

This PR migrates policy module to Server Preset


codegen.mts
```diff
@@ -23,11 +23,9 @@ const config: CodegenConfig = {
           'lab',
           'operations',
           'organization',
-          'policy',
           'schema',
         ],
         scalarsOverrides: {
-          DateTime: { type: 'string' },
```
eddeee888, line 30: By removing `DateTime`, we are relying on Server Preset to wire up `graphql-scalars`'s recommended types, which is `string | Date`
We have to do this now, otherwise there's a mismatch of types in `SchemaPolicy.updatedAt`.

packages/services/api/src/shared/helpers.ts
```diff
@@ -7,7 +7,7 @@ import type {
   OrganizationSelector,
   ProjectSelector,
   TargetSelector,
-} from '../__generated__/types';
+} from '../__generated__/types.next';
```
eddeee888, line 10: I'm seeing a TS error if I use the old generated `types` file here. The reason being there's a type that overrides the Node's `Date` type being generated at the bottom of the file:

```ts
export type Date = Scalars["Date"];
```

Switching to the new types seems to fix this issue, so I think we can start using it

packages/services/api/src/modules/admin/module.graphql.mappers.ts
```diff
@@ -1,10 +1,11 @@
+import type { Scalars } from '../../__generated__/types.next';
 import type { AdminOrganizationStats } from '../../shared/entities';
 
 export type AdminQueryMapper = {};
 export type AdminStatsMapper = {
   period: {
-    from: string;
-    to: string;
+    from: Scalars['DateTime']['output'];
+    to: Scalars['DateTime']['output'];
```
eddeee888, line 8: This is required because I think we have been passing values for `DateTime` scalar slots as `Date`, but typed it as `string`.

n1ru4l (approved):

PR #5268, opened 2024-07-22
fix landing page bg for pricing

dimaMachina:
can't figure why linting doesn't fails locally @kamilkisiela 

https://github.com/user-attachments/assets/2fc1b288-6090-46ab-bb50-a5614ddfaf59


packages/web/docs/src/components/pricing.tsx
```diff
@@ -104,7 +104,7 @@ const OPERATIONS_EXPLAINER = 'GraphQL operations reported to GraphQL Hive';
 
 export function Pricing({ gradient }: { gradient: [string, string] }): ReactElement {
   return (
-    <div className="w-full">
+    <div className="w-full bg-neutral-900">
```
dimaMachina, line 107:
```suggestion
    <div className="w-full bg-dark">
```

dimaMachina:
> Why not change the font to be black on light mode instead? Having a dark background kind of misses the point of light mode? 😅

Landing page is not adapted to light mode, I did it to quick fix this bug

kamilkisiela (approved):

PR #5277, opened 2024-07-24
chore: publish CLI docker images for alpha versions 

n1ru4l:
### Background

Having docker images for alpha releases is nice and we want to use them within our pulumi deployment script for pending features.

### Description

This publishes a docker image with the same tag as the snapshot release listed version.

```
docker run ghcr.io/kamilkisiela/graphql-hive/cli:0.39.0-alpha-20240724164620-5e8722409ceed454769143c20c0b52bbd2fa50b4 \
  --version
@graphql-hive/cli/0.39.0-alpha-20240724164620-5e8722409ceed454769143c20c0b52bbd2fa50b4 linux-arm64 node-v22.4.1
```

~~Problem here is that `--version` shows the wrong version. I think we should change the docker step to use `npm install --global @graphql-hive/cli@version`.~~ Fixed since we now `npm install` the version within the image.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/.github/workflows/release-alpha.yaml
+++ b/.github/workflows/release-alpha.yaml
@@ -108,3 +108,30 @@ jobs:
           tar -xzf *.tar.gz
           # Expect the command to succeed
           ./hive/bin/hive --version || exit 1
+
+  extract-cli-version:
+    needs: [npm]
+    runs-on: ubuntu-22.04
+    name: Extract CLI version
+    outputs:
+      published: ${{ steps.cli.outputs.published }}
+      version: ${{ steps.cli.outputs.version }}
+    steps:
+      - name: Extract published version
+        if:
+          needs.npm.outputs.published && contains(needs.npm.outputs.publishedPackages,
+          '"@graphql-hive/cli"')
+        id: cli
+        run: |
+          echo '${{needs.npm.outputs.publishedPackages}}' > cli-ver.json
+          VERSION=`echo $(jq -r '.[] | select(.name | endswith("@graphql-hive/cli")).version' cli-ver.json)`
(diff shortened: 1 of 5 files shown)


.github/workflows/build-and-dockerize.yaml
```diff
@@ -55,6 +55,10 @@ jobs:
         if: ${{ inputs.build }}
         run: pnpm build
 
+      - name: CLI prepack
+        if: ${{ inputs.build }}
+        run: pnpm run --filter @graphql-hive/cli workspace-fix
```
n1ru4l, line 60: I had to add this step, otherwise, the CLI building fails for the alpha release - however, it does not fail for stable, which uses this same workflow 😕

.github/workflows/release-alpha.yaml
```diff
@@ -4,7 +4,7 @@ on:
 jobs:
   # Release alpha version on NPM for Hive libraries
   npm:
-    uses: the-guild-org/shared-config/.github/workflows/release-snapshot.yml@main
+    uses: the-guild-org/shared-config/.github/workflows/release-snapshot.yml@feat-export-outputs
```
n1ru4l, line 7: See https://github.com/the-guild-org/shared-config/pull/534, required for the `extract-cli-version` step.

kamilkisiela (approved):

PR #5283, opened 2024-07-25
Migrate `schema` module to Server Preset (Part 17)

eddeee888:
### Background

This PR migrates `schema` module to Server Preset


packages/services/api/src/modules/schema/module.graphql.mappers.ts
```diff
@@ -24,7 +23,22 @@ export type SchemaChangeApprovalMapper = SchemaCheckApprovalMetadata;
...
+      target: string;
+    };
+  }>;
+  pageInfo: Readonly<{
+    hasNextPage: boolean;
+    hasPreviousPage: boolean;
+    startCursor: string;
+    endCursor: string;
+  }>;
+}>;
```
eddeee888, line 41: This change in mapper is required because `PaginatedSchemaVersionConnection` comes from storage and the result is transformed in the [manager](https://github.com/kamilkisiela/graphql-hive/blob/53a0804434979f447c59b0a1c54ed668fc1ab808/packages/services/api/src/modules/schema/providers/schema-manager.ts#L370-L372) to have `organization`, `project` and `target`.

Because of this slightly wrong mapper type, server preset static analysis will make us implement `SchemaVersionConnection`, even when we know we don't have to at runtime.

packages/services/api/src/modules/schema/resolvers/FailedSchemaCheck.ts
```diff
@@ -0,0 +1,57 @@
...
+    return injector.get(SchemaCheckManager).getConditionalBreakingChangeMetadata(schemaCheck);
+  },
+  schemaPolicyErrors: ({ schemaPolicyErrors }, _arg, _ctx) => {
+    // FIXME: @eddeee888 check why this doesn't have runtime error
+    return schemaPolicyErrors as any;
+  },
+  schemaPolicyWarnings: ({ schemaPolicyWarnings }, _arg, _ctx) => {
+    // FIXME: @eddeee888 check why this doesn't have runtime error
+    return schemaPolicyWarnings as any;
+  },
```
eddeee888, line 56: The resolvers that are returning `SchemaPolicyWarningConnection` seem to detect potential runtime error because there's a difference between the [mapper](https://github.com/kamilkisiela/graphql-hive/blob/2af71a2bc0d74b0a46cf082a4b67566069dc27b9/packages/services/api/src/modules/schema/providers/models/shared.ts#L81-L89) vs the type from [storage model](https://github.com/kamilkisiela/graphql-hive/blob/7c476f941dd935460aba7e6e6b0b2079032f89d3/packages/services/storage/src/schema-change-model.ts#L797-L804):

| Field | Mapper | Storage | Compatible |
| -- | -- | -- | -- |
| message | `string` | `string` | ✅ |
| source | `string` | N/A | ❌ |
| line | `number` | `numer` or `null` or `undefined` | ❌ |
| column | `number` | `numer` or `null` or `undefined` | ❌ |
| ruleId | `string` | `string` or `null` | ❌ |
| endLine | `number` or `null` | `number` or `null` or `undefined` | ❌ |
| endColumn | `number` or `null` | `number` or `null` or `undefined` | ❌ |

If I try to change the mapper to the storage type, it works but [SchemaPolicyWarningConnection](https://github.com/kamilkisiela/graphql-hive/pull/5283/files#diff-8776f3d9c5421f59077d3af32a5fa8169786fc231630ea3f4ce5c7b1573ff13fR9-R12) would have type errors.
And changing the Storage zod model to match the mapper's might cause issue at runtime.

Either way, I think we'd need to fix some runtime implementation, and I'd need to look deeper to see which way approach is better. So, very open for suggestions how to fix this issue 🙂

Another option is to keep the FIXME and keep going as I don't like holding a big PR like this open for too long as it might have conflicts

codegen.mts
```diff
@@ -50,7 +49,7 @@ const config: CodegenConfig = {
           resolversNonOptionalTypename: {
             interfaceImplementingType: true,
             unionMember: true,
-            excludeTypes: ['TokenInfoPayload'],
+            excludeTypes: ['TokenInfoPayload', 'Schema', 'GraphQLNamedType'],
```
eddeee888, line 52: These abstract types are added here as we don't use `__typename` to handle them.

packages/services/api/src/modules/schema/resolvers/SingleSchema.ts
```diff
@@ -0,0 +1,22 @@
...
+     * This is because `number` can be serialized https://github.com/Urigo/graphql-scalars/blob/7780dac195bc168d6cbfb8c8ba4168a8829a4e65/src/scalars/iso-date/DateTime.ts#L34-L39
+     * But the type only has `Date | string` https://github.com/Urigo/graphql-scalars/blob/7780dac195bc168d6cbfb8c8ba4168a8829a4e65/src/scalars/iso-date/DateTime.ts#L84
+     * I think we should have an input and output type in `graphql-scalars` e.g.
+     * {
+     *   input: 'Date | string',
+     *   output: 'Date | string | number'
+     * }
+     */
+    return date as any;
+  },
```
eddeee888, line 21: As mentioned by the FIXME, this is because the codegen scalar type doesn't handle input/output. 
I'm thinking to leave it like this and update scalars package later 🙂

packages/services/api/src/modules/schema/resolvers/FailedSchemaCheck.ts
eddeee888: This is fixed in https://github.com/kamilkisiela/graphql-hive/pull/5283/commits/d4a46a246bc08a2216dd6a3edddd38d209de5ec1 by updating the `SchemaCheckWarning` type (used as mapper)

kamilkisiela (approved):

PR #5298, opened 2024-08-01
Bump server preset to v0.10.3

eddeee888:
### Background

v0.10.3 comes with a fix to add `__isTypeOf` to Object types that exist in multiple schema (using `extend`).
Without this fix, types that are union members or implementing types cannot use the `__isTypeOf` to handle abstract type. This is required for https://github.com/kamilkisiela/graphql-hive/pull/5294 because `organization.Organization` implements `__isTypeOf`

Note that this is not enforcing only one `__isTypeOf` per Object type. I think that is something we can add to the static analysis later.


eddeee888:
Closing as we are already on 0.10.4

PR #5306, opened 2024-08-02
Add features section (needs polish on images)

hasparus:
### Background

image

### Description

Added the features section. Need to upload proper cropped images so every section has one, but the layout for desktop is done. I imagine the mobile layout would be the same as with the old design, but let's circle back with designers.


hasparus:
I'm marking this one as ready for review. It will get changes to the screenshots and a dropdown instead of tabs on mobile, but feel free to look at the code, @saihaj.

PR #5307, opened 2024-08-03
feat: log requests and retries

n1ru4l:
Reverts kamilkisiela/graphql-hive#5304


packages/libraries/cli/src/commands/artifact/fetch.ts
```diff
@@ -48,6 +48,18 @@ export default class ArtifactsFetch extends Command<typeof ArtifactsFetch> {
       retry: {
         retries: 3,
       },
+      logger: {
+        info: (...args) => {
+          // Note:
+          // We log to std:error here to avoid piped stdout from being polluted.
+          if (this.flags.debug) {
```
kamilkisiela, line 55: Why is info an error?

packages/libraries/core/src/client/http-client.ts
```diff
@@ -96,7 +96,7 @@ export async function makeFetchCall(
 
   return await asyncRetry(
     async (bail, attempt) => {
-      logger.info(
+      logger?.info(
```
kamilkisiela, line 99: console.error(‘request started’)

kamilkisiela (changes requested):
The problem is not with polluted stdout, but printed happy-paths logs in non-debug mode

packages/libraries/cli/src/commands/artifact/fetch.ts
n1ru4l: For this command specifically, because some people run

```
hive artifact:fetch > foo.graphql
```

instead of

```
hive artifact:fetch --outputFile foo.graphql
```

If we write to std:error - we still support `--debug` for these people, as std:error would not be piped to the file. But I am also fine with writing to std:out. The bug earlier was due to us accidentally logging even if the `--debug` flag is not provided.

packages/libraries/core/src/client/http-client.ts
n1ru4l: I changed the logger to optional by default to avoid cluttering console log output if someone missed passing a logger.

packages/libraries/cli/src/commands/artifact/fetch.ts
kamilkisiela: But that’s my problem with it. Debug is special and whoever runs —debug should know what’s the deal with it.
If you curl to a file, you won’t do a verbose mode or something

n1ru4l: ok will change to info then

packages/libraries/core/src/client/http-client.ts
n1ru4l: done!

packages/libraries/cli/src/commands/artifact/fetch.ts
n1ru4l: done!

kamilkisiela (approved):

PR #5314, opened 2024-08-06
Ecosystem management

hasparus:
### Background

https://github.com/user-attachments/assets/a2724657-73ef-4416-8bfb-23038bbf53d1

### Description

Added the graph/chart ecosystem illustration.

There's still room for some polish, but I'm gonna do the next sections first.


packages/web/app/tailwind.config.cjs
```diff
@@ -70,10 +71,29 @@ module.exports = {
         700: '#008c66',
         800: '#007052',
         900: '#005b43',
+        // workaround for ESLint not respecting Tailwind settings paths in CI
```
hasparus, line 74: I'm confused why ESLint doesn't find the proper Tailwind config on CI, and this seems like something that will solve itself in time: I assume that eventually the app and the website will use the same color palette.

saihaj: cc @dimaMachina to help here

PR #5315, opened 2024-08-06
(4) Drop v2: Create Operation Modal

TuvalSimha:
## Summary

- Drop `Modal` component and used `Dialog` instead.
- Align design for all modals
- Added component to Storybook
- Added `toast` for better user experience

### DIFF
--- a/packages/web/app/src/components/target/laboratory/create-operation-modal.tsx
+++ b/packages/web/app/src/components/target/laboratory/create-operation-modal.tsx
@@ -1,14 +1,31 @@
 import { ReactElement } from 'react';
-import { useFormik } from 'formik';
+import { useForm, UseFormReturn } from 'react-hook-form';
 import { useMutation } from 'urql';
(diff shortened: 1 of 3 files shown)


packages/web/app/src/components/target/laboratory/create-operation-modal.tsx
```diff
@@ -50,6 +67,24 @@ const CreateOperationMutation = graphql(`
...
 export type CreateOperationMutationType = typeof CreateOperationMutation;
 
+const createOperationModalFormSchema = z.object({
+  name: z
+    .string({
+      required_error: 'Operation name is required',
+    })
+    .min(2, {
+      message: 'Operation name must be at least 2 characters long',
+    })
```
saihaj, line 77: I see the `yup` one had `3` so let's make it `3`
```suggestion
    .min(3, {
      message: 'Operation name must be at least 3 characters long',
    })
```

PR #5316, opened 2024-08-06
(5) Drop V2: Edit Operation Modal

TuvalSimha:
## Summary

- Drop `Modal` component and used `Dialog` instead.
- Align design for all modals
- Added component to Storybook
- Added `toast` for better user experience


packages/web/app/src/components/target/laboratory/edit-operation-modal.tsx
```diff
@@ -30,13 +46,32 @@ const UpdateOperationNameMutation = graphql(`
   }
 `);
 
+const editOperationModalFormSchema = z.object({
+  name: z
+    .string({
+      required_error: 'Operation name is required',
+    })
+    .min(2, {
+      message: 'Operation name must be at least 2 characters long',
```
saihaj, line 55:
```suggestion
    })
    .min(3, {
      message: 'Operation name must be at least 3 characters long',
```

PR #5319, opened 2024-08-07
(6) Drop V2: Callout component

TuvalSimha:


packages/web/app/src/components/ui/callout.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { cva, type VariantProps } from 'class-variance-authority';
+import { cn } from '@/lib/utils';
+import { InfoCircledIcon } from '@radix-ui/react-icons';
+
+const TypeToEmoji = {
+  default: '💡',
+  error: '🚫',
+  info: <InfoCircledIcon className="h-6 w-auto" />,
+  warning: '⚠️',
+};
```
saihaj, line 11: if we can try to utilize icons from all of these from radix that would be better because on some browsers I have seen emoji's don't render properly

TuvalSimha: Done! 🥇

PR #5325, opened 2024-08-09
Add enterprise focused cards & Get Started

hasparus:
### Background

image
image

### Description

Added Get Started and Enterprise-focused sections.

Text in the third card needs to be changed to fit better and address enterprise requirements.


saihaj (approved):
copy of this should be updated. Started discussion on Slack since not a blocker

PR #5329, opened 2024-08-09
Team section

hasparus:
### Background

image

### Description

Laid out the new team section.

Work in progress:

- [ ] Get new avatar illustrations for everybody
- [ ] Grab everybody's chosen social links
- Integrating it with a CMS like Notion would be an overkill probably, right? Can keep it in code or a CSV file in repo.

### DIFF
--- a/packages/web/docs/src/components/landing-page.tsx
+++ b/packages/web/docs/src/components/landing-page.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement } from 'react';
 import * as Tooltip from '@radix-ui/react-tooltip';
+import { cn } from '../lib';
 import { CallToAction } from './call-to-action';
 import { CheckIcon } from './check-icon';
 import { ArchDecoration, HighlightDecoration, LargeHiveIconDecoration } from './decorations';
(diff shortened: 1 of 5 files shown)


packages/web/docs/src/components/landing-page.tsx
```diff
@@ -76,16 +78,22 @@ export function IndexPage(): ReactElement {
...
   );
 }
 
-function GetStartedTodaySection() {
+function GetStartedTodaySection({ className }: { className?: string }) {
   return (
-    <section className="relative overflow-hidden bg-[#003834] p-12 text-center sm:rounded-3xl sm:p-24">
+    <section
+      className={cn(
+        'relative overflow-hidden bg-[#003834] p-12 text-center sm:rounded-3xl sm:p-24',
```
saihaj, line 93: should we move this `bg` color to tailwind theme?

hasparus: This is something like `green-1000` with a slight blue tint. I tried to use `green-1000` instead of it and it looked worse. I _think_ I'm willing to consider this one to be a one-off to avoid bloat the config.

PR #5330, opened 2024-08-11
(8) Drop V2: Setting User

TuvalSimha:


saihaj (changes requested):
couple things on this one.
1. It would be great UX if we can fill the full name and display name from already what user has in DB
2. The save button should be disabled if we one of the fields is empty. Since we require both to be filled. See video 


https://github.com/user-attachments/assets/3e80d614-3e5f-41aa-85a1-83a237bf0bcc

dotansimha (approved):

TuvalSimha:
Video:

https://github.com/user-attachments/assets/ec5dfb86-6e48-43a9-a3cf-5f6252c4c1f5

Before:
Screenshot 2024-12-09 at 17 32 10

After:
Screenshot 2024-12-09 at 17 31 39

@dotansimha

PR #5358, opened 2024-08-12
Move the rounded border class to a proper place

hasparus:
### Background

I fixed the missing border radius on cards earlier today, but I added it to _all the cards_ instead of just the 3 that needed it.

### Description

This PR moves the className, so now the cards that need to be rounded are still rounded and the cards that need to have a straight border have a straight border.


saihaj (approved):
great work!

PR #5359, opened 2024-08-12
Company testimonials

hasparus:
### Background

image

### Description

Added a desktop version of the testimonials section.

- [x] Mobile UI incoming tomorrow
- [ ] Need content ofc

### DIFF
--- a/packages/web/docs/src/components/company-testimonials.tsx
+++ b/packages/web/docs/src/components/company-testimonials.tsx
@@ -32,7 +32,7 @@ const testimonials: Testimonial[] = [
     href: '#TODO',
   },
   {
-    company: 'Meetup',
+    company: 'Linktree',
     logo: MeetupLogo,
     text: 'Hive offers an impressive suite of tools for managing and monitoring GraphQL schemas. The collaborative features, such as schema sharing and team-based permissions, have streamlined our development process.',
     person: [
@@ -47,7 +47,7 @@ const testimonials: Testimonial[] = [
     href: '#TODO',
   },
   {
-    company: 'Meetup',
+    company: 'Klarna',
     logo: MeetupLogo,
     text: 'Hive offers an impressive suite of tools for managing and monitoring GraphQL schemas. The collaborative features, such as schema sharing and team-based permissions, have streamlined our development process.',
     person: [
@@ -62,7 +62,7 @@ const testimonials: Testimonial[] = [
     href: '#TODO',
   },
   {
-    company: 'Meetup',
+    company: 'Uber',
     logo: MeetupLogo,
     text: 'Hive offers an impressive suite of tools for managing and monitoring GraphQL schemas. The collaborative features, such as schema sharing and team-based permissions, have streamlined our development process.',
     person: [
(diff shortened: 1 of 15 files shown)


packages/web/docs/src/components/company-testimonials.tsx
```diff
@@ -0,0 +1,176 @@
...
+  },
+  {
+    company: 'Linktree',
+    logo: MeetupLogo,
+    text: 'Hive offers an impressive suite of tools for managing and monitoring GraphQL schemas. The collaborative features, such as schema sharing and team-based permissions, have streamlined our development process.',
+    person: [
+      'Ryan Baldwin',
+      'Senior Backend Engineering Manager',
+      'https://github.com/user-attachments/assets/107be3fa-1051-49d5-b3d7-8fe374a3bd03',
+    ],
```
saihaj, line 42: can we just make this object?

```ts
person: {
  name: string,
  title: string,
  avatarUrl: string
}
```

packages/web/docs/src/components/company-testimonials.tsx
```diff
@@ -0,0 +1,176 @@
...
+    data: [
+      ['65M+', 'daily events processed'],
+      ['40%', 'more resource efficient'],
+    ],
```
saihaj, line 91: even this can have TS signature like this instead of just array

```ts
data: Array<{title: string, description: string}>
```

packages/web/docs/src/components/company-testimonials.tsx
hasparus: Sure. This will probably go to a JSON file. I'll switch all tuples to objects to make it less error-prone.

Do you mind if I keep the same keys I used in the named tuple types?

```ts
type _ = {
  person: Array<{title: string, description: string}> 
  data: Array<[numbers: string, description: string]>
}
```

saihaj: also for person I don't think we need it to be an array. We just need to show one pic at a time

saihaj (approved):
looks good. Left a design question on slack for mobile testimonial section

packages/web/docs/src/components/company-testimonials.tsx
```diff
@@ -95,37 +95,74 @@ const testimonials: Testimonial[] = [
...
+    const tabsList = tabsListRef.current;
+    if (!scrollview || !tabsList) return;
+    const scrollLeft = scrollview.scrollLeft;
+    const scrollWidth = scrollview.scrollWidth;
+    const index = Math.round((scrollLeft / scrollWidth) * testimonials.length);
+
+    const tabs = tabsList.querySelectorAll('[role="tab"]');
+    for (const [i, tab] of tabs.entries()) {
+      tab.setAttribute('data-state', i === index ? 'active' : 'inactive');
+    }
```
hasparus, line 112: It feels kinda primitive, but the scroll-snap event is still in the working group, so I'm doing this to sync scrolls to changes in Radix data-state.

packages/web/docs/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es2015",
```
hasparus, line 3: Next.js disregards this anyway, right? We could probably use `es2022` or even `esnext`.

PR #5370, opened 2024-08-13
Conditional Breaking Changes: Change default days

TuvalSimha:


packages/web/app/src/pages/target-settings.tsx
```diff
@@ -621,7 +624,8 @@ const ConditionalBreakingChanges = (props: {
               size="small"
               type="number"
               min="1"
-              max={targetSettings.data?.organization?.organization?.rateLimit.retentionInDays ?? 30}
+              max={retentionInDays}
+              defaultValue={values.period}
```
kamilkisiela, line 628: isn't that controlled by formik? I think it's not needed.

packages/web/app/src/pages/target-settings.tsx
```diff
@@ -488,6 +488,9 @@ const ConditionalBreakingChanges = (props: {
   const isEnabled = settings?.enabled || false;
   const possibleTargets = targetSettings.data?.targets.nodes;
   const { toast } = useToast();
+  const retentionInDays =
+    targetSettings.data?.organization?.organization?.rateLimit.retentionInDays ?? 30;
+  const defaultDays = retentionInDays >= 30 ? 30 : 7;
```
saihaj, line 493: can we make `30` and `7` constants in outside the component? or maybe even put them as a shared constant

PR #5379, opened 2024-08-14
Create Channel Modal: Make error more user friendly

TuvalSimha:


packages/web/app/src/components/project/alerts/create-channel.tsx
```diff
@@ -42,186 +100,258 @@ export const CreateChannelModal = ({
...
-        slackChannel: '',
-        endpoint: '',
-      },
-      validationSchema: Yup.object().shape({
-        name: Yup.string().required('Must enter name'),
-        type: Yup.mixed().oneOf(Object.values(AlertChannelType)).required('Must select type'),
-        slackChannel: Yup.string()
-          .matches(/^[@#]{1}/, 'Must start with a @ or # character')
-          .when('type', ([type], schema) =>
-            type === AlertChannelType.Slack ? schema.required('Must enter slack channel') : schema,
```
dotansimha, line 60: I think `Slack channel is required` is better wording here. Also in the other errors.

packages/web/app/src/components/project/alerts/create-channel.tsx
```diff
@@ -42,186 +100,258 @@ export const CreateChannelModal = ({
...
+            : null,
+      },
+    });
+    // When Data is null, it means that the user has not integrated with Slack App
+    if (!data) {
+      toggleModalOpen();
+      await toast({
+        title: 'Please integrate with Slack App',
+        description:
+          'The channel has been created but you will not receive alerts until you integrate with Slack App',
```
dotansimha, line 155: Hmm, what does it actually means? should the user do something? should we point them somewhere? 🤔 

Also, I think if you just check `!data` first, all Mutation errors will fall into that condition, so the other `else if` will never get called?

packages/web/app/src/components/project/alerts/create-channel.tsx
```diff
@@ -42,186 +100,258 @@ export const CreateChannelModal = ({
...
+        variant: 'destructive',
+        duration: 10_000,
```
dotansimha, line 157: I think we don't need to hide it, user doesn't need to "look for it" in case of an error?

packages/web/app/src/components/project/alerts/create-channel.tsx
```diff
@@ -42,186 +111,266 @@ export const CreateChannelModal = ({
...
-        }
-        if (data?.addAlertChannel.error) {
-          console.error(data.addAlertChannel.error);
-        }
-        if (data?.addAlertChannel.ok) {
-          toggleModalOpen();
-        }
-      },
-    });
+  const [query] = useQuery({
```
kamilkisiela, line 114: I think we should pass a fragment here and fetch it somewhere

packages/web/app/src/components/project/alerts/create-channel.tsx
```diff
@@ -42,186 +111,266 @@ export const CreateChannelModal = ({
...
-          )}
-          <p className="text-sm text-gray-500">
-            This will be displayed on channels list, we recommend to make it self-explanatory.
-          </p>
-        </div>
+  async function onSubmit(values: CreateChannelModalFormValues) {
+    if (
+      values.type === AlertChannelType.Slack &&
+      !values?.slackChannel?.startsWith('#') &&
+      !values?.slackChannel?.startsWith('@')
```
kamilkisiela, line 145: This should be defined in Zod schema

packages/web/app/src/components/project/alerts/create-channel.tsx
```diff
@@ -42,186 +111,266 @@ export const CreateChannelModal = ({
...
-            placeholder="Select channel type"
-            options={[
-              { value: AlertChannelType.Slack, name: 'Slack' },
-              { value: AlertChannelType.Webhook, name: 'Webhook' },
-              { value: AlertChannelType.MsteamsWebhook, name: 'MS Teams Webhook' },
-            ]}
-          />
-          {touched.type && errors.type && <div className="text-sm text-red-500">{errors.type}</div>}
-        </div>
+    if (values.endpoint && !values.endpoint.startsWith('http')) {
```
kamilkisiela, line 151: this as well, in zod schema

packages/web/app/src/components/project/alerts/create-channel.tsx
```diff
@@ -42,186 +111,266 @@ export const CreateChannelModal = ({
...
+                                  Slack - (Requires Integration)
+                                </SelectItem>
+                                <DocsLink
+                                  className="w-1/3"
+                                  href="https://the-guild.dev/graphql/hive/docs/management/organizations#slack"
+                                >
+                                  Read More
+                                </DocsLink>
+                              </div>
+                            )}
```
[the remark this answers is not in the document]
kamilkisiela: maybe we should point to org settings where the slack integration is (if user has access to it and if not a link to docs)?

Screenshot 2024-08-14 at 16 43 59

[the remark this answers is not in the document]
kamilkisiela: We could also add `<a id="integrations"/>` (if we don't have it yet) and add `#integrations` at the end of the link (to settings), so a user is scrolled there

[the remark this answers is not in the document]
kamilkisiela: plus let's do `Read more` instead of `Read More` and even better, `Instructions` or something similar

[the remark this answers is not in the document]
kamilkisiela: and if it's a link to organization settings we could say "Configure" or similar thing

PR #5380, opened 2024-08-14
Mobile feature tabs dropdown

hasparus:
### Background

https://github.com/user-attachments/assets/9443ccb6-d7c0-4d02-a3c3-4736f3bb0cf4

### Description

I'm not proud of this code, but I'm not a fan of importing 2 Radix alternative components and hiding them based on the viewport.

I styled the tabs to be a dropdown select on mobile viewports so we don't have a duplicate state management. (And we ship less JavaScript).


saihaj (changes requested):

hasparus:
Merging as we have https://github.com/kamilkisiela/graphql-hive/pull/5492 with a fix.

PR #5382, opened 2024-08-14
Team section mobile

hasparus:
### Background

image

### Description

It still needs pictures, and I'm not 100% happy with tablet-sized screens, but it matches the new designs for mobiles.

### DIFF
--- a/packages/web/docs/src/components/team-section.tsx
+++ b/packages/web/docs/src/components/team-section.tsx
@@ -14,11 +14,11 @@ export function TeamSection({ className }: { className?: string }) {
         className,
       )}
     >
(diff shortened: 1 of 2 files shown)


packages/web/docs/src/components/team-section.tsx
```diff
@@ -1,3 +1,4 @@
+import { Fragment } from 'react';
```
saihaj, line 1: CleanShot 2024-08-15 at 08 51 44@2x

saihaj (approved):
amazing ✨

packages/web/docs/src/components/team-section.tsx
hasparus: I have to get an autofix for this.

PR #5403, opened 2024-08-16
Compare schema coordinates of composable versions

kamilkisiela:
- [ ] truncate the table before merging

Previously, we compared schemas we received, without looking at composability status of the versions.


dotansimha:
@kamilkisiela what about this one?

PR #5404, opened 2024-08-16
Community section

hasparus:
### Description

Added the section.


packages/web/docs/src/components/masking-scrollview.tsx
```diff
@@ -0,0 +1,159 @@
+import { useEffect, useLayoutEffect, useReducer, useRef, useState } from 'react';
+
+export type MaskingScrollviewFade = {
+  y?: boolean;
+  x?: boolean;
```
saihaj, line 5: let's make it required and just provide `false`


```suggestion
  y: boolean;
  x: boolean;
```

packages/web/docs/src/components/masking-scrollview.tsx
```diff
@@ -0,0 +1,159 @@
...
+};
+
+export interface MaskingScrollviewProps {
+  fade: MaskingScrollviewFade;
+  children?: React.ReactNode;
```
saihaj, line 10: we can just provide `null`

```suggestion
  children: React.ReactNode;
```

packages/web/docs/src/components/masking-scrollview.tsx
```diff
@@ -0,0 +1,159 @@
...
+            ? `0% ${scrolledSides.top ? '-128px' : '0px'}, 0% ${scrolledSides.bottom ? '0px' : '-128px'}`
+            : '',
+        ]
+          .filter(Boolean)
+          .join(', '),
+        transition: shouldTransition
+          ? 'mask-position 0.5s ease, -webkit-mask-position 0.5s ease'
+          : '',
+      }}
+    >
```
saihaj, line 63: I would move this as a variable and memoize it

packages/web/docs/src/components/masking-scrollview.tsx
hasparus: I'm actually thinking this might be `"x" | "y"`, because the mask gradients combined at the corners look sorta off.

hasparus: Oh yeah, building for both `x` and `y` would need a different mask gradient.

image

packages/web/docs/src/components/masking-scrollview.tsx
hasparus: I'd have probably try to argue and provide benchmarks of both options, **but** changing `fade` to `"x" | "y"` trivializes it to 3 ternaries.

🤔 If this was a public library component in a widely used design system, I'd probably move this logic to a className / Tailwind plugin, to allow `"max-sm:scrollview-fade-x sm:scrollview-fade-y"`.

saihaj (approved):

PR #5466, opened 2024-08-19
Ecosystem cards

hasparus:
### Backgrou

https://github.com/user-attachments/assets/e5c869f2-b05b-435f-981b-5f5a0817e5d5

### Description

Finally wrapped this section up. Included some minor fixes and refactors for the other sections too.


packages/web/docs/src/components/community-section.tsx
```diff
@@ -76,7 +71,7 @@ export function CommunitySection({ className }: { className?: string }) {
         fade={{ y: true }}
       >
         {Array.from({ length: 4 }).map((_, i) => (
-          <div className="contents flex-col gap-6 sm:flex [&>:last-child]:flex-1" key={i}>
+          <div className="contents flex-col gap-6 sm:flex max-lg:[&>:last-child]:flex-1" key={i}>
```
saihaj, line 74: makes the key unique enough

```suggestion
          <div className="contents flex-col gap-6 sm:flex max-lg:[&>:last-child]:flex-1" key={`community-section-${i}`}>
```

hasparus: Isn't it exactly the same in terms of uniqueness? There's four columns now, with keys 0, 1, 2, 3.

hasparus: Wait up, I might have learned something.
Aren't keys scoped to the parent — only compared with siblings?

If they were global to the page, it would enable some neat remounts between parents that I usually used `React.createPortal` for.

hasparus: Found this in old React docs: https://legacy.reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings

saihaj: right! my bad I updated my knowledge :))

saihaj (approved):
left a design question on slack other than that good to go

PR #5474, opened 2024-08-20
fix: retry lock and avoid gateway timeout

n1ru4l:
### Background

Use a lock acquire timeout smaller than the gateway HTTP timeout and retry requests from CLI in case lock acquiring failures.

```
try to acquire lock
--- not acquired after 30 seconds ----> send response to CLI to retry
--- acquired --> global timeout of lock for 90 seconds; extend lock every few seconds + run action
```

That way we can prevent requests from the CLI reach the envoy timeout of 60 seconds - after the lock is acquired the actual schema publish usually takes around 5-15 seconds, so even in the worst-case scenario if the lock is acquired after 29 seconds, the total request time would not get close to 60 seconds.

The 90 seconds is a hard limit to avoid a resource being locked forever (such as in the case of the bug we encountered yesterday). After this one is exceeded we force release the lock with the trade-off that race conditions can happen as two actors are mutating the resources.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/shared/__tests__/mutex.spec.ts
+++ b/packages/services/api/src/modules/shared/__tests__/mutex.spec.ts
@@ -15,7 +15,7 @@ it('should allow only one lock at a time', async ({ expect }) => {
   // second lock shouldnt resolve
   await expect(Promise.race([throwAfter(50), lock2])).rejects.toBeTruthy();
 
-  unlock1();
+  await unlock1();
 
   // after the first lock releases, second one resolves
   await expect(lock2).resolves.toBeTruthy();
@@ -97,7 +97,7 @@ it('should keep auto-extending lock until unlock', async ({ expect }) => {
   // second lock still cannot be acquired resolve
   await expect(Promise.race([throwAfter(50), lock2])).rejects.toBeTruthy();
 
-  unlock();
+  await unlock();
 
   // only after unlock can the second lock be acquired
   await expect(lock2).resolves.toBeTruthy();
--- a/packages/services/api/src/modules/shared/__tests__/mutex.spec.ts
+++ b/packages/services/api/src/modules/shared/__tests__/mutex.spec.ts
@@ -31,43 +31,6 @@ it('should allow different locks at any time', async ({ expect }) => {
   await expect(mutex.lock('3', { signal })).resolves.toBeTruthy();
 });
 
-it('should cancel locking on abort signal', async ({ expect }) => {
-  const mutex = new Mutex(new Tlogger(), new Redis(differentPort()));
-
-  const [signal, abort] = createSignal();
-
-  const unlock1 = await mutex.lock('1', { signal });
-
-  const lock2 = mutex.lock('1', { signal });
-
-  abort();
-
-  await expect(lock2).rejects.toMatchInlineSnapshot('[Error: Locking aborted]');
-
-  unlock1();
-
-  // make sure that the aborted lock does not lock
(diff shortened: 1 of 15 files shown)


packages/services/api/src/modules/shared/__tests__/mutex.spec.ts
```diff
@@ -31,43 +31,6 @@ it('should allow different locks at any time', async ({ expect }) => {
...
-  const lock2 = mutex.lock('1', { signal: createSignal()[0] });
-
-  // second lock shouldnt resolve
-  await expect(Promise.race([throwAfter(50), lock2])).rejects.toBeTruthy();
-
-  abort();
-
-  // first lock is aborted, second one should resolve now
-  await expect(lock2).resolves.toBeTruthy();
-});
```
n1ru4l, line 69: if the abort signal/request is canceled, but the action is already being performed, we don't want to release the lock. Instead, we wait until the action is performed or the global resource lock timeout is exceeded.

packages/services/api/src/modules/shared/providers/mutex.ts
```diff
@@ -49,10 +60,10 @@ export class Mutex {
     this.logger = logger.child({ service: 'Mutex' });
     this.redlock = new Redlock([redis]);
     this.redlock.on('error', err => {
-      // these errors will be reported directly by the locking mechanism
       if (err instanceof ResourceLockedError) {
         return;
       }
+      // these errors will be reported directly by the locking mechanism
```
n1ru4l, line 66:
```suggestion
```

packages/services/api/src/modules/shared/providers/mutex.ts
```diff
@@ -49,10 +60,10 @@ export class Mutex {
     this.logger = logger.child({ service: 'Mutex' });
     this.redlock = new Redlock([redis]);
     this.redlock.on('error', err => {
-      // these errors will be reported directly by the locking mechanism
       if (err instanceof ResourceLockedError) {
```
n1ru4l, line 63:
```suggestion
      // these errors will be reported directly by the locking mechanism
      if (err instanceof ResourceLockedError) {
```

packages/services/api/src/modules/shared/providers/mutex.ts
```diff
@@ -40,6 +40,17 @@ export interface MutexLockOptions {
   autoExtendThreshold?: number;
 }
 
+/** Error indicating that a resource is locked and the lock can not be acquired within the provided time frame. */
+export class MutexResourceLockedError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'MutexError';
```
n1ru4l, line 47:
```suggestion
    this.name = 'MutexResourceLockedError';
```

packages/services/api/src/modules/shared/providers/mutex.ts
```diff
@@ -69,66 +80,101 @@ export class Mutex {
...
+        if (
+          err instanceof ExecutionError &&
+          err.message === 'The operation was unable to achieve a quorum during its retry window.'
+        ) {
+          throw new MutexResourceLockedError(`Resource "${id}" is locked.`);
+        }
+
+        throw err;
+      });
+
```
n1ru4l, line 121:
```suggestion

      logger.debug('Acquired lock (id=%s)', id);

```

PR #5479, opened 2024-08-20
Make ecosystem chart responsive

hasparus:
### Background

https://github.com/user-attachments/assets/247097f6-fd26-418b-8e39-b68c81b911b9

### Description

The ecosystem chart is now nice on mobiles, and we're almost ready to ship 🚢 

Used container queries to change CSS vars and text styles.

I considered using the Tailwind plugin https://github.com/tailwindlabs/tailwindcss-container-queries,
but I mostly wanted to change custom properties, not the styles, so that wouldn't make it more convenient.

### DIFF
--- a/packages/web/docs/src/components/ecosystem-management.module.css
+++ b/packages/web/docs/src/components/ecosystem-management.module.css
@@ -1,0 +1,50 @@
+.container {
+  container-type: inline-size;
+}
+
+.vars {
+  --node-w: 96px;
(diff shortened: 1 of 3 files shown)


packages/web/docs/src/components/ecosystem-management.module.css
saihaj, line 1: nice. I forgot we could use module css out of the box in Next

saihaj (approved):

PR #5480, opened 2024-08-21
Hive footer FAQ & Get Your API Right

hasparus:
### Background

https://github.com/user-attachments/assets/261a2ab2-23e2-48a6-95e4-e7349fc2d4ff

https://github.com/user-attachments/assets/126a69ba-7a6b-4fcd-a399-a1d9213346e0

### Description

Added the FAQs, "Get Your API Right" section and made some minor fixes.

### DIFF
--- a/packages/web/docs/src/components/call-to-action.tsx
+++ b/packages/web/docs/src/components/call-to-action.tsx
@@ -20,6 +20,10 @@ const variantStyles = {
     'bg-green-800 hover:bg-green-700 text-white' +
       ' focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-green-800/40',
   ),
+  tertiary: cn(
+    'bg-transparent text-green-1000' +
+      ' focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-green-800',
+  ),
 };
 
 type CallToActionVariant = keyof typeof variantStyles;
--- a/packages/web/docs/src/components/landing-page.tsx
+++ b/packages/web/docs/src/components/landing-page.tsx
@@ -8,6 +8,7 @@ import { CompanyTestimonialsSection } from './company-testimonials';
 import { ArchDecoration, HighlightDecoration, LargeHiveIconDecoration } from './decorations';
 import { EcosystemManagementSection } from './ecosystem-management';
 import { FeatureTabs } from './feature-tabs';
+import { FrequentlyAskedQuestions } from './frequently-asked-questions';
 import { Heading } from './heading';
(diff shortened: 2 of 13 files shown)


packages/web/docs/src/components/call-to-action.tsx
```diff
@@ -20,6 +20,10 @@ const variantStyles = {
     'bg-green-800 hover:bg-green-700 text-white' +
       ' focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-green-800/40',
   ),
+  tertiary: cn(
+    'bg-transparent text-green-1000' +
+      ' focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-green-800',
```
saihaj, line 25:
```suggestion
      'focus-visible:outline-2 focus-visible:outline-offset-1 focus-visible:outline-green-800',
```

packages/web/docs/src/components/landing-page.tsx
```diff
@@ -90,6 +96,8 @@ export function IndexPage(): ReactElement {
         <TeamSection className="mx-4 md:mx-6" />
         <CommunitySection className="mx-4 mt-6 md:mx-6" />
         <ToolsAndLibrariesCards className="mx-4 mt-6 md:mx-6" />
+        <FrequentlyAskedQuestions className="mx-4 md:mx-6" />
+        <GetYourAPIGameRightSection className="mx-4 md:mx-6" />
```
saihaj, line 100: later thing but wondering if the parent container can just add the appropriate margin horizontally and flex gap for vertical

saihaj (approved):

packages/web/docs/src/components/call-to-action.tsx
hasparus: This space is needed there because I use `+`.

I probably should just surrender and let VSCode wrap lines, right?

packages/web/docs/src/components/landing-page.tsx
hasparus: > container can just add the appropriate margin horizontally

Yup, +1. I didn't know before that the margins are consistent before I implemented those sections (because they aren't in feature tabs)

> flex gap for vertical

Those aren't consistent. It's either `mt-6` or 0, and adding a gap and negative margin feels worse.

packages/web/docs/src/components/call-to-action.tsx
saihaj: > I probably should just surrender and let VSCode wrap lines, right?

yeah 🥲

PR #5492, opened 2024-08-21
Update feature dropdown to pin to current

hasparus:
### Background

image
image
image

### Description

We now always display the options in order.
We also close the dropdown when the current option is clicked, what wasn't working before.


saihaj (approved):
nice!

PR #5496, opened 2024-08-23
Use latest version in checksum check

n1ru4l:
Using the latest composable version to produce a checksum can lead to a problem where publishing a new version is ignored, due to unchanged input schema, even though it actually leads to fixing the composition state of the latest version.

This PR uses latest version to produce a checksum (to detect identical versions and return early and avoid doing unnecessary work), and instead of using all subgraphs, it's using the relevant subgraph only.

### DIFF
--- a/packages/services/api/src/modules/schema/providers/models/composite.ts
+++ b/packages/services/api/src/modules/schema/providers/models/composite.ts
@@ -386,9 +386,9 @@ export class CompositeModel {
     }
 
     const checksumCheck = await this.checks.checksum({
-      existing: schemaVersionToCompareAgainst
+      existing: latest
         ? {
-            schemas: schemaVersionToCompareAgainst.schemas,
+            schemas: latest.schemas,
             contractNames: schemaVersionContractNames,
           }
         : null,
--- a/packages/services/api/src/modules/schema/providers/models/composite.ts
+++ b/packages/services/api/src/modules/schema/providers/models/composite.ts
@@ -163,9 +163,9 @@ export class CompositeModel {
     const comparedVersion = compareToPreviousComposableVersion ? latestComposable : latest;
 
     const checksumCheck = await this.checks.checksum({
-      existing: comparedVersion
+      existing: latest
         ? {
-            schemas: comparedVersion.schemas,
+            schemas: latest.schemas,
             contractNames: schemaVersionContractNames,
           }
         : null,
--- a/packages/services/api/src/modules/schema/providers/models/composite.ts
+++ b/packages/services/api/src/modules/schema/providers/models/composite.ts
@@ -98,7 +98,6 @@ export class CompositeModel {
     selector,
     latest,
     latestComposable,
-    schemaVersionContractNames,
     project,
     organization,
     baseSchema,
@@ -119,13 +118,13 @@ export class CompositeModel {
       isComposable: boolean;
       sdl: string | null;
       schemas: PushedCompositeSchema[];
+      contractNames: string[] | null;
     } | null;
     latestComposable: {
       isComposable: boolean;
       sdl: string | null;
       schemas: PushedCompositeSchema[];
     } | null;
-    schemaVersionContractNames: string[] | null;
     baseSchema: string | null;
     project: Project;
     organization: Organization;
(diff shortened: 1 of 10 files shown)


packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -532,9 +532,9 @@ export class SchemaPublisher {
         selector,
       });
 
-    const schemaVersionContracts = comparedSchemaVersion
+    const schemaVersionContracts = latestSchemaVersion
```
n1ru4l, line 535: I don't think we should use latestSchemaVersion here. We decided that schema versions will always be compared against the last composable version

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -720,22 +720,26 @@ export class SchemaPublisher {
...
+        this.schemaVersionHelper.getSchemaCompositionErrors(latestSchemaVersion),
       ]);
 
       schemaCheck = await this.storage.createSchemaCheck({
         schemaSDL: sdl,
         serviceName: input.service ?? null,
         meta: input.meta ?? null,
         targetId: target.id,
-        schemaVersionId: comparedVersion?.version ?? null,
+        schemaVersionId: latestVersion.version ?? null,
```
n1ru4l, line 742: Why change this?

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: The latest schema version contracts should only be used for the checksum.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -1674,12 +1686,9 @@ export class SchemaPublisher {
...
-    const comparedSchemaVersion = compareToPreviousComposableVersion
-      ? latestComposableSchemaVersion
-      : latestSchemaVersion;
-    const schemaVersionContracts = comparedSchemaVersion
+    const schemaVersionContracts = latestSchemaVersion
       ? await this.contracts.getContractVersionsForSchemaVersion({
-          schemaVersionId: comparedSchemaVersion.id,
+          schemaVersionId: latestSchemaVersion.id,
         })
       : null;
```
n1ru4l, line 1693: same as https://github.com/kamilkisiela/graphql-hive/pull/5496#discussion_r1728545765

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -720,22 +720,26 @@ export class SchemaPublisher {
           })) ?? null,
       });
     } else if (checkResult.conclusion === SchemaCheckConclusion.Skip) {
-      if (!comparedVersion || !comparedSchemaVersion) {
+      // SchemaCheckConclusion.Skip is returned by the checksum check
+      // when the schema is the same as the previous one.
+      // To perform the check we take the previous version (not necessarily composable)
+      // and that's why we ensure that `latestVersion` and `latestSchemaVersion` are available.
+      if (!latestVersion || !latestSchemaVersion) {
```
n1ru4l, line 727: Okay I think I now understand that in case there is a skip, the compared version is always the latest version.

packages/services/api/src/modules/schema/providers/models/composite.ts
```diff
@@ -163,9 +163,9 @@ export class CompositeModel {
     const comparedVersion = compareToPreviousComposableVersion ? latestComposable : latest;
 
     const checksumCheck = await this.checks.checksum({
-      existing: comparedVersion
+      existing: latest
         ? {
-            schemas: comparedVersion.schemas,
+            schemas: latest.schemas,
             contractNames: schemaVersionContractNames,
```
n1ru4l, line 169: let's pass the `contractNames` as part of the latest object, so it is better scoped?

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: got it

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l, line 535: Lets rename this to latestSchemaVersionContracts, to be more explicit

kamilkisiela: did that without even reading your comment :D

PR #5500, opened 2024-08-23
Remove console.logs, update cypress gh action, stick with Node v22.6

kamilkisiela:


n1ru4l (approved):

saihaj (approved):
we can add lint rule for this

kamilkisiela:
> we can add lint rule for this

then we will end up using `// eslint-disable-next-line no-console`. The console logs in a coordinates diff were intentional

PR #5504, opened 2024-08-23
feat: introduce timeout to s3 read requests

saihaj:
### Background

Closes https://github.com/kamilkisiela/graphql-hive/issues/5503

### Description

Right now we have the requests taking forever if there is degradation on underlying storage. This will ensure we stop the request so worker is not indefinitely running.

### DIFF
--- a/packages/services/cdn-worker/src/artifact-storage-reader.ts
+++ b/packages/services/cdn-worker/src/artifact-storage-reader.ts
@@ -22,7 +22,7 @@ type SDLArtifactTypes = `sdl${'.graphql' | '.graphqls' | ''}`;
 export type ArtifactsType = SDLArtifactTypes | 'metadata' | 'services' | 'supergraph';
 
 /** Timeout in milliseconds for S3 read calls. */
-const READ_TIMEOUT_MS = 20_000;
+const READ_TIMEOUT_MS = 5_000;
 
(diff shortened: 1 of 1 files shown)


packages/services/cdn-worker/src/artifact-storage-reader.ts
```diff
@@ -21,6 +21,9 @@ type SDLArtifactTypes = `sdl${'.graphql' | '.graphqls' | ''}`;
 
 export type ArtifactsType = SDLArtifactTypes | 'metadata' | 'services' | 'supergraph';
 
+/** Timeout in milliseconds for S3 read calls. */
+const READ_TIMEOUT_MS = 20_000;
```
dotansimha, line 25: I think @n1ru4l mentioned this should be 5s?

dotansimha (approved):
LGTM. Just please verify with @n1ru4l the value for the timeout

dotansimha:
rebased and fixed prettier ✅ @saihaj

packages/services/cdn-worker/src/artifact-storage-reader.ts
kamilkisiela, line 25:
```suggestion
const READ_TIMEOUT_MS = 5_000;
```

PR #5532, opened 2024-08-27
Extra timeout in CDN Worker's fetchText function

kamilkisiela:
Add a timeout of 5s (with 5 retries) to `fetchText` function (used to fetch data from S3/R2 in special cases where we want to read the object and transform it or pass it raw).

### DIFF
--- a/packages/services/cdn-worker/src/handler.ts
+++ b/packages/services/cdn-worker/src/handler.ts
@@ -10,16 +10,12 @@ import {
   MissingTargetIDErrorResponse,
 } from './errors';
 import type { KeyValidator } from './key-validation';
-import { logMsg } from './log';
 import { createResponse } from './tracked-response';
 
 async function createETag(value: string) {
-  logMsg('createETag');
   const myText = new TextEncoder().encode(value);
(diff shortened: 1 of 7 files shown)


n1ru4l (approved):

dotansimha (approved):

packages/services/cdn-worker/src/handler.ts
```diff
@@ -283,13 +283,6 @@ export function createRequestHandler(deps: RequestHandlerDependencies) {
     const kvStorageKey = `target:${targetId}:${storageKeyType}`;
     const rawValueAction = await deps
       .getArtifactAction(targetId, null, storageKeyType, null)
-      .catch(() => {
-        logMsg('ReadArtifactAction::retry');
-        // Do an extra attempt to read the value from the store.
-        // If we see that a single retry does not help, we should do a proper retry logic here.
-        // Why not now? Because we do have a new implementation that is based on R2 storage and this change is simple enough.
```
kamilkisiela, line 282: aws client has built-in retry now, so no need to keep it.

PR #5543, opened 2024-08-28
feat: CDN request retry race from multiple S3 buckets

n1ru4l:
### Background

Follow up for https://github.com/kamilkisiela/graphql-hive/pull/5542.

We are experiencing unreliabilities with R2, where requests hang for long time, even after attempting retries.

This is causing a small amount of requests to the CDN to fail. As a countermeasure, we already introduced writing artifacts to a new S3 bucket in a different region and another cloud provider (see https://github.com/kamilkisiela/graphql-hive/pull/5538).

We synced over the state from the other bucket, so both buckets are now the source of truth of the CDN state.

In this PR we want to use the new bucket (AWS S3) as a fallback in case the first bucket (CF R2) fails to serve a response.

### Description

This introduces the new bucket to the CDN worker and changes the business logic for first attempting a single read from R2 without retries and a timeout of 5 seconds, and then a follow-up race reads from S3 and R2. As soon as one of these buckets responds with an acceptable HTTP response, the other pending request will be canceled.

We also track the metrics, so we can gain more insight into how S3 performs in comparison to R2.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/deployment/services/cf-cdn.ts
+++ b/deployment/services/cf-cdn.ts
@@ -15,10 +15,12 @@ export class CDNSecret extends ServiceSecret<{
 export function deployCFCDN({
   environment,
   s3,
+  s3Mirror,
   sentry,
 }: {
   environment: Environment;
   s3: S3;
+  s3Mirror: S3;
   sentry: Sentry;
 }) {
   const cfConfig = new pulumi.Config('cloudflareCustom');
@@ -35,6 +37,7 @@ export function deployCFCDN({
     sentryDsn: sentry.enabled && sentry.secret ? sentry.secret?.raw.dsn : '',
     release: environment.release,
     s3,
+    s3Mirror,
   });
 
   const deployedCdn = cdn.deploy();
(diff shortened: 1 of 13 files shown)


deployment/services/cf-cdn.ts
```diff
@@ -15,10 +15,12 @@ export class CDNSecret extends ServiceSecret<{
 export function deployCFCDN({
   environment,
   s3,
+  s3Mirror,
   sentry,
 }: {
   environment: Environment;
   s3: S3;
+  s3Mirror: S3;
```
saihaj, line 23: in case someone is using Pulumi to self host they don't need to deploy mirror?

```suggestion
  s3Mirror: S3 | null;
```

kamilkisiela: Yes but Pulumi is really for us, and only a reference for people. Making s3Mirror nullable makes no sense from perspective of our deployment.

kamilkisiela (approved):

PR #5547, opened 2024-08-28
Account for usedOverridden arg in extractSuperGraphInformation

jdolle:
### Background

I have noticed fields provided by interfaces have the wrong subgraph associated with them in Hive's explorer.

### Description

Interfaces are being marked as resolved by the wrong subgraph in Hive. Although very poorly documented in the federation spec, there is a "usedOverridden" argument inside the "join__field" in these cases. Ignore any join__field directives with this argument.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


kamilkisiela:
Thank you! There is indeed such argument, we do cover that in our composition, but we forgot to include it here. Good catch.

Issue #1071, opened 2023-01-19
Audit log


kamilkisiela:
## Background

Audit logs provide a full record of all user activities and system events, scoped under each organization.

Having this kind of log will allow organizations to monitor all actions (and background actions) for security and compliance purposes.

## Implementation 

To achieve this kind of logging, we would need to have an easy-to-use technical mechanism to integrate with the crucial flows. 
Some of the flows might be triggered by a user (for example: CLI) or a machine (for example: CI/CD actions, GraphQL gateway), or by a background job (for example: Hive purge process)

### Events

To get started with this task, we'll need the following events to be fully covered (also see **Compliance** section): 

- User is being invited to an org (when? who was invited? who is the invitee?)
- User joined an org (when? approved? what was the referrer for the request?)
- User tried to join an org with an expired/invalid invite (when? can we correlate to an invite?)
- User has created a project/target (when? what was the project/target name?)
- Admin created a new role
- Admin assigned a new role to a user
- Admin removed a user 
- Admin transferred ownership
- CI made a schema check (when? what target/project?) 
- User did a schema publish (when? what target/project?) 
- Hive background job deleted old schema 
- Changes to project/org/target settings.
- Changes to schema policy settings under an org or a project.

### Storage 

We expect the audit log to contain a lot of records, and we expect to allow users to have time-series views on that data, we want to use ClickHouse. 

### Compliance

To be compliant with requirements, we need to allow admins to:

- View the list of events 
- Export to CSV based on a time range.

The following are nice-to-have, but can help with compliance:
- Filter events by date/time
- Filter by event type
- Filter by actor 

### API 

API needs to expose the list of events, including actor, date & time, and other significant information based on the event (based on the type of the event). 

To achieve such a thing, we can use an approach similar to ActivityLog implementation (a GraphQL `interface` and implementing `type`s).

Access to that part of the API needs to be allowed to the organization's admins. 

### UI

We can begin with a paginated list of recent events, and allow filtering based on event type and/or date/time range. By default, the list should show the N (30?) last events. 

## Definition of Done

Based on the definition above, this task can be split into multiple sub-tasks/standalone PRs:

1. Storage: deals with recording of the logs and actions. 
2. API: Expose the data through the GraphQL API/
3. UI: Show it to admins.

capaj:
I wrote a brief proposal here: https://github.com/kamilkisiela/graphql-hive/blob/audit-log-wip/docs/proposals/proposal-audit-log.md

[...]

PR #5564, opened 2024-08-30
feat: preflight scripts for laboratory

dimaMachina:
closes https://github.com/kamilkisiela/graphql-hive/issues/3612

Preview Link for docs: https://github.com/graphql-hive/console/pull/5564#issuecomment-2406985714


packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -445,14 +482,14 @@ function LaboratoryPageContent(props: {
       {!query.fetching && !query.stale && (
         <GraphiQL
           fetcher={fetcher}
-          showPersistHeadersSettings={false}
-          shouldPersistHeaders={false}
+          shouldPersistHeaders
```
dimaMachina, line 485: I removed it because it's better for UX to save headers in localStorage

packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -242,6 +243,40 @@ function Save(props: {
   );
 }
 
+const onModifyHeaders: ComponentProps<typeof GraphiQL>['onModifyHeaders'] = async (
```
dimaMachina, line 246: I will create PR for graphiql, this is very useful for folks, maybe a better name? this is to modify headers before execution

patches/@graphiql__react.patch
```diff
@@ -0,0 +1,423 @@
...
+ export declare type ExecutionContextType = {
+@@ -45,8 +45,13 @@ export declare type ExecutionContextProviderProps = Pick<UseAutoCompleteLeafsArg
+      * This prop sets the operation name that is passed with a GraphQL request.
+      */
+     operationName?: string;
++    /**
++     * Modify headers before execution
++     * e.g. for interpolating headers values `"myKey": "{{valueToInterpolate}}"`
++     */
++    onModifyHeaders?: (headers?: Record<string, unknown>) => MaybePromise<Record<string, unknown>>;
```
dimaMachina, line 362: cc what do you think about naming of this new prop

packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -242,6 +243,40 @@ function Save(props: {
...
+  headers = {},
+) => {
+  const { environmentVariables, logs } = await executeScript();
```
dimaMachina, line 249: cc @n1ru4l `executeScript` sends a message to the worker, is it fine to do it with `await`? I am not very familiar..

cypress.config.ts
```diff
@@ -13,7 +21,192 @@ export default defineConfig({
...
+        await client.end();
+        return res.rows;
+      }
+
+      on('task', {
+        connectDB,
+        async deleteUser(email = 'test@test.com') {
+          const [user] = await connectDB(`SELECT *
+                                          FROM users
+                                          WHERE email = '${email}';`);
```
saihaj, line 46: nit: style wise we can just have it all in one line
```suggestion
          const [user] = await connectDB(`SELECT * FROM users WHERE email = '${email}';`);
```

packages/services/api/src/modules/preflight-script/module.graphql.ts
```diff
@@ -0,0 +1,53 @@
+import { gql } from 'graphql-modules';
+
+export const typeDefs = gql`
+  type PreflightScript {
+    id: ID!
+    sourceCode: String!
+    createdAt: DateTime!
+    updatedAt: DateTime!
+    # TODO: should I add this field?
+    # createdBy: User!
```
saihaj, line 10: I think we should expose it. In our UI we could show various scripts available in the org for user to pick from. Like I can create a script and tell you go find the script I made

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,45 @@
...
+    { sourceCode }: PreflightScriptModule.CreatePreflightScriptInput,
+  ) {
+    const currentUser = await this.authManager.getCurrentUser();
+
+    return this.storage.createPreflightScript({
+      createdByUserId: currentUser.id,
+      sourceCode,
+      targetSlug,
+    });
+  }
```
saihaj, line 37: I started this convo on slack but we should only allow `Contributor` role and above to be able to create/execute scripts. ideally with #5668 we have a dedicated action for `createScript`, `updateScript`, `deleteScript`

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,45 @@
...
+import { AuthManager } from '../../auth/providers/auth-manager';
+import { Logger } from '../../shared/providers/logger';
+import { Storage } from '../../shared/providers/storage';
+import type { PreflightScriptModule } from './../__generated__/types';
+
+@Injectable({
+  global: true,
+  scope: Scope.Operation,
+})
+export class PreflightScriptProvider {
```
saihaj, line 11: also another place we would need to add Audit logs #1071

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  async createPreflightScript(
+    targetSlug: string,
+    { sourceCode }: PreflightScriptModule.CreatePreflightScriptInput,
+  ) {
+    const currentUser = await this.authManager.getCurrentUser();
+
+    return this.storage.createPreflightScript({
+      createdByUserId: currentUser.id,
+      sourceCode,
```
saihaj, line 34: we need to sanitize this input. Can't rely on FE validations on storing this.

Let's maybe run https://github.com/NodeSecure/js-x-ray or something like this tool so we can throw error on the client. Same logic should apply on updates. At a high level I can think of things that we don't want to store. That library has some good rules and we can apply most of those and test what all rules are unnecessary for us.
- anything that touches `Buffer` someone can do a Buffer overflow attack
- No `eval` or anything around the lines of that can cause eval to happen (yes I know we run in a worker but still)
- we don't want to store invalid JS code of user

We need to limit the size of these scripts. Someone can send in a 100mb minified JS just to spam us. Let's add a hard limit of 100KB (seems reasonable to me even minified [jquery](https://bundlephobia.com/package/jquery@3.7.1) is ~80KB) to begin with for these user scripts. If someone has bigger script we should learn first then increase limit.

packages/web/app/src/lib/preflight-sandbox/allowed-globals.ts
```diff
@@ -0,0 +1,61 @@
+// initial list comes from https://github.com/postmanlabs/uniscope/blob/develop/lib/allowed-globals.js
+export const ALLOWED_GLOBALS = [
```
saihaj, line 2: I don't think someone needs `Buffer` or even `Function`. Not sure if it is a requirement for the crypto lib we added. If it's not needed let's cut down on the globals

packages/services/storage/src/index.ts
```diff
@@ -3975,6 +3975,58 @@ export async function createStorage(
...
+          "id"
+          , "source_code" as "sourceCode"
+          , "target_id" as "targetId"
+          , "created_by_user_id" as "createdByUserId"
+          , to_json("created_at") as "createdAt"
+          , to_json("updated_at") as "updatedAt"
+      `);
+
+      return result && PreflightScriptModel.parse(result);
+    },
```
n1ru4l, line 4029: We want to avoid having this file grow infinitely, so instead you can add these methods and SQL queries within the preflight script provider file.

cypress.config.ts
```diff
@@ -13,7 +21,192 @@ export default defineConfig({
...
+        });
+        await client.connect();
+        const res = await client.query(query);
+        await client.end();
+        return res.rows;
+      }
+
+      on('task', {
+        connectDB,
+        async deleteUser(email = 'test@test.com') {
```
n1ru4l, line 43: In integration tests, we create new users with a new org for each test case. Unless there is a clear reason for doing the clean up we can drop it, as it adds additional stuff to maintain.

cypress.config.ts
```diff
@@ -13,7 +21,192 @@ export default defineConfig({
     POSTGRES_URL: 'postgresql://postgres:postgres@localhost:5432/registry',
   },
   e2e: {
-    setupNodeEvents(on) {
+    setupNodeEvents(on, config) {
```
n1ru4l, line 24: Could we potentially re-use the testkit we have within our integration tests here?

dimaMachina: yeah, we should

cypress.config.ts
dimaMachina: yeah would be nice to have 1 proper seeding

packages/migrations/src/actions/2024.09.26T00.00.00.create-preflight-scripts.ts
```diff
@@ -0,0 +1,22 @@
...
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "updated_at" timestamptz NOT NULL DEFAULT now(),
+  PRIMARY KEY ("id")
+);
+
+CREATE INDEX "document_preflight_scripts_connection_pagination" ON "document_preflight_scripts" (
+  "target_id" ASC,
+  "created_at" DESC,
+  "id" DESC
+);
```
n1ru4l, line 20: We only have one preflight script per organization, so we don't need pagination for it.

We just need an index on "target_id" to quickly find the preflight script for a given target.

Also we should avoid the creation of multiple preflight scripts for the same target. We can set a unique constraint on the `"target_id"` field to avoid duplicate records for the same target on the db level. But we should also have that being handled within our business logic.

packages/services/api/src/modules/preflight-script/module.graphql.ts
dimaMachina: cc @n1ru4l wdyt

packages/web/app/src/lib/preflight-sandbox/allowed-globals.ts
dimaMachina: removed all global variables which ends with `Buffer` and `Array`

packages/services/storage/src/index.ts
dimaMachina: moved

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
dimaMachina: done, limited to 5_000 chars

cypress/e2e/preflight-script.cy.ts
```diff
@@ -0,0 +1,224 @@
...
+beforeEach(() => {
+  cy.task('deleteUser');
+  cy.task<Token>('createUser').then(result => {
+    cy.task('createOrganization', result.sAccessToken);
+    cy.task('createProject', result.sAccessToken);
+    cy.setCookie('sRefreshToken', result.sRefreshToken);
+  });
+  // cy.task<Token>('login').then(result => {
+  //   cy.setCookie('sRefreshToken', result.sRefreshToken);
+  // });
```
saihaj, line 12:
```suggestion
```

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,215 @@
...
+  targetId: z.string(),
+  createdByUserId: z.union([z.string(), z.null()]),
+  createdAt: z.string(),
+  updatedAt: z.string(),
+});
+
+const scanner = new jsxray.AstAnalyser();
+await getLocalLang();
+
+function validateSourceCode(code: string) {
```
saihaj, line 30: I think you can just make this as a [`.superRefine`](https://zod.dev/?id=superrefine) for the zod

saihaj: actually keep it how you have it that way we can return errors early and not have to do a full zod parsing

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    if (error) {
+      const { message } = fromZodError(error);
+      return {
+        error: {
+          __typename: 'PreflightScriptError' as const,
+          message,
+        },
+      };
+    }
```
saihaj, line 136: move this before you are doing a DB insert. Actually now if I think about it on what I was saying about `zod.superRefine` will make much more sense

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,215 @@
...
+      },
+    });
+
+    const target = await this.storage.getTarget({
+      organizationId,
+      projectId,
+      targetId,
+    });
+
+    const result = await this.pool.maybeOne(sql`/* updatePreflightScript */
```
saihaj, line 172: missing zod validation before inserting

dimaMachina: previously in `storage` we did parse after instertion, is something changed? https://github.com/graphql-hive/platform/blob/562a002b2ae02342972477610319229098b1f14f/packages/services/storage/src/index.ts#L598

cypress.config.ts
dimaMachina: I create separated PR for this refactoring, but not sure why can't start docker containers locally
https://github.com/graphql-hive/platform/pull/5881
```
ERROR: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: unsupported frontend capability moby.buildkit.frontend.contexts
/Users/dmytro/Desktop/GUILD/graphql-hive/integration-tests:
```

packages/migrations/src/actions/2024.11.07T00.00.00.create-preflight-scripts.ts
```diff
@@ -0,0 +1,23 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2024.11.07T00.00.00.create-preflight-scripts.sql',
```
n1ru4l, line 4:
```
  name: '2024.11.07T00.00.00.create-preflight-scripts.sql',
```

TODO @dimaMachina use correct file extension

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,215 @@
...
+    logger: Logger,
+    private storage: Storage,
+    private session: Session,
+    private idTranslator: IdTranslator,
+    @Inject(PG_POOL_CONFIG) private pool: DatabasePool,
+  ) {
+    this.logger = logger.child({ source: 'PreflightScriptProvider' });
+  }
+
+  async getPreflightScript(targetSlug: string) {
```
n1ru4l, line 65: TODO @dimaMachina Seems like the variable name is wrong here.

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    if (error) {
+      const { message } = fromZodError(error);
+      return {
+        error: {
+          __typename: 'PreflightScriptError' as const,
+          message,
+        },
+      };
+    }
```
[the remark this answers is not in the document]
n1ru4l: What transaction are you referring to? You should validate the data you insert into the database before you insert, not afterwards.

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
n1ru4l: What you are referring to is to parse the data read from the database. There should still be input validation before inserting into the database.

packages/services/api/src/modules/preflight-script/module.graphql.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  extend type Mutation {
+    createPreflightScript(
+      selector: TargetSelectorInput!
+      input: CreatePreflightScriptInput!
+    ): PreflightScriptResult!
+    updatePreflightScript(
+      selector: TargetSelectorInput!
+      input: UpdatePreflightScriptInput!
+    ): PreflightScriptResult!
```
n1ru4l, line 30: Instead of having two mutation fields for creating and updating the preflight script we could instead just have a single one `Mutation.modifyPreflightScript`. Is there a specific reason for splitting creation and updating?

packages/services/api/src/modules/preflight-script/module.graphql.ts
```diff
@@ -0,0 +1,53 @@
+import { gql } from 'graphql-modules';
+
+export const typeDefs = gql`
```
n1ru4l, line 3: The module preflight-script is kind of part of the laboratory. So we can merge this module into that folder instead of having a new one.

packages/services/api/src/modules/preflight-script/module.graphql.ts
```diff
@@ -0,0 +1,53 @@
...
+      selector: TargetSelectorInput!
+      input: UpdatePreflightScriptInput!
+    ): PreflightScriptResult!
+  }
+
+  """
+  @oneOf
+  """
+  type PreflightScriptResult {
+    ok: PreflightScriptOkPayload
```
n1ru4l, line 37: We do not append `Payload` to the `Ok` type.

The naming convention is

```
MutationFieldNameResult
MutationFieldNameOk
MutationFieldNameError
```

packages/web/app/src/lib/preflight-sandbox/execute-script.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  const messages: LogMessage[] = [];
+
+  const log =
+    (level: 'log' | 'warn' | 'error' | 'info') =>
+    (...args: unknown[]) => {
+      console[level](...args);
+      messages.push(
+        `${level.charAt(0).toUpperCase()}${level.slice(1)}: ${args.map(String).join(' ')}`,
+      );
```
n1ru4l, line 65: The messages should be streamed to the main thread ass they occur not gathered and send to the main thread at the end of the execution of the preflight script.

Imagine a scenario where the worker would freeze. Having logs at least partially delivered to the main thread could give a hint on the issue.

packages/migrations/src/actions/2024.11.07T00.00.00.create-preflight-scripts.ts
dimaMachina: you are right, let's use `const filename = path.parse(import.meta.url).base` in all migrations? and `await import('./actions/...YOUR_MIGRATION_FILE'),` like you did for supertokens?

packages/services/api/src/modules/preflight-script/module.graphql.ts
kamilkisiela: Yes, especially if we ever want to move some pieces of it to GraphiQL as a plugin and embed into Hive Gateway

packages/services/api/src/modules/preflight-script/module.graphql.ts
dimaMachina: ty, btw we can catch these errors with graphql-eslint 😁

packages/services/api/src/modules/preflight-script/module.graphql.ts
dimaMachina: done, should I merge these modules to 1 as well? https://github.com/graphql-hive/console/blob/f02835c164ea71c38b15ffc2b71e67840c876ebf/packages/services/api/src/modules/lab/index.ts#L6-L20

dimaMachina:
> Here a summary of the review. I also commented parts of the code where eligible.

nice review changes!



> Here a summary of the review. I also commented parts of the code where eligible.
> 
> **Code Structure**
> 
> * Merge the folders of the preflight-script and collection module

did you mean `lab` module? or should I merge `collection` module to `lab` in this PR too? also see [one question](https://github.com/graphql-hive/console/pull/5564#discussion_r1847904006)

> **Permissions**
> 
> * replace `laboratory:createPreflightScript`  and `laboratory:updatePreflightScript` with `laboratory:modifyPreflightScript`

replaced

> **GraphQL**
> 
> * There seems no need to have both `Mutation.createPreflightScript` and `Mutation.updatePreflightScript` . Let’s replace both it with a single field `Mutation.updatePreflightScript` utilizing SQL `ON CONFLICT` with the unique target_id index for updating/modifying the existing record.

tbd

> * The types do not follow our convention for naming GraphQL fields:
>   ```shell
>   MutationFieldNameResult
>   MutationFieldNameOk
>   MutationFieldNameError
>   ```

fixed, [I can setup graphql-eslint naming convention to lint them](https://github.com/graphql-hive/console/pull/5564#discussion_r1847899365)!

> **Worker**
> 
> * The console logs should be streamed to and displayed on the main thread UI as soon as they occur. Currently, they are only displayed/sent at the end of the preflight script execution.
> * There is no timeout/check for a hanging worker. We should have a heartbeat with the worker and kill/restart the worker if it hangs + notify the user in case this happens.
> * I could not find anything that indicates how the worker is run on a different subdomain. As discussed at the beginning of this project we do not want to share the cookies and credentials from the web app with the worker for security reasons.

tbd

> **Security**
> 
> * Database input is not sanitised/validated before inserting into the database is some places.

tbd

packages/services/api/src/modules/preflight-script/module.graphql.ts
dimaMachina: > Is there a specific reason for splitting creation and updating?

just common practice of crud operations, changed to `updatePreflightScript` with SQL on conflict

packages/services/api/src/modules/preflight-script/module.graphql.ts
n1ru4l: that is what I meant.

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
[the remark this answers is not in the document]
dimaMachina: fixed https://github.com/graphql-hive/console/blob/ea79bc517c089d9819761c48733d09a9dfbee7b7/packages/services/api/src/modules/lab/providers/preflight-script.provider.ts#L15-L31

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
dimaMachina: thank you, super refine is nice, encapsulate error logic for source code https://github.com/graphql-hive/console/pull/5564#discussion_r1848045797

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
dimaMachina: fixed https://github.com/graphql-hive/console/pull/5564#discussion_r1848045797

packages/services/api/src/modules/preflight-script/module.graphql.ts
dimaMachina: done in https://github.com/graphql-hive/console/pull/5564/commits/c0f6488bad6cd19900aa96b0c23e9e6b293560a9

packages/services/api/src/modules/preflight-script/providers/preflight-script.provider.ts
```diff
@@ -0,0 +1,215 @@
...
+import { sql, type DatabasePool } from 'slonik';
+import { z } from 'zod';
+import { fromZodError } from 'zod-validation-error';
+import { getLocalLang, getTokenSync } from '@nodesecure/i18n';
+import * as jsxray from '@nodesecure/js-x-ray';
+import {
+  CreatePreflightScriptInput,
+  TargetSelectorInput,
+  UpdatePreflightScriptInput,
+} from '../../../__generated__/types';
```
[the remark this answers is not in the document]
dimaMachina: what should I use instead? define new `Entity` in https://github.com/graphql-hive/console/blob/preflight-script/packages/services/api/src/shared/entities.ts ?

packages/web/app/src/lib/preflight-sandbox/execute-script.ts
dimaMachina: thank you, I forgot to implement it, amazing catch, fixed in https://github.com/graphql-hive/console/pull/5564/commits/58739cd697a86039d2fd263590c80f4ab932573b


https://github.com/user-attachments/assets/66e96c96-3bf5-423e-8471-8a3bcf02019c

packages/migrations/src/actions/2024.11.19T00.00.00.create-preflight-scripts.ts
```diff
@@ -0,0 +1,26 @@
+import path from 'node:path';
+import { type MigrationExecutor } from '../pg-migrator';
+
+const filename = path.parse(import.meta.url).base;
+
```
n1ru4l, line 5: TODO @dimaMachina Since we bundle this file, this migration will be ran as `index.js` instead of this file name.

Please hardcode the filename here. as a string

n1ru4l: image

packages/services/api/src/modules/auth/lib/authz.ts
```diff
@@ -350,6 +350,7 @@ const actionDefinitions = {
   'target:modifySettings': defaultTargetIdentity,
   'laboratory:describe': defaultTargetIdentity,
   'laboratory:modify': defaultTargetIdentity,
+  'laboratory:modifyPreflightScript': defaultTargetIdentity,
```
n1ru4l, line 353: TODO @dimaMachina Grant this permission to `TargetAccessScope.SETTINGS` here: https://github.com/graphql-hive/console/blob/5838e821fc86d2f0a44003af775ceec2f05a4acd/packages/services/api/src/modules/auth/lib/supertokens-strategy.ts#L314

packages/services/api/src/modules/lab/module.graphql.ts
```diff
@@ -8,4 +8,45 @@ export default gql`
...
     mocks: JSON
   }
+
+  type PreflightScript {
+    id: ID!
+    sourceCode: String!
+    createdAt: DateTime!
+    updatedAt: DateTime!
+    # TODO: should I add this field?
+    # createdBy: User!
```
n1ru4l, line 18: TODO @dimaMachina remove this. We don't need it since right now you can only have one preflight script per organization. We might reconsider later on.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -0,0 +1,550 @@
+import { ComponentProps, Dispatch, useCallback, useEffect, useRef, useState } from 'react';
+import { clsx } from 'clsx';
+import type { editor } from 'monaco-editor';
+import { useMutation, useQuery } from 'urql';
+import { persist } from 'zustand/middleware';
+import { shallow } from 'zustand/shallow';
+import { createWithEqualityFn } from 'zustand/traditional';
```
n1ru4l, line 7: Lets remove these dependencies and use React.Provider if we need to teleport properties into the plugin.
I don't even understand why we even need to store this in local storage.

packages/services/api/src/modules/lab/module.graphql.ts
dimaMachina: removed

packages/migrations/src/actions/2024.11.19T00.00.00.create-preflight-scripts.ts
dimaMachina: hardcoded

deployment/utils/reverse-proxy.ts
```diff
@@ -14,6 +14,45 @@ export class Proxy {
...
+  ) {
+    new k8s.apiextensions.CustomResource(`internal-proxy-${dns.record}`, {
+      apiVersion: 'projectcontour.io/v1',
+      kind: 'HTTPProxy',
+      metadata: {
+        name: `internal-proxy-metadata-${dns.record}`,
+      },
+      spec: {
+        virtualhost: {
+          fqdn: 'lab-worker.app.graphql-hive.com',
```
dotansimha, line 34: This should be passed as an argument to `registerInternalProxy`. the `reverse-proxy.ts` file should not know about the lab

deployment/utils/reverse-proxy.ts
```diff
@@ -14,6 +14,45 @@ export class Proxy {
...
+          services: [
+            {
+              name: route.service,
+              port: 1234,
+            },
+          ],
+          pathRewritePolicy: {
+            replacePrefix: [
+              {
+                replacement: '/public/preflight-script-worker.js',
```
dotansimha, line 47: same here, pass as arg

deployment/services/proxy.ts
```diff
@@ -7,6 +7,25 @@ import { GraphQL } from './graphql';
...
+  reverseProxy: Proxy;
+  app: App;
+  environment: Environment;
+}) {
+  return reverseProxy.registerInternalProxy({ record: environment.appDns }, [
+    {
+      path: '/worker.js',
+      service: app.service,
+      upstreamUrl: `/public/preflight-script-worker.js?nocache=${Date.now()}`,
+      host: 'lab-worker.app.graphql-hive.com', // use env helpers of course
```
dotansimha, line 24: You can build this using `lab-worker.${environment.appDns}`

deployment/services/proxy.ts
```diff
@@ -7,6 +7,25 @@ import { GraphQL } from './graphql';
...
+  app,
+  environment,
+}: {
+  reverseProxy: Proxy;
+  app: App;
+  environment: Environment;
+}) {
+  return reverseProxy.registerInternalProxy({ record: environment.appDns }, [
+    {
+      path: '/worker.js',
```
dotansimha, line 21: I think you can pass `path` from `index.ts`, because you'll need the `/worker.js` value to build the `PREFLIGHT_SCRIPT_URL` for the app container

deployment/utils/reverse-proxy.ts
```diff
@@ -14,6 +14,45 @@ export class Proxy {
...
+      spec: {
+        virtualhost: {
+          fqdn: 'lab-worker.app.graphql-hive.com',
+        },
+        routes: routes.map(route => ({
+          conditions: [{ prefix: route.path }],
+          services: [
+            {
+              name: route.service,
+              port: 1234,
```
dotansimha, line 41: this should be calculated based on the `Service` you pass here. 

Check how it's done in the rest of the file: 

```
services: [
              {
                name: route.service.metadata.name,
                port: route.service.spec.ports[0].port,
              },
            ],
```

deployment/utils/reverse-proxy.ts
```diff
@@ -14,6 +14,45 @@ export class Proxy {
     private staticIp?: { address?: string; aksReservedIpResourceGroup?: string },
   ) {}
 
+  registerInternalProxy(
```
dotansimha, line 17: In general, let's make this function more generic. You can pass `service: k8s.Service`, `dnsRecord: string`, `url`, `...` and build the `HTTPProxy` in a generic way.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -0,0 +1,550 @@
+import { ComponentProps, Dispatch, useCallback, useEffect, useRef, useState } from 'react';
```
dotansimha, line 1: @dimaMachina this file seems a bit too messy and heavy. I know `graphiql` plugins are not the easiest, but I'm sure we can break it into smaller, simpler pieces. Can you please iterate it a bit to try to make it simpler and add some comments on the important flows?

dotansimha (changes requested):

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
dotansimha: Also, the state management section seems a bit messy? 🤔

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -0,0 +1,550 @@
...
+  },
+} satisfies Record<'script' | 'env', ComponentProps<typeof MonacoEditor>>;
+
+type ConsoleLog = { type: 'log'; message: string };
+
+type PreflightScriptResult =
+  | { environmentVariables: Record<string, unknown> }
+  | ConsoleLog
+  | { error: Error };
+
```
n1ru4l, line 144: @dimaMachina Please add a type property so we don't do all these property in checks. Its a bit messy.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -0,0 +1,550 @@
...
+  const handleRunScript = useCallback(async () => {
+    clearTimeout(timerId);
+    const env = envEditorRef.current?.getValue() ?? '';
+    preflightWorker.postMessage({
+      script: scriptEditorRef.current?.getValue() ?? '',
+      environmentVariables: env ? JSON.parse(env) : {},
+    });
+    const { setEnv } = usePreflightScriptStore.getState();
+    preflightWorker.onmessage = ({ data }: MessageEvent<PreflightScriptResult>) => {
```
n1ru4l, line 387: It seems like two event listeners are assigned to the preflightWorker, how does this even work? Can we have one event listener that handles both scenarios. Having two is a bit confusing and messy.

deployment/utils/reverse-proxy.ts
dimaMachina: there is no `k8s.Service` what did you mean?

packages/services/api/src/modules/auth/lib/authz.ts
dimaMachina: granted

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
dimaMachina: for which reason? just to remove dep? less bundle size? zustand is perfectly fits in current needs
- updates localStorage for disabling preflight script or saving env state in it
- I don't use any providers (I don't need them)
- [`onModifyHeaders`](https://github.com/graphql-hive/console/blob/7cf49048996e1965e2ef4cf34218f63e3fac6fd9/packages/web/app/src/pages/target-laboratory.tsx#L247) is outside of React tree and never rerenders

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
dimaMachina: > It seems like two event listeners are assigned to the preflightWorker

not two but 1, there is reassignment of `onmessage` handler

having one implies a refactoring which I couldn't succeded from first attempt, they do different job:
- `onmessage` from preflight modal handle logs/errors/env
- `onmessage` from execute query handle `errors/env`

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -0,0 +1,550 @@
...
+  const environmentVariables = env ? JSON.parse(env) : {};
+
+  if (disabled) {
+    return { environmentVariables };
+  }
+  clearTimeout(timerId);
+  preflightWorker.postMessage({ script, environmentVariables });
+
+  const { promise, resolve } = Promise.withResolvers<Exclude<PreflightScriptResult, ConsoleLog>>();
+  preflightWorker.onmessage = ({ data }: MessageEvent<PreflightScriptResult>) => {
```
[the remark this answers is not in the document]
dimaMachina: added stopping worker for this case https://github.com/graphql-hive/console/pull/5564/commits/4dde658bd12a770d4def3bf9a9c489024d38dd9d

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
dimaMachina: added

Urigo:
I think this feature deserves a screen cast to review?
maybe separate ones - one for user experience (that could later shared also on docs and release) and one with a walkthrough of the code

.github/workflows/tests-e2e.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: setup environment
         uses: ./.github/actions/setup
         with:
-          codegen: false
+          codegen: true
```
n1ru4l, line 30: we now use the integration test seed here which uses client preset.

deployment/index.ts
```diff
@@ -333,12 +333,10 @@ deployCloudFlareSecurityTransform({
     // Staging
     'staging.graphql-hive.com',
     'app.staging.graphql-hive.com',
-    'lab-worker.staging.graphql-hive.com',
     'cdn.staging.graphql-hive.com',
     // Dev
     'dev.graphql-hive.com',
     'app.dev.graphql-hive.com',
-    'lab-worker.dev.graphql-hive.com',
```
n1ru4l, line 341: no longer needed, we don't need to run on a different domain to isolate the worker.

deployment/utils/reverse-proxy.ts
```diff
@@ -14,6 +14,67 @@ export class Proxy {
...
```
n1ru4l, line 17: @dotansimha we can remove this, right?

docker/.end2end.env
n1ru4l, line 1: I deleted this file since it is almost identical with the existing env file we have in the integration tests folder except the `SUPERTOKENS_API_KEY` xD

I unified both, in the future we can probably even use `docker-compose.integration-tests.yml` for end2end tests instead of having yet another file.

docker/docker-compose.end2end.yml
```diff
@@ -34,5 +34,9 @@ services:
     networks:
       - 'stack'
 
+  supertokens:
+    ports:
+      - '3567:3567'
```
n1ru4l, line 39: needs to be exposed for the seed script

integration-tests/testkit/auth.ts
```diff
@@ -131,6 +131,7 @@ const createSession = async (
      */
     return {
       access_token: data.accessToken.token,
+      refresh_token: data.refreshToken.token,
```
n1ru4l, line 134: we need this in cypress for "logging" in the user

integration-tests/testkit/seed.ts
```diff
@@ -57,21 +50,29 @@ import {
...
-  };
-
   function createConnectionPool() {
+    const pg = {
+      user: ensureEnv('POSTGRES_USER'),
+      password: ensureEnv('POSTGRES_PASSWORD'),
+      host: ensureEnv('POSTGRES_HOST'),
+      port: ensureEnv('POSTGRES_PORT'),
+      db: ensureEnv('POSTGRES_DB'),
+    };
```
n1ru4l, line 74: Delaying this assertion because it is not required for seeding within the e2e tests.

package.json
```diff
@@ -44,7 +44,7 @@
     "seed": "tsx scripts/seed-local-env.ts",
     "start": "pnpm run local:setup",
     "test": "vitest",
-    "test:e2e": "CYPRESS_BASE_URL=$HIVE_APP_BASE_URL cypress run",
+    "test:e2e": "CYPRESS_BASE_URL=$HIVE_APP_BASE_URL cypress run --browser chrome",
```
n1ru4l, line 47:
```suggestion
    "test:e2e": "CYPRESS_BASE_URL=$HIVE_APP_BASE_URL cypress run",
```

TODO @n1ru4l  need to check whether it runs without chrome

packages/web/app/package.json
```diff
@@ -131,7 +133,8 @@
     "vite-tsconfig-paths": "5.1.2",
     "wonka": "6.3.4",
     "yup": "1.4.0",
-    "zod": "3.23.8"
+    "zod": "3.23.8",
+    "zustand": "5.0.1"
```
n1ru4l, line 137: TODO @n1ru4l remove dependency

packages/web/app/src/lib/preflight-sandbox/shared-types.ts
n1ru4l, line 1: Inspired by @jasonkuhrt I tried some namespace action. 😄 

I kind of like that type types are isolated.

cypress.config.ts
```diff
@@ -1,19 +1,38 @@
...
   video: isCI,
   screenshotOnRunFailure: isCI,
   defaultCommandTimeout: 15_000, // sometimes the app takes longer to load, especially in the CI
   retries: 2,
-  env: {
-    POSTGRES_URL: 'postgresql://postgres:postgres@localhost:5432/registry',
-  },
   e2e: {
     setupNodeEvents(on) {
+      on('task', {
```
n1ru4l, line 22: note: this is executed on a different thread than the tests, so we cannot simply inject the whole seed object into the tests. For now, we only need these seeds. So it's fine. Later one we might want to investigate into another solution for injecting the seed.

n1ru4l (approved):

package.json
n1ru4l, line 47: I noticed that the chrome is not used by default on CI. Both locally and on CI chrome seems to be less flaky.

dotansimha (approved):

dotansimha:
@n1ru4l rebased :) 

btw, what about this one? https://github.com/graphql-hive/console/pull/5748

PR #5567, opened 2024-08-30
Hive navigation

hasparus:
### Background

https://github.com/user-attachments/assets/7b2f0c23-e753-4ab7-915b-2d0ccca9db78

image

### Description

Added the new navbar, footer, and made _a lot_ of minor fixes and small changes requested by the design.

**Watch out:** This PR uses a tarballed version of shared components library. We need to publish it before merging.

### To-do

- [ ] I need to update the header height css var so the Introduction in sidebar doesn't clip when we scroll down


packages/web/docs/theguild-components-7.0.0-hasparus.22.tgz
saihaj, line 1: remove this?

saihaj: I approved all the components PRs so let's get a new release out and use that instead

hasparus: Done

packages/web/docs/src/components/landing-page.tsx
```diff
@@ -27,6 +27,8 @@ import { TeamSection } from './team-section';
 export function IndexPage(): ReactElement {
   return (
     <Tooltip.Provider>
+      {/* .nextra-sidebar-container styles here are hacks. the Sidebar component should be pulled out customized and swapped */}
```
hasparus, line 30: The implementation here is _questionable_, but pulling out the Sidebar from `nextra-theme-docs` to `@theguild/components` and changing its styles would be a little bit more effort.

I'm sorry, forgive me, and let's refactor it later?

hasparus: Okay, I pulled it to a CSS file, as I'm adding a .light class to the parent of the sidebar on mount.

PR #5574, opened 2024-09-05
docs: proposal for usage reporting and persisted documents

n1ru4l:
introduces new pages usage-reporting and persisted documents that show how to configure the gateway for the supported providers.


packages/web/docs/src/pages/docs/gateway/persisted-documents.mdx
```diff
@@ -0,0 +1,57 @@
+import { Callout, Cards, Tabs } from '@theguild/components'
+
+# Persisted Documents
```
ardatan, line 3: This skips all the options of providing a custom store etc here;
https://the-guild.dev/graphql/mesh/v1/serve/features/performance/persisted-operations

packages/web/docs/src/pages/docs/gateway/usage-reporting.mdx
```diff
@@ -0,0 +1,128 @@
...
+
+If you want to report usage metrics to a Apollo GraphOS, configure your `gateway.config.ts` file as
+following.
+
+```ts filename="gateway.config.ts"
+import { defineConfig } from '@graphql-hive/gateway'
+
+export const gatewayConfig = defineConfig({
+  reporting: {
+    type: 'graphos',
```
ardatan, line 97: https://the-guild.dev/graphql/mesh/v1/serve/schema-registry/graphos#serve-supergraph-from-apollo-graphos
Where to put this then?

packages/web/docs/src/pages/docs/gateway/persisted-documents.mdx
dotansimha: 2 options here:
1- demonstrate it in a tab or something similar
2- point users to code examples instead of creating a new title for it. it feels too dense and hard to understand when there's a super long page with a lot of titles

ardatan: Since they are using the same plugin, and configuration option. I don't think this is too much;
https://the-guild.dev/graphql/mesh/v1/serve/features/performance/persisted-operations#extracting-client-operations
Other than that, all the other options are just all about one small paragraph as you can see here (similar to allowing arbitrary operations)
the-guild.dev/graphql/mesh/v1/serve/features/performance/persisted-operations

dotansimha (approved):

PR #5581, opened 2024-09-05
Ecosystem Illustration Polish

hasparus:
### Description

Added ring on hover and made nodes cycle through their edges.


packages/web/docs/src/components/ecosystem-management.tsx
```diff
@@ -284,16 +323,50 @@ function EdgeLabel(props: EdgeLabelProps) {
...
+        }
+
+        hovered.current = true;
+
+        if (edges.includes(highlightedEdge!)) return;
+        onHighlight([edges[0]]);
+      }}
+      onPointerOut={event => {
+        if (
+          !event.currentTarget.contains(event.relatedTarget as Node) &&
```
hasparus, line 358: onPointerOut and onPointerOver trigger all the time _inside_ of the chart Node, so we need to check if the `relatedTarget` is inside of `currentTarget`.

dotansimha (approved):
Perfect!

PR #5593, opened 2024-09-08
Ecosystem Management GraphQL Logo Gradient

hasparus:
### Background

image

### Description

The GraphQL logo that was added in Kamil's content branch didn't use the same gradient-ids, so it wasn't included in the Safari-specific fix I made before.

- This PR moves all files related to the Ecosystem Management section out of `/public/` and into `components/ecosystem-management` folder.
- `new URL()` lets the bundler copy the SVG to the static dir as if we imported and give us the resulting path, so we don't have to remember about `basePath` (😢  flashbacks from the bug during prod release).
- The GraphQL logo we use for the client is moved to the `ecosystem-management.svg` to use the same (by id, not just same-looking) gradients.


packages/web/docs/src/components/ecosystem-management/index.tsx
```diff
@@ -6,11 +6,13 @@ import {
...
 } from '@theguild/components';
-import { cn } from '../lib';
-import { BookIcon } from './book-icon';
-import { CheckIcon } from './check-icon';
+import { cn } from '../../lib';
+import { BookIcon } from '../book-icon';
+import { CheckIcon } from '../check-icon';
 import styles from './ecosystem-management.module.css';
 
+const svgHref = new URL('./ecosystem-management.svg', import.meta.url).toString();
```
hasparus, line 14: `new URL` moves the responsibility of remembering about `basePath` to the bundler akin to Webpack's `file-loader`

`public` is nice and works everywhere, but it requires careful handling 😓

saihaj (approved):

saihaj (approved):

PR #5614, opened 2024-09-20
ci-cd.mdx: Revised the document for better clarity.

simnalamburt:
### Background
I am directly fixing the parts that were confusing when initially setting up Hive CI/CD.

### Description
- Updated to use `actions/checkout@v4` instead of `v3`
- Used `HIVE_TOKEN` directly without exporting as an environment variable
- Documented that the `HIVE_ENDPOINT` environment variable is optional
- Clarified that 'schema.graphql' is a file path

### DIFF
--- a/packages/web/docs/src/pages/docs/other-integrations/ci-cd.mdx
+++ b/packages/web/docs/src/pages/docs/other-integrations/ci-cd.mdx
@@ -26,11 +26,11 @@ is active, you may specify an additional `--github` flag to report the results b
 Check Suite (for `schema:check` and `schema:publish` commands):
 
 ```bash
-hive schema:publish schema.graphql --github
+hive schema:publish path/to/schema.graphql --github
 ```
 
 ```bash
-hive schema:check schema.graphql --github
+hive schema:check path/to/schema.graphql --github
 ```
 
 ### GitHub Workflow for CI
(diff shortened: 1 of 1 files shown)


packages/web/docs/src/pages/docs/other-integrations/ci-cd.mdx
```diff
@@ -71,16 +71,15 @@ jobs:
   publish:
     runs-on: ubuntu-22.04
     steps:
-      - name: checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: schema publish
         env:
-          HIVE_TOKEN: ${{ secrets.HIVE_TOKEN }}
```
saihaj, line 78: we should still keep the `HIVE_TOKEN`

saihaj: I guess what you changed would work but I have always used environment and that's how I have seen most docs

simnalamburt: I tried it in my company's CI/CD, and it worked fine without the `$HIVE_TOKEN` environment variable. Is there a specific reason why it's needed? Aren't we already passing it as a command line argument anyway?

PR #5616, opened 2024-09-22
feat: Track clients & operations filters using search params

davidguilherme:
### Background

At Toast, every team has a weekly metric review meeting where teams go through their dashboards and metrics for their services and SPAs. Teams have been using Hive Insights for client metrics. As teams go through a spreadsheet with all dashboard links, it has been a bit annoying to have to set the filters every time.

This PR adds the clients and operations filters to the search params (same done already for the date presets).

### Description

- Created a hook to manage any search params filter.
- Changed clients and operations filters to use `useSearchParamsFilter` instead of `useState`.
- Fixed `useDateRangeController` to not remove current search params when the user changes it.
- Changed Fallback component to display a more generic message as there might not be stats for a given combination of filters (clients, operations, date).
- Changed clients and operations triggers so it ignores values that are not present in the backend response.
- Added a new util helper to pick partial objects from a set of keys

### Checklist

- [ ] Input validation
  - I tested with a couple inputs (invalid ones, with special characters, long strings). Same as below. Should I truncate the list?
- [ ] Memory management
  - Should we truncate to specific size for clients and operations filter?

### DIFF
--- a/packages/web/app/src/components/target/insights/Fallback.tsx
+++ b/packages/web/app/src/components/target/insights/Fallback.tsx
@@ -29,7 +29,7 @@ export function OperationsFallback({
             <AlertCircleIcon className="size-4" />
             <AlertTitle>No stats available yet.</AlertTitle>
             <AlertDescription>
-              There is no information available for the selected date range.
+              There is no information available for the selected filters.
             </AlertDescription>
           </Alert>
         </div>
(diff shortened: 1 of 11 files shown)


kamilkisiela:
I merged your PR into a temporary branch (due to limitations of our CI, we can't test PRs from external contributors on their branch), I will test it there and merge to `main`.

PR #5622, opened 2024-09-27
#2 preflight scripts (db + cypress e2e tests)

dimaMachina:


cypress.config.ts
```diff
@@ -16,21 +22,175 @@ export default defineConfig({
...
+          const [user] = await connectDB(`SELECT *
+                                          FROM users
+                                          WHERE email = '${email}';`);
+          if (user) {
+            await connectDB(`
+BEGIN;
+
+DELETE FROM organizations WHERE user_id = '${user.id}';
+DELETE FROM users WHERE id = '${user.id}';
+DELETE FROM supertokens_emailpassword_user_to_tenant WHERE email = '${email}';
```
dimaMachina, line 53: maybe we should add `ON DELETE CASCADE` for this table? otherwise I was unable to recreate a user with same email after deletion him in `users` table

cypress/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "lib": ["es5", "dom"],
+    "target": "es2021",
+    "lib": ["es2021", "dom"],
```
dimaMachina, line 4: to use `String#replaceAll` and remove type errors

packages/migrations/src/actions/2024.09.26T00.00.00.create-preflight-scripts.ts
```diff
@@ -0,0 +1,22 @@
...
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "updated_at" timestamptz NOT NULL DEFAULT now(),
+  PRIMARY KEY ("id")
+);
+
+CREATE INDEX "document_preflight_scripts_connection_pagination" ON "document_preflight_scripts" (
+  "target_id" ASC,
+  "created_at" DESC,
+  "id" DESC
+);
```
dimaMachina, line 20: we do not need it right?
```suggestion
```

packages/services/api/src/modules/collection/resolvers.ts
```diff
@@ -41,7 +41,7 @@ const OperationValidationInputModel = zod
   .partial()
   .passthrough();
 
-async function validateTargetAccess(
+export async function validateTargetAccess(
```
dimaMachina, line 44: Is it okay to reexport this function from `collections/resolvers`?

packages/services/api/src/modules/preflight-script/module.graphql.ts
```diff
@@ -0,0 +1,53 @@
...
+      input: UpdatePreflightScriptInput!
+    ): PreflightScriptResult!
+  }
+
+  """
+  @oneOf
+  """
+  type PreflightScriptResult {
+    ok: PreflightScriptOkPayload
+    error: PreflightScriptError
```
dimaMachina, line 38: can I just use `interface Error` here?
```suggestion
    error: Error
```

PR #5623, opened 2024-09-27
Navigation border and Components v7 bump

hasparus:
### Description

This PR pulls in newest Components `v7` to add the bottom border to HiveNavigation.
I went through a few versions testing it, fixing merge conflicts with `main` and doing minor fixes.

Added `.hive-search-results` class with a patch to `nextra-theme-docs` to have something stable to select for styling.


saihaj:
Another regression is the `cmd+k` and `esc` in search bar

**Current Prod**

https://github.com/user-attachments/assets/f90d5a7f-59b9-4886-8232-d1894ef3f210


**Preview**

https://github.com/user-attachments/assets/0bdfeb31-e960-44a8-b8c9-9911f27b8b28

saihaj:
Another one in nav is the hover color state 

**Prod**

https://github.com/user-attachments/assets/5cb4a74e-a10b-423b-9146-1aaabe797c87

**Preview**

https://github.com/user-attachments/assets/133b923a-7b8f-47c2-8d08-66b273a9cd99

saihaj:
Cannot open the dropdown on the current preview for schema registry & other options


**Prod**

https://github.com/user-attachments/assets/61d5e5ae-15a3-4d46-b6a8-0175ca380a4e


**Preview**

https://github.com/user-attachments/assets/f4723460-f057-430c-9032-85e01ed03566

saihaj:
On small screens unable to scroll.

**Prod**

https://github.com/user-attachments/assets/0d84b7c8-c3b9-4aba-90e5-9ce30cfd6e52

**Preview**

https://github.com/user-attachments/assets/76908bbe-dfa9-43c2-8bc8-50e983cc896e

saihaj:
on small screens unable to scroll. See video (left preview & right prod)

https://github.com/user-attachments/assets/3c8221d9-0416-4baf-82b7-b617f726e486

saihaj (dismissed):
bunch of things feel regressed with this new upgrade. Let's look into those before merging

dimaMachina:
> Another regression is the `cmd+k` and `esc` in search bar
> 
> **Current Prod**
> 
>  CleanShot.2024-10-08.at.15.39.14.mp4 
> 
> **Preview**
> 
>  CleanShot.2024-10-08.at.15.38.14.mp4

This is new nextra search behaviour

packages/web/docs/src/components/landing-page.tsx
```diff
@@ -29,9 +29,13 @@ const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffec
...
-    // TODO: Accept a className for sidebar in the theme config?
-    const sidebarContainer = document.querySelector('.nextra-sidebar-container');
-    sidebarContainer?.parentElement?.classList.add('light');
+    // We add .light class to body to style the Headless UI
+    // portal containing search results.
+    document.body.classList.add('light');
+
+    return () => {
+      document.body.classList.remove('light');
+    };
```
dimaMachina, line 38: I don't get it, it's a regular practice to add `dark` class on `<html>`, why we need `light` class?

packages/web/docs/src/components/landing-page.tsx
```diff
@@ -29,9 +29,13 @@ const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffec
 
 export function IndexPage(): ReactElement {
   useIsomorphicLayoutEffect(() => {
```
dimaMachina, line 31: I don't get it, why on server we use `useEffect` and on client `useLayoutEffect`? `useEffect` is never called on server

dimaMachina:
double focus 
image
image
image
image

dimaMachina:
here focus is broken
https://github.com/user-attachments/assets/97ab32bd-dd18-43aa-b13d-30591470f246

hasparus:
> I think something has regressed in this nav upgrade because in current prod version when I am making it smaller it scales down whereas on this preview (see video) it does a scroll
> 
>  CleanShot.2024-10-08.at.15.35.49.mp4

Yeah, I recognize this bug. The Searchbox structure changed and it stopped shrinking, and I thought 🤔 I've fixed it already  with

image

https://github.com/user-attachments/assets/af9e778d-d5d2-42cf-9836-0f5fccd92bbb

but it turns out it didn't translate to Safari.

packages/web/docs/src/components/landing-page.tsx
hasparus: We want to keep the same number of hooks so it doesn't count as calling them conditionally.

I really only want this useLayoutEffect, but it warns in server render.

https://github.com/streamich/react-use/blob/master/docs/useIsomorphicLayoutEffect.md

> ⚠️ Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer’s output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.

packages/web/docs/src/components/landing-page.tsx
hasparus: Okay, so we have both `.light` and `.dark` classes on `html`, controlled by Tailwind, right?
And they tell our styles if they're in dark mode or light mode context, usually by affecting the `dark:` prefix.

I switched Tailwind's dark mode variant to compile to `&:not(.light *)` (from `is(.dark *)`), because
1. We have an use case where we have `.light` sectors under `html.dark` page — as the landing page is light mode only by design and with some reasoning behind it (not my decision, I'm just a soldier here, I like dark mode like every other developer).
1. I don't want to mess with Tailwind's control over `dark`/`light` class on `<html>` element.

You might notice they even have this exact code as an example in the docs.

https://tailwindcss.com/docs/dark-mode#overriding-the-dark-variant

Thus, everything is dark mode, unless it's under something with a `.light` class.

Here's the original commit from September 2, where it comes from: https://github.com/kamilkisiela/graphql-hive/commit/c7add31c36acf26c01d70583b92605f912fd9845

hasparus:
> when you click on pricing in header
> 
> ## currently
> image
> ## should be
> image

Isn't this just a hover style? uhh, I think we have the same in prod

hasparus:
> Another one in nav is the hover color state
> 
> **Prod**
> 
>  CleanShot.2024-10-08.at.15.40.51.mp4 
> **Preview**
> 
>  CleanShot.2024-10-08.at.15.40.25.mp4

@saihaj can you check again at https://1be49073.hive-landing-page.pages.dev/

I made fixes for this two days ago, but it seems that the deploy preview was stale. It seems good on the one from today.

dimaMachina:
@hasparus can you refactor this
image

to remove `<a>` and attach id directly on `<h2>`? anchor must have `href`

dimaMachina:
image

these links must have cursor-pointer, since they are clickable ?

hasparus:
> @hasparus can you refactor this image
> 
> to remove `<a>` and attach id directly on `<h2>`? anchor must have `href`

I don't know how that happened. This anchor and id is entirely redundant here. Removed it.

hasparus:
> image
> these links must have cursor-pointer, since they are clickable ?

Those are buttons that open a submenu, do you think `cursor-pointer` makes sense for them still?
The menu will open on hover, and if you click it you close it, so I'm a bit afraid of "prompting" the user to click with `cursor-pointer`.

hasparus:
> > > when you click on pricing in header
> > > ## currently
> > > image
> > > ## should be
> > > image
> > 
> > 
> > Isn't this just a hover style? uhh, I think we have the same in prod
> 
> I mean you must see Heading `Pricing` currently it's out of viewport, behind header, see how it looks like for https://1be49073.hive-landing-page.pages.dev/#community-driven-open-source

Oh, I see it. I wonder how that changed :o

hasparus:
> > > > when you click on pricing in header
> > > > ## currently
> > > > image
> > > > ## should be
> > > > image
> > > 
> > > 
> > > Isn't this just a hover style? uhh, I think we have the same in prod
> > 
> > 
> > I mean you must see Heading `Pricing` currently it's out of viewport, behind header, see how it looks like for https://1be49073.hive-landing-page.pages.dev/#community-driven-open-source
> 
> Oh, I see it. I wonder how that changed :o

Fixed by using `<a>` instead of NextLink.

https://github.com/user-attachments/assets/bc02d282-1f70-45ae-9cd3-ba713708c7c6

patches/nextra-theme-docs.patch
```diff
@@ -1,15 +1,24 @@
 diff --git a/dist/index.js b/dist/index.js
-index 788fff4cb15ddd9ea3d19493cc80dd69951caca6..5efca5acdbf794d5e01d1355d7af37103e46a1c2 100644
```
dimaMachina, line 2: this patch not needed

packages/web/docs/src/components/feature-tabs.tsx
```diff
@@ -123,10 +122,10 @@ export function FeatureTabs({ className }: { className?: string }) {
                 key={tab}
                 value={tab}
                 className={
-                  'rdx-state-active:text-green-1000 rdx-state-active:border-beige-600 rdx-state-active:bg-white' +
```
dimaMachina, line 126: can you use `cn` for these classes?

packages/web/docs/src/components/frequently-asked-questions/index.tsx
```diff
@@ -49,7 +49,7 @@ export function FrequentlyAskedQuestions({ className }: { className?: string })
               >
                 <li>
                   <Accordion.Header>
-                    <Accordion.Trigger className="hover:bg-beige-100/80 -mx-2 my-1 flex w-[calc(100%+1rem)] flex-row items-center justify-between rounded-xl bg-white px-2 py-3 text-left font-medium transition-colors duration-[0.8s] md:my-2 md:py-4">
+                    <Accordion.Trigger className="hive-focus hover:bg-beige-100/80 -mx-2 my-1 flex w-[calc(100%+1rem)] flex-row items-center justify-between rounded-xl bg-white px-2 py-3 text-left font-medium transition-colors duration-[0.8s] md:my-2 md:py-4">
```
dimaMachina, line 52:
```suggestion
                    <Accordion.Trigger className="hive-focus hover:bg-beige-100/80 -mx-2 my-1 flex w-[calc(100%+1rem)] items-center justify-between rounded-xl bg-white px-2 py-3 text-left font-medium transition-colors duration-[.8s] md:my-2 md:py-4">
```
by default

packages/web/docs/src/components/company-testimonials/index.tsx
```diff
@@ -113,7 +113,7 @@ export function CompanyTestimonialsSection({ className }: { className?: string }
                   key={testimonial.company}
                   value={testimonial.company}
                   className={
-                    'flex-grow-0 [&[data-state="active"]>:first-child]:bg-blue-400' +
```
dimaMachina, line 116: also `cn` here

dimaMachina (approved):

dotansimha (approved):

PR #5651, opened 2024-10-08
Remove icons from feature tabs

hasparus:
Removed the icons.

BTW: Schema Management tab is missing a link to docs. We should one.

image


saihaj (approved):

packages/web/docs/src/components/feature-tabs.tsx
```diff
@@ -233,19 +230,17 @@ function SchemaManagementIcon() {
 }
 
 function Feature(props: {
-  icon: ReactNode;
```
saihaj, line 236: remove unused 

CleanShot 2024-10-08 at 15 34 18@2x

PR #5656, opened 2024-10-09
bring back Nextra 2 default focus on landing page

hasparus:
### Background

image

### Description

I like the change to explicitly use `.nextra-focus` in Nextra 3 conceptually. It's clean.

But I also really enjoy having a consistent focus style for landing pages, so I brought back the old default (and inlined its colors).

I might refactor it with Cascade Layers in a bit to make it easier to override, but this PR fixed the regression.

### DIFF
--- a/packages/web/docs/src/focus-styles.css
+++ b/packages/web/docs/src/focus-styles.css
@@ -1,0 +1,18 @@
+/* same as in Nextra 2 */
+
+a,
+summary,
+button,
+input,
+[tabindex]:not([tabindex='-1']) {
+  @apply outline-none;
+
+  &:focus-visible {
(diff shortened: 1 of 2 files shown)


dimaMachina (commented):
I don't think it's a right way of defining `:focus` styles, I would prefer avoid defying global styles

## different focus styles in nav links

image
image

## double focus?

image
image
image
image

## weird focus

image

## missing focus

image

## cut focus

image

## double focus

image
image

## footer, links should not take full width

image


## links in docs now have rounded corners

image

hasparus:
@dimaMachina I disagree with this. If we use just the text as link, with no padding, the hitbox is very small and it's hard to click, especially on mobiles.

image

hasparus:
> I don't think it's a right way of defining :focus styles, I would prefer avoid defying global styles

I'm closing it then, and following up with a PR to components instead.

hasparus:
> links in docs now have rounded corners

I copied this from Nextra 2 styles, so we currently also we rounded corners 

image

they imho look kinda better than square corners

image

though maybe I'd change the border-radius from 4px to 2px

image

PR #5683, opened 2024-10-17
Remove GraphQL Conference Card

hasparus:
### Background

| before | after |
| ------- | ---- |
| image | image |

### Description

Removed the conference card, because the conference already happened a while ago.


hasparus:
Subsumed by https://github.com/the-guild-org/docs/pull/1736

PR #5743, opened 2024-10-24
Avoid make stripe API calls on `development`. Update stripe.tsx

dimaMachina:
https://guild-oss.slack.com/archives/C01E8ATBQGN/p1729766480779259

### DIFF
--- a/packages/web/app/src/lib/billing/stripe.tsx
+++ b/packages/web/app/src/lib/billing/stripe.tsx
@@ -1,21 +1,22 @@
-import { ReactElement, ReactNode, Suspense, useRef } from 'react';
+import { FC, ReactNode, Suspense, useState } from 'react';
+import { env } from '@/env/frontend';
 import { Elements as ElementsProvider } from '@stripe/react-stripe-js';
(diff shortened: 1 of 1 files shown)


packages/web/app/src/lib/billing/stripe.tsx
```diff
@@ -1,21 +1,22 @@
...
+      return loadStripe(stripePublicKey);
     }
-  }
-
-  const stripe = stripeRef.current;
+  });
 
-  if (stripe === null) {
+  if (!stripe) {
     return children as any;
```
saihaj, line 20: unrelated to your changes but do we need this cast?
```suggestion
    return children;
```

saihaj (approved):
looks reasonable to me

packages/web/app/src/lib/billing/stripe.tsx
dimaMachina: Yes, FC expect ReactElement as return type, but children is ReactNode

dotansimha (approved):

PR #5750, opened 2024-10-27
Upgrade fastify to v5 and try to workaround Superokens issues

dotansimha:


packages/services/server/src/index.ts
```diff
@@ -128,7 +128,12 @@ export async function main() {
     },
   );
 
-  server.setErrorHandler(supertokensErrorHandler());
+  let errorHandler = supertokensErrorHandler();
+  server.setErrorHandler((err, req, res) => {
+    // DOTAN: Sorry about "as any" here, but it's the only way to make it work.
```
dotansimha, line 133: @n1ru4l @kamilkisiela wasn't sure about this one. We are using v15 of supertokens-node and the latest is v20, seems like a whole different adventure. 

@n1ru4l How can I test/trigger this error handler during development to make sure it's working?

dotansimha:
will be done in a more safe way in https://github.com/graphql-hive/platform/pull/5769

PR #5827, opened 2024-10-28
Remove old codegen resolvers setup

eddeee888:
### Background

Now that we have migrated to Server Preset, we can remove the old codegen config and generated files:

- Remove old codegen config
- Rename `__generated__/types.next.ts` to `__generated__/types.ts` and all related references
- To avoid having old git ignored generated files lying around which could result in wrong suggestion, try removing generated files and regen:

```bash
pnpm rimraf -g "./packages/services/**/__generated__"
pnpm run graphql:generate
```


n1ru4l (approved):
needs conflicts resolved; otherwise let's ship it!

kamilkisiela (approved):

PR #5843, opened 2024-10-29
fix: permission check bypass for document collections

n1ru4l:
### Background

It was possible to modify collections the user should not have access to. There was no assertion for checking whether the document collections belong to the provided document collection.

Since we use UUIDs the impact of this vulnerability should be pretty low.

In addition, I also took the opportunity to strip out the business logic from the GraphQL resolvers.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/collection/resolvers/Mutation/createDocumentCollection.ts
```diff
@@ -1,19 +1,19 @@
-import { TargetAccessScope } from '../../../auth/providers/scopes';
 import { CollectionProvider } from '../../providers/collection.provider';
-import { validateTargetAccess } from '../../validation';
```
kamilkisiela, line 3: @n1ru4l should we remove it, if no longer used?

kamilkisiela (approved):

PR #5853, opened 2024-10-30
introspection of subgraph schema in Hive CLI

kamilkisiela:
Support introspection of subgraph's schema in the `$ hive introspect` command.

This change allows developers to extract the schema of a subgraph (GraphQL Federation) from a running service.
It is useful if the GraphQL framework used in the subgraph does not expose the schema as `.graphql` file.
A GraphQL framework is clearly lacking features here... but we can't PR them all.

---

This change slightly changes the previous behavior of the `$ hive introspect` command. If the introspected GraphQL API is capable of resolving `{ _service { sdl } }` query, the command will use it to fetch the schema. Otherwise, it will use the regular introspection query.

This change is backward-compatible for most, if not all users and should not affect the existing workflows.

In case it does, you can use the `$ hive introspect --ignore-federation` flag to revert to the previous behavior.


saihaj (approved):

packages/libraries/cli/src/commands/introspect.ts
```diff
@@ -42,10 +46,14 @@ export default class Introspect extends Command<typeof Introspect> {
       {} as Record<string, string>,
     );
 
-    const schema = await loadSchema(args.location, {
-      headers,
-      method: 'POST',
-    }).catch(err => {
+    const schema = await loadSchema(
+      flags['ignore-federation'] === true ? 'introspection' : 'federation-subgraph-introspection',
```
jdolle, line 50: Instead of needing to be configured, could we first try `_service` and if we get back a type error, then fallback to the old introspection query?
I think having this be seamless would provide a better UX.

jdolle (commented):
I think this is a nice feature for our CLI.
I'd like to see it get picked up again.

packages/libraries/cli/src/commands/introspect.ts
kamilkisiela: It tries to fetch _sdl and next it introspects by default. The flag that skips the _sdl is opt-in. It's a breaking change so we need to at least give an option to switch back to the previous behaviour, if needed.

kamilkisiela: but I'm also fine with releasing a new major and remove the opt-in option. If you want to push this PR further, I more than happy to pass it to you

noghartt:
cc @n1ru4l, I will be taking a look and working on a "fork" of this PR. Anything special which I should know before starting on it?

n1ru4l:
@noghartt I am fine with shipping it with the functionality as described in the original changeset!

noghartt:
> @noghartt I am fine with shipping it with the functionality as described in the original changeset!

@n1ru4l @kamilkisiela, just one thing which I thought about:

I believe the correct approach is to provide a flag to enable federation rather than ignore it. Because the current behavior of the CLI is having it without the federation, for 100% of backward compatibility, I do believe that:

Instead of having an `--ignore-federation`, we could have a `--enable-federation`. What do you guys think?

n1ru4l:
@noghartt I don't think we need to make it opt-in, it is fine to do a minor release (since we are in 0.x.x range) and mention it in the changeset that this is a minor breaking change. I don't think it would affect many people anyways.

jdolle:
I think because Federation is a layer on top of GraphQL, and when you hear introspection you think of the built-in introspection query, that it does make sense to have this as an opt-in feature. This also alleviates concerns around it being a breaking change.

packages/services/server/src/index.ts
```diff
@@ -138,7 +138,7 @@ export async function main() {
       }
 
       if (req.headers.origin !== env.hiveServices.webApp.url) {
-        return callback(new CorsError());
+        // return callback(new CorsError());
```
n1ru4l:
```suggestion
        return callback(new CorsError());
```

Issue #5870, opened 2024-11-07
Show total number of unused fields in the schema 


saihaj:
internal context: https://guild-oss.slack.com/archives/C01E8ATBQGN/p1730984772581689

PR #5878, opened 2024-11-08
Show total unused fields

martyganz:
### Background

#5870 

I would like to see the total unused fields so I can see if my team is improving over time with cleaning-up their unused fields/types. If you do not see added value then let me know. 

### Description

Affects the web app only as it is only display changes.

This is how I made it look:
Scherm­afbeelding 2024-11-08 om 15 34 22

### Checklist

Only display changes which should not affect the security of the system in any way.


packages/web/app/src/pages/target-explorer-unused.tsx
```diff
@@ -97,8 +98,29 @@ const UnusedSchemaView = memo(function _UnusedSchemaView(props: {
     return null;
   }
 
+  const unusedTypes = types.length ?? 0;
+  const unusedFields = types.reduce((a, b) => {
+    if (b.__typename === 'GraphQLObjectType') {
```
dotansimha, line 103: I think this counts all fields, not just unused?

PR #5885, opened 2024-11-11
Fix 404 and broken links

dotansimha:
closes https://github.com/the-guild-org/website/issues/1784

### DIFF
--- a/packages/web/docs/next.config.js
+++ b/packages/web/docs/next.config.js
@@ -147,6 +147,11 @@ export default withGuildDocs({
       destination: '/docs/other-integrations/:path*',
       permanent: false,
     },
+    {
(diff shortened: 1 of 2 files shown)


packages/web/docs/next.config.js
```diff
@@ -147,6 +147,11 @@ export default withGuildDocs({
       destination: '/docs/other-integrations/:path*',
       permanent: false,
     },
+    {
+      source: '/docs/api-reference/gateway/cli',
+      destination: 'docs/gateway',
```
n1ru4l, line 152: this redirect destination seems incorrect. https://the-guild.dev/graphql/hive/docs/api-reference/gateway-cli seems to be the better fit.

PR #5888, opened 2024-11-11
chore: upgrade supertokens-node to 16.x.x

n1ru4l:
### Background

https://github.com/supertokens/supertokens-node/blob/master/CHANGELOG.md#1674---2024-03-01

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/server/src/supertokens.ts
+++ b/packages/services/server/src/supertokens.ts
@@ -182,17 +182,19 @@ export const backendConfig = (requirements: {
             ...originalImplementation,
             async createNewSession(input) {
               console.log(`Creating a new session for "${input.userId}"`);
-              const user = await ThirdPartyEmailPasswordNode.getUserById(input.userId);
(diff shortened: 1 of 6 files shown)


packages/services/server/src/supertokens.ts
```diff
@@ -201,14 +203,14 @@ export const backendConfig = (requirements: {
                 version: '1',
                 superTokensUserId: input.userId,
                 externalUserId,
-                email: user.email,
+                email: user.emails[0],
```
n1ru4l, line 206: Since we don't use the enterprise version each user will always only have a single email.

PR #5916, opened 2024-11-13
Make content of landing page visible to crawlers

kamilkisiela:
Tabs, FAQ and the diagram content did not have mounted DOM elements.
The only sacrifice is the animation in FAQ, but honestly, I prefer it without it, as you click and wait a second to read what's in there.

### DIFF
--- a/packages/web/docs/src/components/feature-tabs.tsx
+++ b/packages/web/docs/src/components/feature-tabs.tsx
@@ -128,7 +128,12 @@ export function FeatureTabs({ className }: { className?: string }) {
         </Tabs.List>
         <div className="grid grid-cols-1 lg:grid-cols-2">
           <>
-            <Tabs.Content value="Schema Registry" tabIndex={-1}>
+            <Tabs.Content
+              value="Schema Registry"
+              forceMount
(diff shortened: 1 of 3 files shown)


packages/web/docs/src/components/feature-tabs.tsx
```diff
@@ -128,7 +128,12 @@ export function FeatureTabs({ className }: { className?: string }) {
         </Tabs.List>
         <div className="grid grid-cols-1 lg:grid-cols-2">
           <>
-            <Tabs.Content value="Schema Registry" tabIndex={-1}>
+            <Tabs.Content
+              value="Schema Registry"
+              forceMount
+              className="data-[state=inactive]:hidden"
```
kamilkisiela, line 134:
```suggestion
              {/* Make it accessible to crawlers, otherwise there's no DOM element */}
              forceMount
              className="data-[state=inactive]:hidden"
```

dotansimha (approved):

kamilkisiela:
> should we do that in other places as well?

These components were specific to hive's landing page. I don't know if other pages have issues with content visibility, I know Hive had.

PR #5997, opened 2024-11-18
Recommend @graphql-hive/gateway-runtime for programmatic usage

enisdenjo:
The index of Hive Gateway re-exports stuff that is Node-centric, like the [CLI stuff](https://github.com/graphql-hive/gateway/blob/ae84d49f71a72c331f66ff81cea092a791ddc019/packages/gateway/src/index.ts#L1) and various [CLI built-in plugins](https://github.com/graphql-hive/gateway/blob/ae84d49f71a72c331f66ff81cea092a791ddc019/packages/gateway/src/index.ts#L7-L15).

Because of that, the `createGatewayRuntime` from `@graphql-hive/gateway` is not tailored for the edge. You have to run your edge workers in Node compat mode.

However, using `createGatewayRuntime` from `@graphql-hive/gateway-runtime` works without any Node shims or compat modes. It should be the recommended way of running the Hive Gateway on the edge.

At the moment, all other alternatives of fixing the index exports are breaking changes. So the only thing we can do is simply advise users to use the runtime package instead.

### DIFF
--- a/packages/web/docs/src/pages/docs/gateway/deployment/index.mdx
+++ b/packages/web/docs/src/pages/docs/gateway/deployment/index.mdx
@@ -29,7 +29,7 @@ this case, we will show here how to pass the request information from your envir
 then get the response for your environment back.
 
 ```ts
-import { createGatewayRuntime } from '@graphql-hive/gateway'
+import { createGatewayRuntime } from '@graphql-hive/gateway-runtime'
 import type {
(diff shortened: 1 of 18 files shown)


n1ru4l (approved):
Approved, just one question for clarification: https://guild-oss.slack.com/archives/CSFK5GGAU/p1732183078959109

PR #6021, opened 2024-11-21
refactor: decode user in storage

n1ru4l:
### Background

I will touch this code soon, so I wanted ti first clean it up a bit.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-manager.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-manager.ts
@@ -969,9 +969,8 @@ export class SchemaManager {
       return null;
     }
 
-    return this.storage.getOrganizationUser({
-      organizationId: args.organizationId,
-      userId: args.userId,
+    return this.storage.getUserById({
+      id: args.userId,
     });
   }
 
--- a/packages/services/api/src/modules/shared/providers/storage.ts
+++ b/packages/services/api/src/modules/shared/providers/storage.ts
@@ -69,7 +69,6 @@ export interface Storage {
   isReady(): Promise<boolean>;
   ensureUserExists(_: {
     superTokensUserId: string;
-    externalAuthUserId?: string | null;
     email: string;
     oidcIntegration: null | {
       id: string;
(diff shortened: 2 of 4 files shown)


packages/services/api/src/modules/schema/providers/schema-manager.ts
```diff
@@ -969,9 +969,8 @@ export class SchemaManager {
       return null;
     }
 
-    return this.storage.getOrganizationUser({
-      organizationId: args.organizationId,
-      userId: args.userId,
+    return this.storage.getUserById({
+      id: args.userId,
```
n1ru4l, line 973: Method was only used here and did not provide a real benefit over the exiting one..

packages/services/api/src/modules/shared/providers/storage.ts
```diff
@@ -69,7 +69,6 @@ export interface Storage {
   isReady(): Promise<boolean>;
   ensureUserExists(_: {
     superTokensUserId: string;
-    externalAuthUserId?: string | null;
```
n1ru4l, line 72: We do not use the externalAuthUserId anywhere for a long time now.

dotansimha: This was used with Auth0, it's legacy, and can be removed for sure.

packages/services/storage/src/index.ts
```diff
@@ -477,64 +426,55 @@ export async function createStorage(
...
         email: string;
         fullName: string;
         displayName: string;
-        externalAuthUserId: string | null;
         oidcIntegrationId: string | null;
       },
       connection: Connection,
     ) {
-      const user = await connection.one<users>(
+      await connection.query<unknown>(
```
dotansimha, line 463: why `unknown`? 🤔

packages/services/storage/src/index.ts
```diff
@@ -477,64 +426,55 @@ export async function createStorage(
...
       `);
 
-      if (user) {
-        return transformUser(user);
+      if (!record) {
+        return null;
       }
 
-      return null;
+      return UserModel.parse(record);
```
dotansimha, line 445: thank you 🙇

dotansimha (approved):

packages/services/storage/src/index.ts
n1ru4l: because we don't want anyone to use that 🗡️

PR #6042, opened 2024-11-27
explore better logging in `envoy` access logs 

dotansimha:


deployment/utils/reverse-proxy.ts
```diff
@@ -225,6 +225,11 @@ export class Proxy {
           'upstream_service_time',
           'user_agent',
           'x_forwarded_for',
+          'x_trace_id',
+          // X-API-TOKEN is a custom header, that contains only the token without a prefix.
+          'x_api_token=%REQ(X-API-TOKEN):3%',
+          /// Authorization header contains the token with the prefix "Bearer " (so 7+4 to get the first 4 chars).
+          'authorization=%REQ(AUTHORIZATION):11%',
```
kamilkisiela, line 232: Will it work when the value is below 11 characters?

dotansimha: yeah, I tested it manually and it seems to just output empty string

PR #6045, opened 2024-11-27
Improve error handling in `usage` service and return correct HTTP code

TuvalSimha:

PR #6053, opened 2024-11-28
fix(service): maskToken function throwing an exception

TuvalSimha:
### Background

https://github.com/graphql-hive/console/pull/6045 introduced an exception that can crash the whole service. This PR fixes the implementation of the pull request, adds basic tests and replaces usages of similar functions in the codebase with this shared one.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/service-common/src/helpers.ts
+++ b/packages/services/service-common/src/helpers.ts
@@ -33,6 +33,9 @@ export function invariant(
   throw new Error(value);
 }
 
-export function maskToken(token: string) {
+export function maskToken(token: string): string {
+  if (token.length <= 5) {
+    return '*'.repeat(token.length - 1) + token[token.length - 1];
+  }
   return token.substring(0, 3) + '*'.repeat(token.length - 6) + token.substring(token.length - 3);
 }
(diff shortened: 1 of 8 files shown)


packages/services/service-common/src/helpers.ts
```diff
@@ -33,6 +33,9 @@ export function invariant(
   throw new Error(value);
 }
 
-export function maskToken(token: string) {
+export function maskToken(token: string): string {
```
kamilkisiela, line 36:
```
maskToken('')
// Uncaught RangeError: Invalid count value: -1
//   at String.repeat (<anonymous>)
```

n1ru4l: @TuvalSimha we can return the string as is if the length is below a threshold (e.g. 6 characters) in that case it is not a valid token and thus does not require masking.

dotansimha: i agree with @n1ru4l

kamilkisiela (approved):

n1ru4l (approved):

PR #6054, opened 2024-11-28
fix(api): log project type instead of [object Object]

n1ru4l:
### Background

Small improvement for logging.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Authentication management
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Error handling and logging
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

PR #6055, opened 2024-11-28
improve logging around schema composition and external calls

dotansimha:
related https://github.com/graphql-hive/console/pull/6054

### DIFF
--- a/packages/services/schema/src/orchestrators.ts
+++ b/packages/services/schema/src/orchestrators.ts
@@ -319,7 +319,11 @@ async function callExternalService(
         span.setAttribute('error.message', error.message);
         span.setAttribute('error.type', error.name);
 
-        logger.error('Network error without response. (%s, %s)', error.name, error.message);
+        logger.error(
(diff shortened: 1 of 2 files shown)


packages/services/schema/src/orchestrators.ts
```diff
@@ -318,7 +319,8 @@ async function callExternalService(
         span.setAttribute('error.message', error.message);
         span.setAttribute('error.type', error.name);
 
-        logger.info('Network error without response. (%s)', error.message);
+        logger.error('Network error without response. (%s, %s)', error.name, error.message);
```
n1ru4l, line 322: @dotansimha Let's prefix the string interpolations like others.

```
logger.error('Network error without response. (errorName=%s, errorMessage=%s)', error.name, error.message);
```

n1ru4l (approved):
one comment

packages/services/schema/src/orchestrators.ts
dotansimha: sure, fixed!

PR #6056, opened 2024-11-28
Hive Gateway programmatic usage and API

enisdenjo:
Everyone is advised to use the runtime version of Hive Gateway because it's slimmer and fully JS env agnostic. Aside from putting emphasis on this and other things we've discovered during serverless assistence, some other notable changes are:

- No index pages of deployment sections, they're just repetative
- Talk about programmatic API immediately
- Slimmer naming of deployment destination "Frameworks" and "Runtimes"
- Mention "Hive Gateway" as much as possible
- Remove nestjs framework from deployments completely (was empty)
- Use callout for emphasis instead of citations in "Monitoring and Tracing"
- Fixed a bunch of smaller things that caught my eye during the refactor


EmrysMyrddin (approved):

packages/web/docs/src/pages/docs/gateway/deployment/node-frameworks/_meta.ts
```diff
@@ -1,10 +1,8 @@
 export default {
-  index: 'Introduction',
   express: 'Express',
   fastify: 'Fastify',
   koa: 'Koa',
   hapi: 'Hapi',
-  nestjs: 'NestJS',
```
n1ru4l, line 7: It seems like we should add a re-direct for the removal of this route?

packages/web/docs/src/pages/docs/gateway/deployment/node-frameworks/_meta.ts
```diff
@@ -1,10 +1,8 @@
 export default {
-  index: 'Introduction',
```
n1ru4l, line 2: It seems like we should add a re-direct for the removal of this route?

n1ru4l (changes requested):
We are removing routes that could be linked to, can we add redirects for these?

n1ru4l (approved):

PR #6060, opened 2024-11-29
docs: self-hosting changelog in documentation

n1ru4l:
### Background

This embeds the changelog for self-hosting within the documentation.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


.prettierignore
```diff
@@ -32,3 +32,4 @@ CHANGELOG.md
 
 # File generated by @hive/app#generate-changelog script
 packages/web/app/src/components/ui/changelog/generated-changelog.ts
+packages/web/docs/src/pages/docs/self-hosting/changelog.mdx
```
n1ru4l, line 35: I tried to use a `.tsx` file, but nextra did not work with `RemoteContent` unless the file uses the `.mdx` extension. I could not find any information in the extra documentation about that topic. So I switched to `.mdx`, even though now we don't get TypeScript benefits, I also had to disable prettier for the file, as it would destroy it every time it tries to format it.

packages/web/docs/src/pages/docs/self-hosting/changelog.mdx
```diff
@@ -0,0 +1,35 @@
...
+    },
+  };
+}
+
+# Self-Hosting Changelog
+
+Here you can find all the recent improvements and updates about the self-hosting images.
+The Docker images published to the GitHub registry using the same tags.
+
+<RemoteContent suppressHydrationWarning />
```
n1ru4l, line 35: This constantly causes a rehydration error. But I am not sure if we can just ignore it or whether it actually is a big deal.

packages/web/docs/src/pages/docs/self-hosting/changelog.mdx
```diff
@@ -0,0 +1,35 @@
+import { readFileSync } from 'node:fs';
+import { buildDynamicMDX } from 'nextra/remote';
+import { RemoteContent } from '@theguild/components';
+
+export async function getStaticProps () {
+  const data = readFileSync(
+    __dirname + '/../../../../../../../../deployment/CHANGELOG.md',
```
n1ru4l, line 7: `__dirname` is not what you expect (`packages/web/docs/src/pages/docs/self-hosting`), but `packages/web/docs/.next/server/pages/docs/self-hosting`.

packages/web/docs/src/pages/docs/self-hosting/changelog.mdx
dimaMachina: fine for now, will be fixed with Nextra 4 migration

packages/web/docs/src/pages/docs/self-hosting/changelog.mdx
dimaMachina: `process.cwd(), '..', 'deployment', 'CHANGELOG.md'` maybe more simpler?

PR #6068, opened 2024-12-02
Fix typo in Auth docs

ardatan:
Fix according to https://github.com/ardatan/graphql-mesh/blob/master/packages/plugins/jwt-auth/src/index.ts#L21


packages/web/docs/public/feed.xml
saihaj, line 1: remove this?

n1ru4l (approved):

PR #6070, opened 2024-12-03
Remove Slider transition in Safari

hasparus:
### Description

I removed the CSS transition on the `transform` in the Slider component in Safari, because it jiggles weirdly, it looks bad, and I'm not sure if it's possible to fix fast.

**Alternative solutions that didn't work:** adding `will-change`; changing the easing function and transition time.

#### Chromium

https://github.com/user-attachments/assets/98d48650-7d99-4b5d-b37f-5ec185a1dcc2

#### Safari Before

https://github.com/user-attachments/assets/f55ab279-58a6-4d53-a9bb-53e8e03d786f

#### Safari After

https://github.com/user-attachments/assets/f8102352-7a50-43d2-9f10-eb9e977ed332

### Prior Art

I should make a meme out of this.

image


packages/web/docs/src/components/slider.tsx
```diff
@@ -80,9 +80,11 @@ export function Slider({ counter, className, style, ...rest }: SliderProps) {
           }
         }
 
-        div,
-        div:after {
-          transition: transform var(--ease) 500ms;
+        @supports not (font: -apple-system-body) {
```
hasparus, line 83: This detects Safari, not MacOS. `-apple-system-body` is a "precursor" system font.

image
image

dimaMachina (approved):

PR #6074, opened 2024-12-03
For Bun integration, Recommend Hive GW CLI

ardatan:
Hive GW CLI is fully compatible with Bun just like Node.js.
So this PR does
- Mention Bun as a compatible runtime for CLI
- Recommend to use CLI on Bun instead of programmatic API

Note: This doesn't recommend using Bun over Node. It recommends CLI over programmatic API for Bun. Because previously for Bun runtime, only programmatic API was recommended. Now we mention that CLI is compatible with Bun so if you want to use Bun, you can use CLI directly.

### DIFF
--- a/packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
+++ b/packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
@@ -17,7 +17,7 @@ other JS runtime with Hive Gateway;
 
 You can follow the introduction page directly to use Hive Gateway CLI. [See here](/docs/gateway)
 
-Since Bun has the compatibility layer for Node.js, even Node specific features are available in Bun.
+Since Bun has the compatibility layer for Node.js, all Node specific features are available in Bun.
 
 ## Hive Gateway Runtime (advanced-only)
 
--- a/packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
+++ b/packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
@@ -13,7 +13,16 @@ platform besides Node.js. [Bun](https://bun.sh) is a modern JavaScript runtime l
 and it supports Fetch API as a first class citizen. So the configuration is really simple like any
 other JS runtime with Hive Gateway;
 
-The following code is a simple example of how to use Hive Gateway with Bun.
+## Hive Gateway CLI
+
+You can follow the introduction page directly to use Hive Gateway CLI. [See here](/docs/gateway)
+
+Since Bun has the compatibility layer for Node.js, even Node specific features are available in Bun.
+
+## Hive Gateway Runtime (advanced-only)
+
+Use this method only if you know what you are doing. It is recommended to use Hive Gateway CLI for
+most cases.
 
 ```ts
 import { createGatewayRuntime } from '@graphql-hive/gateway-runtime'
--- a/packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
+++ b/packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
@@ -15,7 +15,7 @@ other JS runtime with Hive Gateway;
 
 ## Hive Gateway CLI
 
-You can follow the introduction page directly to use Hive Gateway CLI. [See here](/docs/gateway)
+You can follow the introduction page directly to [use Hive Gateway CLI](/docs/gateway#installation).
 
 Since Bun has the compatibility layer for Node.js, all Node specific features are available in Bun.
 
--- a/packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
+++ b/packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
@@ -19,7 +19,7 @@ You can follow the introduction page directly to [use Hive Gateway CLI](/docs/ga
 
 Since Bun has the compatibility layer for Node.js, all Node specific features are available in Bun.
 
-## Hive Gateway Runtime (advanced-only)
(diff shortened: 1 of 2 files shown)


packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
```diff
@@ -13,7 +13,16 @@ platform besides Node.js. [Bun](https://bun.sh) is a modern JavaScript runtime l
 and it supports Fetch API as a first class citizen. So the configuration is really simple like any
 other JS runtime with Hive Gateway;
 
-The following code is a simple example of how to use Hive Gateway with Bun.
+## Hive Gateway CLI
+
+You can follow the introduction page directly to use Hive Gateway CLI. [See here](/docs/gateway)
```
kamilkisiela, line 18:
```suggestion
You can follow the introduction page directly to [use Hive Gateway CLI](/docs/gateway#installation).
```

Maybe we should point directly to the instructions?

packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
```diff
@@ -13,7 +13,16 @@ platform besides Node.js. [Bun](https://bun.sh) is a modern JavaScript runtime l
...
+
+Since Bun has the compatibility layer for Node.js, even Node specific features are available in Bun.
```
kamilkisiela, line 20:
```suggestion
Since Bun has the compatibility layer for Node.js, all Node specific features are available in Bun.
```

I guess?

packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
```diff
@@ -13,7 +13,16 @@ platform besides Node.js. [Bun](https://bun.sh) is a modern JavaScript runtime l
...
 other JS runtime with Hive Gateway;
 
-The following code is a simple example of how to use Hive Gateway with Bun.
+## Hive Gateway CLI
+
+You can follow the introduction page directly to [use Hive Gateway CLI](/docs/gateway#installation).
+
+Since Bun has the compatibility layer for Node.js, all Node specific features are available in Bun.
+
+## Hive Gateway Runtime (advanced-only)
```
enisdenjo, line 22: Why do we claim that the programmatic API is for advance usecases an only to be used if "you know what you're doing"?

Its completely ok and really not that advanced to use the prog. API.

ardatan: I copied what we have in Node.js page

ardatan: Removed (Advanced)

packages/web/docs/src/pages/docs/gateway/index.mdx
```diff
@@ -83,6 +83,11 @@ environment. Then, you can install Hive Gateway CLI with your preferred package
 npm i @graphql-hive/gateway
 ```
 
+<Callout>
```
dotansimha, line 86: I think we still don't have a clear recommendation here regarding Bun. It's new and cool, but I feel like it's still not in the same state of Node when it comes to stability. 

I think maybe phrasing it as something like "Hive Gateway is runtime-agnostic and on different runtimes. The Guild's development process uses NodeJS, and the gateway can also run on alternative runtimes like Bun"?

packages/web/docs/src/pages/docs/gateway/deployment/runtimes/bun.mdx
```diff
@@ -13,7 +13,16 @@ platform besides Node.js. [Bun](https://bun.sh) is a modern JavaScript runtime l
 and it supports Fetch API as a first class citizen. So the configuration is really simple like any
 other JS runtime with Hive Gateway;
 
-The following code is a simple example of how to use Hive Gateway with Bun.
+## Hive Gateway CLI
+
+You can follow the introduction page directly to [use Hive Gateway CLI](/docs/gateway#installation).
+
+Since Bun has the compatibility layer for Node.js, all Node specific features are available in Bun.
```
dotansimha, line 20: That's kinda correct, but I think we don't have a decision to promote Bun as the recommended runtime?

ardatan: This is Bun's page, so we don't recommend it but we say you can use Hive Gateway CLI with Bun. Previously it was just runtime.

packages/web/docs/src/pages/docs/gateway/index.mdx
ardatan: We have both unit and E2E tests with Bun. So Bun is tested exactly in the same way Node is tested in the CI. We can rephrase the sentence here but the idea here is to emphasize that "You can use Hive Gateway CLI on Bun just like you do on Node".

Urigo: what is the reason to specifically callout Bun here?

Urigo (changes requested):

packages/web/docs/src/pages/docs/gateway/index.mdx
ardatan: We refer `bun install` with `npm2yarn` in our docs, and in the other parts of this section, it looks like CLI only supports Node.js but it is incorrect. CLI also supports Bun as we test it e2e and run unit tests against its test runner.
Do you think this gives a wrong message?

Urigo: is it also true for other environments?  like Deno?

Urigo: ok I see, thanks, approved by me

PR #6081, opened 2024-12-04
feat(grafana): postgres connection panel, db performance tracking dashboard

saihaj:
It should look something like this:

CleanShot 2024-12-04 at 20 39 03@2x

image

### DIFF
--- a/deployment/grafana-dashboards/Hive-Vitals.json
+++ b/deployment/grafana-dashboards/Hive-Vitals.json
@@ -363,6 +363,170 @@
     },
     {
       "datasource": {
+        "type": "grafana-azure-monitor-datasource",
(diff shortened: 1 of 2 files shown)


deployment/grafana-dashboards/Hive-Vitals.json
```diff
@@ -361,6 +361,170 @@
       "title": "Postgres Overview",
       "type": "timeseries"
     },
+    {
+      "datasource": {
+        "type": "grafana-azure-monitor-datasource",
+        "uid": "u65mwBI4z"
```
kamilkisiela, line 367: Isn't that id specific to an environment?

dotansimha: fixed

PR #6083, opened 2024-12-05
limit memory on usage service

saihaj:
Right now the way it is working in production one pod keeps growing memory forever. This way we restrict pod budget so we don't grow infinitely. I am hoping that it will instead scale pods out when we hit limits.

CleanShot 2024-12-05 at 12 35 52@2x


dotansimha (approved):

kamilkisiela:
Limiting memory can lead to data being dropped

PR #6084, opened 2024-12-05
stop using `unknown-yet/unknown-yet` for skeletons in `/org` and `/org/project` pages

dimaMachina:
don't forget to turn off whitespace while review code

### DIFF
--- a/packages/web/app/src/pages/project.tsx
+++ b/packages/web/app/src/pages/project.tsx
@@ -8,6 +8,7 @@ import { useQuery } from 'urql';
 import { z } from 'zod';
 import { Page, ProjectLayout } from '@/components/layouts/project';
 import { Button } from '@/components/ui/button';
+import { Card } from '@/components/ui/card';
 import { EmptyList } from '@/components/ui/empty-list';
 import { Input } from '@/components/ui/input';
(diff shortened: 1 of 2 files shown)


packages/web/app/src/pages/project.tsx
```diff
@@ -61,146 +59,155 @@ const TargetCard = (props: {
...
+      </div>
+    </TooltipProvider>
+  );
+
+  const { organizationSlug, projectSlug } = useParams({
+    from: '/authenticated/$organizationSlug/$projectSlug',
+  });
+
+  return (
+    <Card className="h-full self-start bg-gray-900/50 p-5 px-0 pt-4 hover:bg-gray-800/40 hover:shadow-md hover:shadow-gray-800/50">
```
dimaMachina, line 196: I copied classes from org page, and used card from new UI components

kamilkisiela (changes requested):
This PR should update 2 components and make them render either `<Link>{children}</Link>` or `<>{children}</>`, but it changes the way the components receive IDs of project and organization and a few other things.
Could be +10 - 10 lines but we get +298 - 290.

Let's keep the change as simple as possible without introducing so many changes.

Issue #5854, opened 2024-10-30
mobile images needs to be responsive


saihaj:
On mobile the images are adding horizontal scroll to the page. Ideal would be if we cna just show the image and on click it can enlarge. 

![IMG_1291](https://github.com/user-attachments/assets/f4d01ff2-c16c-4356-a622-b8ca86f851ca)

PR #6085, opened 2024-12-06
fix images on mobile

dimaMachina:
closes https://github.com/graphql-hive/console/issues/5854
context:

images are broken on mobile in all usages of next/image due `max-w...` classes

image

- [x] block usage `next/image` on mdx files, for future mistake
- [x] refactor all static imports of images
- [x] override `` injected component by adding rounded corners and box-shadow


packages/web/docs/src/pages/docs/schema-registry/external-schema-composition.mdx
```diff
@@ -1,9 +1,4 @@
-import NextImage from 'next/image'
 import { Callout, Steps, Tabs } from '@theguild/components'
-import externalCompositionInvalid from '../../../../public/docs/pages/management/external-composition-invalid.png'
```
hasparus, line 3: Just thinking — would it be possible to colocate the images inside of `docs` (content) and the use relative imports instead of absolute paths to public?

.eslintrc.cjs
```diff
@@ -224,11 +224,37 @@ module.exports = {
         },
         tailwindcss: {
           callees: tailwindCallees,
-          whitelist: ['light', 'hive-focus', 'hive-focus-within'],
+          whitelist: ['light', 'hive-focus', 'hive-focus-within', '_text-primary-600'],
```
kamilkisiela, line 227: why this class has an underscore prefix?

packages/web/docs/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "noEmit": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
kamilkisiela, line 12: why?

packages/web/docs/tsconfig.json
```diff
@@ -21,6 +21,6 @@
     ],
     "strictNullChecks": true
   },
-  "include": ["next-env.d.ts", "src", ".next/types/**/*.ts"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
```
kamilkisiela, line 24: do we have files outside src?

dimaMachina: tailwind.config

packages/web/docs/tsconfig.json
dimaMachina: without this change there is a type errors for some packages

.eslintrc.cjs
dimaMachina: this link is not highlighted for some reason 
image

nextra uses `_` prefix for its classes

jdolle (commented):
The user experience with these changes is much better. Great job on that.

The change to moduleResolution makes me a bit uneasy though. That can have major impacts and is typically not necessary.

This needs updated and needs integration tests/etc to pass also.

saihaj:
I guess we need to bring the fix from this in a different PR. Closing this for inactivity.

PR #6088, opened 2024-12-06
chore: wake up clickhouse before publishing app deployment

n1ru4l:
Closes https://github.com/graphql-hive/deployment/issues/1425 

### Background

Deployments on staging and dev often fail because the app deployment publishes times out due to click house being in rest mode.

### Description

This change will send a dummy query (`SELECT 1`) to the clickhouse instance on dev and staging to make sure it responds in time for the app deployment publish.


dotansimha (approved):
Pushed a few fixes and rebased. LGTM. 
Also verified dependencies.

PR #6090, opened 2024-12-06
Showcase `@defer` and `@stream` in docs/gateway/defer-stream

kamilkisiela:
We have `See more` link at the end that points to Yoga, where we tell people how to configure Yoga and run queries via cURL or GraphiQL.  It's a bit weird given the gateway talks to subgraphs, so those instructions makes no sense.

I copied part of Yoga docs to show example queries and expected behavior.


saihaj (approved):

kamilkisiela:
> Should we still link to yoga ?

@saihaj why?

PR #6098, opened 2024-12-10
docs: resolveUser -> resolveUserFn

jjangga0214:
### Background

스크린샷 2024-12-10 오후 3 12 09

But a type error happens.

스크린샷 2024-12-10 오후 3 11 06

I guess `resolveUser` is refactored to `resolveUserFn`.


kamilkisiela:
#6093

PR #6099, opened 2024-12-10
Fix the watch mode in web/app

kamilkisiela:
I moved us to `tsx`.

It's tested, it works, no need to use latest and greatest.

```
tsx watch --clear-screen=false --exclude \"./**/*.mjs\" src/server/dev.ts
```

`--clear-screen=false` - because it's annoying when things disappear, especially when something does not work and you're debugging.

`--exclude "./**/*.mjs"` - because Vite or something related to Vite complies the config file to `.mjs` and the watch mode freaks out.

`src/server/dev.ts` - because this way we can fully control environment variables and use our good old and reliable `dotenv`.

### DIFF
--- a/packages/web/app/package.json
+++ b/packages/web/app/package.json
@@ -5,11 +5,11 @@
   "scripts": {
     "build": "tsx ../../../scripts/runify.ts src/server/index.ts && vite build --outDir dist/client",
     "build-storybook": "storybook build",
-    "dev": "node --env-file=.env --watch-path src/server --import tsx src/server/index.ts --dev",
+    "dev": "tsx watch --clear-screen=false --exclude \"./**/*.mjs\" src/server/dev.ts",
     "generate-changelog": "node ../../../scripts/generate-changelog.js",
     "postinstall": "pnpm generate-changelog",
     "storybook": "storybook dev -p 6006",
-    "typecheck": "tsc"
+    "typecheck": "tsc --noEmit"
   },
   "devDependencies": {
     "@date-fns/utc": "2.1.0",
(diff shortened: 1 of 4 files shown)


packages/web/app/package.json
```diff
@@ -5,11 +5,11 @@
   "scripts": {
     "build": "tsx ../../../scripts/runify.ts src/server/index.ts && vite build --outDir dist/client",
     "build-storybook": "storybook build",
-    "dev": "node --env-file=.env --watch-path src/server --import tsx src/server/index.ts --dev",
+    "dev": "tsx watch --clear-screen=false --exclude \"./**/*.mjs\" src/server/dev.ts",
```
dimaMachina, line 8:
```suggestion
    "dev": "NODE_ENV=development tsx watch --clear-screen=false --exclude \"./**/*.mjs\" src/server/dev.ts",
```
is this will works? instead of `process.env.NODE_ENV = 'development';` in `server/dev.ts`

kamilkisiela: Why though? It won’t work on Windows and it’s just odd to pass env vars like that for that single purpose.

dimaMachina: just to avoid creating yet another file for dev environment? we can pass this env via `cross-env` package to be platform agnostic

packages/web/app/package.json
```diff
@@ -83,6 +83,7 @@
     "cmdk": "0.2.1",
     "date-fns": "4.1.0",
     "dompurify": "3.2.0",
+    "dotenv": "16.4.5",
```
dimaMachina, line 86: see https://github.com/graphql-hive/console/pull/6099/files#r1878514490

packages/web/app/package.json
kamilkisiela: 200 files or 180 makes zero difference

dotansimha (approved):

PR #6118, opened 2024-12-13
enhance(@graphql-hive/yoga): improvements with Yoga's features

ardatan:
Align with GraphQL Yoga >=5.10.4 which has the relevant features mentioned below;

- Remove \`tiny-lru\` dependency, and use `createLruCache` from `graphql-yoga`
- - Yoga already provides a LRU cache implementation, so we can use that instead of having a separate dependency
- Use Explicit Resource Management of GraphQL Yoga plugins for disposal which already respects Node.js process termination
- - The feature has been implemented in `@whatwg-node/server` which is used by GraphQL Yoga. [Learn more about this feature](https://github.com/ardatan/whatwg-node/pull/1830)
- Use \`context.waitUntil\` which is handled by the environment automatically, if not `@whatwg-node/server` already takes care of it with above.
- - The feature has been implemented in `@whatwg-node/server` which is used by GraphQL Yoga. [Learn more about this feature](
    https://github.com/ardatan/whatwg-node/pull/1830)
- Respect the given `fetchAPI` by GraphQL Yoga(might be Hive Gateway) for the `fetch` function
- - Hive Gateway uses `fetchAPI` given to GraphQL Yoga for the entire Fetch API implementation, so if Hive Client respects that, we have a control over HTTP calls done internally by the gateway
- Respect Yoga's \`logger\` for logging
- - Similar to above, we want to respect the logger provided by Yoga to have a better control over logging


packages/libraries/yoga/src/index.ts
```diff
@@ -206,21 +213,21 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
...
-            if (document) {
-              const record = contextualCache.get(context);
-              if (record) {
-                record.experimental__documentId = key;
-                record.paramsArgs = {
-                  ...record.paramsArgs,
-                  query: document,
-                };
+          getPersistedOperation(key, _request, context) {
+            return mapMaybePromise(experimental__persistedDocuments.resolve(key), document => {
```
kamilkisiela, line 217: I will accept such performance optimization only if you prove me it actually matters.
I know what's the different between Promise and sync return and how it affects performance, but here, to me, it's completely unnecessary and just adds complexity for no reason.

ardatan: image
Any unnecessary `async` affects the performance, that's why for example graphql-js handles async and sync.

packages/libraries/yoga/src/index.ts
```diff
@@ -42,31 +41,30 @@ export function createHive(clientOrOptions: HivePluginOptions) {
...
+      logger = yoga.logger;
+      hive = isHiveClient(clientOrOptions) ? clientOrOptions : createHive({
+        ...clientOrOptions,
+        agent: clientOrOptions.agent ? {
+          logger: yoga.logger,
+          ...clientOrOptions.agent,
+          __testing: {
+            ...clientOrOptions.agent.__testing,
+            // Hive Plugin should respect the given FetchAPI, note that this is not `yoga.fetch`
+            fetch: yoga.fetchAPI.fetch,
```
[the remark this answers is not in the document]
ardatan: I didn't find another way to control `fetch` used in `HiveClient`.

packages/libraries/yoga/src/index.ts
kamilkisiela: I think you ignored what I wrote

kamilkisiela: tell me how 0.004 vs 0.015 matters when it comes to pulling persisted operation from a storage

ardatan: Everytime it gets the requests, any promise in the chain causes the entire operation to be deferred to the next tick.
It is not the numbers there but the exponential latency it adds.

ardatan: Like I said, this is the same why graphql-js also handles resolvers in this way. And this is why we have the same logic in the gateway from bottom to the top.

kamilkisiela: Still not convinced

kamilkisiela: in case of graphql-js, it creates 1020120 promises, here, you have 1.

ardatan: Because I want to avoid to add another promise in the gateway for example.
One extra promise cause the entire chain to be promise.

ardatan: I'd love to avoid any extra latency as much as possible as we care about performance from the lowest level to the highest.

EmrysMyrddin: With a gateway under slight pressure, by waiting for the next tick, you miss the oportunity to start sending the response imediatly. Which means a ready to send response will wait for the other micro-tasks queued before being actually sent.

And that's what can cause very hight latence spike. If a huge sync task is in the queue, the response will seat there for potentially a long time, for no real reason other than the code looking slightly uglier.

kamilkisiela: that's the theory :) What I asked is a proof that this indeed affects the gateway.
We can micro-optimize 100 things and spend 20h on it and see 2ms different on a query that takes 600ms.
What I'm saying is, that I will only allow for optimizations that are actually affecting the gateway and not polluting code, because we think it does.

ardatan: Anyway reverted

.changeset/spotty-toes-help.md
```diff
@@ -0,0 +1,12 @@
...
+'@graphql-hive/yoga': patch
+---
+
+Align with the latest yoga;
+
+- Remove \`tiny-lru\` dependency, and use `createLruCache` from `graphql-yoga`
+- Use Explicit Resource Management of GraphQL Yoga plugins for disposal which already respect Node.js process termination
+- Use \`context.waitUntil\` which is handled by the environment automatically, if not `@whatwg-node/server` already takes care of it with above.
+- Respect the given `fetchAPI` by GraphQL Yoga(might be Hive Gateway) for the `fetch` function
+- Avoid using `async/await` for performance reasons, because in case of cache-hit, it returns synchronously
```
ardatan, line 11:
```suggestion
```

packages/libraries/yoga/src/index.ts
ardatan: The other changes are more important than this :)

ardatan: Also one think I need to make clear, the issue is not to improve the performance of getting this from cache. It is the overall performance it affects because it makes the operation async.
In the entire conversation, you are repeating that it is that specific part, no it is not. My goal is not to make here faster, my goal is to avoid promises as much as we could so we won't have unnecessary promises, and get as much as performance from our gateway/yoga etc as we can. It affects the entire workflow because it adds an extra promise to the pipeline. If this is not your concern as Hive Console team, we can implement our own version of this plugin with our concerns.
> but here, to me, it's completely unnecessary
It might be hard for you to read in that case but readability of us is less important than getting more performance.

.changeset/spotty-toes-help.md
```diff
@@ -0,0 +1,11 @@
+---
+'@graphql-hive/yoga': patch
+---
+
+Align with the latest yoga;
```
dotansimha, line 5: link to "latest yoga" and maybe specify the version here as well.

ardatan, line 5:
```suggestion
Align with GraphQL Yoga >=5.10.4 which has the relevant features mentioned below;
```

ardatan: Updated, good point! Let me know what you think!

packages/libraries/yoga/src/index.ts
```diff
@@ -124,10 +116,12 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
...
-              void record.callback(
-                args,
-                errors.length ? { errors } : {},
-                record.experimental__documentId,
+              args.contextValue.waitUntil(
+                record.callback(
+                  args,
+                  errors.length ? { errors } : {},
+                  record.experimental__documentId,
+                ),
```
n1ru4l, line 124: On Cloudflare Workers, I know what this will do (suspend the worker from being terminated until the promise resolves).

Can you explain to me what implications this has within a Node.js environment? Is `waitUntil` there just a dummy implementation?

ardatan: No with this PR `waitUntil` has been implemented in Node.js fully; https://github.com/ardatan/whatwg-node/pull/1830
All the promises passed to `waitUntil` are kept until they get resolved, and disposal waits for it.
Disposal logic has also been added with this PR so any disposable plugins together with ongoing waitUntil promises are resolved through this disposal logic.
In Node.js and compatible environments, we add a listener for process termination for this disposal logic.

ardatan:
```ts
const deferredForWaitUntil = createDeferred();
const deferredForPlugin = createDeferred();
const yoga = createYoga({
  schema: createSchema({
     // ...
     resolvers: {
        Query: {
            someStuff(_, __, ctx) {
                ctx.waitUntil(
                   deferredForWaitUntil.promise.then(() => {
                     console.log('Wait until')
                   })
                )
            }
        }
     }
  })
  plugins: [
     {
       [Symbol.asyncDispose]() {
          console.log('Plugin disposal')
          return deferredForPlugin.promise;
       }
     }
  ]
}).

const res = await yoga.fetch('/graphql');
console.log(await res.text()) // Not blocked and printed

// await using is better but I want to show the timing
// This is also called when Node.js process terminates
yoga[Symbol.asyncDispose]().then(() => {
  console.log('Fully disposed');
})

deferredForWaitUntil.resolve() // Printed waitUntil but stil lwaiting for promise
deferredForPlugin.resolve() // Printed Plugin disposal and Fully disposed
```

ardatan: Did improvements based on @n1ru4l 's feedback!
https://github.com/ardatan/whatwg-node/pull/1899

Urigo: @ardatan I'm completely clueless about the whole `waitUntil` and `disposble` stuff
where can I read and learn about it?

.changeset/spotty-toes-help.md
```diff
@@ -0,0 +1,17 @@
+---
+'@graphql-hive/yoga': patch
+---
+
+Align with GraphQL Yoga >=5.10.4 which has the relevant features mentioned below;
+
+- Remove \`tiny-lru\` dependency, and use `createLruCache` from `graphql-yoga`
+- - Yoga already provides a LRU cache implementation, so we can use that instead of having a separate dependency
```
Urigo, line 8: what is the difference between Yoga's implementation and `tiny-lru`?

ardatan: Yoga uses `lru-cache` package, and I don't think there is a big difference. My idea is to avoid introducing an extra dependency in the plugin but use the one Yoga provides to reduce the duplicate code.

Urigo: if there is no difference, why does Yoga adds this to itself?

packages/libraries/yoga/src/index.ts
ardatan: This is from my other PR adding documentation about this logic;
This is the section I added to our documentation -> https://github.com/dotansimha/graphql-yoga/blob/4743ee974ed7decba8d89b26251e214d7cf42885/website/src/pages/docs/features/explicit-resource-management.mdx

This is the PR -> https://github.com/dotansimha/graphql-yoga/pull/3567

.changeset/spotty-toes-help.md
ardatan: In Yoga, we use LRU caching for parsing and validation caching, and we preferred `lru-cache` at some point. 
But this plugin introduces a new dependency `tiny-lru` which does the same job. In order to avoid introducing new dependencies doing the same job, and it is an unnecessary dependency especially for the bundle size. But instead, Yoga exports `createLRUCache` function for plugins to re-use the same thing Yoga uses to avoid extra dependencies, code etc.

Urigo: yes I understand
I'm just asking, if the fact that now you expose this functionality means we need to maintain it and people will rely on it, and if so, is it really yoga's job?

Urigo: how important is the bundle size here?

ardatan: > how important is the bundle size here?

In Yoga, we always avoided to introduce new dependencies, duplicate code etc for bundle size. Extra dependencies mean extra bundle size not just for Yoga but also for Gateway which has this package built-in.

> I'm just asking, if the fact that now you expose this functionality means we need to maintain it and people will rely on it, and if so, is it really yoga's job?

We already maintain LRU caching in Yoga. So I don't think it is too much work here. It is already there in Yoga and used by many plugins.

ardatan: For me any extra non-optimal thing is important because when they become together, it becomes an issue that we need to solve later by analyzing each of these. I just want to prevent them growing before.

Urigo (approved):

packages/libraries/yoga/src/index.ts
```diff
@@ -247,5 +252,10 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
         }),
       );
     },
+    onDispose() {
+      if (hive[autoDisposeSymbol]) {
+        return hive.dispose();
+      }
```
n1ru4l, line 258: Why do we add this check here?

ardatan: I think it is future proof to have this call. So anything needs to be done during disposal will be done here. Later on, maybe you have some logic for disposal in the Hive client.

.changeset/spotty-toes-help.md
```diff
@@ -0,0 +1,17 @@
...
+- Use Explicit Resource Management of GraphQL Yoga plugins for disposal which already respect Node.js process termination
+- - The feature has been implemented in `@whatwg-node/server` which is used by GraphQL Yoga. [Learn more about this feature](https://github.com/ardatan/whatwg-node/pull/1830)
+- Use \`context.waitUntil\` which is handled by the environment automatically, if not `@whatwg-node/server` already takes care of it with above.
+- - The feature has been implemented in `@whatwg-node/server` which is used by GraphQL Yoga. [Learn more about this feature](
+    https://github.com/ardatan/whatwg-node/pull/1830)
+- Respect the given `fetchAPI` by GraphQL Yoga(might be Hive Gateway) for the `fetch` function
+- - Hive Gateway uses `fetchAPI` given to GraphQL Yoga for the entire Fetch API implementation, so if Hive Client respects that, we have a control over HTTP calls done internally by the gateway
+- Respect Yoga's \`logger\` for logging
+- - Similar to above, we want to respect the logger provided by Yoga to have a better control over logging
```
n1ru4l, line 17: I tri
```suggestion
- Upgrade to `graphql-yoga` >= `5.10.4`
- Improve graceful process termination on Node.js by leveraging `graphql-yoga`'s [dispose lifecycle hooks](https://the-guild.dev/graphql/yoga-server/docs/features/envelop-plugins#ondispose)
- Improve Cloudflare Worker runtime support by registering pending usage reporting requests using the [`waitUntil` API](https://developers.cloudflare.com/workers/runtime-apis/context/#waituntil)
- Use the logger instance from the `graphql-yoga` instance, for a more unified logging experience
- Use the `fetch` API implementation on the `graphql-yoga` instance for HTTP calls
- Replace the internal \`tiny-lru\` dependency with `graphql-yoga`'s internal LRU cache implementation
```

ardatan: Looks way better! Thanks 🙏🏻

packages/libraries/yoga/src/index.ts
```diff
@@ -147,46 +135,66 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
...
+            ...clientOrOptions,
+            agent: clientOrOptions.agent
+              ? {
+                  logger: {
+                    // Hive Plugin should respect the given Yoga logger
+                    error: (...args) => yoga.logger.error(...args),
+                    info: (...args) => yoga.logger.info(...args),
+                  },
+                  ...clientOrOptions.agent,
+                  __testing: {
```
n1ru4l, line 187: Maybe lets drop the `__testing` thing and add a `fetch` option on the top level instead?

n1ru4l: @kamilkisiela What do you think?

kamilkisiela: Yeah. The `__testing` is just used for testing, it's not meant to be used this way. We can do a top level setting.

ardatan: I changed it to be a top level setting. Let me know what do you think?

n1ru4l (approved):

PR #6119, opened 2024-12-13
docs: add documentation for upstream reliability (retry + timeout)

EmrysMyrddin:
Documentation for new upstream retry and timeout feature https://github.com/graphql-hive/gateway/pull/322

### DIFF
--- a/packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
+++ b/packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
@@ -83,7 +83,7 @@ export const gatewayConfig = defineConfig({
 
 The default delay to wait between tries.
 
-If the subgraph is using HTTP transport, the delay is replaced by the response's `Retry-After`
+If the subgraph is using HTTP protocol, the delay is replaced by the response's `Retry-After`
 header if present.
 
 The `Retry-After` can either contain a number of second or a date.
--- a/packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
+++ b/packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
@@ -7,7 +7,7 @@ description: Improve reliability of Hive Gateway with timeout and retry policy f
 Hive Gateway offers Retry and Timeout features to improve the reliability of the traffic between the
 gateway and the subgraphs.
 
-## Retry
+## Retry mechanism
 
 A large range of errors happening between the gateway and the upstream subgraphs are temporary
 errors. Which means a lot of errors can be solved by issuing the same request again after a short
@@ -30,7 +30,7 @@ sequenceDiagram
   Gateway->>Client: { data: { field: "value" } }
 ```
 
-### Configuration
+### Global configuration
 
 This feature can be enabled for all subgraphs at once by using `upstreamRetry` option.
 
@@ -53,7 +53,7 @@ export const gatewayConfig = defineConfig({
 })
 ```
 
-#### Configuration by subgraph
+### Per-subgraph configuration
 
 Hive Gateway lets you define different retry policy for each subgraph by providing a function to the
 `upstreamRetry` option
@@ -79,7 +79,7 @@ export const gatewayConfig = defineConfig({
 })
 ```
 
-#### Delay
+### Retry delay configuration
 
 The default delay to wait between tries.
 
@@ -99,7 +99,7 @@ export const gatewayConfig = defineConfig({
 })
 ```
 
-#### Backoff
+### Exponential backoff configuration
 
 To avoid adding pressure on a subgraph that is already potentially struggling with load, a standard
 method is to use an exponential backoff.
@@ -127,7 +127,7 @@ export const gatewayConfig = defineConfig({
 })
 ```
 
-#### Skip a request
+### Conditional retry logic
 
 Hive Gateway offer the possibility to skip retrying a specific upstream request, based on its
 context or returned value.
@@ -146,13 +146,13 @@ export const gatewayConfig = defineConfig({
 })
 ```
 
-## Timeout
+## Timeout management
 
 An upstream request timeout limit is the maximum time the gateway waits for a response from an
 upstream subgraph before aborting the request. This ensures the gateway remains responsive and
 prevents it from waiting indefinitely for a slow or unresponsive subgraphs.
 
-### Configuration
+### Global configuration
 
 Hive Gateway allows to configure a global timeout for all subgraphs at once.
 
@@ -165,7 +165,7 @@ export const gatewayConfig = defineConfig({
 })
 ```
 
-#### Configuration by subgraph
+### Per-subgraph configuration
 
 You can finely grain configure the timeout for each upstream execution request by providing a
 function instead of a number.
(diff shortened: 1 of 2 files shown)


ardatan (approved):

ardatan (approved):

ardatan:
After merging this, let's answer here and let them try this
https://github.com/graphql-hive/gateway/issues/303

n1ru4l (approved):
nice!

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
```diff
@@ -0,0 +1,212 @@
...
+import { defineConfig } from '@graphql-hive/gateway'
+
+export const gatewayConfig = defineConfig({
+  upstreamRetry: {
+    maxRetries: 3
+  }
+})
+```
+
+#### Configuration by subgraph
```
dotansimha, line 56: Per-subgraph Configuration?

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+})
+```
+
+#### Delay
+
+The default delay to wait between tries.
+
+If the upstream execution is using `fetch`, the delay is replaced by the response's `Retry-After`
```
dotansimha, line 86: I feel like this section is not 100% clear, what does it mean "If the upstream execution is using `fetch`"?

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
```diff
@@ -0,0 +1,212 @@
...
+```ts filename="gateway.config.ts"
+import { defineConfig } from '@graphql-hive/gateway'
+
+export const gatewayConfig = defineConfig({
+  // The maximum time in milliseconds to wait for a response from the upstream.
+  upstreamTimeout: 1000
+})
+```
+
+#### Configuration by subgraph
```
dotansimha, line 168: Per-subgraph configuration?

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
```diff
@@ -0,0 +1,212 @@
+---
+description: Improve reliability of Hive Gateway with timeout and retry policy for upstream requests
```
dotansimha, line 2: Maybe this article should also include our recommended setup?

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
EmrysMyrddin: If the subgraph is using HTTP transport, or any other transport using `fetch` to communicate with upstream.

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
```diff
@@ -0,0 +1,212 @@
...
+
+export const gatewayConfig = defineConfig({
+  upstreamRetry: {
+    maxRetries: 3,
+    retryDelayFactor: 1.25 // default value. Use a value of 1 to disable backoff.
+  }
+})
+```
+
+#### Skip a request
```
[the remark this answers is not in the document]
EmrysMyrddin: Yes I agree, I was struggling finding a good title for this one... if you have a suggestion, I don't know how to call it

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
EmrysMyrddin: Humm, I mean, the setup is highly variable depending on upstream implementation and client requirements.

The recommendation is to setup retry and timeout, the actual values are up to the user.

Or you want to show of very simple example with both retry and timeout enabled ?

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
ardatan: @EmrysMyrddin We can tell if the upstream execution uses HTTP protocol. `fetch` is too detail maybe?

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
ardatan: I think it is a bit tricky to provide some values because it is very specific to user's configuration/setup.
For retry, `Retry-After` header is the best way to control the gateway's behavior. For timeout, mayyybe we can give a value like 5 seconds or sth but still too tricky to recommend something in  my opinion.

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
```diff
@@ -0,0 +1,212 @@
+---
+description: Improve reliability of Hive Gateway with timeout and retry policy for upstream requests
+---
+
+# Upstream reliability
```
kamilkisiela, line 5: Screenshot 2024-12-17 at 09 20 06

Maybe instead of:
```
Retry
  Configuration
    Configuration by subgraph
    Delay
    Backoff
    Skip a request
Timeout
  Configuration
    Configuration by subgraph
Mixing Retry and Timeout
```

Maybe we should do:

```
Retry mechanism
  Global configuration
  Per-subgraph configuration
  Retry delay configuration
  Exponential backoff settings
  Conditional retry logic
Timeout management
  Global configuration
  Per-subgraph configuration
Combining retry and timeout features
```

@ardatan @EmrysMyrddin @dotansimha 

What do you think?

EmrysMyrddin: Thank you ! I was struggling very hard to find proper titles here ^^'

dotansimha (approved):

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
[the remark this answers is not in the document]
dotansimha: `Conditional Retry`?

packages/web/docs/src/pages/docs/gateway/other-features/upstream-reliability.mdx
```diff
@@ -0,0 +1,212 @@
...
+    }
+  }
+})
+```
+
+### Retry delay configuration
+
+The default delay to wait between tries.
+
+If the subgraph is using HTTP transport, the delay is replaced by the response's `Retry-After`
```
ardatan, line 86:
```suggestion
If the subgraph is using HTTP protocol, the delay is replaced by the response's `Retry-After`
```
It can a transport that uses HTTP not just HTTP transport.

Urigo:
This is great
should you also add an example to the examples to link people to?

ardatan:
> should you also add an example to the examples to link people to?

I don't think we have `examples` in the gateway repository yet.

Urigo:
>  I don't think we have examples in the gateway repository yet.

Should we?

EmrysMyrddin:
@Urigo I don't think we have an e2e test for this feature yet. It's the closest form of "examples" we have in this repo.
Should I add one ? Or we merge this PR first and add this test later ?

Urigo:
@EmrysMyrddin I suggest that you'll add an example, without tests for now
That would means that me and others can use that example to learn and we can decide later on if and how to test it
Seeing and running an example for this would help me, thank you!

ardatan:
@EmrysMyrddin I am on it. I'm creating an E2E then later it will be an example

PR #6122, opened 2024-12-13
feat(@graphql-hive/cli): json output

### LINKED ISSUE #3592: JSON output format for the CLI

jasonkuhrt:
closes #3592

<details>
<summary>Previous description</summary>

This PR introduces a new `--json` flag to the Hive CLI to output
structured data.

One benefit of this feature is that it makes basic scripting in CI with
e.g. bash easier because:

1. User is guarnateed an output adhearing to semver. They can build
   against its data shape with confidence. Compare to the fact that
   human-readable output is not covered by semver.
2. User can apply tools like `jq` to slice and dice and assert on the
   output.

There are also a number of refactors aimed at supporting a uniform rollout of JSON output

1. Removal of config commands.
2. No more try-catch on the entirety of the `run` block. Moved to a `catch` method on `CommandBase` class.
3. Instead of `if-else if-else if-else` chains, `if-return` blocks that end with an `casesExhausted`.
4. A new `runResult` method that avoids the lack of type safety around of `throw` for the purpose of outputting JSON in failure cases

----

### About the implementation 

The approach used in this PR is to have a schema for the CLI. Each command should have a known set of strongly typed outputs, just like any API. This includes failures as well as successes.

OClif takes the [unfortunate] approach of not treating errors as data. This makes achieving schema coverage across all cases harder. A new small method `runResult` has been introduced to introduce the ability to treat failures as errors.

The CLI schema went through multiple iterations until I landed on the following general approach:

1. We introduce the concept of "output data types" that co-locate the representations it has: json (the data, basically), text (derived from the data)
    - This represents a shift away from outputting directly against GraphQL response data. For example when rendering text for schema changes we now render that text against the output data schema, not the graphql response data. This is logical since the output formats contain the same semantics (give or take) and so having each representation be based on the same data makes sense. JSON or Text are both based on the same schema.
2. ~Try to forward our GraphQL schema where reasonable to do so~ (intent to match dropped)
3. Every command outputs a discriminant union. The discriminant property is `__typename` because it aligns with the GraphQL convention. Another convention would be to use `_tag` which is what Effect and fp-ts for example use. For now, it seems that `__typename` is a reasonable choice?
4. Unlike our GrpahQL schema, we wrap an envelope around our types which provides categorization between failure and success.
5. The envelope data field should be reserved for information that was produced while running side-effects e.g. API response data or other sources of data (e.g. disk).

</details>

This PR introduces JSON output capability to Hive CLI

### Use-cases

- Shell scripting, particularly with `jq` (or similar)

### Implementation Points

1. Removal of config commands (based on PR feedback)
2. No more try-catch on the entirety of `run`. Using `catch` method on base class.
3. Removal of `if-else if-else if-else` chains in favour of `if-return` blocks that end with an `casesExhausted`.
4. A `runResult` method alternative to `run` for safe control flow
    - `return`, no `throw` is the goal, not yet reached in all cases
5. A new command output definition and runtime system
    - Union of data types, schemas defined with Typebox
    - Each data type MAY colocate text rendering for that data type
6. New "Texture" library for text manipulation (`Text` is a global variable :/)

##### Output System

The new output system shifts how we think about layers of the application (cli):

```
1 Input parsing 
2 Action
3 Output mapping                   <- new
4 Output formatting (text,json)    <- changed
```

Output mapping: Map from the result of the action to a defined output schema

Output formatting: Run a formatter against the output schema to emit to the environment. We have two formatters now: JSON (new), Text (existing). Both of these receive data of the output schema. **Important**: formatters should work from a consistent source of data. So, to that end, whereas text was previously was based directly on action results, it is now based on the output schema (1). This is a key design decision in the implementation.

The motivation for this approach is thus: If the output schema changes we want it to affect all formatters so that it forces us to consider the UX of each formatter. For example, if we want to add some new data to the JSON output, we may want, and certainly should consider, if the text formatting should also be updated.

Another motivation is that co-locating formatters around one data type groups them by domain rather than technology. It is very likely that changing a particular output data type will require changes to multiple formatters because they have a shared concern of displaying information about the same topic (e.g. `SuccessAppPublish`).

- (1) Aside: Text formatters also have access to the parsed input since that is often useful for text formatters.

##### Testing

I used snapshots for testing. This has tradeoffs. I think its the right approach but here are some points.

1. There is a bug in Vitest (Jest also had this issue) in regards to filtering and snapshots. I've started a [discussion](https://github.com/vitest-dev/vitest/discussions/7106#discussioncomment-11634480) that @hi-ogawa has replied to and seems to have reproduced the issue. I think we will need this bug fixed to have a reasonable story around our current testing.
2. There are some specific Vitest workflows in the terminal that I think are important for managing/fixiing/updating tests in the CLI. They are not particularly obvious and should be documented and optimized further by the team. I will add this to a `CONTRIBUTING.md` which I've been meaning to make a PR for anyways. We will need point (1) resolved though. Its been pretty clumsy without that working.
3. Here's the overall complexity we're dealing with that necessitates some docs for effectively workflows:
    1. Snapshot tests
    2. Many tests
    3. Slow tests
    4. Matrix tests (test.each/describe.each etc.)

    This combination makes the effect of (1) very painful. For example matrix tests along with slow tests mean that we absolutely need to be using `vitest ... --test-name-pattern 'abc'` when developing. But then we cannot use `-u` because that would remove obsolete snapshots (Vitest thinking all the skipped ones are obsolete).

    My PR to `CONTRIBUTING.md` will cover these and other points. For example using `vitest run ... / vitest --run` to avoid watch mode, how to effectively mix `$ git ...` usage with snapshot updates, etc.

##### Areas for Future Improvement

1. The need for us to `throw` failures so that:
    1. It goes through internal (and complicated) `handle` function
    6. It goes through `catch`
    7. It goes through `toErrorJson`
    8. It goes to stdout

    Adds needless complexity in the form of indirection, magic, and lack of type safety. I could not find an obviously reliable way to circumvent it. Ditching OClif would be great. Otherwise, with some more time, we could investigate more a simpler approach (such as deep importing OClif utilities to achieve the effect without doing a throw hop.

2. There are many areas of the commands that are not yet integrated fully to be using a result type approach.

5. The schema design feels unresolved to me. Things I'm still mulling over:
    1. ~Allow exit codes to be expressed in the envelope functions somehow. Then they should be propagated to OClif somehow. [Related discussion](https://github.com/oclif/core/discussions/1270).~ resolved
    2. We could use our data type names as error codes for documentation links and issue reporting.
    3. ~The various modules in `/output` feel too complicated to me still, just a feeling.~ resolved

6. Test coverage and test time performance are in bad shape. This PR makes perf worse by adding more tests. Coverage is not improved by this PR.

----

### TODO

- [x] Try TypeBox instead of Zod
- [x] Do a couple tests to understand how OClif handles JSON output with regards to errors. It is unclear from the docs nor can I find much on the web.
- [x] command schema publish JSON output
- [x] Final pass on JSON output of commands
- [x] Get CI passing
- [x] Integrate review feedback from others, if any
- [x] [Integration tests](https://github.com/graphql-hive/console/pull/6122#issuecomment-2550569846)
- [x] Co-locate text format renders with output schema
- [x] [add a changeset file](https://github.com/graphql-hive/console/pull/6122#issuecomment-2550570829)


jasonkuhrt:
I am looking for early feedback on this draft PR to avoid going in a direction that won't be merged.

The approach I am taking is as follows:
1. Leverage https://oclif.io/docs/json.
2. Bring better type safety to json output with some helpers built into our base command.

.vscode/settings.json
```diff
@@ -28,5 +28,8 @@
     ["cx\\(([^)]*)\\)", "(?:'|\"|`)([^']*)(?:'|\"|`)"],
     ["cn\\(([^)]*)\\)", "(?:'|\"|`)([^']*)(?:'|\"|`)"],
     ["clsx\\(([^)]*)\\)", "(?:'|\"|`)([^']*)(?:'|\"|`)"]
-  ]
+  ],
+  "[typescript]": {
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
+  }
```
jasonkuhrt, line 34: Background not his change:

I use [`dprint`](https://dprint.dev) on my projects and so far VSCode has not being smart enough to automatically figure out when a project is using prettier or dprint. When I told it, it seems that it stored that information into the project local settings.

I assume this is harmless for everyone else on the team.

packages/libraries/cli/src/base-command.ts
```diff
@@ -304,6 +330,29 @@ class ClientError extends Error {
   }
 }
 
+export type Flags<T extends typeof Command> = Interfaces.InferredFlags<
```
jasonkuhrt, line 333: In general I try to bias toward a "literate" style of code flow when possible meaning higher level concepts precede lower ones. In practice that looks like references progressing downward which is what I've done here.

packages/libraries/cli/src/base-command.ts
```diff
@@ -304,6 +330,29 @@ class ClientError extends Error {
...
+    : 'Error: Missing `static successDataSchema = ...` on your command.';
+
+type OptionalSuccessSchema = undefined | SuccessSchema;
+
+type SuccessSchema =
+  | z.SomeZodObject
+  | z.ZodUnion<[z.SomeZodObject, z.SomeZodObject]>
+  | z.ZodUnion<[z.SomeZodObject, z.SomeZodObject, z.SomeZodObject]>
+  | z.ZodUnion<[z.SomeZodObject, z.SomeZodObject, z.SomeZodObject, z.SomeZodObject]>;
+
```
jasonkuhrt, line 355: In general I try to bias toward a "literate" style of code flow when possible meaning higher level concepts precede lower ones. In practice that looks like references progressing downward which is what I've done here.

packages/libraries/cli/src/commands/config/set.ts
n1ru4l, line 1: It's pretty safe to just remove this command instead of investing time in figuring out the JSON output. Nobody uses it.

packages/libraries/cli/src/commands/config/get.ts
n1ru4l, line 1: It's pretty safe to just remove this command instead of investing time in figuring out the JSON output. Nobody uses it.

packages/libraries/cli/src/commands/config/delete.ts
n1ru4l, line 1: It's pretty safe to just remove this command instead of investing time in figuring out the JSON output. Nobody uses it.

packages/libraries/cli/src/helpers/output-schema.ts
```diff
@@ -0,0 +1,41 @@
...
+    export const Stdout = z.object({
+      outputMode: z.literal('stdout'),
+      content: z.string(),
+    });
+    export const File = z.object({
+      outputMode: z.literal('file'),
+      path: z.string(),
+    });
+  }
+}
```
n1ru4l, line 41: Can you give some context on using namespaces here instead of just exporting an object?

```ts
export const Effect = {
  Skipped: Enevlop.extend({ /* ... */ }),
  /* ... */
  DataOutputMode: {
     Stdout: z.object({ /* ... */ }),
     /* ... */
  },
}
```

This is a new pattern introduced to the code base here, and I wonder about its benefits.

jasonkuhrt: The answer to this is subtle.

Conceptually `namespace` model inline modules but in practice they only do so for types whilst for values they actually just compile to objects at runtime (but only _if there are any values exported from them_).

They have a special property in TS which is that they can support exporting types as well. And those type names can overlap with the value names. This _sometimes has advantages for library design_. In practice, what `namespace` can do so can additional ES modules. The "problem" with ES modules is how heavyweight it is to create new files be comparison of inline, single-line modules. And, when ES modules _are_ used, then, you actually deal with a statically analyzable structure, which plain objects are not. This is relevant for Tree shaking for example.

Related to all this is a tc39 proposal for inline modules that, if make it into the language, I suspect will allow the use cases for `namespace` to be entirely replaced by something native to JS, which would be nice https://github.com/tc39/proposal-module-declarations.

In terms of code style, returning to this code block... here is the summary of reasons using `export const Effect ...` is not 1:1:

1. It is true that `namespace` _when there are values exports_ compiles to `export const ...` however that can be seen as an implementation detail. From a readability point of view we can view `namespace` to mean this: "I want this to be a separate namespace but I was too lazy or unsure yet about the direction to warrant creating more files with `* as X` imports etc."

2. As soon as we want to do something as follows then `namespace` (or ESM modules) is the _only way_:

	```ts
	export namespace Effect {
	  export const Skipped = ...
	  export type Skipped = z.infer<typeof Skipped>
	}
    // ...
    // Somewhere else in the codebase:
    // ...
    const foo: Effect.Skipped = ...
	```

In terms of usage, some additional points:

  - There are some cases where IDEs currently have poor intellisense for autocomplete of values/types from ESM namespaces whereas they fair better with `namespace`. This is a consideration.
  - Every single instance of `namespace` is an instance where ESM modules could have been used instead. But `namespace` is a lot easier to change since there's no extra files involved. So sometimes when I'm introducing some new code/ideas, I start with `namespace` because I'm not 100% sure yet how that code will evolve. Mature code in libraries should probably usually go towards ESM modules where tree shaking support is important. For apps, tree-shaking may not matter, in which case ESM modules vs namespaces don't matter much either.

packages/libraries/cli/src/helpers/config.ts
jasonkuhrt, line 1: This is the last place we're using Zod in this package. I considered refactoring it now to have validation be 100% TypeBox but I think it's best to do that as a separate PR. I don't know what this code does and risk adding regressions/complexifying this PR/adding review time etc. All I've done for now is refactor it to use the native Zod namespace.

packages/libraries/cli/src/commands/schema/check.ts
```diff
@@ -86,7 +98,51 @@ const schemaCheckMutation = graphql(/* GraphQL */ `
...
+  column: Typebox.Nullable(Typebox.Number()),
+});
+type Warning = Typebox.Static<typeof Warning>;
+
 export default class SchemaCheck extends Command<typeof SchemaCheck> {
+  static SuccessSchema = Typebox.Union([
+    OutputSchema.Envelope.extend({
+      data: Typebox.Object({
+        // todo is this the right "term" for this check type?
+        checkType: Typebox.Literal('registry'),
```
jasonkuhrt, line 131: Is there a better term we could use here than `registry`?

jasonkuhrt:
https://github.com/oclif/core/discussions/1270

packages/libraries/cli/src/commands/dev.ts
```diff
@@ -83,6 +83,15 @@ type ServiceWithSource = {
...
 };
 
+// TODO add JSON output support.
+// Unlike other commands that return, this common kicks off a long running process in the terminal
+// that outputs messages to the terminal over time.
+// Therefore the OClif framework pattern of returning an object is incompatible.
+// We'll need to inspect the json flag ourselves and return the appropriate output.
+//
+// Presumably users would typically NOT use JSON output for this command. This task appears to be motivated by
+// the principal of simplicity via consistency.
```
jasonkuhrt, line 93: Can we leave this for a future PR?

packages/libraries/cli/src/base-command.ts
```diff
@@ -45,26 +68,66 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
...
+   */
+  logInfo(...args: any[]) {
     this.log(colors.yellow('ℹ'), ...args);
   }
 
-  infoWarning(...args: any[]) {
+  /**
+   * {@link Command.log} with warning styling.
+   */
+  logWarning(...args: any[]) {
```
jasonkuhrt, line 130: I renamed these to make our commands more readable. For example `this.fail`  versus `this.error` have very different behaviour but that's not clear, nor is it easy to remember which is which.

packages/libraries/cli/src/commands/app/create.ts
```diff
@@ -129,21 +145,29 @@ export default class AppCreate extends Command<typeof AppCreate> {
...
+        id: result.createAppDeployment.ok.createdAppDeployment.id,
+        name: flags['name'],
+        version: flags['version'],
+      },
+    });
   }
 }
 
-const ManifestModel = z.record(z.string());
+const ManifestModel = Typebox.Record(Typebox.String(), Typebox.String());
```
jasonkuhrt, line 170: This was a simple Zod use-case so I replaced with Typebox which we're moving towards now.

packages/libraries/cli/src/commands/whoami.ts
```diff
@@ -109,12 +120,31 @@ export default class WhoAmI extends Command<typeof WhoAmI> {
...
+    }
+
+    if (result.tokenInfo.__typename === 'TokenNotFoundError') {
       this.error(`Token not found. Reason: ${result.tokenInfo.message}`, {
         exit: 0,
         suggestions: [`How to create a token? https://docs.graphql-hive.com/features/tokens`],
       });
     }
+
+    casesExhausted(result.tokenInfo);
```
jasonkuhrt, line 147: You'll see this in a few places. It moves towards a stricter switch case pattern where new union cases should definitely raise a static error in our code.

packages/libraries/cli/src/base-command.ts
```diff
@@ -4,16 +4,38 @@ import type { ExecutionResult } from 'graphql';
...
 import { Config, GetConfigurationValueType, ValidConfigurationKeys } from './helpers/config';
+import { OmitNever } from './helpers/general';
+import { Envelope, OutputType } from './helpers/output-schema';
+import { Typebox } from './helpers/typebox/__';
 
-export type Flags<T extends typeof Command> = Interfaces.InferredFlags<
-  (typeof BaseCommand)['baseFlags'] & T['flags']
->;
-export type Args<T extends typeof Command> = Interfaces.InferredArgs<T['args']>;
+export default abstract class BaseCommand<$Command extends typeof Command> extends Command {
```
jasonkuhrt, line 13: I often use `$` prefix type names to maintain readability whilst also separating type parameters from other types.

packages/libraries/cli/src/helpers/typebox/_.ts
jasonkuhrt, line 1: I re-export the Typebox lib for easier importing (namespace) and to seamlessly merge our extensions to it.

jasonkuhrt:
@n1ru4l My day was slowed down a bit by wrangling OClif indirections (I'm not a fan of its DX, but will survive) and getting up to speed on Typebox. After some iterations I think the PR is getting to a solid place. I'm hoping tomorrow I can get it into a mergeable shape. Looking forward to your feedback. And as mentioned in a comment, I'm happy to break this up into multiple PRs if that helps the review here.

n1ru4l:
@jasonkuhrt one PR seems fine for now. I will give my further feedback once we are in a mergeable state!

jasonkuhrt:
@n1ru4l I've updated the PR description with some additional thinking behind this PR.

Tomorrow I'll look forward to your feedback and address it ASAP.

I would feel better if we could add e2e tests before merging this but I'm not sure how feasible that will be. I could see it adding another 1-2 days of work for me. Let me know what you think.

.vscode/settings.json
dotansimha: Checked locally and it seems harmless. I assume most of us are using prettier by default anyway, so it doesn't really change anything ✅

packages/libraries/cli/src/base-command.ts
dotansimha: I think we are not using this approach in the codebase at the moment, but I don't have any strong opinion here. It seems readable and easy to understand.

packages/libraries/cli/src/commands/dev.ts
dotansimha: Yes, it can be handled later (if needed at all). I think the strongest use-case for JSON output is for CIs and for additional processing of the commands output. `dev` is different by nature so it can wait.

packages/libraries/cli/src/fragments/__.ts
```diff
@@ -0,0 +1 @@
+export * as Fragments from './_';
```
dotansimha, line 1: needed? (both `_` and `__`)

packages/libraries/cli/src/base-command.ts
```diff
@@ -1,19 +1,33 @@
...
+  public static enableJsonFlag = true;
 
-type OmitNever<T> = { [K in keyof T as T[K] extends never ? never : K]: T[K] };
+  /**
+   * A description of the action that this command performs.
+   * Used in certain automated error messages.
+   */
+  public static descriptionOfAction = 'perform action';
```
dotansimha, line 22: should be changed?

dotansimha:
Seems like we are in a great direction. One thing that is worth adding here is integration tests.

I have these scenarios in mind: run all relevant commands with the JSON output flags, and verify all outputs (success/error). Also verify that the JSON output is parsable (just `JSON.parse` it, I guess?) 

Today we already have integration tests for CLI, so we can just add a few more: https://github.com/graphql-hive/console/tree/main/integration-tests/tests/cli

dotansimha:
Also @jasonkuhrt , can you please add a changeset file (`pnpm changeset` in the root and it will guide you). It will trigger CI to also release an alpha version of the CLI so we can try it out.

packages/libraries/cli/src/fragments/__.ts
jasonkuhrt: AFAIK Two modules is the only way to get a set module exports under one exported namespace. The names `_` and `__` could be anything. Do you know of a way to only need one module? And/or do you have alternate name suggestions?

dotansimha: `index.ts`? I'm fine with `_` / `__`, no real strong opinions here.

packages/libraries/cli/src/commands/dev.ts
jasonkuhrt: Agreed!

integration-tests/testkit/cli.ts
```diff
@@ -35,40 +37,18 @@ async function exec(cmd: string) {
...
-}
-
-async function dev(args: string[]) {
-  const registryAddress = await getServiceHost('server', 8082);
-
-  return await exec(
-    ['dev', `--registry.endpoint`, `http://${registryAddress}/graphql`, ...args].join(' '),
-  );
-}
+export const hiveCLI = HiveCLI.create({
```
n1ru4l, line 40: @jasonkuhrt  Can we please keep this out of the scope of this PR, I feel like it is getting diverged from the original intent.

n1ru4l:
> Every command outputs a discriminant union. The discriminant property is __typename because it aligns with the GraphQL convention. Another convention would be to use _tag which is what Effect and fp-ts for example use. For now, it seems that __typename is a reasonable choice?

I am fine with using `__typename`, even though it is super graphql specific. Maybe just using the name `type` is good enough? Not sure if we need that prefix underscore on whatever naming we choose.

integration-tests/testkit/cli.ts
jasonkuhrt: I can definitely understand that and was partly expecting a comment like this :). In my defense, the inferred library is arguably in scope in the sense that it provides the base to rapidly test each command, many of which are currently not tested.

There are a few approaches I could take here:

1. Test JSON using the existing pattern of re-creating wrappers for each command.
2. Test JSON using the inferred library, but do not update any existing tests to use that library yet.
3. Do not test JSON on any commands beyond those that are currently tested.

Please tell me which you prefer! 🙏

n1ru4l: For now, lets use the existing method of running the commands. We don't need to extensively test every single command.

Testing the schema publish, schema check and app publish flow is crucial and we can have json tests for these.

jasonkuhrt:
> I am fine with using __typename, even though it is super graphql specific. Maybe just using the name type is good enough? Not sure if we need that prefix underscore on whatever naming we choose.

Few thoughts:
- If inheriting styles from GraphQL here feel awkward let's drop that consideration from discussion. So let's forget `__typename` key name.
- IMO The commands should be returning discriminant unions, but currently they are nested in an envelope which breaks the concept. I now think we should raise the branding to the envelope level. This is predictable simple and consistent with how we'd generally think about data in an application.
- If we lift the branding from `data` then there are no namespace issues, and `type` works fine. Otherwise we need to consider naming more carefully.

packages/libraries/cli/src/fragments/__.ts
jasonkuhrt: `index.ts` for `_` or `__`? And what do we call the other one (Or, perhaps you also feel code shouldn't be responsible for its own namespace in which case __ modules go away)?

In case its not clear: I'm keeping in mind that once we are using ESM internally `index` will no longer have any special behaviour and so imports will need to do `./a/b/c/index.js` rather than `./a/b/c`. That reduces advantage of `index`. One annoying thing about `index` is that in a directory its special status doesn't typically "rise to the top" of the directory listing yet its unique among all those modules and is better (IMO) locked in place consistently e.g. top/bottom whatever.

Having shared some of my thinking here, I'll stick with `_/__` since it sounds like you're open to it here, but also I want to continue to discuss this over time as we have more feelings and ideas about it too etc.

jasonkuhrt:
https://github.com/vitest-dev/vitest/discussions/7106

jasonkuhrt:
I have been struggling to get OClif to be consistent between my machine and ci. Stack traces are shown in CI. I don't know why yet. Trying various things.

jasonkuhrt:
Maybe related https://github.com/oclif/oclif/issues/229

jasonkuhrt:
https://github.com/oclif/core/discussions/1277

jasonkuhrt:
https://github.com/oclif/core/discussions/1277#discussioncomment-11632434

.github/workflows/tests-integration.yaml
```diff
@@ -71,8 +71,6 @@ jobs:
 
       - name: run integration tests
         timeout-minutes: 10
-        env:
-          HIVE_DEBUG: 1
```
jasonkuhrt, line 75: This causes stack traces to be rendered by OClif which breaks our CLI snapshot tests.

integration-tests/package.json
```diff
@@ -34,9 +36,9 @@
     "human-id": "4.1.1",
     "ioredis": "5.4.1",
     "slonik": "30.4.4",
-    "strip-ansi": "7.1.0",
+    "strip-ansi": "6.0.1",
```
jasonkuhrt, line 39: For our CLI output snapshot serializer.

integration-tests/package.json
```diff
@@ -34,9 +36,9 @@
...
     "tslib": "2.8.1",
-    "vitest": "2.0.5",
+    "vitest": "2.1.8",
```
jasonkuhrt, line 41: Thought this might be the version that improves terminal watch mode flickering. Doesn't seem to have worked but no matter, upgrade vitest all the same.

integration-tests/package.json
```diff
@@ -15,12 +15,14 @@
...
     "@esm2cjs/execa": "6.1.1-cjs.1",
     "@graphql-hive/apollo": "workspace:*",
+    "@graphql-hive/cli": "workspace:*",
     "@graphql-hive/core": "workspace:*",
     "@graphql-typed-document-node/core": "3.2.0",
     "@hive/rate-limit": "workspace:*",
     "@hive/schema": "workspace:*",
     "@hive/server": "workspace:*",
     "@hive/storage": "workspace:*",
+    "@sinclair/typebox": "^0.34.12",
```
jasonkuhrt, line 25: I like this library :)

integration-tests/package.json
```diff
@@ -15,12 +15,14 @@
     "@aws-sdk/client-s3": "3.693.0",
     "@esm2cjs/execa": "6.1.1-cjs.1",
     "@graphql-hive/apollo": "workspace:*",
+    "@graphql-hive/cli": "workspace:*",
```
jasonkuhrt, line 18: It would be great to get our coverage to _not_ be child-process based because each CLI invocation is taking 1-2 seconds which is terrible. Ideally we can reserve child process spawning tests for only a select few cases that really need that. I briefly played around with that while waiting for CI a few times. Currently this dep remains unused, we could remove.

integration-tests/testkit/cli.ts
```diff
@@ -1,38 +1,27 @@
...
     env: {
       OCLIF_CLI_CUSTOM_PATH: cliDir,
       NODE_OPTIONS: '--no-deprecation',
     },
   });
 
-  if (outout.failed) {
-    throw new Error(outout.stderr);
+  if (result.failed) {
+    throw new Error('CLI execution marked as "failed".', { cause: result.stderr });
```
jasonkuhrt, line 21: While trying to understand where stack traces and errors were coming from, I adjusted this to be more explicit.

integration-tests/tests/cli/__snapshots__/cli-error-user-input.spec.ts.snap
jasonkuhrt, line 1: This test suite helped me develop some better JSON coverage for CLI user input errors. My real motivation today though was to get a quick iteration loop going while debugging the stack traces issue. Anyways, there's use in having this coverage so I kept it here. Its not complete of course. To really take this all the way we're going to need to address https://github.com/graphql-hive/console/pull/6122/files#r1894362099 I think. Imagine dozens, into the hundreds, of these tests.

integration-tests/tests/cli/cli-error-user-input.spec.ts
jasonkuhrt, line 1: This test suite helped me develop some better JSON coverage for CLI user input errors. My real motivation today though was to get a quick iteration loop going while debugging the stack traces issue. Anyways, there's use in having this coverage so I kept it here. Its not complete of course. To really take this all the way we're going to need to address https://github.com/graphql-hive/console/pull/6122/files#r1894362099 I think. Imagine dozens, into the hundreds, of these tests.

package.json
```diff
@@ -117,7 +117,6 @@
       "@graphql-eslint/eslint-plugin@3.20.1": "patches/@graphql-eslint__eslint-plugin@3.20.1.patch",
       "got@14.4.4": "patches/got@14.4.4.patch",
       "slonik@30.4.4": "patches/slonik@30.4.4.patch",
-      "@oclif/core@3.26.6": "patches/@oclif__core@3.26.6.patch",
```
jasonkuhrt, line 120: Was there an OClif patch before my PR? Confused by this.

jasonkuhrt:
FYI I have rewritten the PR description.

.changeset/olive-ducks-fold.md
jasonkuhrt, line 1: Changelog here

packages/libraries/cli/src/helpers/tex/__.ts
n1ru4l, line 1: Why `tex`? To me it looks like a typo of `text`

jasonkuhrt: @n1ru4l Agree, unfortunately `Text` is a global variable. I'll try `texture` which doesn't look like a typo and is a bit playful fwiw. Probably will be easier to auto-import too.

n1ru4l: I don't think using `Text` is a big deal since the global variable is only relevant to Browser environments. `TextFormatTools` or something similar would also come to my mind. `Texture` reminds me more of video games.

integration-tests/testkit/test.ts
```diff
@@ -0,0 +1,102 @@
...
+  cliFederation: CLI;
+  //
+  // "stitching" branch
+  //
+  projectStitching: ProjectSeed;
+  targetAccessTokenStitching: TargetAccessTokenSeed;
+  cliStitching: CLI;
+}
+
+export const test = testBase.extend<Context>({
```
kamilkisiela, line 35: Why this and not:

```ts
test('...', async () => {
  const project = await createProject(...);
});
```

What's the reason to provide a custom `test` function?

integration-tests/testkit/test.ts
```diff
@@ -0,0 +1,102 @@
...
+  seed: async ({}, use) => {
+    const seed = await initSeed();
+    await use(seed);
```
kamilkisiela, line 38: The `use` function tells me nothing, I use this function and I have no idea what it does. I need to dig in and it increases complexity of the setup by 10x.

integration-tests/tests/cli/dev.spec.ts
```diff
@@ -1,33 +1,9 @@
...
 
 describe('dev', () => {
-  test('composes only the locally provided service', async () => {
-    const { createOrg } = await initSeed().createOwner();
-    const { createProject } = await createOrg();
-    const { createTargetAccessToken } = await createProject(ProjectType.Federation);
-    const { secret } = await createTargetAccessToken({});
-    const cli = createCLI({ readwrite: secret, readonly: secret });
-
+  test('composes only the locally provided service', async ({ cliFederation: cli }) => {
```
kamilkisiela, line 6: The `cliFederation` is weird to me. We have one CLI.

integration-tests/testkit/test.ts
jasonkuhrt: - Helps us write consistent test fixtures that are centrally managed, and think harder about cases where we deviate from standard fixtures, which IMO is the right balance because deviations increase maintenance cost.

- Fixtures build upon one another. When a test brings in a fixture, the dependencies of that fixture are automatically run. The result is focused test code in majority cases while maintaining possibility for any test to begin from the base fixture whereupon it builds up resources in some custom way unique to itself.

integration-tests/testkit/test.ts
jasonkuhrt: Vitest fixtures are powerful and worth learning, and you only need to learn it once. If you didn't learn it, it makes sense you don't know what it does. It's unfortunate that there's no JSDoc there. The gist of `use` is that it makes the given value available to a test e.g. `test('...', ({ seed }) => ...`.

kamilkisiela: To me, this:

```
test(..., async () => {
  const seed = seedProject()
})
```

is clear to everyone that will ever see the integration test. They can click on it, find it easily, and if we ever migrate from `vitest` to `jest`, we won't have to move this extra thing that is very specific to vitest.

Your solution means that something in the test case context is added magically and we need to explain it to every person that will ever encounter it.
If that person is not aware of this vitest feature, searching for `seed` will mean 120 search results with that keyword and somewhere in those results, there's will be a correct answer.

jasonkuhrt: Personally I would advocate for effective use of the tools we use. I wasn't aware of any plans to switch to Jest, is that what you're implying (confused because my understanding up until now was the opposite, moving away from Jest)? If your point is just that it is specific to Vitest, that's only partly correct. Playwright test framework also has a similar system in core (while Node test does not). 

Anyways I don't have more points to offer than here https://github.com/graphql-hive/console/pull/6122#discussion_r1905441725. I'll remove the use of fixtures so we can move on to other topics in this PR.

integration-tests/tests/cli/__snapshot_serializers__/cli-output.ts
```diff
@@ -0,0 +1,73 @@
...
+  test: (value: unknown) => {
+    if (typeof value === 'string') {
+      return variableReplacements.some(replacement => replacement.pattern.test(value));
+    }
+    return isExecaError(value);
+  },
+  serialize: (value: unknown) => {
+    if (typeof value === 'string') {
+      let valueSerialized = '';
+      valueSerialized += ':::::::::::::::: CLI SUCCESS OUTPUT :::::::::::::::::\n';
```
jdolle, line 15: I really like the format you went with for cli output. It captures all the important data and is very readable.
My only recommendation again would be that this be in a separate PR to isolate the json output work.

jasonkuhrt: Do you have a proposal about how to test the changes in this PR? Removing this would reduce confidence in the PR changes would it not?

jasonkuhrt: Internal team discussion led to a decision to make other smaller PRs. Let's see where that goes. May affect this thread.

jasonkuhrt:
We will be making a PR per command. In light of that, this PR is no longer valid. Though its changes are still relevant, and will be used piecemeal in new PRs.

PR #6123, opened 2024-12-13
fix(migrations): encode input to ensure valid postgres urls

Intellicode:
### Background

When passwords (or other parameters) contain special characters (such as `%` in a password), it will result in invalid postgres url's

### Description

The change encodes all injected input parameters, I can remove some if you think its unnecessary.

### Checklist

- [x] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


n1ru4l (changes requested):
Hey @Intellicode, thank you for this contribution. There are currently multiple `createConnectionString` copies within our code base e.g.

```
packages/services/storage/src/db/utils.ts
packages/migrations/src/connection-string.ts
```

Could you adjust all of them to use the same function definition with your improvements?

Also could you please add a changeset for `hive` using the command `pnpm changeset`, for describing this improvement?

Intellicode:
@n1ru4l Thanks for your feedback! I've modified the other places to include the logic, I've also made password optional, since some authentication mechanisms allow disabling of password based login (such as in Azure with Workload ID's on Kubernetes)

packages/migrations/src/environment.ts
```diff
@@ -43,7 +43,7 @@ const PostgresModel = zod.object({
   POSTGRES_PORT: NumberFromString,
   POSTGRES_DB: zod.string(),
   POSTGRES_USER: zod.string(),
-  POSTGRES_PASSWORD: zod.string(),
+  POSTGRES_PASSWORD: zod.string().optional(),
```
n1ru4l, line 46:
```suggestion
  POSTGRES_PASSWORD: emptyString(zod.string().optional()),
```

packages/services/storage/src/environment.ts
```diff
@@ -36,7 +36,7 @@ const PostgresModel = zod.object({
   POSTGRES_PORT: NumberFromString,
   POSTGRES_DB: zod.string(),
   POSTGRES_USER: zod.string(),
-  POSTGRES_PASSWORD: zod.string(),
+  POSTGRES_PASSWORD: zod.string().optional(),
```
n1ru4l, line 39:
```suggestion
  POSTGRES_PASSWORD: emptyString(zod.string().optional()),
```

n1ru4l (changes requested):
@Intellicode Can you also update the configuration of other services that access postgres and the READMEs (mark password as optional env variable).

Please also add a changeset for `graphql-hive`

n1ru4l (changes requested):
Updates to the following files are missing:
- https://github.com/graphql-hive/console/blob/main/packages/services/rate-limit/src/environment.ts
- https://github.com/graphql-hive/console/blob/main/packages/services/stripe-billing/src/environment.ts
- https://github.com/graphql-hive/console/blob/main/packages/services/tokens/src/environment.ts
- https://github.com/graphql-hive/console/blob/main/packages/services/usage-estimator/src/environment.ts
- https://github.com/graphql-hive/console/blob/main/packages/services/server/src/environment.ts

PR #6128, opened 2024-12-16
docs: Rework the Custom Plugins page

EmrysMyrddin:
Closes https://github.com/graphql-hive/gateway/issues/126


packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
```diff
@@ -0,0 +1,576 @@
...
+
+    SubgraphDelegation: Subgraph Delegation
+    state SubgraphDelegation {
+      onDelegationPlan: <a href="#ondelegationplan">onDelegationPlan</a>
+      onDelegationStageExecute: <a href="#ondelegationstageexecute">onDelegationStageExecute</a>
+      onSubgraphExecute: <a href="#onsubgraphexecute">onSubgraphExecute</a>
+      onSubgraphExecuteDone: <a href="#onsubgraphexecutedone">onSubgraphExecuteDone</a>
+
+      [*] --> onDelegationPlan
+      onDelegationPlan --> onDelegationStageExecute
```
ardatan, line 228: Not sure if worth to mention but plan can be memoized so this hook is not always called.

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
```diff
@@ -0,0 +1,576 @@
...
+It is the very first hook called just before the underlying GraphQL Yoga instance starts listening
+for HTTP requests.
+
+##### Possible usages
+
+You can use this hook to customize the Yoga instance.
+
+##### API
+
+- `yoga`: the Graphql Yoga `YogaServer` instance that is about to start.
```
ardatan, line 98:
```suggestion
- `yoga`: `YogaServer` instance that is about to start with the gateway.
```
Maybe

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
```diff
@@ -0,0 +1,576 @@
...
+      onDelegationPlan: <a href="#ondelegationplan">onDelegationPlan</a>
+      onDelegationStageExecute: <a href="#ondelegationstageexecute">onDelegationStageExecute</a>
+      onSubgraphExecute: <a href="#onsubgraphexecute">onSubgraphExecute</a>
+      onSubgraphExecuteDone: <a href="#onsubgraphexecutedone">onSubgraphExecuteDone</a>
+
+      [*] --> onDelegationPlan
+      onDelegationPlan --> onDelegationStageExecute
+      onDelegationStageExecute --> onSubgraphExecute
+      onSubgraphExecute --> onSubgraphExecuteDone
+
```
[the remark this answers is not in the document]
EmrysMyrddin: I don't know, I removed it because it depends on the transport you are using. `onFetch` is not called for all subgraphs

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
EmrysMyrddin: Ho, so only the `onDelegationStageExecute` is mandatory ?

If it's cached, `onDelegationPlan` is not called at all ?

ardatan: >Ho, so only the onDelegationStageExecute is mandatory ?

It is always called

>If it's cached, onDelegationPlan is not called at all ?

No it is not if I don't miss anything.

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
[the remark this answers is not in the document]
ardatan: We can put (HTTP only) maybe?

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
```diff
@@ -0,0 +1,576 @@
...
+have access to all components they need.
+
+```ts filename="gateway.config.ts"
+import { useGraphQLJit } from '@envelop/graphql-jit'
+import { defineConfig } from '@graphql-hive/gateway'
+import useNewRelic from '@graphql-mesh/plugin-newrelic'
+
+export const gatewayConfig = defineConfig({
+  plugins: ctx => [
+    useGraphQLJit(),
```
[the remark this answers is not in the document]
EmrysMyrddin: Do you have an idea of which plugin I could add ?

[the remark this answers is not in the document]
ardatan: SOFA?

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
```diff
@@ -0,0 +1,628 @@
...
+A plugin is an object that can define methods to hook into different phases or events of the gateway
+lifecycle. Each hook will receive a single parameter, the hook payload, containing data related to
+the hooked phase and functions allowing you to modify the behavior of this phase.
+
+There are 2 categories of hooks:
+
+- Gateway lifecycle hooks: they are related to the lifecycle or events of the server itself and is
+  not related to a specific request.
+- Request lifecycle hooks: they are tight to a specific phase of a specific request. Those hooks
+  always give access to the related request in there payload.
```
Urigo, line 80: I think a mermid chart of the hooks would be nice here

Urigo (changes requested):

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
ardatan: Like the one below in 184th line?

ardatan: It is possible to see with `View File`;
https://github.com/graphql-hive/console/blob/3ed148ed148ef8677be8e4d927e35fd1951be88b/packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
GitHub is able to show mermaid graphs.

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
```diff
@@ -0,0 +1,628 @@
...
+most of the feature are built around a plugin system.
+
+A plugin can hook into each important steps of the lifecycle of the gateway, including the graphql
+request execution.
+
+The gateway is built using Graphql Yoga (a GraphQL HTTP server) and Envelop (a GraphQL execution
+orchestrator), which are also highly modular technologies. Hive Gateway is extending Yoga and
+Envelop plugin system by providing gateway specific hooks.
+
+This means that if a plugin exists in multiple variants, you should follow this priority order:
```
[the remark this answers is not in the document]
EmrysMyrddin: It's because if they want to use a plugin, it can exists in multiple variants. The Prometheus plugin is a good example of this. You have to choose the correct variant to have access to the best of it.

In the other hand, we have plugins that exists only in the envelop variant, like Sentry. So we have to some way say that you can use Envelop or Yoga plugins too.

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
EmrysMyrddin: There is a mermaid chart for the Request lifecycle. I was not sure yet if a Gateway lifecycle graph would be helpfull, but I can make one if you think so

ardatan:
@EmrysMyrddin Let's make sure to be aligned with https://github.com/dotansimha/graphql-yoga/pull/3567#issue-2745119672

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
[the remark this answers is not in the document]
Urigo: Maybe this is also a technical details that we don't need them to be aware of?
why not create a gateway variant for any plugin users might need?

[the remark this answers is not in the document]
EmrysMyrddin: I do agree. IMO, we should just have one package for a plugin. Then, it should be up to the plugin to handle the fact that some hooks are optionally called, depending on where they are used.

Yet, it's not the case today, so what do we do ? Should we document this point until we refactor plugins ?

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
EmrysMyrddin: I've added a diagram for the gateway lifecycle, let me know what you think about it

packages/web/docs/src/pages/docs/gateway/other-features/plugins.mdx
```diff
@@ -0,0 +1,618 @@
...
+  listening: Listening for HTTP requests
+
+  [*] --> onYogaInit
+  onYogaInit --> onPluginInit
+  onPluginInit --> listening
+  listening --> onDispose: Gateway is shutting down
+
+  note left of listening: <a href="#onschemachange">onSchemaChange</a> called for each new schema loaded
+  note left of listening: <a href="#onfetch">onFetch</a> called for each outgoing HTTP request
+```
```
EmrysMyrddin, line 106: Not sure about how much this diagram is clear.

@Urigo what do you think ?

I'm not sure how to represent the fact that both `onFetch` and `onSchemaChange` can be called at any time during the lifetime of the gateway.

EmrysMyrddin:
Globally, i think we real lack an API reference, generated from TSDoc somewhere. In this page, It would have been nice to have links for the types of each payload fields, instead of requiring the user to manual go to source code.

I thought about adding a github permalink, but I'm afraid that it would be fragile and break if we refactor and move code around.

packages/web/docs/src/pages/docs/gateway/other-features/_meta.ts
```diff
@@ -4,5 +4,5 @@ export default {
   security: 'Security',
   'header-propagation': 'Header Propagation',
   testing: 'Testing & Debugging',
-  'custom-plugins': 'Custom Plugins',
+  plugins: 'Plugins',
```
dotansimha, line 7: fyi, this will cause 404. add redirect or keep the same URL

dotansimha (changes requested):

packages/web/docs/src/pages/docs/gateway/other-features/_meta.ts
ardatan: I'd prefer to have it as custom plugins because this page explains how to write a plugin and how the hooks work.

dotansimha: so let's revert it, it will also eliminate the 404

Urigo (approved):

packages/web/docs/src/pages/docs/gateway/other-features/_meta.ts
ardatan: Reverted

ardatan (approved):

PR #6129, opened 2024-12-16
Multi-Tier caching in the tokens service

kamilkisiela:
I will describe the Pull Request tomorrow, but in short we use Redis and in-memory as L1 and L2 cache.

When cache entry (in-memory) is stale, we fetch it from Redis, if it's missing, we fetch it from Postgres.
If Postgres is down, we return stale data.

Of course there are things to improve when PG is down, like the tokens service being scaled up and a new replica pops in, but Redis has partial or no data.

But at least it's no longer reliant on Postgres.


packages/services/tokens/src/multi-tier-storage.ts
```diff
@@ -0,0 +1,395 @@
...
+          cache.delete(hashedToken);
+        },
+      });
+    }),
+    invalidateTokens: tracker.wrap(async hashedTokens => {
+      if (hashedTokens.length === 0) {
+        return;
+      }
+
+      await redis.del(
```
dotansimha, line 265: Should we run these in parallel? or at least make sure to invalid cache even if Redis failed? 🤔 
I mean, if this function is called, we know that we need to clean the token for sure.

dotansimha (approved):

packages/services/tokens/src/multi-tier-storage.ts
kamilkisiela: Redis **needs** to be invalidated, because it stores data for much much longer, in-memory cache **can** as the token will expire in less than 5 minutes

packages/services/tokens/src/multi-tier-storage.ts
```diff
@@ -0,0 +1,395 @@
...
+    tokenAlias: item.token_alias,
+    name: item.name,
+    date: item.created_at as any,
+    lastUsedAt: item.last_used_at as any,
+    organization: item.organization_id,
+    project: item.project_id,
+    target: item.target_id,
+    scopes: item.scopes || [],
+  };
+}
```
n1ru4l, line 287: just a nit, but instead of transforming the tokens after reading from db, we should probably do it as part of the db logic already 🤔

kamilkisiela: done

PR #6142, opened 2024-12-19
Update docker compose PG version

TuvalSimha:
- [x] Update all Docker compose files to use PG16
- [x] Verify development workflow 
- [x] Write a `changeset` for `hive`
- [x] Added docs recommendation to use current version

### DIFF
--- a/.changeset/dirty-schools-call.md
+++ b/.changeset/dirty-schools-call.md
@@ -2,7 +2,8 @@
 'hive': major
 ---
 
-Docker Compose: Upgrade PostgreSQL version for Docker compose from version 14.13 to use 16.4. For
-self-hosters, we recommend upgrading PostgreSQL based on your Cloud provider's or IT's
-recommendation. The Hive data that was previously created with PostgreSQL v14 is compatible with
-v16.
+Docker Compose: Upgrade PostgreSQL version for Docker compose from version 14.13 to use 16.4.
+
+For self-hosters, we recommend upgrading PostgreSQL based on your Cloud provider's or IT's
+recommendation. The Hive data that was previously created with PostgreSQL v14 is compatible with v16.
+You can read about upgrading PostgreSQL in a Docker container more here - https://helgeklein.com/blog/upgrading-postgresql-in-docker-container/
--- a/.changeset/dirty-schools-call.md
+++ b/.changeset/dirty-schools-call.md
@@ -2,8 +2,11 @@
 'hive': major
 ---
 
-Docker Compose: Upgrade PostgreSQL version for Docker compose from version 14.13 to use 16.4.
+Docker Compose: Upgrade the PostgreSQL version for Docker Compose from version 14.13 to use 16.4. This change is published as major, as it requires attention based on your setup. 
 
-For self-hosters, we recommend upgrading PostgreSQL based on your Cloud provider's or IT's
-recommendation. The Hive data that was previously created with PostgreSQL v14 is compatible with v16.
-You can read about upgrading PostgreSQL in a Docker container more here - https://helgeklein.com/blog/upgrading-postgresql-in-docker-container/
+For self-hosters with a managed database, we recommend upgrading PostgreSQL based on your Cloud provider's or IT's
+recommendation. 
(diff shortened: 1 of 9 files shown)


n1ru4l (approved):

n1ru4l:
Is an existing database backwards compatible or are additional steps necessary?

TuvalSimha:
> Is an existing database from v16.4 backwards compatible or are additional steps necessary?

@n1ru4l 

If your DB was created with PostgreSQL version 14 or earlier, it is not compatible with version 16.4, and you will see the error:
```
DETAIL:  The data directory was initialized by PostgreSQL version 14, which is not compatible with this version 16.4.
```

to resolver this need to run migration all over again with the new version.

packages/web/docs/src/pages/docs/self-hosting/get-started.mdx
```diff
@@ -75,6 +75,12 @@ The following diagram shows the architecture of a self-hosted GraphQL Hive insta
 
 - **PostgreSQL database**: stores all core data, including: users, GraphQL schemas, schema checks,
   projects, targets, organizations and configurations.
+
+<Callout type="info">
+  Recommendation: We currently use PostgreSQL version **16.14** and recommend using this version as
```
dotansimha, line 80: should be `16.4` and not `16.14`

dotansimha: can be improved to:
```
We recommend using PostgreSQL version 16. While older versions (14, 15) may still work, we recommend using 16 as this is the version we using for the cloud version of Hive.
```

.changeset/flat-glasses-complain.md
```diff
@@ -0,0 +1,6 @@
+---
+'hive': minor
+---
```
n1ru4l, line 3: This is a major change, since it will require the user to manually upgrade the database.

.changeset/flat-glasses-complain.md
```diff
@@ -0,0 +1,6 @@
...
+
+Docker Compose: Upgrade PostgreSQL version for Docker compose from version 14.13 to use 16.4
```
dotansimha, line 5: This could be improved. 
Maybe we can add something like:

```
For self-hosters, we recommend upgrading PostgreSQL based on your Cloud provider's or IT's recommendation. The Hive data that was previously created with PostgreSQL v14 is compatible with v16. 
```

packages/web/docs/src/pages/docs/self-hosting/get-started.mdx
TuvalSimha: done

.changeset/flat-glasses-complain.md
dotansimha: Yes, they will need to do it based on their setup (ideally, they shouldn't use the community file... and use a PG instance in Cloud service, or one that was provided by their IT)

n1ru4l: Yeah but the changeset refers to the docker-compose file, so it kinda is a breaking change in terms of that.
Maybe we can also provide simple instructions for upgrading the local existing database (also relevant for our existing dev environments locally)

.changeset/dirty-schools-call.md
```diff
@@ -0,0 +1,8 @@
+---
+'hive': major
+---
+
+Docker Compose: Upgrade PostgreSQL version for Docker compose from version 14.13 to use 16.4. For
+self-hosters, we recommend upgrading PostgreSQL based on your Cloud provider's or IT's
+recommendation. The Hive data that was previously created with PostgreSQL v14 is compatible with
+v16.
```
n1ru4l, line 8: >The Hive data that was previously created with PostgreSQL v14 is compatible with
v16.

If that is the case, it would be a minor upgrade.

.changeset/flat-glasses-complain.md
dotansimha: rephrased, thanks

dotansimha (approved):

PR #6143, opened 2024-12-19
Prepare Vite and Fastify for more than one entry point

kamilkisiela:
Decorates fastify request object with `viteHtmlFile` to determine the html file.
With those changes, the preflight-worker-embed and the app itself both have hot reloading working, logging etc.

I tried to keep it as simple as possible.

### DIFF
--- a/packages/web/app/src/server/index.ts
+++ b/packages/web/app/src/server/index.ts
@@ -22,6 +22,11 @@ const server = Fastify({
   },
 });
 
+const preflightWorkerEmbed = {
+  path: '/__preflight-embed',
+  htmlFile: 'preflight-worker-embed.html',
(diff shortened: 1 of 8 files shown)


packages/web/app/src/server/index.ts
```diff
@@ -85,6 +99,18 @@ async function main() {
   connectGithub(server);
   connectLab(server);
 
+  server.get(preflightWorkerEmbed.path, (_req, reply) => {
+    if (isDev) {
+      // If in development mode, return the Vite index.html.
```
kamilkisiela, line 104:
```suggestion
      // If in development mode, return the Vite preflight-worker-embed.html.
```

PR #6161, opened 2024-12-24
Improve documentation of `@cacheControl`

ardatan:
Related https://github.com/graphql-hive/gateway/pull/351


Urigo (changes requested):
Thank you @ardatan but this is not understandable enough in my opinion.
I would suggest to start with an example of the goal and what is the purpose.
Then an example of each way you describe and a flow chart of what's happening on each option.

ardatan:
@Urigo I added charts and more details. Let me know what do you think?

Urigo:
Thank you @ardatan 
so if I understand correctly, any time the @cachecontrol directive is configured on a subgraph, the gateway will cache according to the directive except for when the subgraph is Yoga?

ardatan:
@Urigo for the third usecase, the subgraph server doesn't matter. It can be Yoga, Apollo anything that exposes the SDL with composeDirective as I showed there.
The other two shows how directive can be used on the specific servers regardless the third usecase

Urigo:
why does Yoga acts differently then apollo server?

ardatan:
Yoga's response caching plugin is different than Apollo's response caching plugin. I mentioned the behavior of both in the PR even if they are explained in their own documentation.

Urigo (approved):
great documentation, thank you

PR #6224, opened 2024-12-27
Extend debug information of the CLI

kamilkisiela:
Prints a list of files of the script directory and a file path to the included node binary.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
       "got@14.4.5": "patches/got@14.4.5.patch",
       "slonik@30.4.4": "patches/slonik@30.4.4.patch",
       "@oclif/core@3.26.6": "patches/@oclif__core@3.26.6.patch",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -118,7 +118,7 @@
       "got@14.4.5": "patches/got@14.4.5.patch",
       "slonik@30.4.4": "patches/slonik@30.4.4.patch",
       "@oclif/core@3.26.6": "patches/@oclif__core@3.26.6.patch",
-      "oclif@4.13.6": "patches/oclif@4.13.6.patch",
+      "oclif": "patches/oclif.patch",
```
dotansimha, line 121: ah cool, I didn't know we could point to a patch without a version. it can solve some of the failures in ci with Renovate

dotansimha (approved):

PR #6227, opened 2024-12-27
docs: finalize Custom Plugins documentation

EmrysMyrddin:
This is a follow up of the #6128 PR.

It contains:
 - [x] A generalization of the links style fix of mermaid diagrams (to avoid having a dirty `div` in the middle of the MDX files)
 - [x] Docs for all missing `*Done` hooks.
 - [ ] Fix of some wording in the description of payload fields

> [!NOTE]
>  I've modified the Tailwind configuration to give access to the primary color of the nextra theme. Before, it was entirely overridden by the `hive-yellow` color. 
> 
> I've named it `nextra-primary` instead of changing the `primary` color value to avoid breaking change, and to maintain coherence with Figma.

### DIFF
--- a/packages/web/docs/src/mermaid.css
+++ b/packages/web/docs/src/mermaid.css
@@ -3,7 +3,7 @@
     @apply text-nextra-primary-600;
     text-decoration-line: underline;
     text-decoration-thickness: from-font;
-    text-underline-position:from-font;
+    text-underline-position: from-font;
 
     &:hover {
       text-decoration-line: none;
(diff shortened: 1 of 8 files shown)


EmrysMyrddin:
This time it should be done !

I was not very inspired for the description of the hooks or the use cases. If you have ideas, don't hesitate to make suggestions !

ardatan (approved):

n1ru4l (approved):

Urigo (approved):

PR #6230, opened 2024-12-30
try to workaround `got` issue with `unhandledRejection`

dotansimha:
Closes https://github.com/graphql-hive/console/issues/6220


patches/p-cancelable@4.0.1.patch
n1ru4l, line 1: Let's keep the original formatting of the file, so the diff only shows the lines we adjusted

kamilkisiela: Otherwise it’s impossible to update the patch when a new release is out

dotansimha: Yeah sure, sorry about that, VSCode messed it up. Fixed :)

n1ru4l (approved):

patches/p-cancelable@4.0.1.patch
```diff
@@ -0,0 +1,23 @@
...
+-					throw new Error(`The \`onCancel\` handler was attached after the promise ${this.#state.description}.`);
++					console.error(
++						`The \`onCancel\` handler was attached after the promise ${this.#state.description}.`,
++					);
++
++					reject(
++						new Error(
++							`The \`onCancel\` handler was attached after the promise ${this.#state.description}.`,
++						),
++					);
```
n1ru4l, line 18: I don't think it makes sense to reject again if the Promise is already rejected? It cannot reject twice, right?

dotansimha: Good point. So, this will lead to nothing happening, right? 🤔 
Should I just log it? This edge case happens when a request is resolved after several failures, meaning that the relevant `Promise` is already resolved.

dotansimha: fixed

Issue #6050, opened 2024-11-28
chore: introduce eslint rule `no-unnecessary-condition`


n1ru4l:
[This rule can improve the code climate and remove a lot of unnecessary checks.](https://typescript-eslint.io/rules/no-unnecessary-condition/)

Example: https://typescript-eslint.io/play/#ts=5.7.2&fileType=.tsx&code=FAYw9gdgzgLgBAMzGOBeOBvOBDAXJuAI3wHY4BfC4USWOCAVwFs1FkB%2BAOmy8OvGjwANmCGtGQodQCWCOAAoRY1CrgwATgwCmASkzVy1IA&eslintrc=N4KABGBEBOCuA2BTAzpAXGUEKQAIBcBPABxQGNoBLY-AWhXkoDt8B6Jge1tiacTJTIAhtEK0yHJgBNK%2BSpPRRE0aB2iRwYAL4gtQA&tsconfig=N4KABGBEDGD2C2AHAlgGwKYCcDyiAuysAdgM6QBcYoEEkJemy0eAcgK6qoDCAFutAGsylBm3TgwAXxCSgA&tokens=false

Introducing it would require revisiting/adjusting/eslint-ignoring quite some statements.

```
✖ 513 problems (513 errors, 0 warnings)
  295 errors and 0 warnings potentially fixable with the `--fix` option.
```

jdolle:
Recommend closing because this plugin has a number of false positives that could cause us to incorrectly remove optionals i.e. from array indexes. We can reevaluate adding at a later point.

PR #6238, opened 2025-01-01
chore: fix linting errors

jdolle:
### Background

There currently are a number of linting errors that can be resolved via `lint --fix`. This creates noise for future linting improvements like https://github.com/graphql-hive/console/issues/6050 

### Description

This fixes the linting issues a minor react key uniqueness issue found during testing.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/web/app/src/components/organization/Permissions.tsx
```diff
@@ -150,19 +150,17 @@ export const PermissionScopeItem = <
                   : false;
 
               return (
-                <>
```
jdolle, line 153: During local testing I found that this extra wrapper caused a warning because the list of permission scope items needs the root element to have a unique key, but the key attribute is set on the SelectItem.
Removing this wrapper lets the key be on the root element and resolves this.

n1ru4l:
Can we within this PR enable the rule for new files and ignore the rule for already existing files that were not fixable?

jdolle:
I could add `lint-staged` but that won't work for running linting as a PR action. It also creates some friction for future changes to these files.
I was attempting to break it apart to keep the PRs small, but I can work through resolving all errors and adding the rule in a single PR.

jdolle:
Doing more research, it looks like this specific config also has a number of false positives that could be an issue https://github.com/typescript-eslint/typescript-eslint/issues?q=is%3Aissue+is%3Aopen+no-unnecessary-condition

One that surfaced for me was https://github.com/typescript-eslint/typescript-eslint/pull/1534
This specific instance looks fixed in a newer version of eslint, but I could foresee https://github.com/typescript-eslint/typescript-eslint/issues/1798 being an issue.

jdolle:
I've reviewed these changes that the linter applied and adjusted the ones for the array indexing false positives.

My preference for now would be to not enable this linter config given the false positives it can have, but to apply these changes in this PR to prepare for a future when this config could be added.
Another option may be to add the config as a `warn`, but I think that could get noisy and could still encourage people to make changes based on false positives.

jdolle:
Closing this due to concerns around the nullability being wrong and the linter plugin having false positives.
Creating a separate PR for the permissions page change.

PR #6242, opened 2025-01-02
override `ip` package with `neoip`

dimaMachina:
https://guild-oss.slack.com/archives/C01E8ATBQGN/p1735797170636169

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
       "ws@^8.0.0": "^8.18.0",
       "ws@*": ">=8.18.0 || >=7.5.10 || >=6.2.3 || >=5.2.4",
       "cookie@<0.7.0": "0.7.2",
-      "ip": "https://registry.npmjs.org/neoip/-/neoip-2.1.0.tgz"
+      "ip": "npm:neoip@2.1.0"
     },
     "patchedDependencies": {
       "mjml-core@4.14.0": "patches/mjml-core@4.14.0.patch",
--- a/package.json
+++ b/package.json
@@ -101,13 +101,15 @@
   "pnpm": {
     "overrides.csstype": "To fix type check error https://github.com/JedWatson/react-select/issues/5952",
     "overrides.mjml": "Our patch removes html-minifier from mjml-core, we dont need this dependency and its causing false-alarm (CVE https://nvd.nist.gov/vuln/detail/cve-2022-37620)",
+    "overrides.ip": "There is no update with fix for ip package, we use fork https://github.com/indutny/node-ip/issues/150#issuecomment-2325961380",
     "overrides": {
       "csstype": "3.1.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -101,13 +101,15 @@
...
+    "overrides.ip": "There is no update with fix for ip package, we use fork https://github.com/indutny/node-ip/issues/150#issuecomment-2325961380",
     "overrides": {
       "csstype": "3.1.2",
       "mjml-core@4.14.0>html-minifier": "-",
       "mjml-cli@4.14.0>html-minifier": "-",
       "ws@^8.0.0": "^8.18.0",
       "ws@*": ">=8.18.0 || >=7.5.10 || >=6.2.3 || >=5.2.4",
-      "cookie@<0.7.0": "0.7.2"
+      "cookie@<0.7.0": "0.7.2",
+      "ip": "https://registry.npmjs.org/neoip/-/neoip-2.1.0.tgz"
```
n1ru4l, line 112: What is the benefit of pointing to the `tgz` archive instead of letting npm resolve it through the version string `2.1.0`?

dimaMachina, line 112: just copied from github comment, you want like this?
```suggestion
      "ip": "npm:neoip@2.1.0"
```

dotansimha: I think this way it's better (to point to a version and package, instead of a URL)

dimaMachina: changed, this syntax should work according to their docs https://pnpm.io/package_json#pnpmoverrides
image

dotansimha (approved):

package.json
```diff
@@ -101,13 +101,15 @@
...
+    "overrides.ip": "There is no update with fix for ip package, we use fork https://github.com/indutny/node-ip/issues/150#issuecomment-2325961380",
     "overrides": {
       "csstype": "3.1.2",
       "mjml-core@4.14.0>html-minifier": "-",
       "mjml-cli@4.14.0>html-minifier": "-",
       "ws@^8.0.0": "^8.18.0",
       "ws@*": ">=8.18.0 || >=7.5.10 || >=6.2.3 || >=5.2.4",
-      "cookie@<0.7.0": "0.7.2"
+      "cookie@<0.7.0": "0.7.2",
+      "ip": "npm:neoip@2.1.0"
```
dotansimha, line 112: Do we need to limit this to a version, or to a specific path in the dependencies tree?

dimaMachina: I think we can limit it with `ip@<=2.0.1` (last published version), wdyt?

n1ru4l (approved):
We do not really use this package, so safe to merge.

```
devDependencies:
@apollo/gateway 2.9.3
└─┬ make-fetch-happen 11.1.1
  └─┬ socks-proxy-agent 7.0.0
    └─┬ socks 2.7.1
      └── ip 2.0.1
```

Issue #207, opened 2022-06-29
Error codes + dedicated pages


kamilkisiela:
Try to explain the most popular issues with this structure:
- Symptom
- Diagnosis
- Solution

Define error codes (like in TS) for errors.

kamilkisiela:
> When the token is missing, explain how to create and apply a token.
> _#185_

kamilkisiela:
#6252 introduced part of it, it did not "fix/close" the issue.

jdolle:
I propose that the error returned from our CLI take the form:

```
Error message.  [CODE]
```

E.g.

```
The command, "badcommand", does not exist.  [1001]
```

The error codes are TBD.

And we may consider adding a message that can be disabled that tells people where to find more
information. E.g.

```
The command, "badcommand", does not exist.  [1001]
> See https://the-guild.dev/graphql/hive/docs/api-reference/cli#errors for a complete list of error codes and recommended fixes.
To disable this message set HIVE_NO_ERROR_TIP=1
```

---

The format this document takes is:

## Error Name

### `example hive command`

```
Error message returned from CLI.
```

**solution:**

```
Recommendation provided by cli error code docs.
```

---

## Invalid Config

### `hive schema:fetch`

```
The provided "hive.json" is invalid.
```

**solution:**

```
A configuration file was found but the format does not match what is expected. See https://github.com/graphql-hive/console/blob/main/packages/libraries/cli/README.md#config-file-hivejson for structure details and try updating to the latest version if contents appear valid.
```

## Invalid Command

### `hive badcommand`

```
The command, "badcommand", does not exist.
```

**solution:**

```
Use "hive help" for a list of available commands.
```

## Missing Argument(s)

### `hive schema:delete`

```
schema:delete requires argument(s):
   SERVICE  name of the service
```

**solution:**

```
Use "hive help [command]" for usage details.
```

## Access Denied (Missing Registry Token)

### `HIVE_TOKEN='' hive schema:fetch`

```
A registry token is required to execute "hive schema:fetch".
```

**solution:**

```
A registry token can be set using the environment variable "HIVE_TOKEN", argument "--registry.accessToken", or config file "hive.json". For help generating a token, see https://the-guild.dev/graphql/hive/docs/management/targets#registry-access-tokens
```

## Access Denied (Missing CDN Key)

### `hive artifact:fetch --artifact sdl`

```
A CDN key is required to execute "hive artifact:fetch".
```

**solution:**

```
[...]

PR #6252, opened 2025-01-02
Add detailed error message for Hive CLI missing access token

jdolle:
### Background

Fixes https://github.com/graphql-hive/console/issues/207

> When the token is missing, explain how to create and apply a token.


saihaj (approved):
LGTM. Needs a changeset though

packages/libraries/cli/src/helpers/errors.ts
```diff
@@ -0,0 +1 @@
+export const ACCESS_TOKEN_MISSING = `--registry.accessToken is required in order to perform this command on the schema registry. To create an registry access token with the necessary permissions, use the Hive console. Select the target under your Hive project and then choose the Settings tab. On the Settings screen, you can manage your target's settings and access tokens. Under the Registry Access Tokens section, click the Create new registry token button. Give your access token a name and select Read & Write for the permissions.`;
```
dotansimha, line 1: I think we can add a link to the docs? Maybe to this page: https://the-guild.dev/graphql/hive/docs/management/targets#registry-access-tokens ?

dotansimha: I mean, instead of having a long CLI message, we can have a short, concise message, with a link to the docs.

jdolle: I've adjusted the message to reflect this.

dotansimha (approved):

Issue #3018, opened 2023-10-09
Static total request for Conditional Breaking Change setting


Samjin:
Hi team, it'd be nice to have an alternative static threshold for breaking changes, such as total requests in a period of time, Considering how diverse the services and total operations per service can be, it could be beneficial to introduce a static threshold. This would help to mitigate a potential problem with percentage where it's too high for service with a high traffic, too low for one with low traffic. With static total request, service can do graceful deprecation to ensure there are no breaking change at client side. 
Thanks

kamilkisiela:
Makes sense, we will try to prioritize it soon.

jdolle:
I'll take this.

PR #6254, opened 2025-01-03
Add breaking changes total operations option to target settings page

jdolle:
### Background

https://github.com/graphql-hive/console/issues/3018

### Description

Screenshot 2025-01-02 at 4 07 25 PM

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/web/app/src/pages/target-settings.tsx
+++ b/packages/web/app/src/pages/target-settings.tsx
@@ -34,6 +34,7 @@ import {
   SubPageLayoutHeader,
 } from '@/components/ui/page-content-layout';
 import { QueryError } from '@/components/ui/query-error';
+import { RadioGroup, RadioGroupItem } from '@/components/ui/radio-group';
 import { Spinner } from '@/components/ui/spinner';
 import { TimeAgo } from '@/components/ui/time-ago';
 import { useToast } from '@/components/ui/use-toast';
@@ -50,6 +51,7 @@ import { subDays } from '@/lib/date-time';
 import { useToggle } from '@/lib/hooks';
 import { cn } from '@/lib/utils';
 import { zodResolver } from '@hookform/resolvers/zod';
+import { RadioGroupIndicator } from '@radix-ui/react-radio-group';
 import { Link, useRouter } from '@tanstack/react-router';
 
 const RegistryAccessTokens_MeFragment = graphql(`
@@ -511,12 +513,23 @@ const ConditionalBreakingChanges = (props: {
     enableReinitialize: true,
     initialValues: {
       percentage: settings?.percentage || 0,
+      count: settings?.count || 1,
       period: settings?.period || 0,
+      unit: settings?.unit ?? 'percentage',
       targetIds: settings?.targets.map(t => t.id) || [],
       excludedClients: settings?.excludedClients ?? [],
     },
     validationSchema: Yup.object().shape({
-      percentage: Yup.number().min(0).max(100).required(),
+      percentage: Yup.number().when('unit', {
+        is: 'percentage',
+        then: schema => schema.min(0).max(100).required(),
+        otherwise: schema => schema.nullable(),
+      }),
+      count: Yup.number().when('unit', {
+        is: 'count',
+        then: schema => schema.min(1).required(),
+        otherwise: schema => schema.nullable(),
+      }),
       period: Yup.number()
         .min(1)
         .max(targetSettings.data?.organization?.organization?.rateLimit.retentionInDays ?? 30)
@@ -530,6 +543,7 @@ const ConditionalBreakingChanges = (props: {
           return Number(num.toFixed(2)) === num;
         })
         .required(),
+      unit: Yup.string().oneOf(['percent', 'count']),
       targetIds: Yup.array().of(Yup.string()).min(1),
       excludedClients: Yup.array().of(Yup.string()),
     }),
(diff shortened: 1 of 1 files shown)


jdolle:
There are a lot of ways we could potentially take this request. I.e. we could say if the total traffic is 1% OR greater than 10 total operations.
I've opted to keep it simple for now and have only one option available at a time.

And for the schema, is there a preference for how to represent the percent or count? Again there’s a bunch of possibilities, but I think I like:
```
type TargetValidationSettings {
  percentage: Float! @deprecated(reason: "Use 'TargetValidationSettings.breakingAmount'. This can represent a count or percent")
  breakingAmount: String! # A string representation of either a floating point percentage or an integer count.
  breakingAmountUnit: BreakingChangeAmountUnit!
}

enum BreakingChangeAmountUnit {
  COUNT
  PERCENT
}
```
I appreciate any feedback.

dotansimha:
> There are a lot of ways we could potentially take this request. I.e. we could say if the total traffic is 1% OR greater than 10 total operations. I've opted to keep it simple for now and have only one option available at a time.


I agree with this take at this stage. It can be "and" / "or" and it will become much more complicated. At the moment, it's ok to have only one selected. 

> And for the schema, is there a preference for how to represent the percent or count? Again there’s a bunch of possibilities, but I think I like:
> 
> ```
> type TargetValidationSettings {
>   percentage: Float! @deprecated(reason: "Use 'TargetValidationSettings.breakingAmount'. This can represent a count or percent")
>   breakingAmount: String! # A string representation of either a floating point percentage or an integer count.
>   breakingAmountUnit: BreakingChangeAmountUnit!
> }
> 
> enum BreakingChangeAmountUnit {
>   COUNT
>   PERCENT
> }
> ```
> 
> I appreciate any feedback.

Regarding the schema structure, I think you can maybe introduce a new common type the can be used? something like:

```graphql
type TargetValidationSettings {
  percentage: Float! @deprecated(reason: "Use 'TargetValidationSettings.breakingAmount'. This can represent a count or percent")
  breakingSettings: UnitAndAmount!
  breakingAmount: String! # A string representation of either a floating point percentage or an integer count.
  breakingAmountUnit: BreakingChangeAmountUnit!
}

type UnitAndAmount {
   unit: AmountUnit!
   amount: Float! # or should this be "String" to make it easier? I mean, `Float` can represent percentage (0-1) AND numeric value
}

enum AmountUnit {
  COUNT
  PERCENT
}
```

No strong opinions here, I'm also fine with addressing it specifically here and not deal with adding something as common.

n1ru4l:
Do we want to store both values (percentage and count) in the database and display them to the user? Or do we only want to store one number?

For the former case, we would need to add both values to the schema as separate fields and a property to indicate which mode is active.

```diff

enum BreakingChangeAmountUnit {
  percentage
  count
}

  type TargetValidationSettings {
    enabled: Boolean!
    period: Int!
    percentage: Float!
    requestCount: Float!
    breakingChangeAmountUnit: BreakingChangeAmountUnit!
    targets: [Target!]!
    excludedClients: [String!]!
  }

```

In the latter case what @dotansimha proposed makes sense.

jdolle:
Thank you both for the input. In the interest of keeping this option simple, I'm in favor of @n1ru4l 's idea -- having discrete values and saving both. That maintains our current functionality without having to migrate and avoids ambiguous types. And any attempt to make this feature future proof is probably in vain since we aren't sure what additional behavior will be added later, if any.

packages/libraries/apollo/src/version.ts
```diff
@@ -1 +1 @@
-export const version = '0.36.2';
+export const version = '0.36.3';
```
jdolle, line 1: not sure why these auto bumped...

packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -110,6 +110,8 @@ export default gql`
     targetSlug: String!
     period: Int!
     percentage: Float!
+    requestCount: Int! = 1
+    breakingChangeFormula: BreakingChangeFormula! = PERCENTAGE
```
jdolle, line 114: to support migrating, I added a default.

jdolle, line 114: To support migrating the old clients.

packages/services/storage/src/index.ts
```diff
@@ -423,6 +423,8 @@ export async function createStorage(
       | 'validation_percentage'
       | 'validation_period'
       | 'validation_excluded_clients'
+      | 'validation_request_count'
```
jdolle, line 426: I still need to generate the `targets` type using `schemats`. Working on the migration now.

packages/web/app/src/pages/target-settings.tsx
```diff
@@ -540,6 +559,8 @@ const ConditionalBreakingChanges = (props: {
           projectSlug: props.projectSlug,
           targetSlug: props.targetSlug,
           ...values,
+          requestCount: values.requestCount ?? 1,
+          percentage: values.percentage ?? 0,
```
jdolle, line 563: I opted to default these here instead of in `Yup` so that users dont see these values appear when clicking submit.

Alternatively, I think it's an okay experience to set the default value onBlur like I did for the [requestCount](https://github.com/graphql-hive/console/pull/6254/files#diff-40476c14d0063812ffd974b7048fe7d14f916776fcca4a67fc371821bc63c7f5R675)

packages/libraries/apollo/src/version.ts
dotansimha: yeah it happens when you `build`, I think it's ok to ignore or commit at this point. 
@n1ru4l @kamilkisiela  should we ignore this file?

packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -178,11 +181,35 @@ export default gql`
   type TargetValidationSettings {
     enabled: Boolean!
     period: Int!
+
+    """
+    If TargetValidationSettings.breakingChangeFormula is PERCENTAGE, then this
+    is the percent of the total operations over the TargetValidationSettings.period
+    required for a change to be considered breaking.
+    """
     percentage: Float!
```
dotansimha, line 190: Should this be `Float` now? 🤔 (without `!`) 
I mean, in case it's being queried and the user is using `REQUEST_COUNT`, this should be empty? 

same for `requestCount: Int!`?

packages/migrations/src/actions/2025.01.07T00.00.00.breaking-changes-request-count.ts
```diff
@@ -0,0 +1,19 @@
...
+export default {
+  name: '2025.01.07T00.00.00.breaking-changes-request-count.ts',
+  run: ({ sql }) => sql`
+CREATE TYPE
+  breaking_change_formula AS ENUM('PERCENTAGE', 'REQUEST_COUNT');
+
+ALTER TABLE
+  targets
+ADD COLUMN
+  validation_request_count INT NOT NULL DEFAULT 1;
```
n1ru4l, line 12: I just wanted to comment that we should not add a default value but, but I double checked 😄 
> From PostgreSQL 11, adding a column with a constant default value no longer means that each row of the table needs to be updated when the ALTER TABLE statement is executed. Instead, the default value will be returned the next time the row is accessed, and applied when the table is rewritten, making the ALTER TABLE very fast even on large tables.

packages/services/api/src/modules/target/module.graphql.ts
n1ru4l: We have a default value and both values are stored in the database, so having them non-nullable is fine.

jdolle: I wanted to avoid making a breaking change to the schema

n1ru4l (changes requested):
For this one we will need an integration test as it is very critical core feature

jdolle:
> For this one we will need an integration test as it is very critical core feature

I've added an integration test, but noticed that the total percentage and counts are considered per-operation.
E.g. if `Query.a` is used by two separate operations, and each operation makes 1 call, then this would be considered breaking with a request count of 1, but not a request count of 2, because each operation was only made once even though the total number of requests for the field was 2.

The same logic applies to the percent based approach. Is that how it was intended to work?
If this is a bug, I think the work is outside the scope of this issue and we should adjust this logic in a separate PR.

n1ru4l (approved):

n1ru4l:
> > For this one we will need an integration test as it is very critical core feature
> 
> I've added an integration test, but noticed that the total percentage and counts are considered per-operation. E.g. if `Query.a` is used by two separate operations, and each operation makes 1 call, then this would be considered breaking with a request count of 1, but not a request count of 2, because each operation was only made once even though the total number of requests for the field was 2.
> 
> The same logic applies to the percent based approach. Is that how it was intended to work? If this is a bug, I think the work is outside the scope of this issue and we should adjust this logic in a separate PR.

That is a good catch, let's fix it in a follow-up pull request.

Issue #3855, opened 2024-01-22
hive cli command for listing services (subgraphs)


n1ru4l:
A simple command that will list the services of a project.

List with:
- name
- url
- link to service on Hive app

_________

Inspiration: https://www.apollographql.com/docs/rover/commands/subgraphs/#listing-subgraphs-in-a-supergraph

jdolle:
I can take this.

jdolle:
I evaluated fetching from the CDN and it was missing some of the data we wanted and would have required more work. I opted to allow schema:fetch to target the latest schema if no actionId is passed in. This should satisfy the requirement and I think is the most intuitive to use.

PR #6255, opened 2025-01-03
Add subgraphs type to cli schema:fetch

jdolle:
Fixes https://github.com/graphql-hive/console/issues/3855

### DIFF
--- a/.changeset/big-experts-melt.md
+++ b/.changeset/big-experts-melt.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/cli': minor
+---
+
+Added subgraph type to schema:fetch cmd to print subgraph details
--- a/packages/libraries/cli/src/helpers/print-table.ts
+++ b/packages/libraries/cli/src/helpers/print-table.ts
@@ -1,0 +1,29 @@
+type TableCell = string | number | Date;
+type Table = TableCell[][];
+
+function printCell(cell: TableCell): string {
+  if (cell instanceof Date) {
+    return cell.toISOString();
+  }
+  return String(cell);
+}
+
+function mapTable<T, O>(table: T[][], map: (cell: T, row: number, col: number) => O): O[][] {
+  return table.map((row, r) => row.map((cell, c) => map(cell, r, c)));
+}
+
+export function printTable(table: Table) {
+  const printedTable = mapTable(table, printCell);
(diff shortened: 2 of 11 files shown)


n1ru4l (changes requested):
lets add a simple integration snapshot test?

packages/libraries/cli/src/helpers/print-table.ts
```diff
@@ -0,0 +1,29 @@
...
+    return cell.toISOString();
+  }
+  return String(cell);
+}
+
+function mapTable<T, O>(table: T[][], map: (cell: T, row: number, col: number) => O): O[][] {
+  return table.map((row, r) => row.map((cell, c) => map(cell, r, c)));
+}
+
+export function printTable(table: Table) {
```
jdolle, line 15: If there's a package we've used and prefer for printing tables, then I'm happy to replcae this. The few I found had fairly large dependency trees and I wanted to avoid that.

jdolle:
> lets add a simple integration snapshot test?

Added. Given the length of the output and how many types are ran in the `each`, I decided to use regexp to match the format instead of doing an exact string match on specific cases.
Let me know if you had a different form of test in mind.

.changeset/big-experts-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-hive/cli': minor
+---
+
+Added subgraph type to schema:fetch cmd to print subgraph details
```
n1ru4l, line 5:
```suggestion
Added subgraph type to `schema:fetch` command to print subgraph details
```

Let's also add a sample call + output here.

packages/libraries/cli/src/helpers/print-table.ts
n1ru4l: We can consider it at some point, for now I think its fine to have it here

n1ru4l (approved):
Approved, the changeset could be a bit more detailed and have an example command usage + output.

dotansimha (approved):

saihaj (approved):

Urigo:
awesome @jdolle 
do you think it would be worth while to do a short 1-2 minute screencast of you demonstrating the new command?
that we can post somewhere on social media or maybe even on the docs?

PR #6276, opened 2025-01-08
chore(prettier): ignore snapshot files

jasonkuhrt:
Progresses toward: https://github.com/graphql-hive/console/pull/6122

Vitest supports snapshot files like "foo.json". Regardless of the file type, prettier should always ignore snapshot files.


jdolle (approved):
Makes sense. Best case, formatting snapshots is a waste of cpu since they should already be formatted.

dotansimha (approved):

PR #6279, opened 2025-01-08
refactor(cli): env helpers

jasonkuhrt:
Progresses toward #6122 

Simplify `BaseCommand` class by removing a function-like method.

### DIFF
--- a/packages/libraries/cli/src/helpers/env.ts
+++ b/packages/libraries/cli/src/helpers/env.ts
@@ -1,0 +1,17 @@
+/**
+ * Look up a value in the environment if the given value is undefined or null.
+ */
+export const orEnvar = <$Value>(
+  value: $Value,
+  envarName: string,
+): Exclude<$Value, null | undefined> | string | null => {
+  if (value !== null && value !== undefined) return value as any;
+
+  // eslint-disable-next-line no-process-env
+  if (envarName && process.env[envarName]) {
+    // eslint-disable-next-line no-process-env
+    return process.env[envarName];
+  }
+
+  return null;
+};
--- a/packages/libraries/cli/src/base-command.ts
+++ b/packages/libraries/cli/src/base-command.ts
@@ -77,28 +77,6 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
       .replace(findDoubleQuotes, (_: string, value: string) => colors.bold(value));
   }
 
-  maybe<TArgs extends Record<string, any>, TKey extends keyof TArgs>({
-    key,
-    env,
-    args,
-  }: {
-    key: TKey;
-    env: string;
-    args: TArgs;
-  }) {
-    if (args[key] != null) {
-      return args[key];
-    }
-
-    // eslint-disable-next-line no-process-env
-    if (env && process.env[env]) {
-      // eslint-disable-next-line no-process-env
-      return process.env[env];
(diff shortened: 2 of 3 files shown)


packages/libraries/cli/src/helpers/env.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Look up a value in the environment if the given value is undefined or null.
+ */
+export const orEnvar = <$Value>(
+  value: $Value,
+  envarName: string,
+): Exclude<$Value, null | undefined> | string | null => {
+  if (value !== null && value !== undefined) return value as any;
```
jdolle, line 8: minor: To avoid using `any` you could have a third argument of a function to deserialize the env var into the `$Value` type.
```
<T>(value: T, envName: string, parser: (envValue: string) => T = (v => v)) => ...
```

Alternatively, it may be worth using something like yup or whatever else package we already use to validate the env vars and/or command arguments so that we have an already parsed and valid env to reference

jasonkuhrt: The main goal of this PR was to stop class method use.

There are definitely more robust approaches to configuration management. The types of this function at least make the developer aware that a string is always part of the return due to the env lookup, and whatever value they give also adds to the type set. E.g. if their value is `null | Date` they'll end with `null | Date | string` which hopefully forces them to think, or simply make this abstraction better per your comment :)

There are CLI tools (e.g. `@molt/command`, probably `@effect/cli`) that support automatic environment lookup for arguments too, which would be better then the current ad-hoc approach I imagine.

jasonkuhrt: Some people on the team view any work on the CLI that isn't Rust based as a waste of time because we want to get to a slim binary for our CLI which isn't easily possible with JS today. Let's see where those conversations go because a lot depends on that I guess. No one on the team likes Oclif for example, but removing it would require addressing the aforementioned point.

kamilkisiela:
I'm fine with this change, but I'm just not sure what's the value here. We move from a class method to a function and yet this logic runs always in a class (oclif command).

packages/libraries/cli/src/helpers/env.ts
n1ru4l: > every method has the potential to use this, etc., degrades the encapsulation boundaries/scope/purpose etc. of the class to have methods that do not relate to that

Every method could invoke `orEnvar` as well.

> if their value is null | Date they'll end with null | Date | string

I just wanted to let you know that this hypothetical scenario does not need to be addressed right now.

n1ru4l (approved):
I am approving this. 

But I don't really think that it was necessary as a prerequisite to solve https://github.com/graphql-hive/console/issues/3592. In the future I would rather prefer to first solve the actual issue without refactors, to immediately provide an user with an improvement. The refactoring can still happen in a second step.

jasonkuhrt:
IMO the `BaseClass` has a lot going on between OClif hooks and non-hooks, things that use `this` and others that just live there for no stateful reason. It's difficult to understand and was among other changes made in trying to make the Command class easier to reason about.

I think the amount of discussion and non-alignment for an attempted straight forward change doesn't make me feel good about merging this. I'll close this to avoid further bike shedding. The history remains for future reference or merging if sentiment changes.

PR #6280, opened 2025-01-08
refactor(cli): texture lib

jasonkuhrt:
progresses #6122


packages/libraries/cli/src/helpers/texture/__.ts
```diff
@@ -0,0 +1 @@
+export * as Texture from './texture';
```
jasonkuhrt, line 1: `Text` is a global variable.

packages/libraries/cli/src/helpers/texture/texture.ts
```diff
@@ -0,0 +1,146 @@
+import { inspect as nodeInspect } from 'node:util';
+import colors from 'colors';
+
+export { colors };
```
jasonkuhrt, line 4: Integrate colours into `Texture` namespace.

packages/libraries/cli/src/helpers/texture/texture.ts
```diff
@@ -0,0 +1,146 @@
...
+  prefixedInspect(colors.yellow('ℹ'))(...values) + newline;
+
+export const warning = (...values: unknown[]) =>
+  prefixedInspect(colors.yellow('⚠'))(...values) + newline;
+
+/**
+ * A text builder. Its methods mutate an internal string value.
+ * Useful for quickly building up content.
+ */
+export interface Builder {
```
jasonkuhrt, line 59: This permits string construction in a way that looks like logging. Especially useful for side-effect style programming e.g. `forEach` loops etc.

packages/libraries/cli/src/helpers/schema.ts
jasonkuhrt, line 1: Changes in here are about stopping `this` oriented functions-as-methods that rely on side-effect loggers that are in the `BaseCommand` class. The approach is pure-function oriented instead (with a but of non-pure within the builder but that's contained to the text-renderer functions so encapsulated).

packages/libraries/cli/src/base-command.ts
```diff
@@ -53,28 +53,19 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
   }
 
   success(...args: any[]) {
-    this.log(colors.green('✔'), ...args);
+    this.log(Texture.success(...args));
```
jdolle, line 56: My personal opinion is that this class was performing the abstraction of formatted based on state and that an additional class is probably unnecessary. But if we have more areas beyond the cli commands that will benefit from formatting, then this makes sense.

packages/libraries/cli/src/base-command.ts
```diff
@@ -53,28 +53,19 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
...
   info(...args: any[]) {
-    this.log(colors.yellow('ℹ'), ...args);
+    this.log(Texture.info(...args));
   }
 
   infoWarning(...args: any[]) {
-    this.log(colors.yellow('⚠'), ...args);
-  }
-
-  bolderize(msg: string) {
```
jdolle, line 71: I like these functions being a higher level abstraction and agree with your decision to remove `bolderize` from here.

jdolle (commented):
Overall I think there are some nice changes here. I like your mindset of leveraging abstraction, but would recommend only introducing new layers when it benefits us.

packages/libraries/cli/src/helpers/texture/texture.ts
```diff
@@ -0,0 +1,146 @@
...
+
+export const space = ' ';
+
+export const indent = space.repeat(3);
+
+export const newline = '\n';
+
+export const header = (value: string) => colors.dim('=== ') + colors.bold(value);
+
+export const plural = (value: unknown[]) => (value.length > 1 ? 's' : '');
```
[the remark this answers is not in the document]
jasonkuhrt: It's used in #6122 which goes further in refactoring with this library.

My 0.02 on it is that at bottom the scope of `Texture` could be thought of as general conversion, styling, and layout of data into text, and since its a local micro lib I view its bar to meet on abstraction lower than that of published packages. Its mean to speed up and simplify application code, expected to change often etc. So I'm ok with this here but if ya'll prefer it out let's do that then.

packages/libraries/cli/src/commands/config/delete.ts
```diff
@@ -14,6 +15,6 @@ export default class DeleteConfig extends Command<typeof DeleteConfig> {
   async run() {
     const { args } = await this.parse(DeleteConfig);
     this._userConfig!.delete(args.key);
-    this.success(this.bolderize(`Config flag "${args.key}" was deleted`));
+    this.success(Texture.boldQuotedWords(`Config flag "${args.key}" was deleted`));
```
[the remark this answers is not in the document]
jasonkuhrt: I'm not sure what the original intent and strategy of text for our CLI was/is. I renamed this function to make it more clear what it does (for example I only later realized it actually strips quotes, even).

I intentionally tried to keep the existing functions and styling in place so that the work could move quickly through a PR process on technical discussion rather than terminal styling bike shedding. That said, I share questions like yours and would welcome opportunities later to go further with our terminal design system.

packages/libraries/cli/src/base-command.ts
jasonkuhrt: More motivation for this change arrives in #6122 where formatters (json, text) are co-located and completely separate from IO (e.g. `this.log`).

packages/libraries/cli/src/helpers/texture/__.ts
n1ru4l, line 1: I am not a fan of these `__.ts` files.

Would it be possible to do `export * as Texture from './texture'` within `packages/libraries/cli/src/helpers/texture/texture.ts` instead?

I quickly tried this locally and seems to work fine. 
image

Alternatively, I would also be fine with using TypeScript namespaces or exporting an object or static class instead.

n1ru4l (changes requested):
See https://github.com/graphql-hive/console/pull/6280/files#r1908450410

packages/libraries/cli/src/helpers/texture/__.ts
jasonkuhrt: > Would it be possible to do export * as Texture from './texture' within packages/libraries/cli/src/helpers/texture/texture.ts instead?

Isn't that recursively exporting itself? If so I'm not sure how this could work. Will give it a whirl locally, never tried this.

> Alternatively, I would also be fine with using TypeScript namespaces or exporting an object or static class instead.

Classes don't nest, so there are other places, like `Output` in #6122 that wouldn't fit the pattern of of a class. Then we'd end up with multiple patterns which I think is worse, but don't know how you feel about it.

If you're first idea doesn't work, I'll try TypeScript namespaces, although those have downsides as I've mentioned at length in other comments, since this is a project, the tree shaking downside is not relevant and Node 23 does have an opt-in flag that will compile them too.

jasonkuhrt: We compile to CJS, not ESM, in which a recursive export works. When we switch to ESM we'll see if it continues to work or not.

n1ru4l (approved):
✅

PR #6281, opened 2025-01-08
chore: update pnpm to latest minor

jasonkuhrt:
Progresses #6122 

It seems that the lock file changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "packageManager": "pnpm@9.15.3",
   "engines": {
     "node": ">=22",
-    "pnpm": ">=9.14.4"
+    "pnpm": ">=9.15.3"
   },
   "scripts": {
     "build": "pnpm turbo build --color",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   },
   "license": "MIT",
   "private": true,
-  "packageManager": "pnpm@9.15.1",
+  "packageManager": "pnpm@9.15.3",
   "engines": {
     "node": ">=22",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
   },
   "license": "MIT",
   "private": true,
-  "packageManager": "pnpm@9.15.1",
+  "packageManager": "pnpm@9.15.3",
```
jdolle, line 15: should we lock the engines version?

jasonkuhrt: Engines spec violations lead to process exit which can be pretty aggressive. The `packageManager` field already forces user's machine to dynamically use the specified package manager and version (assuming `corepack` is enabled).

We could discuss engines in another PR maybe, but don't think it has too much impact. I'm not certain what the point of pnpm engine spec with `packageManager` spec is. I guess if developer or CI is not using `corepack` properly then the engines spec is a kind of fallback/sanity check. Seems like a flakey reason though.

jdolle (approved):

dotansimha:
Thanks @jasonkuhrt ! Usually, we give Renovate the liberty to update these gradually, but I see the need for a manual change because of the lock file. 
I've pushed a tiny fix to also verify `engines` here.

package.json
dotansimha: I've noticed this one after pushing a change to `engines`. I think we also need `engines` here in the monorepo root because it affects local development (only). We had before cases where developers had older versions locally, and then it led to weird lock file diffs.

dotansimha (approved):

PR #6282, opened 2025-01-08
Fix editor state and operation handling in Laboratory + e2e tests

kamilkisiela:
**Fixes** 

When opening a new tab or selecting one of the saved operations a wrong query was populated. It was always the default query or the one that's active. Meaning, you couldn't select and see the saved operation :)

When saving the operation, the submit button of the form was always disabled, even when the state of the form was valid.

**e2e tests**

Added tests for CRUD of collections and their operations.
The scenario where a user visits a shared link to an operation is also now tested.

### DIFF
--- a/.changeset/young-peas-battle.md
+++ b/.changeset/young-peas-battle.md
@@ -1,0 +1,7 @@
+---
+'hive': patch
+---
+
+Fix editor state and operation handling in Laboratory.
+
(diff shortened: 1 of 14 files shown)


jdolle:
This seems like a pretty big bug atm for this feature. Should we add a changeset to make sure it's noted in our changelog?

kamilkisiela:
Yes, I completely forgot about the changeset...

PR #6284, opened 2025-01-08
tests(integration): graphql file fixture

jasonkuhrt:
Leverage Vitest fixtures for standard access to test utilities.

### DIFF
--- a/integration-tests/testkit/test.ts
+++ b/integration-tests/testkit/test.ts
@@ -4,7 +4,7 @@
  */
 
 import { test as testBase } from 'vitest';
-import { TmpFileController, createTmpFileController } from './fs';
+import { createTmpFileController, TmpFileController } from './fs';
 
 interface Context {
   graphqlFile: TmpFileController;
--- a/integration-tests/testkit/test.ts
+++ b/integration-tests/testkit/test.ts
@@ -1,0 +1,18 @@
+/**
+ * This module uses Vitest's fixture system to make common usage patterns
+ * of our testkit easily consumable in test cases. @see https://vitest.dev/guide/test-context.html#test-extend
+ */
+
+import { test as testBase } from 'vitest';
+import { TmpFileController, createTmpFileController } from './fs';
+
+interface Context {
+  graphqlFile: TmpFileController;
+}
+
+export const test = testBase.extend<Context>({
+  graphqlFile: async ({}, use) => {
+    const graphqlFile = createTmpFileController({ extension: 'graphql' });
+    await use(graphqlFile);
+  },
+});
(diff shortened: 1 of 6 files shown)


integration-tests/testkit/test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { test as testBase } from 'vitest';
+import { createTmpFileController, TmpFileController } from './fs';
+
+interface Context {
+  graphqlFile: TmpFileController;
+}
+
+export const test = testBase.extend<Context>({
+  graphqlFile: async ({}, use) => {
```
kamilkisiela, line 14: How is this different from:

```ts
const tmpFile = createTmpFile({ extension: 'graphql' });
```

and how does it improve the testing?

kamilkisiela: The way to discover `graphqlFile` in the `test` function arguments is even harder then looking at `testkit/**` and trying to find something that already exists and may be useful.

kamilkisiela: Also, `graphqlFile` is always created and can't be reused

integration-tests/tests/cli/__snapshots__/failure-user-input.spec.ts.snap
```diff
@@ -0,0 +1,412 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`FailureUserInput - { command: 'app:create' } > OUTPUT FORMAT: JSON 1`] = `
```
kamilkisiela, line 3: I think it belongs to a different PR, because I can't find relevant test

jasonkuhrt:
If Vitest test fixtures aren't speaking for themselves yet, in addition to the prior discussion this week already about them, then it isn't time. I predict this topic will resurface organically in the coming months.

integration-tests/testkit/test.ts
jasonkuhrt: > Also, graphqlFile is always created and can't be reused (can't create another temporary graphql file)

I don't understand this comment. Fixtures are created for each test they are used in. So that would suggest you're wrong. Am I missing something?

kamilkisiela: Try to use 2 temporary files in a test

kamilkisiela: We would have to redo the fixture. That’s why passing a function to create a temporary file is imo future proof

jasonkuhrt: > Try to use 2 temporary files in a test

That's a different use-case. Since you brought it up, I will mention that one pattern I like for tests that need to run against a project is a project instance fixture that, when used, creates a temporary directory from which all other methods are relative to. Here is an example https://github.com/graffle-js/graffle/blob/main/tests/e2e/e2e.test.ts#L20-L44. It also shows using a service bound to a test.

But this is getting off topic.

PR #6285, opened 2025-01-08
chore(cli): idiomatic build script with watch

jasonkuhrt:
progresses #6122 

Integration tests run against the _built_ packages. This PR introduces a watch script for CLI to make it easy to work on CLI and integration tests at the same time.

Also, this PR updates the build script to use the idiomatic `--clean` flag. It will remove the output directories according to what is configured in the `tsconfig.json`. This keeps config DRY. Our current approach uses `rm -rf` duplicating config about which dir is for the TS output.

### DIFF
--- a/packages/libraries/cli/package.json
+++ b/packages/libraries/cli/package.json
@@ -30,11 +30,11 @@
     "graphql"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "tsc --build --clean && tsc --build",
+    "build:watch": "tsc --build --watch",
     "oclif:pack": "npm pack && pnpm oclif pack tarballs --no-xz",
     "oclif:upload": "pnpm oclif upload tarballs --no-xz",
     "postpack": "rm -f oclif.manifest.json",
-    "prebuild": "rimraf dist",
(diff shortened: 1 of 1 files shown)


packages/libraries/cli/package.json
```diff
@@ -30,11 +30,11 @@
     "graphql"
   ],
   "scripts": {
-    "build": "tsc",
+    "build": "tsc --build --clean && tsc --build",
+    "build:watch": "tsc --build --watch",
     "oclif:pack": "npm pack && pnpm oclif pack tarballs --no-xz",
     "oclif:upload": "pnpm oclif upload tarballs --no-xz",
     "postpack": "rm -f oclif.manifest.json",
-    "prebuild": "rimraf dist",
```
jdolle, line 37: minor: I prefer using pre- scripts in this case.
`"prebuild": : "tsc --build --clean"`

n1ru4l (approved):

saihaj (approved):

packages/libraries/cli/package.json
jasonkuhrt: Whoops, missed this, will make a follow up PR.

jasonkuhrt: https://github.com/graphql-hive/console/pull/6331

PR #6286, opened 2025-01-08
tests(integration/cli): snapshot cli output

jasonkuhrt:
progresses #6122.

CLI output snapshots permit detecting unexpected changes as large changes are undertaken such as #6122.

### DIFF
--- a/.github/workflows/tests-integration.yaml
+++ b/.github/workflows/tests-integration.yaml
@@ -71,8 +71,6 @@ jobs:
 
       - name: run integration tests
         timeout-minutes: 10
-        env:
-          HIVE_DEBUG: 1
         run: |
           VITEST_MAX_THREADS=${{ steps.cpu-cores.outputs.count }} pnpm --filter integration-tests test:integration --shard=${{ matrix.shardIndex }}/3
 
--- a/integration-tests/testkit/cli-snapshot-serializer.ts
+++ b/integration-tests/testkit/cli-snapshot-serializer.ts
@@ -1,0 +1,73 @@
+import stripAnsi from 'strip-ansi';
+import type { SnapshotSerializer } from 'vitest';
+import { ExecaError } from '@esm2cjs/execa';
+
+export const cliOutputSnapshotSerializer: SnapshotSerializer = {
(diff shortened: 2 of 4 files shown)


.github/workflows/tests-integration.yaml
```diff
@@ -71,8 +71,6 @@ jobs:
 
       - name: run integration tests
         timeout-minutes: 10
-        env:
-          HIVE_DEBUG: 1
```
jasonkuhrt, line 75: Removing this because it causes OCLif to alter output (show stack traces) which will cause output snapshots to not match.

jdolle: Odd -- I see no reference to this env var in our codebase other than here.

Maybe @kamilkisiela remembers why this was first added
https://github.com/graphql-hive/console/pull/930/files#diff-35ee6748a6115f20e5aed37a43d75360daa460ae5c4e5a73e62f63b4f5c3be95R67

integration-tests/testkit/cli-snapshot-serializer.ts
```diff
@@ -0,0 +1,73 @@
...
+  // and we don't care enough about CLI output styling to fork our snapshots for it.
+  value = stripAnsi(value);
+  for (const replacement of variableReplacements) {
+    value = value.replaceAll(replacement.pattern, replacement.mask);
+  }
+  return value;
+};
+
+/**
+ * The esm2cjs execa package we are using is not exporting the error class, so use this.
```
jdolle, line 68: It looks like vitest supports esm modules. It may be possible to migrate to the unpatched `execa`.
If it hasnt been explored already, it may be worth creating an issue to migrate.

jasonkuhrt: ESM migration would be great, OClif supports it now too fwiw. But details like a deep dep in some odd package can derail the effort. Would be keen to try though and then track what if any reasons it couldn't be done, to tackle the effort systematically.

Made an issue: https://github.com/graphql-hive/console/issues/6288

n1ru4l (approved):

PR #6332, opened 2025-01-09
Assign custom roles to members without a role

kamilkisiela:
This migration is going to create a new role for each group of members that have the same scopes but no role assigned.

Users won't be affected by this change, as they will still have the same scopes.

The role will be named `Auto Role {counter}`.
The counter will be reset for each organization.

The script will create 1 or 2 roles for 97% of organizations that haven't complete the migration.

Completes https://the-guild.dev/graphql/hive/product-updates/2023-12-05-member-roles

### DIFF
--- a/packages/migrations/src/actions/2025.01.09T00-00-00.legacy-member-scopes.ts
+++ b/packages/migrations/src/actions/2025.01.09T00-00-00.legacy-member-scopes.ts
@@ -25,7 +25,7 @@ export default {
   async run({ sql, connection }) {
     const queryResult = await connection.query(sql`
       SELECT
-        organization_id as "organizationId"
+        organization_id as "organizationId",
         sorted_scopes as "sortedScopes",
         ARRAY_AGG(user_id) AS "userIds"
       FROM (
--- a/packages/migrations/src/actions/2025.01.09T00-00-00.legacy-member-scopes.ts
+++ b/packages/migrations/src/actions/2025.01.09T00-00-00.legacy-member-scopes.ts
@@ -1,0 +1,98 @@
+import { z } from 'zod';
+import { type MigrationExecutor } from '../pg-migrator';
+
+const QUERY_RESULT = z.array(z.object({
+  organization_id: z.string(),
+  sorted_scopes: z.array(z.string()),
+  user_ids: z.array(z.string()),
+}));
+
+/**
+ * This migration is going to create a new role for each group of members
+ * that have the same scopes but no role assigned.
+ * 
+ * The role will be named "Auto Role {counter}".
+ * The counter will be reset for each organization.
+ * 
+ * Completes:
+ * https://the-guild.dev/graphql/hive/product-updates/2023-12-05-member-roles
+ * 
+ * Users won't be affected by this change, as they will still have the same scopes.
+ */
+export default {
+  name: '2025.01.09T00-00-00.legacy-member-scopes.ts',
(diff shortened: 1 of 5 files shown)


packages/migrations/src/actions/2025.01.09T00-00-00.legacy-member-scopes.ts
```diff
@@ -0,0 +1,98 @@
...
+
+      await connection.query(sql`
+        WITH new_role AS (
+          INSERT INTO organization_member_roles (
+            organization_id, name, description, scopes
+          )
+          VALUES (
+            ${row.organization_id},
+            ${`Auto Role ${counter}`},
+            'Auto generated role to assign to members without a role',
```
kamilkisiela, line 83: We can use it to communicate something

packages/migrations/src/actions/2025.01.09T00-00-00.legacy-member-scopes.ts
```diff
@@ -0,0 +1,98 @@
...
+      const startedAt = Date.now();
+
+      await connection.query(sql`
+        WITH new_role AS (
+          INSERT INTO organization_member_roles (
+            organization_id, name, description, scopes
+          )
+          VALUES (
+            ${row.organization_id},
+            ${`Auto Role ${counter}`},
```
n1ru4l, line 82: just a thought: Let's say the migration fails midway through and we need to manually restart it, wouldn't the role name cause a conflict if we try to insert the same name again as `counter` is reset? Maybe we should something more "unique"?

kamilkisiela: You're right about the uniqueness. It's enforced at pg level.

packages/migrations/src/actions/2025.01.09T00-00-00.legacy-member-scopes.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+export default {
+  name: '2025.01.09T00-00-00.legacy-member-scopes.ts',
+  noTransaction: true,
+  async run({ sql, connection }) {
+    const queryResult = await connection.query(sql`
+      SELECT
+        organization_id,
+        sorted_scopes,
+        ARRAY_AGG(user_id) AS user_ids
```
n1ru4l, line 30:
```suggestion
        organization_id AS "organizationId", 
        sorted_scopes AS "sortedScopes",
        ARRAY_AGG(user_id) AS "userIds"
```

NIT: We could use quotes  and aliases so we don't need to use snake-case in JS 😄

kamilkisiela: NIT

n1ru4l (approved):

packages/migrations/src/actions/2025.01.09T00-00-00.legacy-member-scopes.ts
```diff
@@ -41,6 +41,7 @@ export default {
                   user_id,
                   UNNEST(scopes) AS scope
               FROM organization_member
+              WHERE role_id IS NULL
```
n1ru4l, line 44: 👁️

kamilkisiela: ran the tests and discovered this one

PR #6334, opened 2025-01-09
refactor(cli): move table printer into Texture lib

jasonkuhrt:
A recent PR introduces a local text library. Another PR introduced a text table renderer. This PR merges those two efforts.


packages/libraries/cli/src/helpers/texture/table.ts
jasonkuhrt, line 1: @jdolle If you prefer the previous code I'll revert this. I've seen function expression and statement styles both used. We could have an ESLint rule for that but don't. I'm of the mind to use expressions.

jdolle: This is good.

n1ru4l (approved):

PR #6336, opened 2025-01-09
refactor(cli): introduce hive data layer lib

jasonkuhrt:
Progresses #6122 

When we add JSON format output there will be another schema introduced.

This PR prepares for that by organizing the Hive data source such that it can be easily understood apart from the forthcoming output schema.

Later, data mappers (aka. Data Transformation Objects (DTOs)) that bridge the hive data source to the CLI schema will be introduced, and can be co-located within Hive modules if we wish. In particular the Hive fragments can each have their own mapper, which will replace  their current printers (printers will be owned by the CLI schema layer, since the text will be based off the CLI schema, not its data source).


packages/libraries/cli/src/helpers/hive/_.ts
jasonkuhrt, line 1: @n1ru4l I'm aware you don't like this files. However it is not possible to re-export from other modules.

jasonkuhrt:
Related to these changes is that I think the `graphql` method on `BaseCommand` should be refactored into a client constructor under the `Hive` lib. I think this because it is _more_ specific to the Hive data source than it is to the CLI base command. This change would result in co-locating a domain.

However I have refrained from doing the above in this PR to reduce what must be discussed.

PR #6338, opened 2025-01-09
feat: cli schema:fetch targets latest if actionId argument is missing

jdolle:
### Background

Resolves https://github.com/graphql-hive/console/issues/3855

### Description

Allows all schema:fetch commands to target the latest schema if no actionId is provided.


packages/libraries/cli/src/commands/schema/fetch.ts
```diff
@@ -39,6 +39,37 @@ const SchemaVersionForActionIdQuery = graphql(/* GraphQL */ `
   }
 `);
 
+const LatestSchemaVersionQuery = graphql(/* GraphQL */ `
+  query LatestSchemaVersion(
+    $includeSDL: Boolean!
+    $includeSupergraph: Boolean!
+    $includeSubgraphs: Boolean!
+  ) {
+    latestValidVersion {
```
jdolle, line 48: I tried to use a shared fragment fragment in `SchemaVersionForActionId` and `LatestSchemaVersionQuery` queries, but couldnt get the `graphql` parser to understand it.
I also considered creating a single document with multiple queries but `registryApi(...). request` doesn't support passing an operation name.
So I think a little duplication is the simplest.

n1ru4l (approved):

kamilkisiela (approved):

PR #6344, opened 2025-01-10
feat(cli): json fmt output for cmd whoami

jasonkuhrt:
Progresses #6122

This PR brings JSON format output to `whoami`.

It is large because it introduces the system to support JSON format output in general.

Future PRs for bringing JSON format output to commands should be much smaller.


integration-tests/testkit/cli-output-snapshot.ts
jasonkuhrt, line 1: Changed the file name to match the namespace.

integration-tests/tests/cli/__snapshots__/failure-user-input.spec.ts.snap
jasonkuhrt, line 1: This snapshot captures some proof of JSON for error cases working.

integration-tests/testkit/cli-output-snapshot.ts
```diff
@@ -0,0 +1,328 @@
...
+   *
+   * ```ts
+   * import { expect } from 'vitest';
+   * import { CliOutputSnapshot } from '...';
+   *
+   * expect.addSnapshotSerializer(CliOutputSnapshot.serializer);
+   * CliOutputSnapshot.valueCleaners.push(/foo/, /bar/);
+   * ```
+   */
+  export const register = (parameters?: {
```
jasonkuhrt, line 270: TODO mention which kinds of values can be snpahoted.

packages/libraries/cli/src/output/_exports.ts
jasonkuhrt, line 1: @n1ru4l I think the `output` lib will be too large for putting everything into one `namespace` keyword. Note seen here yet are common output schema types. They will accumulate gradually as commands receive JSON format support. Do you agree? If so, what do you think should be done with this module, and https://github.com/graphql-hive/console/pull/6344/files#diff-40d04b721165c406f8d1cc0f63e9dcbff96bbbe20c3be71dde77acb4376966e8.

packages/libraries/cli/src/helpers/typebox/_exports.ts
jasonkuhrt, line 1: @n1ru4l you mentioned wanting this file removed. However it is not possible to re-export package entrypoints inside `namespace` (`export * from '@sinclair/typebox'`). Also, I think merging all the modules within our typebox directory into one would be a mistake for code readability. Can you please share your vision for the outcome you want to see  here beyond "remove this file"?

PR #6348, opened 2025-01-13
Customizable default member role for OIDC users

### LINKED ISSUE #6346: Assign default role for SSO users
It should be possible to select a default role that will be assigned to users that log in via SSO for the first time.

The assignment UI could either happen within the OIDC provider settings dialog.

image

Or somewhere within the member settings.

image

kamilkisiela:
Closes #6346

- [ ] follow-up PR with the product update

This feature is only relevant to organizations with OIDC enabled.
Every new OIDC users is added to the organization with the default role.
If a default role has not been yet configured, the `Viewer` system role is used instead.

Default role has `default` badge and instructions how to change it
Default role tooltip

Actual setting
OIDC Settings

### DIFF
--- a/packages/migrations/src/actions/2025.01.13T10-08-00.default-role.ts
+++ b/packages/migrations/src/actions/2025.01.13T10-08-00.default-role.ts
@@ -1,0 +1,11 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.01.13T10-08-00.default-role.ts',
+  // Adds a default role to OIDC integration
+  run: ({ sql }) => sql`
+    ALTER TABLE "oidc_integrations"
+    ADD COLUMN "default_role_id" UUID REFERENCES organization_member_roles(id)
+    ON DELETE SET NULL;
+  `,
+} satisfies MigrationExecutor;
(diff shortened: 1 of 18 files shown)


packages/migrations/src/actions/2025.01.13T10-08-00.default-role.ts
```diff
@@ -0,0 +1,11 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.01.13T10-08-00.default-role.ts',
+  // Adds a default role to OIDC integration
+  run: ({ sql }) => sql`
+    ALTER TABLE "oidc_integrations"
+    ADD COLUMN "default_role_id" UUID REFERENCES organization_member_roles(id)
```
n1ru4l, line 8: Since it has a `ON DELETE` trigger we should set an index on `"oidc_integrations"."default_role_id"`.

n1ru4l (approved):
Just one more index https://github.com/graphql-hive/console/pull/6348/files#r1918652362

packages/migrations/src/actions/2025.01.13T10-08-00.default-role.ts
kamilkisiela: Done

Issue #2747, opened 2023-08-17
Feature Request: subgraph metadata (team, contact etc)


kamilkisiela:
Allow to pass a custom directive with contact/team informations, at subgraph level.
This way we could display it in the UI.

jdolle:
I can take this. :)

For passing the data I see a few options:
1. Use `@tag`. We can match prefixes such as `team-` in [this example](https://www.apollographql.com/docs/graphos/reference/federation/directives#applying-metadata). This has the benefit of not introducing a new directive or anything custom just for our platform.
2. Add a new directive for key/value metadata, like `@meta(key: "team", value: "foo")`. This has the benefit of not having to be parse a single tag value to determine key/value. But a negative of duplicating some of the behavior of  federation since`@tag` is intended to handle metadata already.
3. Add metadata via our API. This would keep our API more similar between schema types, since non-federated schemas already support metadata. We can also be very specific about what we support and how we display it in our UI.

I don't think any are terrible options but I like (1) using `@tag` most because it fits with our focus on federation support and is a more generic solution. So if people ever wanted to switch platforms, their metadata would be in their schemas instead of needing to extract it from our API somehow. Also, it doesn't prevent us from adding options to require certain tags based on prefixes in the future.

jdolle:
My plan is to:
1. look for `@tag` directives on the `schema`
2. parse the tag values to have keyed and unkeyed metadata. I.e. `@tag(name: "team-customers")` => `{ key: "team", value: "customers" }`. And `@tag(name: "tier1")` => unkeyed "tier1"
3. save those in our database along with other subgraph data
4. allow fetching this metadata via hive cli https://the-guild.dev/graphql/hive/docs/high-availability-cdn#hive-metadata . This is so that others don't need to worry about parsing or standardizing the tag data.

Since using the `@tag`, I think our current schema UI is sufficient for displaying this information for now. But that we should revisit our schema page as a followup issue. I think in the long run we will want to show metadata per-field in our explorer and possibly other locations.

jdolle:
Speaking with the team it seems `@tag` is already used for [contracts](https://the-guild.dev/graphql/hive/docs/schema-registry/contracts) and overloading it may not be the best option.
I'm looking into enabling metadata via the CLI to unify our metadata model across schema types, and then will look into supporting a custom directive like `@meta(name: String!, value: String!)` that can be added specifically to federated graphs

jdolle:
[...]

PR #6350, opened 2025-01-14
Add a way to view subgraph metadata on target schemas page

jdolle:
### Background

Starts https://github.com/graphql-hive/console/issues/2747

### Description

This displays the metadata we have saved in our database in a more convenient way that is consistent across schema types. Although I don't love the nesting and popover together, I couldn't come up with a better solution that fits in the current UI.
Screenshot 2025-01-13 at 9 10 15 PM


packages/services/api/src/modules/schema/providers/schema-manager.ts
```diff
@@ -233,7 +233,9 @@ export class SchemaManager {
   @atomic(stringifySelector)
   async getMaybeSchemasOfVersion(schemaVersion: SchemaVersion) {
     this.logger.debug('Fetching schemas (schemaVersionId=%s)', schemaVersion.id);
-    return this.storage.getSchemasOfVersion({ versionId: schemaVersion.id });
+    // !note: The decorators are doing something to prevent passing in an additional boolean for "includeMetadata"...
+    //  so for now always include metadata
+    return this.storage.getSchemasOfVersion({ versionId: schemaVersion.id, includeMetadata: true });
```
jdolle, line 238: I'm not sure why the metadata was excluded initially from some calls. I assume the push was for fetching metadata from the CDN and therefore it may have been removed from the schema? Any additional context anyone has on this would help.

jdolle:
Closing since this is not the approach we will be taking for this feature.

PR #6358, opened 2025-01-14
fix: check request threshold against total request count

jdolle:
### Background

resolves a bug discovered in https://github.com/graphql-hive/console/pull/6254#issuecomment-2589472801

### Description

This change adds another DB call to check the total request count against the request threshold before getting the top operations and top clients.


integration-tests/tests/api/target/usage.spec.ts
```diff
@@ -2509,6 +2509,135 @@ test.concurrent(
...
+      .then(r => r.expectNoGraphQLErrors());
+
+    if (above.schemaCheck.__typename !== 'SchemaCheckError') {
+      throw new Error(`Expected SchemaCheckError, got ${above.schemaCheck.__typename}`);
+    }
+
+    expect(above.schemaCheck.errors).toEqual({
+      nodes: [
+        {
+          message: "Field 'a' was removed from object type 'Query'",
```
jasonkuhrt, line 2633: Without your change, is it the case that 'a' would be allowed to be removed when it should not be?

.changeset/nine-meals-raise.md
```diff
@@ -0,0 +1,5 @@
+---
+'hive': patch
```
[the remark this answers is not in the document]
kamilkisiela: it's in `deployment`. We don't version `packages/services/*`, we version Hive as a whole package.

n1ru4l:
Thinking about this, the following would probably the most optimal:

(1.) Get the total request count of requests for each schema coordinate
If the amount is higher than the `requestCountThreshold`, for the affected schema coordinates we get
(2.) the top 10 operations without the current "total >= requestCountThreshold" condition
(3.) the top 10 clients without the current "total >= requestCountThreshold" condition

integration-tests/tests/api/target/usage.spec.ts
jdolle: yeah

.changeset/nine-meals-raise.md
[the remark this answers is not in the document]
dotansimha: Also, @jasonkuhrt , we are using this psuedo-package to manage the versioning of the entire solution (also for self-hosting)

packages/services/api/src/modules/operations/providers/operations-reader.ts
```diff
@@ -1578,7 +1622,7 @@ export class OperationsReader {
                 FROM ${aggregationTableName('operations')}
                 ${this.createFilter({ target: targets, period: roundedPeriod })}
                 GROUP BY target, date
-                ORDER BY
+                ORDER BY
```
n1ru4l, line 1625: 👀

n1ru4l (approved):

packages/services/api/src/modules/operations/providers/operations-reader.ts
kamilkisiela: 👁️ 👁️

kamilkisiela (approved):

PR #6361, opened 2025-01-15
case study content

saihaj:
Update content for case studies

### DIFF
--- a/packages/web/docs/src/app/case-studies/(posts)/wealthsimple/page.mdx
+++ b/packages/web/docs/src/app/case-studies/(posts)/wealthsimple/page.mdx
@@ -152,7 +152,7 @@ The implementation has delivered several significant benefits:
 
 3. **Gateway Evolution**
 
-   Wealthsimple is improving their its API gateway infrastructure by focusing on migrating from
+   Wealthsimple is improving its API gateway infrastructure by focusing on migrating from
    their current Apollo server with schema stitching to Hive Gateway. Having already tested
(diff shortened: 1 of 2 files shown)


packages/web/docs/src/app/case-studies/(posts)/wealthsimple/page.mdx
```diff
@@ -128,15 +152,26 @@ The implementation has delivered several significant benefits:
 
 3. **Gateway Evolution**
 
-   Wealthsimple is improving their API gateway infrastructure by focusing on migrating from their
-   current Apollo server with schema stitching to Hive Gateway. Having already tested
+   Wealthsimple is improving their its API gateway infrastructure by focusing on migrating from
```
hasparus, line 155:
```suggestion
   Wealthsimple is improving its API gateway infrastructure by focusing on migrating from
```

packages/web/docs/src/app/case-studies/(posts)/wealthsimple/page.mdx
```diff
@@ -4,18 +4,19 @@ excerpt:
...
   paramount, leading to the adoption of GraphQL with Hive as their central API management solution.'
 category: Finance
+# We would need to go through approval to get author images. I guess we can leave this as is for now.
 authors:
-  - name: Author
-    position: Position
-    avatar: https://placedog.net/24/24
+  - name:
+    position:
+    avatar:
```
hasparus, line 11: We could also put an empty array there, but I just confirmed that this works — avatar is optional, so when it's falsy we don't render it, and the name and position end up in empty spans.

PR #6374, opened 2025-01-16
Adjust Kafka message estimator on MESSAGE_TOO_LARGE

kamilkisiela:
Adjust the Kafka message size estimation only when Kafka gives back `MESSAGE_TOO_LARGE` error. Until now it included all errors, even connection error.


packages/services/usage/src/usage.ts
```diff
@@ -155,6 +162,9 @@ export function createUsage(config: {
     interval: config.kafka.buffer.interval,
     limitInBytes: 990_000, // 1MB is the limit of a single request to EventHub, let's keep it below that
     useEstimator: config.kafka.buffer.dynamic,
+    isTooLargePayloadError(error) {
+      return error instanceof KafkaJSProtocolError && error.type === 'MESSAGE_TOO_LARGE';
```
kamilkisiela, line 166:
```suggestion
      return error instanceof Error && 'type' in error && error.type === 'MESSAGE_TOO_LARGE';
```

packages/services/usage/src/usage.ts
```diff
@@ -1,4 +1,11 @@
-import { CompressionTypes, Kafka, logLevel, Partitioners, RetryOptions } from 'kafkajs';
+import {
+  CompressionTypes,
+  Kafka,
+  KafkaJSProtocolError,
+  logLevel,
+  Partitioners,
+  RetryOptions,
+} from 'kafkajs';
```
kamilkisiela, line 8:
```suggestion
import { CompressionTypes, Kafka, logLevel, Partitioners, RetryOptions } from 'kafkajs';
```

n1ru4l (approved):

PR #6378, opened 2025-01-16
feat(laboratory/preflight-sandbox): allow appending headers

### LINKED ISSUE #6269: Allow to manipulate outgoing http (graphql) request in Laboratory's Preflight Script
Some kind of API to manipulate the headers at least, as step one.
It's not meant to modify the input and textarea elements of GraphiQL.

```
lab.request.headers.set(...)
```

jasonkuhrt:
### Background

closes #6269

### Description

Enhance the Laboratory Preflight Script to support appending headers to the request.

The approach taken makes no changes to the existing system design. See PR comments for notable points.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Testing

### DIFF
--- a/packages/web/app/src/lib/kit/headers.ts
+++ b/packages/web/app/src/lib/kit/headers.ts
@@ -1,0 +1,14 @@
+export namespace Headers {
+  /**
+   * Take given HeadersInit and append it (mutating) into given Headers.
+   *
+   * @param headers - The Headers object to append to.
+   * @param headersInit - The HeadersInit object to append from.
+   */
+  export const appendInit = (headers: Headers, headersInit: HeadersInit): void => {
+    const newHeaders = new globalThis.Headers(headersInit);
+    newHeaders.forEach((value, key) => {
+      headers.append(key, value);
+    });
+  };
+}
--- a/packages/web/app/src/lib/kit/types/json.ts
+++ b/packages/web/app/src/lib/kit/types/json.ts
@@ -1,0 +1,19 @@
+export namespace JSON {
+  export const encode = <value extends Value>(value: value): string => {
+    return globalThis.JSON.stringify(value);
+  };
+
+  export const encodePretty = <value extends Value>(value: value): string => {
+    return globalThis.JSON.stringify(value, null, 2);
+  };
+
+  export const decode = (value: string): Value => {
+    return globalThis.JSON.parse(value);
+  };
+
+  export type Value = PrimitiveValue | NonPrimitiveValue;
+
+  export type NonPrimitiveValue = { [key: string]: Value } | Array<Value>;
+
+  export type PrimitiveValue = string | number | boolean | null;
+}
--- a/packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
+++ b/packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
@@ -38,9 +38,12 @@ import {
   TriangleRightIcon,
 } from '@radix-ui/react-icons';
 import { useParams } from '@tanstack/react-router';
+import { Kit } from '../kit';
 import { cn } from '../utils';
 import type { LogMessage } from './preflight-script-worker';
 import { IFrameEvents } from './shared-types';
+
+type Result = Omit<IFrameEvents.Outgoing.ResultEventData, 'type' | 'runId'>;
 
 export const preflightScriptPlugin: GraphiQLPlugin = {
   icon: () => (
@@ -141,9 +144,11 @@ const PreflightScript_TargetFragment = graphql(`
 
 type LogRecord = LogMessage | { type: 'separator' };
 
-function safeParseJSON(str: string): Record<string, unknown> | null {
+function safeParseJSON<$UnsafeCast extends Kit.JSON.Value = Kit.JSON.Value>(
+  str: string,
+): $UnsafeCast | null {
   try {
-    return JSON.parse(str);
+    return Kit.JSON.decode(str) as $UnsafeCast;
   } catch {
     return null;
   }
@@ -165,23 +170,47 @@ export function usePreflightScript(args: {
     'hive:laboratory:isPreflightScriptEnabled',
     false,
   );
-  const [environmentVariables, setEnvironmentVariables] = useLocalStorage(
-    'hive:laboratory:environment',
-    '',
-  );
+
+  // ------------
+  // Result State
+  // ------------
+  //
+  // todo: Probably better to store the result as a single JSON object value.
+  // Use a proper versioned schema with codecs for coercing, defaults, validation, etc.
+  //
+  // todo: Improve `useLocalStorage` by allowing passing a codec? Then we can co-locate
+  // the codec with the data and have it applied transparently, use a decoded value
+  // for the default, etc. ?
+
+  const [environmentVariables, setEnvironmentVariables] = useLocalStorage('hive:laboratory:environment', '{}'); // prettier-ignore
   const latestEnvironmentVariablesRef = useRef(environmentVariables);
-  useEffect(() => {
-    latestEnvironmentVariablesRef.current = environmentVariables;
-  });
+  useEffect(() => { latestEnvironmentVariablesRef.current = environmentVariables; }); // prettier-ignore
+  const decodeEnvironmentVariables = (encoded: string) => safeParseJSON<Result['environmentVariables']>(encoded) ?? {}; // prettier-ignore
+
+  const [headers, setHeaders] = useLocalStorage('hive:laboratory:headers', '[]');
+  const latestHeadersRef = useRef(headers);
+  useEffect(() => { latestHeadersRef.current = headers; }); // prettier-ignore
+  const decodeHeaders = (encoded: string) => safeParseJSON<Result['headers']>(encoded) ?? [];
+
+  const decodeResult = (): Result => {
+    return {
+      environmentVariables: decodeEnvironmentVariables(latestEnvironmentVariablesRef.current), // prettier-ignore
+      headers: decodeHeaders(latestHeadersRef.current),
+    };
+  };
+  // -----------
 
   const [state, setState] = useState<PreflightWorkerState>(PreflightWorkerState.ready);
   const [logs, setLogs] = useState<LogRecord[]>([]);
 
   const currentRun = useRef<null | Function>(null);
 
-  async function execute(script = target?.preflightScript?.sourceCode ?? '', isPreview = false) {
+  async function execute(
+    script = target?.preflightScript?.sourceCode ?? '',
+    isPreview = false,
+  ): Promise<Result> {
     if (isPreview === false && !isPreflightScriptEnabled) {
-      return safeParseJSON(latestEnvironmentVariablesRef.current);
+      return decodeResult();
     }
 
     const id = crypto.randomUUID();
@@ -201,7 +230,7 @@ export function usePreflightScript(args: {
           type: IFrameEvents.Incoming.Event.run,
           id,
           script,
-          environmentVariables: (environmentVariables && safeParseJSON(environmentVariables)) || {},
+          environmentVariables: decodeEnvironmentVariables(environmentVariables),
         } satisfies IFrameEvents.Incoming.EventData,
         '*',
       );
@@ -257,16 +286,20 @@ export function usePreflightScript(args: {
         }
 
         if (ev.data.type === IFrameEvents.Outgoing.Event.result) {
-          const mergedEnvironmentVariables = JSON.stringify(
-            {
-              ...safeParseJSON(latestEnvironmentVariablesRef.current),
-              ...ev.data.environmentVariables,
-            },
-            null,
-            2,
-          );
-          setEnvironmentVariables(mergedEnvironmentVariables);
-          latestEnvironmentVariablesRef.current = mergedEnvironmentVariables;
+          const mergedEnvironmentVariablesEncoded = Kit.JSON.encodePretty({
+            ...decodeEnvironmentVariables(latestEnvironmentVariablesRef.current),
+            ...ev.data.environmentVariables,
+          });
+          setEnvironmentVariables(mergedEnvironmentVariablesEncoded);
+          latestEnvironmentVariablesRef.current = mergedEnvironmentVariablesEncoded;
+
+          const mergedHeadersEncoded = Kit.JSON.encodePretty([
+            ...decodeHeaders(latestHeadersRef.current),
+            ...ev.data.headers,
+          ]);
+          setHeaders(mergedHeadersEncoded);
+          latestHeadersRef.current = mergedHeadersEncoded;
+
           setLogs(logs => [
             ...logs,
             `> End running script. Done in ${(Date.now() - now) / 1000}s`,
@@ -299,6 +332,16 @@ export function usePreflightScript(args: {
           setLogs(logs => [...logs, log]);
           return;
         }
+
+        if (ev.data.type === IFrameEvents.Outgoing.Event.ready) {
+          return;
+        }
+
+        if (ev.data.type === IFrameEvents.Outgoing.Event.start) {
+          return;
+        }
+
+        Kit.neverCase(ev.data);
       }
 
       window.addEventListener('message', eventHandler);
@@ -317,7 +360,7 @@ export function usePreflightScript(args: {
       window.removeEventListener('message', eventHandler);
 
       setState(PreflightWorkerState.ready);
-      return safeParseJSON(latestEnvironmentVariablesRef.current);
+      return decodeResult();
     } catch (err) {
       if (err instanceof Error) {
         setLogs(prev => [
@@ -329,7 +372,7 @@ export function usePreflightScript(args: {
           },
         ]);
         setState(PreflightWorkerState.ready);
-        return safeParseJSON(latestEnvironmentVariablesRef.current);
+        return decodeResult();
       }
       throw err;
     }
--- a/packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts
+++ b/packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts
@@ -1,8 +1,14 @@
 import CryptoJS from 'crypto-js';
 import CryptoJSPackageJson from 'crypto-js/package.json';
+import { Kit } from '../kit';
 import { ALLOWED_GLOBALS } from './allowed-globals';
 import { isJSONPrimitive } from './json';
 import { WorkerEvents } from './shared-types';
+
+interface WorkerData {
+  headers: Headers;
+  environmentVariables: Record<string, unknown>;
+}
 
 export type LogMessage = string | Error;
 
@@ -47,11 +53,14 @@ async function execute(args: WorkerEvents.Incoming.EventData): Promise<void> {
     return;
   }
 
-  const { environmentVariables, script } = args;
+  const { script } = args;
 
-  // When running in worker `environmentVariables` will not be a reference to the main thread value
-  // but sometimes this will be tested outside the worker, so we don't want to mutate the input in that case
-  const workingEnvironmentVariables = { ...environmentVariables };
+  const workerData: WorkerData = {
+    headers: new Headers(),
+    // When running in worker `environmentVariables` will not be a reference to the main thread value
+    // but sometimes this will be tested outside the worker, so we don't want to mutate the input in that case
+    environmentVariables: { ...args.environmentVariables },
+  };
 
   // generate list of all in scope variables, we do getOwnPropertyNames and `for in` because each contain slightly different sets of keys
   const allGlobalKeys = Object.getOwnPropertyNames(globalThis);
@@ -116,15 +125,42 @@ async function execute(args: WorkerEvents.Incoming.EventData): Promise<void> {
     },
     environment: {
       get(key: string) {
-        return Object.freeze(workingEnvironmentVariables[key]);
+        return Object.freeze(workerData.environmentVariables[key]);
       },
       set(key: string, value: unknown) {
         const validValue = getValidEnvVariable(value);
         if (validValue === undefined) {
-          delete workingEnvironmentVariables[key];
+          delete workerData.environmentVariables[key];
         } else {
-          workingEnvironmentVariables[key] = validValue;
+          workerData.environmentVariables[key] = validValue;
         }
+      },
+    },
+    /**
+     * Helpers for manipulating the request before it is sent.
+     */
+    request: {
+      /**
+       * Helpers for manipulating the request headers.
+       */
+      headers: {
+        /**
+         * Add one header to the request.
+         *
+         * @param header - The name of the header.
+         * @param value - The value of the header.
+         */
+        add: (header: string, value: string) => {
+          workerData.headers.append(header, value);
+        },
+        /**
+         * Add multiple headers to the request.
+         *
+         * @param headersInit - The {@link HeadersInit} to add.
+         */
+        addMany: (headersInit: HeadersInit) => {
+          Kit.Headers.appendInit(workerData.headers, headersInit);
+        },
       },
     },
     /**
@@ -167,9 +203,12 @@ ${script}})()`;
     sendMessage({ type: WorkerEvents.Outgoing.Event.error, error: error as Error });
     return;
   }
+
   sendMessage({
     type: WorkerEvents.Outgoing.Event.result,
-    environmentVariables: workingEnvironmentVariables,
+    // todo: We need to more precisely type environment value. Currently unknown. Why?
+    environmentVariables: workerData.environmentVariables as any,
+    headers: Array.from(workerData.headers.entries()),
   });
 }
 
--- a/packages/web/app/src/lib/preflight-sandbox/preflight-worker-embed.ts
+++ b/packages/web/app/src/lib/preflight-sandbox/preflight-worker-embed.ts
@@ -1,3 +1,4 @@
+import { Kit } from '../kit';
 import PreflightWorker from './preflight-script-worker?worker&inline';
 import { IFrameEvents, WorkerEvents } from './shared-types';
 
@@ -103,9 +104,9 @@ function handleEvent(data: IFrameEvents.Incoming.EventData) {
 
         if (ev.data.type === WorkerEvents.Outgoing.Event.result) {
           postMessage({
+            ...ev.data,
             type: IFrameEvents.Outgoing.Event.result,
             runId,
-            environmentVariables: ev.data.environmentVariables,
           });
           terminate();
           return;
@@ -129,6 +130,8 @@ function handleEvent(data: IFrameEvents.Incoming.EventData) {
           terminate();
           return;
         }
+
+        Kit.neverCase(ev.data);
       },
     );
 
--- a/packages/web/app/src/lib/preflight-sandbox/shared-types.ts
+++ b/packages/web/app/src/lib/preflight-sandbox/shared-types.ts
@@ -28,12 +28,10 @@ export namespace IFrameEvents {
       log: string | Error;
     };
 
-    type ResultEventData = {
+    export interface ResultEventData extends Omit<WorkerEvents.Outgoing.EventData.Result, 'type'> {
       type: Event.result;
       runId: string;
-      environmentVariables: Record<string, unknown>;
-    };
-
+    }
     type ErrorEventData = {
       type: Event.error;
       runId: string;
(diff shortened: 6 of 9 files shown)


packages/web/app/src/lib/kit/headers.ts
jasonkuhrt, line 1: I have added a "Kit" lib to house standard library like building blocks. An alternative to this approach is ad-hoc module-scope functions.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -165,23 +170,47 @@ export function usePreflightScript(args: {
...
   const [state, setState] = useState<PreflightWorkerState>(PreflightWorkerState.ready);
   const [logs, setLogs] = useState<LogRecord[]>([]);
 
   const currentRun = useRef<null | Function>(null);
 
-  async function execute(script = target?.preflightScript?.sourceCode ?? '', isPreview = false) {
+  async function execute(
+    script = target?.preflightScript?.sourceCode ?? '',
+    isPreview = false,
+  ): Promise<Result> {
```
jasonkuhrt, line 211: My implementation approach boils down to having `execute` return an extensible set of data. In this PR it includes environment variables as before and, newly, headers.

The current code returns environment variables directly.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -257,16 +286,20 @@ export function usePreflightScript(args: {
...
-          latestEnvironmentVariablesRef.current = mergedEnvironmentVariables;
+          const mergedEnvironmentVariablesEncoded = Kit.JSON.encodePretty({
+            ...decodeEnvironmentVariables(latestEnvironmentVariablesRef.current),
+            ...ev.data.environmentVariables,
+          });
+          setEnvironmentVariables(mergedEnvironmentVariablesEncoded);
+          latestEnvironmentVariablesRef.current = mergedEnvironmentVariablesEncoded;
+
+          const mergedHeadersEncoded = Kit.JSON.encodePretty([
+            ...decodeHeaders(latestHeadersRef.current),
```
jasonkuhrt, line 297: This code copies the pattern that was in use for environment variables. I do not understand it. Specifically, when and why would the executed script append to an existing value in local storage rather than be expected to produce the total value upon every execution?

packages/web/app/src/lib/preflight-sandbox/shared-types.ts
```diff
@@ -28,12 +28,10 @@ export namespace IFrameEvents {
       log: string | Error;
     };
 
-    type ResultEventData = {
+    export interface ResultEventData extends Omit<WorkerEvents.Outgoing.EventData.Result, 'type'> {
```
jasonkuhrt, line 31: I have opted to encode the passthrough relationship from the worker to the iframe with regards to result data.

packages/web/app/src/lib/preflight-sandbox/shared-types.ts
jasonkuhrt, line 1: This module diff is bigger than needed. At the start of my implementation I mistakenly thought I would need to introduce a new event type. Upon attempting to do so, I noticed there was no enforcement between the enum members and the event data. I improved the code to remedy that. However later I realized no new event type was needed, just an augmentation of the existing result event. I could revert the changes here but they remain anyways an improvement. However if we keep the changes I would recommend applying the same pattern to the `IFrame` types to keep our code consistent.

Let me know the step you prefer being taken in this PR.

packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -328,8 +328,26 @@ function LaboratoryPageContent(props: {
         });
         try {
           const result = await preflightScript.execute();
-          if (result && headers) {
-            headers = substituteVariablesInHeader(headers, result);
+          // todo: Why check `result` truthiness here if it can never by falsy?
+          if (result) {
```
jasonkuhrt, line 332: Can we remove the `if` check here?

packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -328,8 +328,26 @@ function LaboratoryPageContent(props: {
...
+            // so that users can have a predictable experience working with headers regardless of the place those
+            // headers come from.
+            //
+            // todo: GraphiQLFetcher appears to only support record-shaped headers which seems wrong because it
+            // precludes complete usage of Headers data structure, namely where there are multiple values for one
+            // header. We could try to hack a solution here by doing merges of such cases but that seems
+            // likely to introduce more bugs given the different formats that different kinds of headers use to
+            // delineate multiple values.
+            //
+            // What should really happen is that GraphiQLFetcher accepts a HeadersInit type.
```
jasonkuhrt, line 343: Is my analysis here mistaken?

packages/web/app/src/lib/preflight-sandbox/preflight-worker-embed.ts
```diff
@@ -129,6 +130,8 @@ function handleEvent(data: IFrameEvents.Incoming.EventData) {
           terminate();
           return;
         }
+
+        Kit.neverCase(ev.data);
```
jasonkuhrt, line 134: This is an exhaustiveness check. It helps improve the static type safety of our code by ensuring we don't accidentally forget to handle cases.

jasonkuhrt (commented):
There are still things to do on this PR, hence it being a draft. Those things are:

1. Website docs
2. changeset entry
3. tests

Before taking on that work, I would like to receive feedback that the PR is the right direction. Thanks!

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -165,23 +170,47 @@ export function usePreflightScript(args: {
...
+  // ------------
+  //
+  // todo: Probably better to store the result as a single JSON object value.
+  // Use a proper versioned schema with codecs for coercing, defaults, validation, etc.
+  //
+  // todo: Improve `useLocalStorage` by allowing passing a codec? Then we can co-locate
+  // the codec with the data and have it applied transparently, use a decoded value
+  // for the default, etc. ?
+
+  const [environmentVariables, setEnvironmentVariables] = useLocalStorage('hive:laboratory:environment', '{}'); // prettier-ignore
```
jasonkuhrt, line 185: An empty string will fail JSON parsing. The default value is changed to be something that would parse into an empty state of the value type. Now if JSON parsing fails, it should only be because a stored value was actually invalid. A further improvement might be to notify our error monitoring about it. However given these values also include userland interaction (e.g. user could muck around in their browser local storage) we would need to apply an appropriate filter on incoming errors such as error rate analysis and only keep the most clearly egregious signals for off-work alerting.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -141,9 +144,11 @@ const PreflightScript_TargetFragment = graphql(`
 
 type LogRecord = LogMessage | { type: 'separator' };
 
-function safeParseJSON(str: string): Record<string, unknown> | null {
+function safeParseJSON<$UnsafeCast extends Kit.JSON.Value = Kit.JSON.Value>(
+  str: string,
+): $UnsafeCast | null {
   try {
-    return JSON.parse(str);
+    return Kit.JSON.decode(str) as $UnsafeCast;
```
jdolle, line 151: Would it make sense to move `safeParseJSON` into the `JSON` namespace?

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
jdolle: I think the using spread on the object like the environmentVariables is fine because if it's called multiple times then the keys will override the prior value. This change to headers may be an issue though because you are spreading an array. There may be a bug where `ev.data.headers` keeps getting appended to the list of headers unless that's deduped elsewhere.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
```diff
@@ -299,6 +332,16 @@ export function usePreflightScript(args: {
           setLogs(logs => [...logs, log]);
           return;
         }
+
+        if (ev.data.type === IFrameEvents.Outgoing.Event.ready) {
+          return;
+        }
+
+        if (ev.data.type === IFrameEvents.Outgoing.Event.start) {
```
jdolle, line 340: I like the explicitness of handling all cases. Is it worth adding debug logging here in case e.g. start is called but not ready? I'm not sure what the exact flow is or what could go wrong.

packages/web/app/src/pages/target-laboratory.tsx
jdolle: If it throws when there is no result, then yes

packages/web/app/src/lib/kit/types/json.ts
```diff
@@ -0,0 +1,19 @@
+export namespace JSON {
+  export const encode = <value extends Value>(value: value): string => {
```
kamilkisiela, line 2: Can you explain why we need to make `encode` type safe? and how is it different from `JSON.stringify`?

packages/web/app/src/lib/kit/types/json.ts
```diff
@@ -0,0 +1,19 @@
...
+    return globalThis.JSON.stringify(value);
+  };
+
+  export const encodePretty = <value extends Value>(value: value): string => {
+    return globalThis.JSON.stringify(value, null, 2);
```
kamilkisiela, line 7: Same, what's the difference between:

```js
Hive.kit.JSON.encodePretty(value):
// vs
JSON.stringify(value, null, 2)
```

Imo it's unnecessary layer on top of another layer just to perform JSON serialization

packages/web/app/src/lib/kit/types/json.ts
```diff
@@ -0,0 +1,19 @@
...
+  };
+
+  export const decode = (value: string): Value => {
+    return globalThis.JSON.parse(value);
```
kamilkisiela, line 11: Same with decoding, if we expect deserialized json to have some shape, we can just slap Zod on it and use JSON.parse

packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts
```diff
@@ -116,17 +125,44 @@ async function execute(args: WorkerEvents.Incoming.EventData): Promise<void> {
...
+         */
+        add: (header: string, value: string) => {
+          workerData.headers.append(header, value);
+        },
+        /**
+         * Add multiple headers to the request.
+         *
+         * @param headersInit - The {@link HeadersInit} to add.
+         */
+        addMany: (headersInit: HeadersInit) => {
```
kamilkisiela, line 161: Can't we just let people access `Headers` that they are familiar with? With your approach we need to add methods like `delete`/`remove`, `get` and `has` and they need to learn new API.

n1ru4l: We should probably give them access to a [Headers object](https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers). 

We can then serialize it to a JSON object and forward it to the main thread.

kamilkisiela: Yes, entries -> Headers -> entries

packages/web/app/src/pages/target-laboratory.tsx
n1ru4l: image

`safeParseJSON` returns `null`, thus the check

packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -328,8 +328,26 @@ function LaboratoryPageContent(props: {
         });
         try {
           const result = await preflightScript.execute();
-          if (result && headers) {
-            headers = substituteVariablesInHeader(headers, result);
+          // todo: Why check `result` truthiness here if it can never by falsy?
+          if (result) {
+            // We merge the result headers into the fetcher headers before performing header variable substitution
+            // so that users can have a predictable experience working with headers regardless of the place those
+            // headers come from.
```
n1ru4l, line 335: I think any behaviour is fine as long as it is explained within the documentation.

The reason people asked for this feature is to avoid having to substitute headers and can set these directly.
So to me it might even make sense to apply these headers after substitution.

packages/web/app/src/pages/target-laboratory.tsx
n1ru4l: I am pretty sure it would support passing `HeadersInit`, as it is forwarded to `fetch`. https://github.com/graphql/graphiql/blob/main/packages/graphiql-toolkit/src/create-fetcher/createFetcher.ts

While technically we should support multi-header values, I don't think it would be used by anyone right now and we could improve supporting it later on.

packages/web/app/src/lib/preflight-sandbox/shared-types.ts
n1ru4l: I am fine with the changes in this file

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
jasonkuhrt: > This change to headers may be an issue though because you are spreading an array. There may be a bug where ev.data.headers keeps getting appended to the list of headers unless that's deduped elsewhere.

Great call out. Oddly enough it is deduped later but in a way that I consider a bug wherein multiple header values could be lost.

On the other hand, I agree with you of course that here we could have kind of the opposite problem where we turn what were single-value headers into duplicated/multi-value headers.

In order to solve this, I would like to understand the system better. Why/when are there local storage values merged here.

packages/web/app/src/lib/preflight-sandbox/shared-types.ts
jasonkuhrt: I will revert.

However I would like to clarify that DRY was not the point here.

This was about encoding a data relationship. This code would tell the reader that Iframe Result type is tied to the Worker Result type. It would imply that there is a passthrough relationship.

I am interpreting your feedback to mean the following: Any apparent data symmetry between these layers should be treated as incidental and not to be encoded.

packages/web/app/src/pages/target-laboratory.tsx
jasonkuhrt: In my PR a fallback value is now returned. I seemed to have forgotten that the original code did this and confused my self. I will remove the `if` in this PR.

packages/web/app/src/pages/target-laboratory.tsx
jasonkuhrt: If you look at the types of the fetcher, it does not agree. So at the very least there is a TS bug.

I will make sure to mention this limitation in our documentation so that we help avoid frustrating users who IMO rightly expect such primitives to work.

In light of this limitation,  perhaps it is a mistake to ship `HeadersInit` support in the `lab` library then, since it would lead users to expect that multiple value headers are supported when in fact they are not.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
jasonkuhrt: I think so, however based on past experience I am concerned with mixing refactors and feature work.

Once a refactor begins, it either continues to its natural conclusion (consistent approach throughout codebase) which can be _huge_, or, an arbitrary stop point is chosen (maybe something semi-clean like a scope of module, or namespace, or package) leading to two styles in the codebase for an intermediary time window (which I think we know to mean, more often than not, without a habitual practice of refactoring, long term).

A theoretically "clean" approach is taking all the desired refactor work out into its own  PR but there are many problems with this in practice too.

In a team setting, it is difficult for me personally to navigate this.

I am prepared to automatically follow the PR-by-PR reviewer feedback that sanctions a refactor.

So for example I am fine to respond to your comment by doing that refactor. But I do not want to discover another reviewer disagrees with it after. If you had been here for a long time, I guess I would have a higher trust level.

Anyways, those are my sloppy on-the-fly thoughts as I juggle other work in my day etc.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
jasonkuhrt: When you say debug logging, are we talking console.log or something more specialized?

packages/web/app/src/lib/kit/types/json.ts
jasonkuhrt: The types of `JSON.stringify` and `encode` are different. The `encode` function is type safe on its inputs. When using `encode`, it would yield a static type error if you passed an `Error`, `Date` etc. anywhere in the value. `JSON.stringify` does not do this. By being more strict, it forces the developer to think about the data they are serializing, reducing bugs. For example its unlikely this is what a developer would ever intend:

```ts
> JSON.stringify({data: new Error('foobar')})
'{"data":{}}'
```

If a developer doesn't care, the solution is easy:

````ts
// @ts-expect-error: I couldn't care less about your type safety
JSON.encode({ data: new Error('foobar') })
````

Or

````ts
// I couldn't care less about your type safety
JSON.encode({ data: new Error('foobar') } as any)
````

Now the code is explicitly not type safe, which I think is objectively superior to implicitly not type safe.

Do you disagree?

File: packages/web/app/src/lib/kit/types/json.ts (hunk 11 above)
Comment (jasonkuhrt) [reply packages/web/app/src/lib/kit/types/json.ts:7]: https://github.com/graphql-hive/console/pull/6378#discussion_r1920311753

Comment (jasonkuhrt) [reply packages/web/app/src/lib/kit/types/json.ts:7]: We don't need two threads for this conversation. I am resolving this one.

File: packages/web/app/src/lib/kit/types/json.ts (hunk 10 above)
Comment (kamilkisiela) [reply packages/web/app/src/lib/kit/types/json.ts:2]: I don't think I agree. At the point where you stringify something, you're most likely to already have something typed anyway, and validated.

We never had an issue with serializing a function or a class.

File: packages/web/app/src/lib/kit/types/json.ts (hunk 12 above)
Comment (jasonkuhrt) [reply packages/web/app/src/lib/kit/types/json.ts:11]: `JSON.parse` results in an `any` return type. This is generally even worse than the problems with `any` on the input that I mentioned in https://github.com/graphql-hive/console/pull/6378#discussion_r1920311753, because that `any` propagates into all downstream code.

I do not think `any` should be a casually accepted type in our code base. I think its removal will result in code that is easier for us to read and write, fewer bugs for end-users, and lowered time to delivery of features.

File: packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts (hunk 13 above)
Comment (jasonkuhrt) [reply packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts:161]: Yes we can give them access to data to mutate.

Given this [comment](https://github.com/graphql-hive/console/pull/6378#discussion_r1920285682) I am no longer sure it should be `Headers` however.

Based on the resolution there, we can expose a reference of type `Record<string, string>` or `Headers`. TBD.

File: packages/web/app/src/lib/kit/types/json.ts (hunk 12 above)
Comment (kamilkisiela) [reply packages/web/app/src/lib/kit/types/json.ts:11]: We could turn `JSON.parse` type definition to `unknown` globally and go from there.

Comment (kamilkisiela) [reply packages/web/app/src/lib/kit/types/json.ts:11]: even more type safe

File: packages/web/app/src/pages/target-laboratory.tsx (hunk 14 above)
Comment (jasonkuhrt) [reply packages/web/app/src/pages/target-laboratory.tsx:335]: Despite the comment, I do not have a strong opinion on the product behaviour here.

Please let me know your final position and I will make it so.

> I think any behaviour is fine as long as it is explained within the documentation.

Agreed, I have made a note about documenting this in the PR task list.

File: packages/web/app/src/pages/target-laboratory.tsx (hunk 5 above)
Comment (kamilkisiela) [reply packages/web/app/src/pages/target-laboratory.tsx:343]: Are you sure GraphiQL does not accept arrays? I wouldn't trust `Fetcher` type

Comment (jasonkuhrt) [reply packages/web/app/src/pages/target-laboratory.tsx:343]: I am not sure. All I know so far is as I said:

> If you look at the types of the fetcher, it does not agree (is Record<... | undefined). So at the very least there is a TS bug.

May I investigate this first so that this PR can proceed optimally?

Comment (kamilkisiela) [reply packages/web/app/src/pages/target-laboratory.tsx:343]: If we can support `HeadersInt` then it will solve the Lab API part and we won't have to duplicate it or come up with our own API, so yes. I think it would be nice to double check if it's true.

File: packages/web/app/src/lib/kit/types/json.ts (hunk 10 above)
Comment (jasonkuhrt) [reply packages/web/app/src/lib/kit/types/json.ts:2]: I think your approach prefers an unsafe implicit style presupposing optimal conditions which is a point of view I cannot see myself aligning with. Its good that our positions are clear though. To resolve this let's just defer to your ownership role here. These changes will be reverted.

File: packages/web/app/src/lib/kit/types/json.ts (hunk 12 above)
Comment (jasonkuhrt) [reply packages/web/app/src/lib/kit/types/json.ts:11]: It is not more type safe, it just more annoying. The result of `JSON.parse` has a known type, which I have defined in this PR.

File: packages/web/app/src/pages/target-laboratory.tsx (hunk 14 above)
Comment (jasonkuhrt) [reply packages/web/app/src/pages/target-laboratory.tsx:335]: https://github.com/graphql/graphiql/blob/main/packages/graphiql-toolkit/src/create-fetcher/types.ts#L107

File: packages/web/app/src/pages/target-laboratory.tsx (hunk 5 above)
Comment (jasonkuhrt) [reply packages/web/app/src/pages/target-laboratory.tsx:343]: Type Level (not supported):

https://github.com/graphql/graphiql/blob/main/packages/graphiql-toolkit/src/create-fetcher/types.ts#L107

Value Level (not supported):

https://github.com/graphql/graphiql/blob/d217a65efcf9d4c19368f71310339c831556f26b/packages/graphiql-toolkit/src/create-fetcher/lib.ts#L63-L67

Comment (jasonkuhrt) [reply packages/web/app/src/pages/target-laboratory.tsx:343]: I have made a PR to address the upstream issue https://github.com/graphql/graphiql/pull/3854.

I am not sure how long it will take for that change to ship. Options now:

1. Wait
2. Ship with an interface of `Record<string, string>`. When the upstream minor is released, break this interface by shipping `Headers`.
3. Ship `Headers` and document the limitation (website, changelog, JSDoc). When the upstream minor is resolved, we update our system and the limitation is gone. Users do not need to migrate their code.

I think (3) is a good option because this series of events edge case is unlikely:
1. We ship this before the upstream fix and
2. A user wants to use multiple headers feature before the upstream fix and
3. That user has not seen the caveat.

That will be my direction now. If you disagree let me know.

**Comment by jasonkuhrt** [comment]
Thank you all for the draft feedback. I will have a reviewable version of this PR by the end of next Monday.

File: packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx (hunk 9 above)
Comment (jdolle) [reply packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx:340]: Yeah like `console.debug` or send an alert to sentry if we're worried about a specific case.

File: packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx (hunk 8 above)
Comment (jdolle) [reply packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx:151]: I appreciate keeping the changes to a minimum.
A followup sounds fine. I think for larger refactors we could also open new issues and tackle them as we have time.

File: packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts (hunk 13 above)
Comment (jasonkuhrt) [reply packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts:161]: Give this feedback on headers, why then were environment variables treated differently?

https://github.com/graphql-hive/console/blob/main/packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts#L117-L129

File: packages/web/app/src/lib/preflight-sandbox/shared-types.ts
Comment (jasonkuhrt) [reply packages/web/app/src/lib/preflight-sandbox/shared-types.ts:1]: Ok. I will apply the same refactor to `IFrameEvents` then.

File: packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts (hunk 13 above)
Comment (kamilkisiela) [reply packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts]: We wanted to merge Preflight Script asap and it was very chaotic process

Comment (kamilkisiela) [reply packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts]: Now we want to incrementally add new features and improve logic there

**Comment by jasonkuhrt** [comment]
Given [this feedback](https://github.com/graphql-hive/console/pull/6378#discussion_r1919791727), what was the rationale for the existing environment variables?

https://github.com/graphql-hive/console/blob/main/packages/web/app/src/lib/preflight-sandbox/preflight-script-worker.ts#L117-L129

File: packages/web/app/src/pages/target-laboratory.tsx (hunk 14 above)
Comment (jasonkuhrt) [reply packages/web/app/src/pages/target-laboratory.tsx]: I took your "might" as a preference and made the change in that direction.

File: packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx (hunk 7 above)
Comment (jasonkuhrt) [reply packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx:192]: Turned these thoughts into a todo.

File: packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx (hunk 8 above)
Comment (jasonkuhrt) [reply packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx:151]: I like your suggestion and in this case the diff effect is minimal so I've done it in this PR.

File: packages/web/app/src/pages/target-laboratory.tsx [hunk 15]
```diff
@@ -251,17 +253,17 @@ function Save(props: {
...
       if (typeof value === 'string') {
         // Replace all occurrences of `{{keyName}}` strings only if key exists in `environmentVariables`
-        value = value.replaceAll(/{{(?<keyName>.*?)}}/g, (originalString, envKey) => {
+        value = value.replaceAll(/{{(?<keyName>.*?)}}/g, (originalString, envKey: string) => {
           return Object.hasOwn(environmentVariables, envKey)
-            ? (environmentVariables[envKey] as string)
+            ? String(environmentVariables[envKey])
             : originalString;
         });
       }
```
jasonkuhrt, line 269: Having specified the index type of environment variables beyond `unknown` upstream, it led me here. This code now explicitly handles non-strings.

packages/web/app/src/lib/preflight-sandbox/graphiql-plugin.tsx
jasonkuhrt: I've come to better understand the system. Local storage is not needed in the first place for headers.

packages/web/app/src/lib/kit/types/json.ts
jasonkuhrt, line 1: I have only just discovered https://github.com/graphql-hive/console/blob/231f7ad4af64342efde5cc14fb1a4baf73f9a6fc/packages/web/app/src/lib/preflight-sandbox/json.ts

I will refactor this PR to use that instead of the `Kit` namespace that I introduced. 

Also though, I propose the value of a centralized place for easily discoverable standard library like functionality. `Kit` aims to move towards such a thing. It might help reduce situations like this where an existing abstraction is missed by a colleague whom goes on to recreate it in their own PR later.

packages/web/app/src/pages/target-laboratory.tsx
jasonkuhrt: Following up on this point, because we have gone with `Headers`, we actually avoid any issue here. This is because `Headers` performs concatenation eagerly:

```ts
> Object.fromEntries(new Headers([['bar','1'],['bar','2']]))
{ bar: '1, 2' }
```

None the less, the GraphiQL PR is an improvement for the wider community. I will see that through albeit now less urgently.

jasonkuhrt:
fyi https://github.com/graphql-hive/console/pull/6378#discussion_r1924328387

cypress/e2e/laboratory-preflight-script.cy.ts
```diff
@@ -1,5 +1,16 @@
...
+const selectors = {
+  buttonModal: 'preflight-script-modal-button',
+  buttonToggle: 'toggle-preflight-script',
+  graphiql: {
+    buttonExecute: '.graphiql-execute-button',
+  },
+  modal: {
+    buttonSubmit: 'preflight-script-modal-submit',
+  },
+};
```
jasonkuhrt, line 12: If you like this pattern I can apply it consistently elsewhere now or later.

cypress/e2e/laboratory-preflight-script.cy.ts
```diff
@@ -186,6 +197,66 @@ throw new TypeError('Test')`,
...
+    );
+    cy.dataCy(selectors.modal.buttonSubmit).click();
+    cy.intercept({
+      method: 'POST',
+      headers: {
+        [preflightHeaders.accept.name]: preflightHeaders.accept.value,
+      },
+    }).as('post');
+    cy.get(selectors.graphiql.buttonExecute).click();
+    cy.wait('@post');
```
jasonkuhrt, line 255: I found a bug wherein:
- if GraphiQL is executed
- and then the preflight script is enabled, setup with a headers altering script that overrides some base headers
- the effect of the preflight script on request is nil (the base headers are not overridden)

If the preflight headers are something not in base headers, then it works.

I'm not sure what's going on, investigating.

cypress/e2e/laboratory-preflight-script.cy.ts
```diff
@@ -186,6 +202,81 @@ throw new TypeError('Test')`,
...
+    cy.dataCy(selectors.buttonModalCy).click();
+    setEditorScript(`lab.request.headers.append('accept', '${preflightHeaders.accept}')`);
+    cy.dataCy(selectors.modal.buttonSubmitCy).click();
+    // Run GraphiQL
+    cy.intercept({ headers: preflightHeaders }).as('request');
+    cy.get(selectors.graphiql.buttonExecute).click();
+    cy.wait('@request');
+  });
+
+  it.only('result.request.headers are NOT substituted with environment variables', () => {
```
n1ru4l, line 244:
```suggestion
  it('result.request.headers are NOT substituted with environment variables', () => {
```

packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -356,6 +356,16 @@ function LaboratoryPageContent(props: {
           hasFinishedPreflightScript = true;
         }
 
+        const headers = {
+          // todo: test case covering point below
+          // We want to prevent users from interpolating environment variables into
+          // their preflight script headers. So, apply substitution BEFORE merging
+          // in preflight headers.
+          //
```
n1ru4l, line 364: Seems like we can remove this TODO?

n1ru4l (approved):
Looks good to me! 🥳 

I just noted the `.only` and TODO comment for the header substitution.

dotansimha (approved):

PR #6382, opened 2025-01-17
Improvements to OTEL traces in usage

dotansimha:
- [x] Remove redundant child span for usage requests 
- [x] Better type-safety for cases where OTEL is disabled 
- [x] Added more breadcrumbs and improve the existing 
- [ ] question: https://github.com/graphql-hive/console/pull/6382#issuecomment-2598040827


dotansimha:
@kamilkisiela @n1ru4l 
In most cases, we are using `400` / `401` and `403` in specific cases where we do have enough info. 
Should we ignore all traces that ended up with these status codes? At the moment we ignore only `200` and `429`.

PR #6389, opened 2025-01-17
Show Impact metric in the Operations list on the Insights page.

kamilkisiela:
Impact equals to the total time spent on this operation in the selected period in seconds.
It helps assess which operations contribute the most to overall latency.

```
Impact = Requests * avg/1000
```

The result of that formula is formatted number:

```
1                = 1
1.5              = 1.5
10               = 10
10.5             = 11
100              = 100
554              = 554
1000             = 1K
1240             = 1.2K
10000            = 10K
10712            = 11K
100000           = 100K
321120.15        = 321K
1000000          = 1M
10000000         = 10M
100000000        = 100M
1000000000       = 1B
3123400000       = 3.1B
10000000000      = 10B
100000000000     = 100B
1000000000000    = 1T
10000000000000   = 10T
100000000000000  = 100T
```

A time unit could be used instead, but comparing seconds, minutes, hours, days, months, years... is awkward and much harder (9 weeks vs 4 months) than 23K vs 1M.

---

I also used a newer Tooltip component, so it looks much better.

### DIFF
--- a/.changeset/short-donkeys-live.md
+++ b/.changeset/short-donkeys-live.md
@@ -1,0 +1,12 @@
+---
+'hive': minor
+---
+
+Show Impact metric in the Operations list on the Insights page.
+Impact equals to the total time spent (p95) on this operation in the selected period in seconds.
+It helps assess which operations contribute the most to overall latency.
+
+
+```
+Impact = Requests * p95/1000
+```
--- a/.changeset/short-donkeys-live.md
+++ b/.changeset/short-donkeys-live.md
@@ -3,10 +3,10 @@
 ---
 
 Show Impact metric in the Operations list on the Insights page.
-Impact equals to the total time spent (p95) on this operation in the selected period in seconds.
+Impact equals to the total time spent on this operation in the selected period in seconds.
 It helps assess which operations contribute the most to overall latency.
 
 
 ```
-Impact = Requests * p95/1000
+Impact = Requests * avg/1000
 ```
--- a/packages/web/app/src/components/target/insights/List.tsx
+++ b/packages/web/app/src/components/target/insights/List.tsx
@@ -59,7 +59,9 @@ function OperationRow({
   const p90 = useFormattedDuration(operation.p90);
   const p95 = useFormattedDuration(operation.p95);
   const p99 = useFormattedDuration(operation.p99);
-  const impact = useFormattedNumber(operation.impact);
+  const impact = useFormattedNumber(
+    operation.impact < 1000 ? Math.round(operation.impact * 100) / 100 : operation.impact,
+  );
 
   return (
     <>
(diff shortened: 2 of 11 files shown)


kamilkisiela:
@erikwrede

packages/web/app/src/components/target/insights/List.tsx
```diff
@@ -56,6 +59,7 @@ function OperationRow({
   const p90 = useFormattedDuration(operation.p90);
   const p95 = useFormattedDuration(operation.p95);
   const p99 = useFormattedDuration(operation.p99);
+  const impact = useFormattedNumber(operation.impact);
```
kamilkisiela, line 62:
```suggestion
  const impact = useFormattedNumber(
    operation.impact < 1000 ? Math.round(operation.impact * 100) / 100 : operation.impact,
  );
```

erikwrede (approved):
Thanks for tagging me! The current use of the p95 * request count metric is already quite helpful. However, for operations with p95s caused by load spikes, the "Impact" metric may be misleading because it only reflects the top 5% of requests. In such cases, "Impact" might be more accurately represented by the total sum of latency across all requests.
Both the present and suggested metrics are incredibly useful, so this is a great start!

kamilkisiela:
> Thanks for tagging me! The current use of the p95 * request count metric is already quite helpful. However, for operations with p95s caused by load spikes, the "Impact" metric may be misleading because it only reflects the top 5% of requests. In such cases, "Impact" might be more accurately represented by the total sum of latency across all requests. Both the present and suggested metrics are incredibly useful, so this is a great start!

ok, I moved it to `(averageInSeconds) * count`

.changeset/short-donkeys-live.md
```diff
@@ -0,0 +1,12 @@
...
+'hive': minor
+---
+
+Show Impact metric in the Operations list on the Insights page.
+Impact equals to the total time spent (p95) on this operation in the selected period in seconds.
+It helps assess which operations contribute the most to overall latency.
+
+
+```
+Impact = Requests * p95/1000
```
kamilkisiela, line 11:
```suggestion
Impact = Requests * avg/1000
```

packages/web/app/src/components/target/insights/List.tsx
```diff
@@ -222,29 +240,47 @@ function OperationsTable({
...
+                        <Tooltip>
+                          <TooltipTrigger>
+                            <InfoIcon className="size-4 text-gray-400" />
+                          </TooltipTrigger>
+                          <TooltipContent className="max-w-[300px] text-left text-sm">
+                            <p className="mb-4">
+                              Equals to the total time spent (p95) on this operation in the selected
+                              period in seconds.
+                            </p>
+                            <code>Impact = Requests * p95/1000</code>
```
kamilkisiela, line 274:
```suggestion
                            <p className="mb-4">
                              Equals to the total time spent on this operation in the selected
                              period in seconds.
                            </p>
                            <code>Impact = Requests * avg/1000</code>
```

n1ru4l (approved):

PR #6391, opened 2025-01-18
Standardize CLI Errors

jdolle:
### Background

resolves https://github.com/graphql-hive/console/issues/207

### Description

Created an error class for all possible errors from the CLI.


packages/libraries/cli/src/base-command.ts
```diff
@@ -98,7 +100,7 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
    * @param key
    * @param args all arguments or flags
    * @param defaultValue default value
-   * @param message custom error message in case of no value
```
jdolle, line 101: Instead of a message as an argument, I standardized the message format and if a more custom message or suggested fix would be useful, then capture the error and throw the custom error.

packages/libraries/cli/src/commands/app/create.ts
```diff
@@ -123,7 +132,7 @@ export default class AppCreate extends Command<typeof AppCreate> {
               );
             }
           }
-          this.error(result.addDocumentsToAppDeployment.error.message);
```
jdolle, line 126: instead of using the `this.error` function, each error has an `exitCode` property that sets the exit code.

packages/web/docs/src/pages/docs/api-reference/cli.mdx
```diff
@@ -453,3 +453,15 @@ hive schema:check schema.graphql --github
 
 List of all available CLI commands and their options can be found
 [here](https://github.com/graphql-hive/platform/blob/main/packages/libraries/cli/README.md#commands)
+
```
jdolle, line 456: any recommendations for auto generating this content?

packages/libraries/cli/src/base-command.ts
```diff
@@ -127,38 +139,34 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
...
-    const userConfigValue = this._userConfig!.get(key);
-
-    if (userConfigValue != null) {
-      return userConfigValue;
-    }
+    let value: GetConfigurationValueType<TKey>;
 
-    if (defaultValue) {
-      return defaultValue;
+    if (args[key] != null) {
```
jdolle, line 147: These changes are so that empty strings dont fallback to the default.
I believe empty strings for a required argument indicates there is something wrong, like a variable wasnt set within a script correctly

packages/web/docs/src/pages/docs/api-reference/cli.mdx
kamilkisiela: I guess you could use `export async function getStaticProps() { }` API to render content based on the `packages/libraries/cli/src/helpers/errors.ts` in some way.

jdolle: Im moving the documentation portion to a separate PR and am looking into this. ty

PR #6395, opened 2025-01-20
Docs: Comparison table for different use cases of `@cacheControl`

ardatan:
Adds comparison table related to https://github.com/ardatan/graphql-mesh/pull/8268


jdolle (approved):
Great synopsis and useful reference links. Love it

PR #6400, opened 2025-01-21
Show output of preflight script

### LINKED ISSUE #6229: log output for preflight scripts when executing operation
We only show log output within the preflight script editor, not when executing the preflight script before sending a GraphQL request.

We should have a box with the log output in the area marked on the screenshot, with the ability to toggle the visibility, quickly enable/disable the script, and clear the logs.

Screenshot 2024-12-30 at 10 43 58

See https://github.com/graphql-hive/console/pull/5564

kamilkisiela:
Closes #6229

- [ ] product update

https://github.com/user-attachments/assets/28148db9-f00e-40dc-b1d7-17a03959e91b

### DIFF
--- a/packages/web/app/src/pages/target-laboratory.tsx
+++ b/packages/web/app/src/pages/target-laboratory.tsx
@@ -57,6 +57,7 @@ import 'graphiql/style.css';
 import '@graphiql/plugin-explorer/style.css';
 import { PromptManager, PromptProvider } from '@/components/ui/prompt';
 import { useRedirect } from '@/lib/access/common';
+import { captureException } from '@sentry/react';
 
 const explorer = explorerPlugin();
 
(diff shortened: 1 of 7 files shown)


packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -655,3 +692,108 @@ function useApiTabValueState(graphqlEndpointUrl: string | null) {
...
+              let logMessage = '';
+
+              if (log instanceof Error) {
+                logType = 'error';
+                logMessage = `${log.name}: ${log.message}`;
+              } else if (typeof log === 'string') {
+                logType = log.split(':')[0].toLowerCase() as 'error' | 'warn' | 'info' | 'log';
+                logMessage = log.substring(log.indexOf(':') + 1).trim();
+              } else {
+                return null;
```
jdolle, line 785: maybe json stringify and spit out a warning in this case? And/or send to sentry since I dont think this should ever happen.

kamilkisiela: True, I wanted to turn those string-based lines into structured format in the following PR, but we can log a warning here now

PR #6418, opened 2025-01-24
chore: version updates

jasonkuhrt:
This happened after installing and building the project locally from the current trunk.


dotansimha:
@jasonkuhrt I think the next release (https://github.com/graphql-hive/console/pull/6406/files) will fix that?

PR #6420, opened 2025-01-24
chore(services/emails): fix development watch mode

jasonkuhrt:
When working locally on Emails service I reliably experience tsup being unable to wait for or force exit the previously running process for whatever reason.

CleanShot 2025-01-24 at 10 47 46@2x

When I switch to `tsx watch` I no longer experience this issue.

- This changes dev script for one service, but it would be confusing to only change one of our services. I have not investigated if other services have this issue. Maybe its not needed, if we are already aligned on just switching all to `tsx watch`.

- There is significantly more configuration for our `tsup` development setup which has not been ported over to `tsx` usage. Is it needed? I see some include paths that could be ported to `tsx`. Maybe even those are not needed though.

### DIFF
--- a/packages/services/emails/package.json
+++ b/packages/services/emails/package.json
@@ -5,7 +5,7 @@
   "private": true,
   "scripts": {
     "build": "tsx ../../../scripts/runify.ts",
-    "dev": "tsup-node --config ../../../configs/tsup/dev.config.node.ts src/dev.ts",
+    "dev": "pnpm tsx watch src/dev.ts | pino-pretty --ignore time,pid,hostname",
     "postbuild": "copyfiles -f \"node_modules/bullmq/dist/esm/commands/*.lua\" dist && copyfiles -f \"node_modules/bullmq/dist/esm/commands/includes/*.lua\" dist/includes",
     "typecheck": "tsc --noEmit"
   },


dotansimha (approved):
lgtm, but let's wait for @n1ru4l 's approval, i know he had some local issues with tsup/tsx before?<br><br>edit: let's think about it even more, I think using a different tool means that the dev environment is different than what we deploy.

jasonkuhrt:
> I think using a different tool means that the dev environment is different than what we deploy.

I'm curious to learn more about what this means. I am generally familiar with building TypeScript source code for deployment. Given that recent versions of Node support stripping TypeScript types, one can even begin to imagine TypeScript development with `node` itself.

I will make more effort to learn what we are doing at compile time beyond stripping types and eager to hear others' thoughts.

jasonkuhrt:
@dotansimha I tried "re-request a review..." from you, being not 100% sure if that would restore the state of merging this PR to being blocked. That didn't seem to work. Anyways its just a detail, we're clear, of course, but it would be nice to not have this PR be even technically mergable.

dotansimha:
`tsup-node` is using `node` under the hood, and we are also using it to build the services when preparing the Docker images. Using the same this locally means that we can have some level of trust on this process (if we do the same locally, it's probably going to work the same). 

I think this also related to a broader topic the team has discussed before. We do want better tooling and more modern toolings, but it requires some extra effort to make the validation that it's actually working correctly, and can be used for both development and production runtime.

jasonkuhrt:
fwiw It seems we already have `tsx` in the project but I haven't reviewed for what use-cases.

Am I the first person to hit this process-not-exiting issue?

dotansimha:
> fwiw It seems we already have `tsx` in the project but I haven't reviewed for what use-cases.

Yeah, I think mostly for scripts and things that don't go to production. I also prefer to have it unified, but it must be with a thinking process that also takes under consideration the fact that production env could be affected. 

> Am I the first person to hit this process-not-exiting issue?

I think so? Let's take this async or on Slack and figure out if there's an underlying issue that caused that?

jdolle:
Should we close this or address the issues to improve the dev experience?

dotansimha:
Closed for now, we'll address this if it becomes annoying later?

PR #6449, opened 2025-01-30
allow non target access actors to check, compose, delete and publish schemas

n1ru4l:
### Background

We will soon introduce a new access token not bound to a single target. When we ship that change, users must specify which target they want to act on for those new access tokens.

We want to introduce this pattern early by introducing an optional `--target` flag for the CLI. In the future, we will make this flag mandatory. 

**Example: input target selector**

```bash
hive schema:publish --target the-guild/graphql-hive/production ./my-schema.graphql
```

**Example: target UUID**

```bash
hive schema:publish --target a0f4c605-6541-4350-8cfe-b31f21a4bf80 ./my-schema.graphql
```

**Note:** A follow-up PR will adjust the documentation to use this pattern.

### Description

This change adds passing an optional target slug selector or target UUID to the `Mutation.schemaCheck`, `Mutation.schemaCompose`, `Mutation.schemaDelete`, `Mutation.schemaPublish`, `Mutation.createAppDeployment`, `Mutation.addDocumentsToAppDeployment`, `Mutation.activateAppDeployment`  fields input types and some `Query` fields used by the CLI.

The selector is used to resolve the affected target instead of the target bound to the registry access token (if provided).
If the target resolved by the selector does not match the permissions of the registry access token, a permission error is raised.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/.changeset/wet-eggs-rule.md
+++ b/.changeset/wet-eggs-rule.md
@@ -1,0 +1,30 @@
+---
+'@graphql-hive/cli': minor
+---
+
+Add `--target` flag for commands `app:create`, `app:publish`, `operations:check`, `schema:check`,
+`schema:delete`, `schema:fetch`, `schema:publish` and `dev`.
+
+The `--target` flag can be used to specify the target on which the operation should be performed.
+Either a slug or ID of the target can be provided.
+
+A provided slug must follow the format `$organizationSlug/$projectSlug/$targetSlug` (e.g.
+`the-guild/graphql-hive/staging`).
+
+**Example using target slug**
+
+```bash
+hive schema:publish --target the-guild/graphql-hive/production ./my-schema.graphql
+```
+
+A target id, must be a valid target UUID.
+
+**Example using target id**
+
+```bash
+hive schema:publish --target a0f4c605-6541-4350-8cfe-b31f21a4bf80 ./my-schema.graphql
+```
+
+**Note:** We encourage starting to use the `--target` flag today. In the future the flag will become
+mandatory as we are moving to a more flexible approach of access tokens that can be granted access
+to multiple targets.
--- a/integration-tests/tests/cli/schema.spec.ts
+++ b/integration-tests/tests/cli/schema.spec.ts
@@ -357,7 +357,7 @@ describe.each([ProjectType.Stitching, ProjectType.Federation, ProjectType.Single
 );
 
 test.concurrent(
-  'schema:publish with --target parameter matching the access token',
+  'schema:publish with --target parameter matching the access token (slug)',
   async ({ expect }) => {
     const { createOrg } = await initSeed().createOwner();
     const { inviteAndJoinMember, createProject, organization } = await createOrg();
@@ -388,7 +388,36 @@ test.concurrent(
 );
 
 test.concurrent(
-  'schema:publish fails with --target parameter not matching the access token',
+  'schema:publish with --target parameter matching the access token (UUID)',
+  async ({ expect }) => {
+    const { createOrg } = await initSeed().createOwner();
+    const { inviteAndJoinMember, createProject, organization } = await createOrg();
+    await inviteAndJoinMember();
+    const { createTargetAccessToken, project, target } = await createProject();
+    const { secret } = await createTargetAccessToken({});
+
+    await expect(
+      schemaPublish([
+        '--registry.accessToken',
+        secret,
+        '--author',
+        'Kamil',
+        '--target',
+        target.id,
+        'fixtures/init-schema.graphql',
+      ]),
+    ).resolves.toMatchInlineSnapshot(`
+      :::::::::::::::: CLI SUCCESS OUTPUT :::::::::::::::::
+
+      stdout--------------------------------------------:
+      ✔ Published initial schema.
+      ℹ Available at http://__URL__
+    `);
+  },
+);
+
+test.concurrent(
+  'schema:publish fails with --target parameter not matching the access token (slug)',
   async ({ expect }) => {
     const { createOrg } = await initSeed().createOwner();
     const { inviteAndJoinMember, createProject } = await createOrg();
@@ -413,7 +442,7 @@ test.concurrent(
       exitCode------------------------------------------:
       2
       stderr--------------------------------------------:
-       ›   Error: No access (reason: "Missing permission for performing 
+       ›   Error: No access (reason: "Missing permission for performing
        ›   'schemaVersion:publish' on resource")  (Request ID: __REQUEST_ID__)  [115]
        ›   > See https://__URL__ for
        ›    a complete list of error codes and recommended fixes.
--- a/integration-tests/tests/cli/schema.spec.ts
+++ b/integration-tests/tests/cli/schema.spec.ts
@@ -328,7 +328,7 @@ describe.each([ProjectType.Stitching, ProjectType.Federation, ProjectType.Single
         const { inviteAndJoinMember, createProject } = await createOrg();
         await inviteAndJoinMember();
         const { createTargetAccessToken } = await createProject(projectType);
-        const { secret, latestSchema } = await createTargetAccessToken({});
+        const { secret } = await createTargetAccessToken({});
 
         const cli = createCLI({
           readonly: secret,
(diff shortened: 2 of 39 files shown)


n1ru4l:
@coderabbitai review

integration-tests/tests/cli/schema.spec.ts
```diff
@@ -355,3 +355,72 @@ describe.each([ProjectType.Stitching, ProjectType.Federation, ProjectType.Single
...
+      stderr--------------------------------------------:
+       ›   Error: No access (reason: "Missing permission for performing 
+       ›   'schemaVersion:publish' on resource")  (Request ID: __REQUEST_ID__)  [115]
+       ›   > See https://__URL__ for
+       ›    a complete list of error codes and recommended fixes.
+       ›   To disable this message set HIVE_NO_ERROR_TIP=1
+       ›   Reference: __ID__
+      stdout--------------------------------------------:
+      __NONE__
+    `);
```
n1ru4l, line 424: @jdolle Is there a difference between `Reference: __ID__` and `__REQUEST_ID__`? Do you see the need to introduce a special error code for permission errors?

jdolle: I don't see a big difference in the id types. I think it was convenient to match on the word prior but they are all IDs.

And if this error is returned from our api then technically no need. But i did envision standardized api errors in the future and i did add unique errors for authorization which obviously is from the api...

n1ru4l: Oh I understand now, Reference is the error code, Request ID is the actual request id 😓

packages/libraries/cli/src/commands/app/create.ts
```diff
@@ -66,6 +72,15 @@ export default class AppCreate extends Command<typeof AppCreate> {
...
     }
 
+    let target: GraphQLSchema.TargetSelectorInput | null = null;
+    if (flags.target) {
+      const result = TargetSlug.parse(flags.target);
+      if (result.type === 'error') {
+        throw new InvalidTargetError();
+      }
+      target = result.data;
+    }
```
n1ru4l, line 82: Not sure if already too aggressive: We could start logging a warning here.

kamilkisiela: `process.exit(1)` to make sure people see it

jasonkuhrt:
> We will soon introduce a new access token not bound to a single target. When we ship that change, users must specify which target they want to act on for those new access tokens.

Few questions, maybe already documented, if not, maybe things to clarify in docs at some point:

1. Can a token be scoped to multiple targets?
2. Can a token be scoped to "all current targets"?
3. Can a token be scoped to "all current + future targets"?

.changeset/wet-eggs-rule.md
```diff
@@ -0,0 +1,30 @@
...
+```bash
+hive schema:publish --target the-guild/graphql-hive/production ./my-schema.graphql
+```
+
+A target id, must be a valid target UUID.
+
+**Example using target id**
+
+```bash
+hive schema:publish --target a0f4c605-6541-4350-8cfe-b31f21a4bf80 ./my-schema.graphql
```
dotansimha, line 25: side-node: should we add an issue to improve how the dashboard shows IDs to the user? 🤔

.changeset/wet-eggs-rule.md
```diff
@@ -0,0 +1,30 @@
...
+The `--target` flag can be used to specify the target on which the operation should be performed.
+Either a slug or ID of the target can be provided.
+
+A provided slug must follow the format `$organizationSlug/$projectSlug/$targetSlug` (e.g.
+`the-guild/graphql-hive/staging`).
+
+**Example using target slug**
+
+```bash
+hive schema:publish --target the-guild/graphql-hive/production ./my-schema.graphql
```
dotansimha, line 17: nice!

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -267,40 +264,53 @@ export class SchemaPublisher {
   async check(input: CheckInput) {
     this.logger.info('Checking schema (input=%o)', lodash.omit(input, ['sdl']));
 
+    const selector = await this.idTranslator.resolveTargetReference({
+      reference: input.target ?? null,
+      onError() {
+        throw new InsufficientPermissionError('schemaCheck:create');
+      },
+    });
```
jasonkuhrt, line 272: I see we have three terms: `selector`, `target`, `reference`. Would it make more sense if we consolidated the naming some? E.g.

- `Target` / `Resolve Target` / `Fully Qualified Resource Name` (fqrn)
- `Target` / `Resolve Target` / `Fully Qualified Resource Target` (fqrt)
- `Target Expression` / `Resolve Target Expression` / `Target`

The `fqrn`/`fqrt` idea rifts off of the well known fqdm (fully qualified domain name) terminology. And it reflects me wondering if in our system `selector` is/can be generalized to being just an explicit resource target data structure which potentially shows up in other places too.

jasonkuhrt (approved):
I like the approach of introducing the feature before its breaking variant.

Aside:

Have we considered an approach based on future flags? Popularized maybe by Remix. Could be a pattern we use repeatedly. For example introduce a `--future` flag that when present, enforces everything coming in the next breaking change. Two issues I see now with this: overkill until we have a stable release of CLI (>0 major); expecting up hill struggle to bend OClif toward this pattern.

.changeset/wet-eggs-rule.md
jdolle: I think that's a good idea. Also having a set of commands displayed for publishing your first schema would be helpful:
Screenshot 2025-02-03 at 8 40 50 AM

packages/libraries/cli/src/helpers/errors.ts
```diff
@@ -137,6 +137,16 @@ export class MissingCdnEndpointError extends HiveCLIError {
   }
 }
 
+export class InvalidTargetError extends HiveCLIError {
+  constructor() {
+    super(
+      ExitCode.BAD_INIT,
+      errorCode(ErrorCategory.GENERIC, 9),
```
jdolle, line 144: If this is a generic error (meaning it applies to multiple commands) then it'll be number 20. `errorCode(ErrorCategory.GENERIC, 20),` https://the-guild.dev/graphql/hive/docs/api-reference/cli#errors-119

packages/libraries/cli/src/helpers/target-input.ts
```diff
@@ -0,0 +1,52 @@
...
+        byId: uuidResult.data,
+      },
+    };
+  }
+
+  const parts = str.split('/');
+
+  const organizationSlug = parts.at(0);
+  const projectSlug = parts.at(1);
+  const targetSlug = parts.at(2);
```
jdolle, line 34: i think we should validate `parts.length === 3` to make sure there isnt confusion with additional `/stuff`

packages/services/api/src/modules/auth/lib/authz.ts
```diff
@@ -185,7 +185,7 @@ export abstract class Session {
               args.organizationId,
               args.params,
             );
-            throw new AccessError(`Missing permission for performing '${args.action}' on resource`);
+            throw new InsufficientPermissionError(args.action);
```
jdolle, line 188: :+1: make sure to update the documentation if we're replacing this error with another.

packages/services/api/src/modules/schema/providers/schema-manager.ts
```diff
@@ -99,46 +102,60 @@ export class SchemaManager {
...
+      name: string;
+    }>;
+    target: GraphQLSchema.TargetReferenceInput | null;
+  }) {
     this.logger.debug('Composing schemas (input=%o)', lodash.omit(input, 'services'));
+
+    const selector = await this.idTranslator.resolveTargetReference({
+      reference: input.target ?? null,
+      onError() {
+        throw new InsufficientPermissionError('schema:compose');
```
jdolle, line 127: Could other errors happen?

integration-tests/tests/cli/schema.spec.ts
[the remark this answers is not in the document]
n1ru4l: I am not a big fan of having snapshots not being inline in the test file, but since we use matrixes there seems to be no way around that aside from deduplication. Aside from that, I like this approach for the CLI!

packages/services/api/src/modules/auth/lib/authz.ts
n1ru4l: It is still the same error, this just allows to instantiate it somewhere else

packages/services/api/src/modules/schema/providers/schema-manager.ts
n1ru4l: Internally a few things can happen, but we catch these log them, and then treat all of those as "InsufficientPermissionsError" to the requester.  E.g. if someone provides `99999` as the ID, we don't throw a "please send an UUID" error, but this one instead.

packages/services/api/src/modules/schema/resolvers/Mutation/schemaPublish.ts
```diff
@@ -27,9 +18,6 @@ export const schemaPublish: NonNullable<MutationResolvers['schemaPublish']> = as
     {
       ...input,
       service: input.service?.toLowerCase(),
-      organizationId: organization,
-      projectId: project,
-      targetId: target,
```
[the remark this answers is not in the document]
n1ru4l: lean resolvers ftw

packages/services/api/src/modules/schema/resolvers/Query/latestValidVersion.ts
```diff
@@ -1,13 +1,32 @@
...
+  args,
+  { injector, session },
 ) => {
-  const target = await injector.get(TargetManager).getTargetFromToken();
+  const { targetId, projectId, organizationId } = await injector
+    .get(IdTranslator)
+    .resolveTargetReference({
+      reference: args.target ?? null,
+      onError() {
+        throw new InsufficientPermissionError('project:describe');
```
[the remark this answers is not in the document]
n1ru4l: same as https://github.com/graphql-hive/console/pull/6449/files/de44df447600e1e547052175cb6056c9eb866878#r1940877678

packages/libraries/cli/src/helpers/target-input.ts
n1ru4l: 116575b28 🫡

n1ru4l:
> > We will soon introduce a new access token not bound to a single target. When we ship that change, users must specify which target they want to act on for those new access tokens.
> 
> Few questions, maybe already documented, if not, maybe things to clarify in docs at some point:
> 
> 1. Can a token be scoped to multiple targets?
> 2. Can a token be scoped to "all current targets"?
> 3. Can a token be scoped to "all current + future targets"?

Theses seem to be questions we will answer in the product update when we release the new access tokens and deprecate the old ones 🤔

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: Currenlty, we have the following:
**Target**: A target resource as stored within our database in the `targets` table
**Selector**: An object containing the id breadcrumb of a resource and its parent resources

New in this PR
**Reference**: Input from GraphQL either slug or UUID (see `TargetReferenceInput` input type), that we translate into a `Selector`.

.changeset/wet-eggs-rule.md
n1ru4l: also created https://github.com/graphql-hive/console/issues/6466

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: @jasonkuhrt Let's move this conversation to another issue to discuss and see if it makes sense!

packages/libraries/cli/src/commands/app/create.ts
```diff
@@ -28,6 +31,9 @@ export default class AppCreate extends Command<typeof AppCreate> {
       description: 'app version',
       required: true,
     }),
+    target: Flags.string({
+      description: 'The target in which the app deployment will be created (slug or ID).',
```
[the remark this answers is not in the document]
n1ru4l: [`f881812` (#6449)](https://github.com/graphql-hive/console/pull/6449/commits/f8818122c255572b03333b9bd32623afec32fbba) ✅

packages/services/api/src/modules/schema/providers/schema-publisher.ts
jasonkuhrt: Thanks for the details!

.changeset/wet-eggs-rule.md
```diff
@@ -0,0 +1,30 @@
...
+---
+
+Add `--target` flag for commands `app:create`, `app:publish`, `operations:check`, `schema:check`,
+`schema:delete`, `schema:fetch`, `schema:publish` and `dev`.
+
+The `--target` flag can be used to specify the target on which the operation should be performed.
+Either a slug or ID of the target can be provided.
+
+A provided slug must follow the format `$organizationSlug/$projectSlug/$targetSlug` (e.g.
+`the-guild/graphql-hive/staging`).
```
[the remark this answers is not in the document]
n1ru4l: The current registry tokens have this. However, we want to deprecate those in favor of access tokens on the organization level, that will require people to provide a target for the CLI oeprations.

PR #6450, opened 2025-01-30
fix(laboratory): Preflight Script environment variables no parse storage read

jasonkuhrt:
It would be nice to refactor our hooks that interact with local storage. I briefly explored that but it led to a larger diff. This technically solves the problem. A refactor could be done later.

Failing test shipped with this PR, now passes:
CleanShot 2025-01-30 at 14 01 36@2x


jasonkuhrt:
The AI's suggestion would bring back the bug. I am ignoring it:

CleanShot 2025-01-30 at 17 14 28@2x

Given it got one thing wrong, I am ignoring the the other thing too, I'm not sure its correct. That's how our code was before, and it worked AFAIK.

jasonkuhrt:
@dotansimha can we please turn of this AI thing.

packages/web/app/src/lib/hooks/use-local-storage-json.ts
```diff
@@ -0,0 +1,23 @@
+import { useCallback, useState } from 'react';
+
+export function useLocalStorageJson<T>(key: string, defaultValue: T) {
+  const [value, setValue] = useState<T>(() => {
+    const json = localStorage.getItem(key);
+    try {
+      const result = json ? JSON.parse(json) : defaultValue;
```
n1ru4l, line 7: We can also allow providing a zod model here, what do you think?

jasonkuhrt: That would be great. I have a git stash with that work started as a matter of fact. I would be happy to finish it. While working I realized that `useLocalStorageJson` is outside the path of the error and so took a technically simpler solution.

How about I put up a refactor PR for `useLocalStorageJson` after we merge this one?

n1ru4l: sounds good!

jasonkuhrt:
@n1ru4l I just realized I didn't include a changeset for this. Its a minor thing but I would still prefer to include it next time. If you think there definitely shouldn't have been a changeset for this PR though, let me know, because then I might be confused about something.

PR #6465, opened 2025-02-04
:memo: Fix self-hosted changelog link

arnaud:
### Background

A README link was unfairly mispelled.


n1ru4l:
@arnaud thank you!

PR #6470, opened 2025-02-04
refactor(laboratory): rename preflight-script to just preflight

jasonkuhrt:
### Background

I am going to be working on some more preflight tasks, and saw an opportunity to simplify and make consistent our references to it.

### Description

For the most part "Preflight Script" is used but not always:

- preflight-sandbox
- preflight-script-worker vs preflight-worker-emebd

Changes in this PR suggest a rename towards proper noun of just "Preflight" instead of descriptive title "preflight script".

Some areas of code are not changed by this PR:

- user facing app copy (product discussion that can be in a future PR rather than this one, which is just technical implementation)
- local storage key names (needs some more code to make it a seamless migration)
- api schema (Could I change this?)

There are some other bits of code not yet updated, outside the immediate scope of web app code, like the inject provder stuff. I'll tackle that if appropriate if there is team alignment about this PR.


cypress/e2e/laboratory-preflight-script.cy.ts
```diff
@@ -1,7 +1,7 @@
 import { dedent } from '../support/testkit';
 
 const selectors = {
-  buttonPreflightScript: '[aria-label*="Preflight Script"]',
+  buttonPreflight: '[aria-label*="Preflight Script"]',
```
jdolle, line 4: Should we change the label value also?

jdolle (approved):
subtle, but i like these changes.

packages/web/app/src/lib/preflight/graphiql-plugin.tsx
```diff
@@ -130,7 +127,7 @@ const UpdatePreflightScriptMutation = graphql(`
   }
 `);
 
-const PreflightScript_TargetFragment = graphql(`
+const Preflight_TargetFragment = graphql(`
   fragment PreflightScript_TargetFragment on Target {
```
n1ru4l, line 131: Please keep the variable name consistent with the fragment name

cypress/e2e/laboratory-preflight-script.cy.ts
jasonkuhrt: After looking into this, its coming from GraphiQL and it derives it from the given title  I think. I think changing that would enter into that product realm of something users see/could see, so will leave this as is for now.

jasonkuhrt:
The suggestion to change how stack is hidden is off topic because that did not change in this PR and it is not related to updating the identifier of preflight script.

jasonkuhrt:
There was a e2e failure, I lost some time trying to get e2e running locally against development services. Upon coming back here for more detail I noticed that the e2e failure cause might not be me, but flake due to I don't know what.

CleanShot 2025-02-05 at 12 34 02@2x

Another test run did make it pass.

jasonkuhrt:
It would have been useful to have that feedback before. Feel free to revert the commit.

n1ru4l:
@jasonkuhrt I definitely see value in unifying the usages of sandbox to `preflight-script`. This PR does not need to be reverted for me, it feels like it just got merged too fast without me having the time to have a proper look and give feedback.

jasonkuhrt:
@laurin understood. I probably would have waited had there been no comment from you but given there was, it wasn't clear if that was "it", or there would be more, or… etc. To help avoid this in the future, would it be possible to make yourself a required reviewer on this repository? Anything to automate our process standards really helps me because even with the best of intentions it can be hard for me to track each detail of our process. Sorry for the surprise this time around!

PR #6475, opened 2025-02-05
chore(tests/e2e): script to run locally

jasonkuhrt:
This PR aims to make running e2e tests locally easy by having a script
that encodes how to do it.


cypress.config.ts
```diff
@@ -8,6 +8,14 @@ if (!process.env.RUN_AGAINST_LOCAL_SERVICES) {
   dotenv.config({ path: import.meta.dirname + '/integration-tests/.env' });
 }
 
+if (process.env.RUN_AGAINST_LOCAL_SERVICES === '1') {
+  process.env.SUPERTOKENS_API_KEY = process.env.SUPERTOKENS_API_KEY ?? 'bubatzbieber6942096420';
```
dotansimha, line 12: can we read these from `packages/services/server/.env.template` instead of having it hardcoded?  same for the url

dotansimha (changes requested):

PR #6476, opened 2025-02-05
improve(laboratory): validate editor content with TypeScript

jasonkuhrt:
### Background

This brings TypeScript validation to the laboratory editor so that users have a better experience writing code, because the compiler will catch more of their mistakes for them.

### Description

Monaco React has built in support for this so not much doing was needed.

### DIFF
--- a/packages/web/app/src/lib/preflight/graphiql-plugin.tsx
+++ b/packages/web/app/src/lib/preflight/graphiql-plugin.tsx
@@ -104,6 +104,12 @@ const monacoProps = {
     defaultLanguage: 'javascript',
     options: {
       ...sharedMonacoProps.options,
+      quickSuggestions: true,
+      suggestOnTriggerCharacters: true,
+      acceptSuggestionOnEnter: 'on',
+      tabCompletion: 'on',
+      folding: true,
+      foldingStrategy: 'indentation',
     },
   },
 } satisfies Record<'script' | 'env', ComponentPropsWithoutRef<typeof MonacoEditor>>;
@@ -639,6 +645,29 @@ function PreflightModal({
   }, []);
 
   const handleMonacoEditorBeforeMount = useCallback((monaco: Monaco) => {
+    // Configure JavaScript defaults for TypeScript validation
+    monaco.languages.typescript.javascriptDefaults.setDiagnosticsOptions({
+      noSemanticValidation: false,
+      noSyntaxValidation: false,
+      noSuggestionDiagnostics: false,
+      diagnosticCodesToIgnore: [], // Can specify codes to ignore
+    });
+
+    // Enable modern JavaScript features and strict checks
+    monaco.languages.typescript.javascriptDefaults.setCompilerOptions({
+      target: monaco.languages.typescript.ScriptTarget.ESNext,
+      allowNonTsExtensions: true,
+      moduleResolution: monaco.languages.typescript.ModuleResolutionKind.NodeJs,
+      module: monaco.languages.typescript.ModuleKind.ESNext,
+      noEmit: true,
+      lib: ['es2021', 'dom'],
+      strict: true,
+      noUnusedLocals: true,
+      noUnusedParameters: true,
+      noImplicitReturns: true,
+      noFallthroughCasesInSwitch: true,
+    });
+
     // Add custom typings for globalThis
     monaco.languages.typescript.javascriptDefaults.addExtraLib(
       `


jasonkuhrt:
I will write a test case or two for this tomorrow.

cypress.config.ts
```diff
@@ -8,6 +8,14 @@ if (!process.env.RUN_AGAINST_LOCAL_SERVICES) {
...
 }
 
+if (process.env.RUN_AGAINST_LOCAL_SERVICES === '1') {
+  process.env.SUPERTOKENS_API_KEY = process.env.SUPERTOKENS_API_KEY ?? 'bubatzbieber6942096420';
+  process.env.SUPERTOKENS_CONNECTION_URI =
+    process.env.SUPERTOKENS_CONNECTION_URI ?? 'http://localhost:3567';
+  // It seems that this has to be set in the environment that the cypress cli is executed from.
+  // process.env.CYPRESS_BASE_URL = process.env.CYPRESS_BASE_URL ?? 'http://localhost:3000';
+}
+
```
jasonkuhrt, line 18: Remove this (and package script) and rebase once #6475 is merged.

cypress/e2e/laboratory/_cy.ts
jasonkuhrt, line 1: I tried creating a laboratory domain of its sub sections and co-located the Cypress helpers.

I can undo this change if its out of scope or worse than the previous setup.

cypress/support/dedent.ts
jasonkuhrt, line 1: Moved this to its own module since its just its own lib.

cypress/support/monaco.ts
jasonkuhrt, line 1: Created a monaco lib to keep this generic utility away from specific monaco instances.

cypress/e2e/laboratory/preflight.cy.ts
```diff
@@ -85,8 +51,22 @@ describe('Preflight Script Modal', () => {
...
+    cyp.setEditorContent(script);
+    cy.wait(1000); // :(
+    cy.dataCy(s.modal.editorCy)
+      .find('textarea')
+      .focus()
+      // Followed instructions but does not work https://github.com/dmtrKovalenko/cypress-real-events?tab=readme-ov-file#installation
+      // @ts-expect-error
+      .realPress(['Alt', 'F8']);
+    cy.contains(tsErrorMessage);
+  });
```
jasonkuhrt, line 66: Given the noisy diff caused by refactoring, highlighting that here is the new test for this feature.

cypress/e2e/laboratory/preflight.cy.ts
```diff
@@ -85,8 +51,21 @@ describe('Preflight Script Modal', () => {
...
+    const script = 'let a = 1; a = ""';
+    cyp.setEditorContent(script);
+    cy.wait(1000); // :(
+    cy.dataCy(s.modal.editorCy)
+      .find('textarea')
+      .focus()
+      // @ts-expect-error - Followed instructions but does not work https://github.com/dmtrKovalenko/cypress-real-events?tab=readme-ov-file#installation
+      .realPress(['Alt', 'F8']);
+    cy.contains(tsErrorMessage);
+  });
```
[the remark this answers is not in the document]
jasonkuhrt: I tried this but it doens't work. Getting this to work in Cypress required hacking around. I'll leave a comment making the point that we know its hacky.

n1ru4l:
I know this was not introduced in this PR but would it be easily possible to disable these suggestions and limit it to the APIs available within the preflight script context?

image

n1ru4l:
The editor does not flag TypeScript annotation as an error, is it possible to support typescript linting without allowing type annotations/TypeScript specific syntax?

Otherwise, we must go down the rabbit hole of transpiling the code.

jasonkuhrt:
Great point, I'll look into this to see what our options are.

jasonkuhrt:
> I know this was not introduced in this PR but would it be easily possible to disable these suggestions and limit it to the APIs available within the preflight script context?

@n1ru4l oh yeah, currently `Dom` has been added, which is of course a large set of global types. What triggered me adding it was the fact that `console` was giving a type error which is allowed in Preflight. Do we have documented somewhere what the allowed globals are in Preflight? Without knowing that I'm just guessing about what types to allow.

n1ru4l:
packages/web/app/src/lib/preflight/allowed-globals.ts is the file you are looking for. I hope there is a way for adding these without having to pluck this file together 😢

jasonkuhrt:
@n1ru4l awesome thanks. My hope is we'll just need to list  them in the tsconfig lib settings given to Monaco, and that there is a 1:1 between dom lib names and those globals. tbd...

jasonkuhrt:
Hey @n1ru4l, I made some progress.

1. TS syntax is no longer allowed statically. ✅ 

2. Custom globals. This is WIP. We don't get the granular dom lib types I had for some reason hoped for but should have known wouldn't be there. So I've had to go brute force on this. Luckily, most of the types IIUC in your list are not from dom but ES standard. I have therefore included all the ES2020 types. If that's not acceptable then we're going to have to do the same thing for that which I'm doing for DOM types, which is tricky/error prone IMO so not great.

  For the DOM I've finished console and setTimeout as proof of concept. Fetch is not done yet, and that is the huge one, because of all the stream and http types it drags in. It is many hundreds of lines of type code that we'll need to copy paste in, and I don't think its just one contiguous block of type code either, but I have to go and trace the type paths to cherry pick just whats needed. I feel this is something that the community may have dealt with before so will pause to look for external solutions, e.g. a types package that already slimmed down to just fetch.

  Another idea is that there is a package `@types/web`, and I might be able to import and then type a manual global type with just what we want. That would be ideal, however, unless we can put the declaration file through some kind of tree shaking bundle step, we would be pulling in the whole of web declaration into the preflight bundle which must be MBs big... Maybe we could serve that on an agressively cached cdn to mitigate the issue... or...

  I have that feeling of there "must be a simpler way", maybe I'll think of something over the weekend or have some new energy next week that better thinks about this.
  
  Welcome any ideas from ya'll about this!

CleanShot 2025-02-07 at 12 30 54@2x

CleanShot 2025-02-07 at 12 31 02@2x

n1ru4l (changes requested):
It seems like this PR does a lot of things aside from adding TypeScript validation to the editor. Please adjust the PR to only reflect this and move other changes to another PR for discussion.

packages/web/app/src/lib/MonacoEditorReact/index.ts
```diff
@@ -0,0 +1,3 @@
+import * as MonacoEditorReact from '@monaco-editor/react';
+
+export { MonacoEditorReact };
```
kamilkisiela, line 3: What's the purpose of it? Why to re-export this?

packages/web/app/src/lib/preflight/components/EditorTitle.tsx
```diff
@@ -0,0 +1,10 @@
+import { ReactNode } from 'react';
+import { cn } from '@/lib/utils';
+
+export function EditorTitle(props: { children: ReactNode; className?: string }) {
```
kamilkisiela, line 4: Can't we just do `shared-components.ts` instead of creating 10 files where each contains 5LOC.
If we ever need to rename the component, we will have to rename a file, we will loose a diff if not done carefully etc.

I see only downsides.

packages/web/app/src/lib/preflight/components/PreflightModal.tsx
```diff
@@ -0,0 +1,305 @@
...
+    monaco.languages.typescript.javascriptDefaults.addExtraLib(
+      `
+        /// <reference lib="es2020" />
+
+        ${labApiDefinitionRaw}
+        declare const lab: LabAPI;
+
+
+        // ------------------------------------------------------------------------------------------------
+        // The following declarations are taken from:
```
kamilkisiela, line 97: Let's move it to a file, just like the LabAPI, so we could format it, see what it defines easier, now it's just string that is super hard to read.

packages/web/app/src/lib/preflight/shared-types.ts
```diff
@@ -2,6 +2,14 @@
 
 import { Kit } from '../kit';
 
+// todo stop using enums
```
kamilkisiela, line 5: why are enums bad?

jasonkuhrt: Its non-standard ES syntax and less type safe than unions in various situations.

Its a pretty well documented topic so maybe have a search about it, or even ask AI. If you find reasons to still use them, happy to discuss.

jasonkuhrt: Per @n1ru4l's request this will be moved to another PR for discussion.

jasonkuhrt:
@n1ru4l I was consistently getting lost in the massive graphiql module so tried to refactor that. Now yeah the PR has many changes. Once I achieve the goal of this PR, I'll re-work it into the minimal diff and create a new PR for the refactor parts. I'll revert this PR into a draft state until that happens.

cypress/support/dedent.ts
```diff
@@ -0,0 +1,52 @@
+export function dedent(strings: TemplateStringsArray, ...values: unknown[]): string {
```
kamilkisiela, line 1: What's the benefit of having `dedent.ts` that exports `dedent` function?
Why not just have `libs.ts` or `shared.ts` or similar?

kamilkisiela: Now when we change it's name to `dedentString` we rename a file

jasonkuhrt: I find large modules that accumulate dozens of imports and thousands of lines that span various concerns difficult to cognitively navigate. Since the contents of these files change significantly more often than their file names, the readability gains have outweighed the diff loss for me.

kamilkisiela: To me, a file with 500LOC and 20 functions is fine

kamilkisiela: the change of content and 10 imports is not something I can't handle

kamilkisiela: jumping between many small files just to see what they do, I find annoying

jasonkuhrt: Oh another reason for me to keep dedent separated is that it is vendor code rather than our code. A comment there states that the code was brought in for module error reasons. In that case, it would be more clear for that reason alone to me to keep it as a separate package, like it was before being copy and pasted into our project.

Since this is your codebase and you find my style annoying, I will try to undo it. However, you will find my changes less annoying if you provide more details about your requirements, documented in notion or whatever, because this is not the first time we discuss this now and its not my intention to annoy you. It is annoying for me as well to lack confidence in what work of mine will not be reverted in the end. So right now this is lose/lose.

kamilkisiela: We don't need to have a style guide, because it's not like we use X, but not Y, everywhere, always and that's the rule. We had 200 eslint rules, because somebody wanted to enforce patterns, and it did not work, was blocking progress, because the amount of time we spent on fighting the missing object destruction, was too annoying.
In general, we just start simple, see when it breaks, we do not try to predict future (imo it's always a failure, no matter how long you are on this world) and improve over time.

We had so many refactors, especially on the frontend that having to rename files was just super hard to track and most of the time the git history for a file, was gone. Maybe it did not happen to you, but it did to me, a few times in past 2 years.

You say that the content of the file changes more frequently that the file names. That's your assumption and based on my experience from this codebase, it's not true, not for frontend and not for tests.

Regarding the style. When I enter a new codebase, I look how people write it. You did see a chunk of the codebase already, and you can tell what the style is (I assume, because you worked on the CLI and probably looked at other pieces).
You see how Laboratory is written, yes it's not the best thing in the world, but not because it has 1 file instead of 5, or 3 components instead of 7, but rather the logic is hard to follow.

kamilkisiela: I don't mind refactors when they are necessary (like when you can't achieve something and it's blocking a new feature). Even when they are nice-to-have and we already delivered something to customers, but when we discuss it.

jasonkuhrt: Thanks for all your insights on this project. You describe thresholds that I see crossed and you don't. For example the graphiql module's range of concerns is a some kind of blocking like factor for me. most things aren't actually blocking, we're talking about what it takes for a human to develop sustainably. What I see ahead is an approach leaning on code ownership rather than principals, which for me personally is probably the more confusing way to work.<br><br>In regards to to asking for permission to refactor things, I've taken note of that I understand the point. One downside for me is that there's need and opportunity for improvement in this project and a lot of that doesn't cross a need-threshold that matches the energy I have to summon to negotiate permission, especially in an async multi-timezone setup.

kamilkisiela: To discuss it first, I mean, show it to the group to see if people like or not, or they see it necessary, or not. You mentions guidelines and rules, but then you have issues discussing something or to get an approval, to me it's conflicting.

jasonkuhrt: I mention a downside I see to permission processes which is a kind of group level social organizing rule in lieu of craft principals or rules that would obviate need for permission process in some degree of cases. I see no contradiction, hopefully that's clear.

packages/web/app/src/lib/preflight/components/PreflightModal.tsx
jasonkuhrt: Agreed upon solution of sing webworker lib removes need for this code in the first place now. 🎉

jasonkuhrt:
@n1ru4l Please let me know if refactors to the Cypress tests should also be undone in this PR, or if undoing just the app refactor is enough.

jasonkuhrt:
I just found a bug while writing the changeset. After saving, upon re-opening, Monaco seems to retain previous validation state such that the content will be seen as, for example, duplicating identifiers.

I will write a test for this and fix it.

CleanShot 2025-02-12 at 10 25 21@2x

jasonkuhrt:
This may be related https://github.com/suren-atoyan/monaco-react/issues/517

cypress/support/dedent.ts
jdolle: For this specific instance, perhaps a utils directory with an index file that exports the smaller files would meet both your needs.

cypress/support/monaco.ts
```diff
@@ -0,0 +1,34 @@
...
+      const editor = win.monaco.editor.getEditors().find(e => {
+        const parentElement = e.getContainerDomNode().parentElement;
+        return Cypress.$(parentElement).is(editorSelector);
+      });
+
+      // If Monaco instance is found
+      if (editor) {
+        editor.setValue(text);
+      } else {
+        throw new Error('Monaco editor not found on the window or frames[0]');
```
jdolle, line 19: `frames[0]`?
Im curious what the expected outcome of this error is. Is it something we can recover from?

cypress/support/dedent.ts
```diff
@@ -0,0 +1,52 @@
...
+    result = lines.map(l => (l[0] === ' ' || l[0] === '\t' ? l.slice(m) : l)).join('\n');
+  }
+
+  // dedent eats leading and trailing whitespace too
+  result = result.trim();
+  // handle escaped newlines at the end to ensure they don't get stripped too
+  result = result.replace(/\\n/g, '\n');
+
+  return result;
+}
```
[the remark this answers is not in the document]
jasonkuhrt: To be clear, I didn't add it to the codebase, just moved it to its own module. My understanding was that this code was copy-pasted from an external package for incidental reasons. In that case I would say it should not be unit tested since its code we did not write and presumably do not want to change. If we start to change it, and thus "own" it, then I think unit tests could make sense.

cypress/support/monaco.ts
jasonkuhrt: To be clear this was on trunk as well [https://github.com/graphql-hive/console/blob/2507a9b9e07f6ae850f447b6aa90cb4d7cad8f4f/cypress/e2e/laboratory-preflight.cy.ts#L49](https://github.com/graphql-hive/console/blob/2507a9b9e07f6ae850f447b6aa90cb4d7cad8f4f/cypress/e2e/laboratory-preflight.cy.ts#L49).<br><br>I never noted that but now that you mention I'm curious too 🙂. I suspect frames\[0\] is referring to iframes. In regards to what can be done about it, looks like a fatal error to me that would imply our editor setup in the app is somehow broken.<br><br>Note that once we have [CONSOLE-1065](https://linear.app/the-guild/issue/CONSOLE-1065/replace-cypress-with-playwright) this code goes away because Playwright is able to enter text in ways that Cypress fundamentally cannot and thus resorts to hacks of looking for editor instances on window for programatic manipulation.

PR #6477, opened 2025-02-06
fix(cdn-worker): improve logging and ignore analytics engine errors

n1ru4l:
### Description

Cloudflare analytics only allows [collecting a total of 25 data points per request](https://developers.cloudflare.com/analytics/analytics-engine/limits/). Any additional calls will throw an error (`Analytics Engine write limit exceeded.`). 

**Note:** This adds a simple try/catch statement to prevent this error. We should probably keep track of the calls anyways and then log that we are skipping them.

Fixes https://github.com/graphql-hive/console/issues/6479 

Ref CONSOLE-1006

### DIFF
--- a/packages/services/cdn-worker/src/analytics.ts
+++ b/packages/services/cdn-worker/src/analytics.ts
@@ -89,54 +89,59 @@ export function createAnalytics(
         return;
       }
 
-      switch (event.type) {
-        case 'artifact':
-          return engines.usage.writeDataPoint({
-            blobs: [event.version, event.value, targetId],
-            indexes: [targetId.substring(0, 32)],
-          });
-        case 'error':
-          return engines.error.writeDataPoint({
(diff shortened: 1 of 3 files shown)


packages/services/cdn-worker/src/analytics.ts
```diff
@@ -89,54 +89,59 @@ export function createAnalytics(
...
+                  indexes: [targetId.substring(0, 32)],
+                });
+              case 's3-key-validation':
+                return engines.keyValidation.writeDataPoint({
+                  blobs: ['s3-key-validation', event.value.version, event.value.status],
+                  indexes: [targetId.substring(0, 32)],
+                });
+            }
+        }
+      } catch (err) {
```
dotansimha, line 142: no rethrow needed?

dotansimha (approved):

PR #6484, opened 2025-02-06
WIP: Use federation-link-util to extract and implement federation tags dur…

jdolle:
…ing orchestration

### Background

To remove redundant logic and make sure our link implementations have compatibility checks in place, we're exploring a utility library for extracting links and applying versions to schemas.

### Description

This is a first pass at using the library to extract and apply tags and inaccessible directives for contracts.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/schema/src/lib/federation-tag-extraction.spec.ts
+++ b/packages/services/schema/src/lib/federation-tag-extraction.spec.ts
@@ -2,67 +2,8 @@ import { parse, print } from 'graphql';
 import {
   applyTagFilterOnSubgraphs,
   applyTagFilterToInaccessibleTransformOnSubgraphSchema,
-  extractTagsFromFederation2SupergraphSDL,
-  getFederationTagDirectiveNameForSubgraphSDL,
-  getTagDirectiveNameFromFederation2SupergraphSDL,
   type Federation2SubgraphDocumentNodeByTagsFilter,
 } from './federation-tag-extraction';
-
-describe('getFederationTagDirectiveNameForSubgraphSDL', () => {
-  test('subgraph specification without link directive (Federation 1) -> "tag"', () => {
-    const sdl = parse(/* GraphQL */ `
-      schema {
-        query: Query
-      }
-    `);
-
-    expect(getFederationTagDirectiveNameForSubgraphSDL(sdl)).toEqual('tag');
-  });
-
-  test('subgraph specification with link subgraph directive (Federation 2) -> "federation__tag"', () => {
-    const sdl = parse(/* GraphQL */ `
-      schema @link(url: "https://specs.apollo.dev/federation/v2.0") {
-        query: Query
-      }
-    `);
-
-    expect(getFederationTagDirectiveNameForSubgraphSDL(sdl)).toEqual('federation__tag');
-  });
-
-  test('subgraph specification with link subgraph directive granular import (Federation 2) -> "tag"', () => {
-    const sdl = parse(/* GraphQL */ `
-      schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@tag"]) {
-        query: Query
-      }
-    `);
-
-    expect(getFederationTagDirectiveNameForSubgraphSDL(sdl)).toEqual('tag');
-  });
-
-  test('subgraph specification with link subgraph directive granular import object (Federation 2) -> "tag"', () => {
-    const sdl = parse(/* GraphQL */ `
-      schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: [{ name: "@tag" }]) {
-        query: Query
-      }
-    `);
-
-    expect(getFederationTagDirectiveNameForSubgraphSDL(sdl)).toEqual('tag');
-  });
-
-  test('subgraph specification with link subgraph directive granular import object lias (Federation 2) -> "label"', () => {
-    const sdl = parse(/* GraphQL */ `
-      schema
-        @link(
-          url: "https://specs.apollo.dev/federation/v2.0"
-          import: [{ name: "@tag", as: "@label" }]
-        ) {
-        query: Query
-      }
-    `);
-
-    expect(getFederationTagDirectiveNameForSubgraphSDL(sdl)).toEqual('label');
-  });
-});
 
 describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
   describe('correct @inaccessible directive usage based on subgraph version/imports', () => {
@@ -87,7 +28,8 @@ describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
       `);
 
       const outputSdl = print(
-        applyTagFilterToInaccessibleTransformOnSubgraphSchema(sdl, filter).typeDefs,
+        applyTagFilterToInaccessibleTransformOnSubgraphSchema(sdl, filter, 'inaccessible', 'tag')
+          .typeDefs,
       );
       expect(outputSdl).toMatchInlineSnapshot(`
       schema {
@@ -126,7 +68,12 @@ describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
       `);
 
       const outputSdl = print(
-        applyTagFilterToInaccessibleTransformOnSubgraphSchema(sdl, filter).typeDefs,
+        applyTagFilterToInaccessibleTransformOnSubgraphSchema(
+          sdl,
+          filter,
+          'federation__inaccessible',
+          'federation__tag',
+        ).typeDefs,
       );
       expect(outputSdl).toMatchInlineSnapshot(`
       schema @link(url: "https://specs.apollo.dev/federation/v2.0") {
@@ -165,7 +112,12 @@ describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
       `);
 
       const outputSdl = print(
-        applyTagFilterToInaccessibleTransformOnSubgraphSchema(sdl, filter).typeDefs,
+        applyTagFilterToInaccessibleTransformOnSubgraphSchema(
+          sdl,
+          filter,
+          'inaccessible',
+          'federation__tag',
+        ).typeDefs,
       );
       expect(outputSdl).toMatchInlineSnapshot(`
       schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@inaccessible"]) {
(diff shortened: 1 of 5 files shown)


packages/services/schema/src/lib/federation-tag-extraction.spec.ts
```diff
@@ -126,7 +68,12 @@ describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
       `);
 
       const outputSdl = print(
-        applyTagFilterToInaccessibleTransformOnSubgraphSchema(sdl, filter).typeDefs,
+        applyTagFilterToInaccessibleTransformOnSubgraphSchema(
+          sdl,
+          filter,
+          'federation__inaccessible',
+          'federation__tag',
+        ).typeDefs,
```
jdolle, line 76: The need to pass in these directive names vs extract then _again_ from the typedefs accounts for `205` additional lines, which makes it look like the utility didnt help simplify much.... but after taking this fact into account, we realize this change saved approx 220 lines of meaningful code, and makes it possible to support multiple versions of implementations should it ever be necessary.

packages/services/schema/src/lib/federation-tag-extraction.ts
```diff
@@ -472,10 +353,44 @@ export function applyTagFilterOnSubgraphs<
...
+
+  // for every subgraph, get the spec and apply it.
+  // Current behavior is to treat fed1 and fed2 the same and support all versions.
+  const federationTagFilterSpec = new LinkableSpec('https://specs.apollo.dev/federation', {
+    [FEDERATION_V1]: specImpl,
+    v1_1: specImpl,
+    v2_9: specImpl,
+  });
+
+  let filteredSubgraphs = subgraphs.map(subgraph => {
```
jdolle, line 379: I originally intended the LinkableSpec to handle all the implementation for a directive, but since this directive impacts other subgraphs, the logic has to be split.

I admit this makes it a bit more complicated to read and comprehend. There may be a few options to simplify. E.g.:
* expose an API for interacting with multiple subgraphs
* compose first, then apply the spec logic on the one supergraph

packages/services/schema/src/lib/federation-tag-extraction.ts
```diff
@@ -472,10 +353,44 @@ export function applyTagFilterOnSubgraphs<
...
+    };
+    return filteredSubgraph;
+  };
+
+  // for every subgraph, get the spec and apply it.
+  // Current behavior is to treat fed1 and fed2 the same and support all versions.
+  const federationTagFilterSpec = new LinkableSpec('https://specs.apollo.dev/federation', {
+    [FEDERATION_V1]: specImpl,
+    v1_1: specImpl,
+    v2_9: specImpl,
```
jdolle, line 376: Currently, this configuration will support v1.0, v1.1, v2.0 - v2.9. If 2.10 gets released and used, then an error will be thrown because the spec does not support 2.10.

This is intentional because 2.10 may introduce a new feature that is not implemented. But I want to make sure that's okay with everyone since the current approach would assume it is still supported based on the name and therefore not throw.

packages/services/schema/src/lib/federation-tag-extraction.ts
n1ru4l: > compose first, then apply the spec logic on the one supergraph

Transforming `@tag` to `@inaccessible` and passing it to the composition library was the way of doing things here, because I rather trust the federation library giving me a valid supergraph than post-processing the supergraph and potentially generating an invalid supergraph

packages/services/schema/src/lib/federation-tag-extraction.ts
```diff
@@ -472,10 +353,44 @@ export function applyTagFilterOnSubgraphs<
     name: string;
   },
 >(subgraphs: Array<TType>, filter: Federation2SubgraphDocumentNodeByTagsFilter): Array<TType> {
-  const filteredSubgraphs = subgraphs.map(subgraph => ({
-    ...subgraph,
-    ...applyTagFilterToInaccessibleTransformOnSubgraphSchema(subgraph.typeDefs, filter),
-  }));
+  const specImpl = (resolveName: (name: string) => string) => (subgraph: TType) => {
```
n1ru4l, line 356: I overall find these changes here quite complicated, which makes me doubt whether there is an abstraction that can solve all potential directives. 🤔 I need to digest this a bit more.

jdolle: I understand this concern. I'll take another look at the approach and try and simplify the library and this implementation if possible.

packages/services/schema/src/orchestrators.ts
```diff
@@ -669,11 +673,37 @@ const createFederation: (
...
+
+              return Array.from(tags);
+            };
+          const tagSpec = new LinkableSpec('https://specs.apollo.dev/tag', {
+            [FEDERATION_V1]: tagImpl,
+            v0_1: tagImpl,
+            v0_2: tagImpl,
+          });
           const supergraphSDL = parse(tempResult.result.supergraph);
-          const tags = extractTagsFromFederation2SupergraphSDL(supergraphSDL);
```
jdolle, line 673: I'm going through to simplify this implementation and I can't figure out why we want to collect all tags used by a supergraph.
I was thinking it may be to optimize the contract lookup but these collected tags dont get used there.

I think I can simplify this implementation by only writing a spec for federation to find included/excluded tags and apply `@inaccessible`.

packages/libraries/federation-link-utils/src/index.ts
```diff
@@ -39,20 +38,27 @@ export function extractLinkImplementations(typeDefs: DocumentNode): {
...
       if (!supportsFederationV2) {
-        // @note identities dont matter for Federation v1. There are no links to reference.
+        // Identities dont matter for Federation v1. There are no links to reference.
+        // So return the name without the identity's namespace
         return name.startsWith('@') ? name.substring(1) : name;
       }
 
       const matchingLink = linkByIdentity[identity];
       if (!matchingLink) {
-        throw new Error(
```
jdolle, line 53: Instead of worrying about handling errors for this function, I think it makes more sense to return what the name would be assuming it was not imported, and nudge people to use `matchesImplementation` if they want to guarantee links are used. This is simpler and more intuitive for basic usage, but also more explicit for complicated cases.

packages/libraries/federation-link-utils/src/index.ts
```diff
@@ -39,20 +38,27 @@ export function extractLinkImplementations(typeDefs: DocumentNode): {
   const linkByIdentity = Object.fromEntries(
     FederatedLink.fromTypedefs(typeDefs).map(l => [l.identity, l]),
   );
-  const supportsFederationV2 = linkByIdentity[FEDERATION_IDENTITY] !== undefined;
+  // Any schema with a `@link` directive present is considered federation 2
+  // although according to federation docs, schemas require linking specifically
+  // the federation 2.x spec. The reason for not being so picky is that supergraphs also
+  // use @link, but do not necessarily link to the federation 2.x spec.
+  const supportsFederationV2 = Object.keys(linkByIdentity).length > 0;
```
jdolle, line 45: I'm curious what others think about this.

Another option is to not support federation 1 at all, or have a separate name resolver function for fed1.

packages/services/schema/src/lib/__tests__/add-inaccessible-to-unreachable-types.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { parse } from 'graphql';
+import { extractLinkImplementations } from '@graphql-hive/federation-link-utils';
+import { addInaccessibleToUnreachableTypes } from '../add-inaccessible-to-unreachable-types';
+import { composeFederationV2 } from '../compose';
+
+test('Filters based on tags', async () => {
```
jdolle, line 6: i know we have integration tests covering this stuff. But personally I find unit tests to be a lot quicker to write and run.

packages/services/schema/src/lib/compose.ts
```diff
@@ -0,0 +1,427 @@
+import { createHmac } from 'node:crypto';
```
jdolle, line 1: This file was all existing code in `orchestrator.ts`, but I found that file overwhelming and wanted to split it up a bit.
If we don't want these changes alongside the util library usage, then I can submit a separate PR or we can opt not to move this code..

packages/services/schema/src/lib/errors.ts
```diff
@@ -0,0 +1,33 @@
+import { GraphQLError } from 'graphql';
```
jdolle, line 1: this file is also all from code that existed already in `orchestrator.ts`

packages/services/schema/src/lib/add-inaccessible-to-unreachable-types.ts
```diff
@@ -0,0 +1,60 @@
...
+    '@inaccessible',
+  );
+  const federationTypes = new Set([
+    resolveName('https://specs.apollo.dev/join', 'FieldSet'),
+    resolveName('https://specs.apollo.dev/join', 'Graph'),
+    resolveName('https://specs.apollo.dev/link', 'Import'),
+    resolveName('https://specs.apollo.dev/link', 'Purpose'),
+    resolveName('https://specs.apollo.dev/federation', 'Policy'),
+    resolveName('https://specs.apollo.dev/federation', 'Scope'),
+    resolveName('https://specs.apollo.dev/join', 'DirectiveArguments'),
```
jdolle, line 31: I'd prefer this not being a hardcoded list, but this is how it was before also.

And I'm not sure how we'd avoid it unless it is guaranteed all of these are namespaced...

packages/services/schema/src/lib/federation-tag-extraction.spec.ts
jdolle: I moved the link extraction into this federation tag extraction function because for those typedefs it wasnt being used elsewhere and it made this diff much nicer.

packages/services/schema/src/lib/compose.ts
n1ru4l: It makes it definitely harder to review the actual relevant changes. 😓

n1ru4l (approved):

PR #6486, opened 2025-02-07
send request id from cli

n1ru4l:
### Background

If a request from a customer fails, we have to dig our logs in order to find the affected request. Instead, we should specify the request id on the CLI level and log it so a customer can easily share it with us.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/libraries/cli/src/base-command.ts
+++ b/packages/libraries/cli/src/base-command.ts
@@ -189,7 +189,7 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
       'Content-Type': 'application/json',
       Accept: 'application/json',
       'User-Agent': `hive-cli/${this.config.version}`,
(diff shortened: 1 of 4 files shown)


n1ru4l:
@jdolle What I want to achieve here is to have the request-id, being logged for each request happening, so people can easily share it with us. I am unsure whether we should attach the request ID only to the errors, or in general log them as requests are being made. 🤔

n1ru4l:
closed in favor of https://github.com/graphql-hive/console/pull/6488

Issue #6487, opened 2025-02-07
CLI: initialise request id send to server


n1ru4l:
If a request from a customer fails, we have to dig our logs to find the affected request. Instead, we should specify the request-id on the CLI level and log it so a customer can easily share it with us.

_____

On all Hive services we use the `x-request-id` header to identify requests. If not sent within the HTTP request, the header is automatically set on the entry point service.

We should start generating the request ID on the CLI level and log it as a reference for clients to share with us in case they experience any unexpected errors.

PR #6488, opened 2025-02-07
include request id header in http-client

n1ru4l:
### Background

If a request from a customer fails, we have to dig into our logs to find the affected request. Instead, we should specify the request-id on the CLI/library level and log it so a customer can easily share it with us.

The main reason for this was the CLI, but it is useful for all our SDKs.

### Description

Instead of setting a request id on our gateway entry level, we can already start setting the request-id on the SDK/CLI level. This allows users to share the request id with us even if we fail to send a proper response due to unexpected errors.

Closes https://github.com/graphql-hive/console/issues/6487

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/libraries/core/src/client/http-client.ts
+++ b/packages/libraries/core/src/client/http-client.ts
@@ -96,8 +96,10 @@ export async function makeFetchCall(
 
   return await asyncRetry(
     async (bail, attempt) => {
+      const requestId = crypto.randomUUID();
+
       logger?.info(
-        `${config.method} ${endpoint}` +
+        `${config.method} ${endpoint} (x-request-id=${requestId})` +
           (retries > 0 ? ' ' + getAttemptMessagePart(attempt, retries + 1) : ''),
       );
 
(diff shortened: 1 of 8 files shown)


packages/libraries/core/src/client/http-client.ts
```diff
@@ -107,12 +109,16 @@ export async function makeFetchCall(
       const response = await (config.fetchImplementation ?? fetch)(endpoint, {
         method: config.method,
         body: config.body,
-        headers: config.headers,
+        headers: {
+          'x-request-id': requestId,
+          ...config.headers,
```
n1ru4l, line 114: technically someone could overwrite the request ID, I ignored that edge case.

dotansimha (approved):
Thank you, this one was so annoying

packages/libraries/core/src/client/http-client.ts
```diff
@@ -107,12 +109,16 @@ export async function makeFetchCall(
...
+        },
         signal,
       }).catch((error: unknown) => {
         const logErrorMessage = () =>
           logger?.error(
```
jdolle, line 119: In a future improvement, I think we should add the a request ID to the standard logger to avoid having to remember to include it in logs.
This would require we pass a context object into the logger or use another tool such as datadog to maintain a context for the request.

jdolle (approved):
Assuming we maintain the request id on the backend systems, then this very valuable.

PR #6493, opened 2025-02-10
feat: organization access tokens

n1ru4l:
### Background

Adds organization access tokens which will supersede target access tokens.

Organization access tokens allow selecting scopes and permissions, allowing granular resource-based access control.

CONSOLE-950

### Description

This PR introduces:
- Create and delete organization access tokens using GraphQL Mutations
- List available organization access tokens using the GraphQL Query
- Authentication and authorization on the GraphQL API using the organization access tokens

**Other notes:**

It adds the dependency [bentocache](https://bentocache.dev/docs/introduction), simplifying the code we have to write for multi-layer access token caching in memory and redis.

The bento cache implementation will be re-used within the `usage` reporting service (follow-up PR), so organization access tokens will not require the `tokens` service. This makes it possible to remove the service once we sunset target access tokens (or migrate them over), which reduces the overall number of services we need to run Hive.

### Checklist

- [x] Input validation
- [x] Session management
- [x] Access control
- [x] Cryptographic practices


packages/services/api/src/modules/organization/providers/resource-assignments.ts
n1ru4l, line 1: The logic here is shared between organization members and organization access tokens, so I moved it to a new shared file.

packages/services/api/src/modules/schema/providers/schema-manager.ts
```diff
@@ -1007,12 +1007,11 @@ export class SchemaManager {
     });
 
     await this.session.assertPerformAction({
-      action: 'schema:loadFromRegistry',
+      action: 'project:describe',
```
n1ru4l, line 1010: `schema:loadFromRegistry` is too specific. Instead ,it should be `project:describe` access that must be granted to the access token. Since we do not expose the permission to the outside world, this change is safe to perform.

patches/bentocache.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/package.json b/package.json
+index 26e97ac450795f4595d06b7b4b9d1d0f6b16700e..cabc9439217790a325c042f2931fe8c3761ccaa3 100644
+--- a/package.json
++++ b/package.json
+@@ -15,6 +15,8 @@
+   ],
+   "exports": {
+     ".": "./build/index.js",
++    "./build/src/drivers/redis": "./build/src/drivers/redis.js",
++    "./build/src/drivers/memory": "./build/src/drivers/memory.js",
```
n1ru4l, line 10: we have an esm <-> ts module resolution conflict.  Adding these files to the exports map unblocked me from importing the modules without typescript errors.

packages/services/api/src/modules/organization/providers/organization-members.ts
n1ru4l, line 1: Almost all changes here are related to moving the resource assignment logic to a new standalone file `providers/resource-assignments.ts`

n1ru4l:
@coderabbitai review

packages/migrations/src/actions/2025.01.30T00-02-03.organization-access-tokens.ts
```diff
@@ -0,0 +1,24 @@
...
+      "id" UUID PRIMARY KEY NOT NULL DEFAULT uuid_generate_v4()
+      , "organization_id" UUID NOT NULL REFERENCES "organizations" ("id") ON DELETE CASCADE
+      , "created_at" timestamptz NOT NULL DEFAULT now()
+      , "title" text NOT NULL
+      , "description" text NOT NULL
+      , "permissions" text[] NOT NULL
+      , "assigned_resources" jsonb
+      , "hash" text
+      , "first_characters" text
+    );
```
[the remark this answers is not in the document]
n1ru4l: We already have constraints on the business logic layer using zod validation.

packages/services/storage/src/db/types.ts
```diff
@@ -156,6 +156,18 @@ export interface oidc_integrations {
   userinfo_endpoint: string | null;
 }
 
+export interface organization_access_tokens {
+  assigned_resources: any | null;
```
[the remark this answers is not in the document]
n1ru4l: This file is auto-generated.

packages/migrations/src/run-pg-migrations.ts
```diff
@@ -158,5 +158,6 @@ export const runPGMigrations = async (args: { slonik: DatabasePool; runTo?: stri
       await import('./actions/2025.01.17T10-08-00.drop-activities'),
       await import('./actions/2025.01.20T00-00-00.legacy-registry-model-removal'),
       await import('./actions/2025.01.30T00-00-00.granular-member-role-permissions'),
+      await import('./actions/2025.01.30T00-02-03.organization-access-tokens'),
```
n1ru4l, line 161: TODO: update dates

packages/services/api/src/modules/auth/lib/authz.ts
```diff
@@ -366,8 +367,8 @@ const permissionsByLevel = {
     z.literal('laboratory:describe'),
     z.literal('laboratory:modify'),
     z.literal('laboratory:modifyPreflightScript'),
-    z.literal('schema:loadFromRegistry'),
     z.literal('schema:compose'),
+    z.literal('usage:report'),
```
n1ru4l, line 371: TODO: remove for now; will introduce it when we implement organization access token support in usage reporting service

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -0,0 +1,263 @@
...
+    });
+
+    const assignedResources =
+      await this.resourceAssignments.transformGraphQLResourceAssignmentInputToResourceAssignmentGroup(
+        organizationId,
+        args.assignedResources ?? { mode: 'granular' },
+      );
+
+    const permissions = Array.from(
+      new Set(args.permissions.filter(permission => assignablePermissions.has(permission as any))),
```
n1ru4l, line 134: TODO: any cast

packages/services/api/src/modules/organization/module.graphql.ts
n1ru4l, line 1: TODO: implement `Organization.accessTokens` field

packages/services/api/src/modules/organization/resolvers/Mutation/deleteOrganizationAccessToken.ts
```diff
@@ -0,0 +1,17 @@
...
+    organizationAccessTokenId: args.input.organizationAccessTokenId,
+  });
+
+  return {
+    ok: {
+      __typename: 'DeleteOrganizationAccessTokenResultOk',
+      deletedOrganizationAccessTokenId: result.organizationAccessTokenId,
+    },
+  };
+};
```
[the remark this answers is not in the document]
n1ru4l: nope we already handle errors in the business logic layer.

packages/services/api/src/modules/organization/lib/organization-access-token-permissions.ts
```diff
@@ -0,0 +1,67 @@
+import { type PermissionGroup } from './permissions';
+
+export const permissionGroups: Array<PermissionGroup> = [
```
n1ru4l, line 3: This currently only lists the permissions used by the CLI, as we add more functionality to the public, we can add the permissions here. This list is queryable via the GraphQL API, so we can easily show it on the dashboard when creating an organization access token.

packages/migrations/src/actions/2025.02.20T00-00-00.organization-access-tokens.ts
```diff
@@ -0,0 +1,24 @@
...
+  run: ({ sql }) => sql`
+    CREATE TABLE IF NOT EXISTS "organization_access_tokens" (
+      "id" UUID PRIMARY KEY NOT NULL DEFAULT uuid_generate_v4()
+      , "organization_id" UUID NOT NULL REFERENCES "organizations" ("id") ON DELETE CASCADE
+      , "created_at" timestamptz NOT NULL DEFAULT now()
+      , "title" text NOT NULL
+      , "description" text
+      , "permissions" text[] NOT NULL
+      , "assigned_resources" jsonb
+      , "hash" text
```
n1ru4l, line 14:
```suggestion
      , "hash" text NOT NULL
```

packages/services/api/src/modules/audit-logs/providers/audit-logs-types.ts
```diff
@@ -327,6 +327,18 @@ export const AuditLogModel = z.union([
       scriptContents: z.string(),
     }),
   }),
+  z.object({
+    eventType: z.literal('ORGANIZATION_ACCESS_TOKEN_CREATED'),
+    metadata: z.object({
+      organizationAccessTokenId: z.string().uuid(),
+    }),
```
n1ru4l, line 334: @dotansimha Do you see it necessary to store more data here
e.g. 
- assigned permissions
- specified resources

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -0,0 +1,363 @@
...
+      params: { organizationId: record.organizationId },
+    });
+
+    await this.pool.query(sql`
+      DELETE
+      FROM
+        "organization_access_tokens"
+      WHERE
+        "id" = ${args.organizationAccessTokenId}
+    `);
```
n1ru4l, line 211: Instead, we could also do a soft delete. We never did that AFAIK, what do you think?

packages/services/server/src/index.ts
```diff
@@ -413,10 +415,14 @@ export async function main() {
...
               new OrganizationMemberRoles(storage.pool, logger),
-              storage,
               logger,
             ),
           }),
+        (logger: Logger) =>
+          new OrganizationAccessTokenStrategy({
+            logger,
+            cache: registry.injector.get(OrganizationAccessTokensCache),
+          }),
```
n1ru4l, line 425: This registers authentication and authorization to the GraphQL API using the new organization access tokens.

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
dotansimha: hahaha 😆 i was scrolling through the migration file and thought "should this be soft-deleted? feel like a sensitive info that might need to be kept" -> searched for this piece of code -> found your comment 😆

dotansimha: And yeah, i do feel like we need to soft-delete

packages/services/api/src/modules/audit-logs/providers/audit-logs-types.ts
n1ru4l: https://github.com/graphql-hive/console/pull/6493/commits/8c38327a078b12d24c6a73e91478a982d2c575a2

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
n1ru4l: @dotansimha What information should we store about the soft-delete?

We could store, the deletion date, user info (email; userId) etc. or we could just store a boolean `isDeleted`.

This information would be redundant with the audit log, but the audit log retention period is only a year.
The purpose of doing the soft delete would be to have more context if that one-year period has passed?

kamilkisiela (approved):

dotansimha (approved):

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
jdolle: If the goal is to save information for auditing or in case we need to find if a token was set for someone in the past, then I think it'd be better to move this to a separate storage mechanism.
I may be missing something, but I dont see a need to have this information so available. And soft deletes add complexity / are easy to mistakenly not check.

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -0,0 +1,363 @@
...
+        "organization_id" = ${args.organizationId}
+        ${
+          cursor
+            ? sql`
+              AND (
+                (
+                  "created_at" = ${cursor.createdAt}
+                  AND "id" < ${cursor.id}
+                )
+                OR "created_at" < ${cursor.createdAt}
```
jdolle, line 267: was this intended for the tertiary's else? Having this and and or in the same sql doesnt make sense to me.

packages/services/api/src/modules/schema/providers/schema-manager.ts
jdolle: Agreed. I dont see a case where someone should be able to access the project but not the schema.

packages/services/api/src/modules/shared/providers/id-translator.ts
```diff
@@ -156,6 +156,58 @@ export class IdTranslator {
...
+    onError(): never;
+  }): Promise<{ organizationId: string }> {
+    let selector: {
+      organizationId: string;
+    };
+
+    if (args.reference.bySelector) {
+      const organizationId = await this.translateOrganizationId(args.reference.bySelector).catch(
+        error => {
+          this.logger.debug(error);
```
jdolle, line 172: Not as part of this PR, but I think we should standardize our logs to include a unique code so that we can easily track log lines back to their source.
I think as they are now, if i saw these lines in a log that I wouldn't know what went wrong.

packages/services/api/src/modules/auth/lib/organization-access-token-strategy.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    // TODO: we should probably cache this verification
+    const isHashMatch = await OrganizationAccessKey.verify(
+      result.accessKey.privateKey,
+      organizationAccessToken.hash,
+    );
+
+    if (!isHashMatch) {
+      this.logger.debug('Provided private key does not match hash.');
```
jdolle, line 106: Do we want to capture this case to return more specific error messages?

packages/services/api/src/modules/organization/lib/permissions.ts
```diff
@@ -0,0 +1,16 @@
+import type { Permission } from '../../auth/lib/authz';
+
+export type PermissionRecord = {
+  id: Permission;
+  title: string;
+  description: string;
+  dependsOn?: Permission;
+  isReadyOnly?: true;
```
jdolle, line 8: should this be `isReadOnly`?

packages/services/api/src/modules/organization/providers/organization-access-tokens-cache.ts
```diff
@@ -0,0 +1,69 @@
...
+    logger: Logger,
+  ) {
+    this.findById = findById({ pool, logger });
+    this.cache = new BentoCache({
+      default: 'organizationAccessTokens',
+      stores: {
+        organizationAccessTokens: bentostore()
+          .useL1Layer(
+            memoryDriver({
+              maxItems: 10_000,
```
jdolle, line 35: Should this be configurable? Why 10k?

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
n1ru4l: if we fetch additional records we want to fetch:

1. all records that share the same `created_at` date but have a "lower" id (records could have the same `created_at` value if created at the same time)
2. all records that have a lower `created_at` date

packages/services/api/src/modules/organization/providers/organization-access-tokens-cache.ts
n1ru4l: 10k is a magic number I choose that seems big enough for the beginning. 😅
Making it configurable rn might be an overkill. However, I think we should monitor the numbers and adjust if needed.

packages/services/api/src/modules/audit-logs/providers/audit-logs-types.ts
```diff
@@ -327,6 +327,18 @@ export const AuditLogModel = z.union([
       scriptContents: z.string(),
     }),
   }),
+  z.object({
+    eventType: z.literal('ORGANIZATION_ACCESS_TOKEN_CREATED'),
+    metadata: z.object({
+      organizationAccessTokenId: z.string().uuid(),
```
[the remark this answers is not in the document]
n1ru4l: I think it does not really hurt to have it here, it's an additional safety layer when reading from the database

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
n1ru4l: I agree with the forgetting to do the check part when querying the database. 😅 @dotansimha we already store metadata within clickhouse, is the one-year retention insufficient?

packages/services/api/src/modules/auth/lib/organization-access-token-strategy.ts
n1ru4l: Hmm, I think we do not really want to tell a potential brute-forcer that he correctly guessed the access key id, but not the private ket part?

packages/services/api/src/modules/organization/providers/organization-access-tokens-cache.ts
jdolle: Consider adding some more monitoring for this then. E.g. It might be nice to know if this ever exceeds the threshold.
Or are there other ways we could easily determine this is a bottleneck?

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
jdolle: tyvm. I'm not used to cursor pagination in sql. That makes sense.

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
dotansimha: if we store it all in the audit log, and you think that there's no value in making it a soft-delete, then it's fine. <br>My main reason for writing that was because of audit log and trace for user actions. 

As long as users know it's permanently deleting it, we can skip soft-delete.

packages/services/api/src/modules/organization/providers/organization-access-tokens-cache.ts
n1ru4l: @jdolle I wanted to add https://bentocache.dev/docs/plugin-prometheus here to have more insights.

jdolle: That looks promising. I like that they provide a grafana configuration. :)

PR #6498, opened 2025-02-11
refactor(laboratory): preflight editor components

jasonkuhrt:
- **improve(laboratory): validate editor content with TypeScript**
- **group laboratory e2e tests**
- **doc rationale**
- **lint**
- **pin**
- **fix**
- **fix type cast any**
- **feedback**
- **lock**
- **mention hack**
- **wip**
- **just refactor**


jasonkuhrt:
The changes to Cypress could be its own PR. If you prefer that, I'll do it. They were pulled in here from the previous PR https://github.com/graphql-hive/console/pull/6476

packages/web/app/src/lib/preflight/shared-types.ts
```diff
@@ -2,6 +2,14 @@
 
 import { Kit } from '../kit';
 
+// todo stop using enums
+export const enum PreflightWorkerState {
+  running,
+  ready,
+}
```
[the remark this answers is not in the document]
jasonkuhrt: @kamilkisiela another reason given by AI.

jasonkuhrt:
There's an issue with reactivity, fixing e2e test.

package.json
```diff
@@ -86,6 +86,7 @@
     "graphql": "16.9.0",
     "gray-matter": "4.0.3",
     "jest-snapshot-serializer-raw": "2.0.0",
+    "monaco-editor": "0.52.2",
```
jasonkuhrt, line 89: Added for static typing

packages/web/app/src/lib/MonacoEditorReact/index.ts
jasonkuhrt, line 1: Responding to https://github.com/graphql-hive/console/pull/6476#discussion_r1948558744

To have a standard namespace for terms such as `Editor` and `EditorProps`. Easy autoimport.

cypress/e2e/laboratory/_cy.ts
n1ru4l, line 1: Can we name this file `cypress.ts`? I don't understand why we need to use an abbreviation, let alone why we need to prefix it with a underscore.

cypress/e2e/laboratory/preflight.cy.ts
```diff
@@ -1,21 +1,9 @@
...
-  modal: {
-    buttonSubmitCy: 'preflight-modal-submit',
-  },
-};
+import { dedent } from '../../support/dedent';
+import { cyLaboratory } from './_cy';
+
+const s = cyLaboratory.preflight.selectors;
+
+const cyp = cyLaboratory.preflight;
```
n1ru4l, line 6: If requered to re-assign these namespaces/object whatever `s` and `cyp` are not expressive enough for me. Alternative proposal:
- `selectors`
- `cypress`

packages/web/app/src/lib/MonacoEditorReact/index.ts
n1ru4l, line 1: This can be `packages/web/app/src/lib/monaco-editor-react.ts`, the `index.ts` is unnecessary. In general we do not use camel cases for file names (there are a few remaining ones from the past tho)

packages/web/app/src/lib/preflight/components/EditorTitle.tsx
```diff
@@ -0,0 +1,10 @@
+import { ReactNode } from 'react';
+import { cn } from '@/lib/utils';
+
+export function EditorTitle(props: { children: ReactNode; className?: string }) {
+  return (
+    <div className={cn('cursor-default text-base font-semibold tracking-tight', props.className)}>
+      {props.children}
+    </div>
+  );
+}
```
n1ru4l, line 10: Is it really necessary to have a file for these 10 lines?

n1ru4l (changes requested):
I added some comments. 

Summed up:
- We do not use camel case for file names or folder names
- We do not use `_` for file names
- We use expressive variable names and no abbreviations

I still do not understand why this PR is treated as necessary for moving forward https://github.com/graphql-hive/console/pull/6476, which should be the priority instead.

jasonkuhrt:
> I still do not understand why this PR is treated as necessary for moving forward https://github.com/graphql-hive/console/pull/6476, which should be the priority instead.

Hey, two things:

1. It was prioritized. It was blocked on product feedback that you or someone else needs to decide on. If I did something wrong, please be specific about what you would like to see next time.
2. I explained in the previous PR https://github.com/graphql-hive/console/pull/6476#issuecomment-2648269164

packages/web/app/src/lib/MonacoEditorReact/index.ts
jasonkuhrt: > In general we do not use camel cases for file names (there are a few remaining ones from the past tho)

Noted!

I saw quite a few modules in the codebase with pascal case naming. I assume that this is still a valid pattern in the codebase, not deprecated. Is that correct?

cypress/e2e/laboratory/_cy.ts
jasonkuhrt: Oh happy to explain:

1. The directory contains test files. I would like to keep this file from being interleaved in them since its not a test file. That is what `_` achieves. It could also be nested directory or `$` prefix, whereas `cypress.ts` would not achieve this. Can you propose something that would? Alternatively clarify that you don't care about about my goal here.
2. The abbreviation is consistent with APIs from Cypress. Cypress uses a global `cy` namespace and what this module does is provide a kind of `cy` dedicated for laboratory. If that is not logical to you that's fine, but I think its not as random as your reaction to my change might suggest.

Please let me know how the above changes, if at all, your feedback.

packages/web/app/src/lib/MonacoEditorReact/index.ts
kamilkisiela: I mean, there are 10 files that are pascal cased, the rest is auto generated by codegen

packages/web/app/src/lib/preflight/components/EditorTitle.tsx
jasonkuhrt: Maybe its time to add some rules.

cypress/e2e/laboratory/_cy.ts
jdolle: The `.cy.ts` extension does this though right? If this is a utility library for testing rather than a test file, it can have a different extension which is excluded from the test file matching.

cypress/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "compilerOptions": {
     "target": "es2021",
     "lib": ["es2021", "dom"],
-    "types": ["node", "cypress"]
+    "moduleResolution": "node",
+    "types": ["node", "cypress", "monaco-editor"]
```
jdolle, line 6: I think `types` here are for global definitions.
Is that necessary for your case? Or could you import these on the few files where the monaco-editor is referenced?

packages/web/app/src/lib/preflight/components/EditorTitle.tsx
jdolle: if it can be enforced via a linter then I think it'd be fine to move to a stricter separation of components by file.
But otherwise we can keep it loose b/c i dislike enforcing these sorts of things manually.

I agree that if things become unruly then split.

packages/web/app/src/lib/preflight/hooks/usePreflight.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+  const abortExecutionRef = useRef<null | (() => void)>(null);
+
+  async function execute(
+    script = target?.preflightScript?.sourceCode ?? '',
+    isPreview = false,
+  ): Promise<PreflightResultData> {
+    const resultEnvironmentVariablesDecoded: PreflightResultData['environmentVariables'] =
+      Kit.tryOr(
+        () => JSON.parse(latestEnvironmentVariablesRef.current),
```
jdolle, line 56: I only see a few options:
1. Let wipe env vars (current approach)
1. prompt the user to fix the env vars or wipe. I.e. `Continuing will clear your environment variables because they are in an invalid state. Press "OK" to proceed or "Cancel" to stop`
1. Try to use string manipulation to append the new values without correcting the issue.

I think your approach is fine for now. This is an edge case. But prompting is probably nicer?

packages/web/app/src/lib/preflight/hooks/usePreflight.tsx
```diff
@@ -0,0 +1,323 @@
...
+              }
+
+              openedPromptIds.delete(promptId);
+              contentWindow?.postMessage(
+                {
+                  type: IFrameEvents.Incoming.Event.promptResponse,
+                  id,
+                  promptId,
+                  value,
+                } satisfies IFrameEvents.Incoming.EventData,
```
jdolle, line 153: This is cool. I need to read up more on the newer typescript features
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator

packages/web/app/src/lib/preflight/hooks/usePreflight.tsx
```diff
@@ -0,0 +1,323 @@
...
+            ...logs,
+            {
+              level: 'error',
+              message: error.message,
+              line: error.line,
+              column: error.column,
+            },
+            {
+              level: 'log',
+              message: 'Script failed',
```
jdolle, line 205: Should we include the duration here also?

cypress/e2e/laboratory/_cy.ts
jasonkuhrt: > it can have a different extension which is excluded from the test file matching.


@jdolle I think we are talking about different things. At issue in my mind was keeping the utility file in the directory tree in a prefix or suffix position in the directory. In other words it is purely about the visual parsing in the file explorer. In regards to test file execution, as you said, that works fine. Does that clear up for you what I was trying to do?

cypress/tsconfig.json
jasonkuhrt: Good question, I can't recall now how it ended up like this. I will double check and either fix or motivate.

packages/web/app/src/lib/preflight/hooks/usePreflight.tsx
jasonkuhrt: Prompting sounds nice, but maybe doesn't catch all cases, like user closing the tab.

I think a fourth idea we could add to your list is to introduce the concept of "staged" versus "committed" data.

- Committed Persistence: for the last successfully parsed editor content
  - Used by: preflight execution; request execution
- Staging Persistence: for the current content of editor
  - Used by: editor
- If parsing fails on staging persistence data, then it does not get synchronized to committed persistence.

Let me know what you think about this idea.

I think this idea would work well in combination with a prompt/banner/callout whatever that you mentioned. My only contribution here is to allow state stability around this user flow regardless of how they close their tabs, navigate, etc.

packages/web/app/src/lib/preflight/hooks/usePreflight.tsx
jasonkuhrt: `satisfies` was a nice ergonomic addition because it finally allowed the ability to have type-safe `const` types (whereas previously one had to choose between those two things). Not revolutionary but one of the many paper cut improvements over the past year or two.

While various details of TS have improved I think you might not be as far behind as you might feel :) in that the core type system has not significantly changed since perhaps the introduction of template literal types.

It feels to me like TS has significantly improved in meta ways, for example it made milestone improvements on ESM support.

jasonkuhrt:
Note: Until this question is answered https://github.com/graphql-hive/console/pull/6476#issuecomment-2653960708 I think it is best to hold off on further feedback, because a scope of this PR (E2E tests) would be removed if we agree to have those refactors maintained in the other PR.

cypress/e2e/laboratory/_cy.ts
jdolle: yeah.
I'm not very concerned with this visual distinction. Unless the directory gets very large, it's still easy enough to distinguish

jasonkuhrt: It is true that I struggle parsing visuals and navigating inconsistency more than the average person.

jasonkuhrt: If co-locating at the edges of the directory are hurting others' usability of the project, which is the sense I'm getting but not sure, then I will try to find another way, perhaps just back into the catch all directory.

PR #6506, opened 2025-02-12
Fix og:url (this should work out of the box later)

hasparus:
### Description

I updated `og:url` and `canonical` to work properly, but this should really be fixed upstream so we don't have to remember about it, because doing it in each page and layout is a bit error prone.


saihaj (approved):
🐐

Issue #2818, opened 2023-09-14
Feature: Schema Ownership


kamilkisiela:
**Option 1:**  We can use the codebase itself to parse `CODEOWNERS` files.

**Option 2:** Allow parsing the schemas with something like:

```graphql
# CODEOWNERS: dotansimha
type User {
  id: ID!
}
```

**Option 3**: we can use the Hive CLI for scanning CODEOWNERS in the `fs` and then annotate the schema with the GraphQL comment from option 2.

This way we can introduce: 

* Visibility of ownership of each type/field/service 
* Build a tool for schema approvals - we can communicate it to the Reviewers of a PR back to GitHub, or even add a schema publication check and look for the owners approvals. 
* Build a tool for changes visibility - when a schema adds a new reference to a GraphQL `type`, we can notify the owner of that type/service that the type was used. 

> I guess we need GH integration first, but it's a good chance to add that.

n1ru4l:
Related: https://github.com/kamilkisiela/graphql-hive/issues/2747

n1ru4l:
I think we should probably solve this with schema directives, something like the following:

```graphql
type User @codeOwners(teams: ["team-1"]) {
  id: ID!
  posts: PostConnection! @codeOwners(teams: ["team-2", "team-3"])
}
```

n1ru4l:
Apollo Federation also supports schema ownership through directives: https://www.apollographql.com/docs/technotes/TN0036-owner-pattern/

comatory:
> Apollo Federation also supports schema ownership through directives: https://www.apollographql.com/docs/technotes/TN0036-owner-pattern/
> 
> ```graphql
> directive @owner(team: String!) on OBJECT
> ```

Thanks, I see that it's only for Federation v2 though but it's definitely good to know. Do you happen to know if GraphQL Fusion spec will suport such directive?

Anyway I think we're talking about two related things here but maybe it's important to make the distinction

1. Having visibility into ownership via UI in Hive
2. Adding the ownership metadata

I think for `#2`, there's few more things I'd like to consider:

## 2a. leveraging `CODEOWNERS` file

Our setup often has multiple subschemas defined in single repository, they all get published separately to Hive. The "codeowners" file then contains entry like this:

```
src/module-1/schema.graphql @team-1
```

[...]

PR #6517, opened 2025-02-13
feat: add metadata to hive ui with linked directive support

jdolle:
### Background

We need a way to provide and visualize metadata for subgraphs.

https://github.com/graphql-hive/console/issues/2747
https://github.com/graphql-hive/console/issues/2818

### Description

This adds metadata to fields via a directive.
In this PR, I added a way to view metadata on fields via the explorer. But to close these issues, we will need additional views and behaviors. E.g. filtering the explorer based on metadata, and viewing metadata on a field's insights page.
Screenshot 2025-02-13 at 12 54 30 PM

### Checklist

- [x] Testing


packages/libraries/federation-link-utils/src/index.ts
```diff
@@ -46,12 +46,6 @@ export function extractLinkImplementations(typeDefs: DocumentNode): {
 
   return {
     resolveImportName: (identity, name) => {
-      if (!supportsFederationV2) {
-        // Identities dont matter for Federation v1. There are no links to reference.
-        // So return the name without the identity's namespace
-        return name.startsWith('@') ? name.substring(1) : name;
```
jdolle, line 52: Im hitting issues with this logic. It isnt completely accurate...
To know for sure I need to add a check for the `@core` directive also.
WIP

jdolle:
I've adjusted this to save metadata to a column on the schemaVersion table.
Since we're capable of parsing the SDL at runtime, looking up the metadata in an object should be no worse.

packages/services/api/src/modules/schema/resolvers/SchemaMetadata.ts
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * When a mapper is used, it is possible to hit runtime errors in some scenarios:
+ * - given a field name, the schema type's field type does not match mapper's field type
+ * - or a schema type's field does not exist in the mapper's fields
+ *
+ * If you want to skip this file generation, remove the mapper or update the pattern in the `resolverGeneration.object` config.
+ */
+export const SchemaMetadata: SchemaMetadataResolvers = {
+  /* Implement SchemaMetadata resolver logic here */
+};
```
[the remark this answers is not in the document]
jdolle: identity resolver is correct in this case. There's no need to implement anything.

packages/libraries/cli/examples/federation.products.graphql
```diff
@@ -1,60 +1,74 @@
-enum CURRENCY_CODE {
```
jdolle, line 1: I assume the prior version was for Federation 1? But I couldn't get the two example schemas working locally. I've opted to update them to Federation 2. Let me know if I should make a new version instead.

integration-tests/tests/schema/metadata.spec.ts
```diff
@@ -0,0 +1,152 @@
+import { getServiceHost } from 'testkit/utils';
+import type { SchemaBuilderApi } from '@hive/schema';
+import { createTRPCProxyClient, httpLink } from '@trpc/client';
+
+const host =
```
dotansimha, line 5: nit: should this go somewhere into the test kit? 😄

packages/migrations/src/actions/2025.02.14T00-00-00.schema-versions-metadata.ts
```diff
@@ -0,0 +1,10 @@
+import type { MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.02.14T00.00.00.schema-versions-metadata.ts',
```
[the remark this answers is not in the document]
jdolle: No one would be using `@meta` -- since we're not just looking for that directive here, but we're also making sure it's linked to our spec url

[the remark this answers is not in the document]
dotansimha: great!

integration-tests/tests/schema/metadata.spec.ts
jdolle: possibly... I dont see anywhere in the codebase where we use `SCHEMA_SERVICE_HOST_OVERRIDE`... If anything I assume `RUN_AGAINST_LOCAL_SERVICES` is more practical for us to use for overriding the hosts.
I'd be in favor of removing the env var condition in this file and in the `contracts.spec.ts` before moving this into the test kit.

packages/services/schema/src/lib/metadata-extraction.ts
n1ru4l, line 1: Let's have some tests for this implementation.

packages/services/schema/src/orchestrators.ts
```diff
@@ -214,6 +277,16 @@ const createFederation: (
...
         if (composed.type === 'success') {
+          // merge all metadata from every subgraph by coordinate
+          let metadata: Record<string, Array<{ name: string; content: string }>> | null = null;
+          try {
+            const subgraphMetadatas = subgraphs.map(({ typeDefs }) => extractMetadata(typeDefs));
+            metadata = mergeMetadata(...subgraphMetadatas);
+          } catch (e: unknown) {
+            // warn and proceed because metadata isn't considered critical
+            logger.warn(`Cannot extract metadata from subgraphs because ${e}`);
+          }
```
n1ru4l, line 288: What could be thrown there?

packages/services/schema/src/orchestrators.ts
```diff
@@ -273,6 +348,7 @@ const createFederation: (
               filter,
             );
             return {
+              // @note Although it can differ from the supergraph's, ignore metadata on contracts.
               ...subgraph,
               typeDefs: filteredSubgraph.typeDefs,
```
n1ru4l, line 353: What is the reason here for ignoring the metadata?

n1ru4l (approved):
Overall, I think the implementation looks solid to me. ✅

I would like some more unit or integration tests around the metadata extraction to have a bit more confidence and make sure we are not breaking something unexpectedly.

I want @kamilkisiela to review this as well before merging things, to double check this covers what our clients need.

packages/migrations/test/2024.07.23T09.36.00.schema-cleanup-tracker.test.ts
```diff
@@ -93,7 +93,8 @@ await describe('migration: schema-cleanup-tracker', async () => {
             github_sha,
             tags,
             has_contract_composition_errors,
-            conditional_breaking_change_metadata
+            conditional_breaking_change_metadata,
+            schemaMetadata,
```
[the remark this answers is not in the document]
jdolle: yup!

jdolle:
> Could there be a use case where people want to specify meta-data but not have it appear within the supergraph?

`@composeDirective` was required in the first iteration, but isn't anymore. Sorry, I forgot to update the test.

I've updated the tests to cover more cases and removed the `@composeDirective` to clarify that it isnt required.

packages/services/schema/src/orchestrators.ts
jdolle: Metadata will only differ between contracts and supergraph for `@shared` fields or entities, where one subgraph could impact another.
I think this is an edge case and the additional complexity of calculating and displaying metadata doesn't provide much benefit to the users.

packages/services/schema/src/lib/metadata-extraction.ts
jdolle: Covered via the integration test. I'm happy to add unit tests too if preferred.

packages/services/schema/src/orchestrators.ts
jdolle: This is a remnant of a previous iteration. Ive removed the try/catch. It shouldn't throw.

kamilkisiela:
I will review the code tomorrow, but in terms of UI, I think we could pass the metadata to the subgraph badge.
Metadata is bound to a subgraph anyway, and when hovering on the subgraph, we could show extra information (like contact details).

If at least two subgraphs share a field and those team add metadata (like slack channel or just the name of the team), a dedicated icon for metadata will fall short. It won't be able to display info about two teams. If we merge it into the subgraph badge, it connects nicely.

jdolle:
That's an interesting point. I like the idea of linking it to the subgraph, and I agree that fewer icons the better.
I'll make that adjustment.

jdolle:
Screenshot 2025-02-19 at 4 23 33 PM
I've updated the API to add an additional "source" field to indicate the origin of the metadata.
I felt "source" was more accurate than "subgraph" if ever this metadata can be generated in another way.

dotansimha (approved):

PR #6518, opened 2025-02-14
Ensure response body is read before timeout to avoid abort errors

kamilkisiela:
Reading the response body (res.text()) after a fetch timeout can lead to an AbortError, making retries impossible. This change ensures the body is read immediately after the response is received, allowing it to be included in the retry logic.

By decoupling the response body from AbortSignal.timeout(), we prevent unexpected failures when the body is accessed later due to application logic delays. While this slightly increases memory usage, it ensures reliability when handling AWS S3 responses.

- [x] changeset

Closes CONSOLE-1064

### DIFF
--- a/packages/services/cdn-worker/src/aws.ts
+++ b/packages/services/cdn-worker/src/aws.ts
@@ -211,7 +211,6 @@ export class AwsClient {
             status: response.status,
             statusText: response.statusText,
             headers: response.headers,
-            cf: response.cf,
           });
         }
 
--- a/packages/services/cdn-worker/src/aws.ts
+++ b/packages/services/cdn-worker/src/aws.ts
@@ -196,6 +196,10 @@ export class AwsClient {
             result: { type: 'success', response },
           });
 
+          if (!response.body) {
+            return response;
+          }
+
           // Read the body first, to make sure it won't be aborted by the signal (timeout)
           // If we get a timeout error during the body reading, we can't retry,
           // as it's out of the control of that function.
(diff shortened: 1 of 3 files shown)


packages/services/cdn-worker/src/aws.ts
```diff
@@ -196,7 +196,26 @@ export class AwsClient {
...
+          // Read the body first, to make sure it won't be aborted by the signal (timeout)
+          // If we get a timeout error during the body reading, we can't retry,
+          // as it's out of the control of that function.
+          // One example of it is when the res.text() is called with a delay due to application logic,
+          // and the timeout is reached at that point. The read will be aborted.
+          //
+          // For this reason, I prefer to read the body first, make it part of the retry logic,
+          // and only then return the response to the consumer,
+          // even at the cost of higher memory footprint (like it matters...).
+          const bodyText = await response.text();
```
jdolle, line 212: Awaiting makes sense to handle the errors, but should a `TypeError` be handled and not retried?
https://developer.mozilla.org/en-US/docs/Web/API/Response/text

And if you want to return a copy of the response, could you use `response.clone()`?

kamilkisiela: Cloning a response has the same behavior, it's still affected by signals, so when a body is read the signal timeout is still part of it

kamilkisiela: Sometimes the body is corrupted, retrying I think makes sense in this situation

jdolle (approved):

PR #6527, opened 2025-02-15
chore(deps): migration to sentry v9

dotansimha:


dotansimha:
will iterate on this one later.

PR #6531, opened 2025-02-18
In-memory fallback queue and no manual reconnecting of Kafka in Usage service

kamilkisiela:
Completes CONSOLE-1027

- Made `readinessProbe` much more sensitive.
- No manual reconnection of Kafka client based on event listeners - those are for the initial connection, the reconneciton happens automatically in KafkaJS, no need to do that manually.
- Increased retry attempts of Kafka client.
- Added fallback queue to handle failed Kafka producer sends.
- fallback queue size is considered when reporting readiness.

The idea here is to store Kafka messages that failed to be sent, in an in-memory queue, and retry them later.
When Kafka broker is down, we stop accepting http requests and wait for the queue to be flushed entirely.


kamilkisiela:
@coderabbitai review

dotansimha (approved):

dotansimha:
@kamilkisiela how do we test it? can we test on staging and produce some load to make sure it's all working correctly?

PR #6536, opened 2025-02-19
Adds bloom filter index to coordinates agg views

kamilkisiela:
Improves the performance of the queries that filter the rows by the type's name.

For example, when looking for `Member.*` coordinates we eliminate the need to scan the whole table, by leveraging the `idx_typename` index. We filter rows by the first part of the coordinate field (`substringIndex(coordinate, '.', 1)`).

It helps ClickHouse to make 10% of the original work :)

Some numbers. Big queries that took 8-17s are now around 2s.

### DIFF
--- a/packages/migrations/src/clickhouse-actions/012-coordinates-typename-index.ts
+++ b/packages/migrations/src/clickhouse-actions/012-coordinates-typename-index.ts
@@ -29,7 +29,7 @@ export const action: Action = async (exec, query) => {
       table String,
       idx_created Bool DEFAULT false,
       idx_materialized Bool DEFAULT false
-    ) ENGINE = MergeTree()
+    ) ENGINE = MergeTree() ORDER BY tuple()
   `);
 
   const tables = await query(`
--- a/packages/migrations/src/clickhouse-actions/012-coordinates-typename-index.ts
+++ b/packages/migrations/src/clickhouse-actions/012-coordinates-typename-index.ts
@@ -67,10 +67,13 @@ export const action: Action = async (exec, query) => {
     } else {
       console.log(`Creating idx_typename for table ${name}`);
       await exec(
-        `ALTER TABLE "${innerTable}" ADD INDEX idx_typename (substringIndex(coordinate, '.', 1)) TYPE ngrambf_v1(4, 1024, 2, 0) GRANULARITY 1`,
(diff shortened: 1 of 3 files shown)


packages/migrations/src/clickhouse-actions/012-coordinates-typename-index.ts
```diff
@@ -0,0 +1,52 @@
...
+    throw new Error('Expected 3 tables');
+  }
+
+  for (const { uuid, name } of tables) {
+    console.log(`Creating idx_typename for table ${name}`);
+    await exec(
+      `ALTER TABLE ".inner_id.${uuid}" ADD INDEX idx_typename (substringIndex(coordinate, '.', 1)) TYPE ngrambf_v1(4, 1024, 2, 0) GRANULARITY 1`,
+    );
+    const indexes = await query(`
+      SELECT name FROM system.data_skipping_indices WHERE table = ${'.inner_id.' + uuid} AND name = 'idx_typename'
```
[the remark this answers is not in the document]
kamilkisiela: The `sql` tag does exactly that, it extracts the raw values and replaces them with `{paramN:data-type}` format.

packages/migrations/src/clickhouse-actions/012-coordinates-typename-index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  if (tables.length !== 3) {
+    throw new Error('Expected 3 tables');
+  }
+
+  for (const { uuid, name } of tables) {
+    console.log(`Creating idx_typename for table ${name}`);
+    await exec(
+      `ALTER TABLE ".inner_id.${uuid}" ADD INDEX idx_typename (substringIndex(coordinate, '.', 1)) TYPE ngrambf_v1(4, 1024, 2, 0) GRANULARITY 1`,
+    );
```
n1ru4l, line 40: Would this be resilient enough for the migration to terminate unexpectedly and be restarted?

n1ru4l (approved):
approved with small question

dotansimha (approved):
lgtm. let's just make sure to address @n1ru4l 's comment before merging

packages/migrations/src/clickhouse-actions/012-coordinates-typename-index.ts
kamilkisiela: There's no way to reliably check if materialisation of an index was executed successfully

We can only create an index if not exists, other than that nothing else.

kamilkisiela: https://github.com/graphql-hive/console/pull/6536/commits/1768dfa94f777240d8de6795f0d02d0ee1c3e807

PR #6537, opened 2025-02-19
improve(integrations/slack): sentry alert on error

jasonkuhrt:
### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/web/app/src/lib/fetch-json.ts
+++ b/packages/web/app/src/lib/fetch-json.ts
@@ -1,0 +1,134 @@
+import { z } from 'zod';
+import { Kit } from './kit';
+
+export async function fetchJson<schema extends undefined | z.ZodType = undefined>(
+  url: string,
+  requestInit?: RequestInit,
+  schema?: schema,
+): Promise<
+  (schema extends z.ZodType ? z.infer<schema> : Kit.Json.Value) | FetchJsonErrors.FetchJsonErrors
+> {
+  const response = await fetch(url, requestInit)
+    // @see https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#exceptions
+    .catch(Kit.oneOf(error => error instanceof TypeError || error instanceof DOMException));
+
+  if (response instanceof TypeError) {
+    return new FetchJsonErrors.FetchJsonRequestTypeError({ requestInit }, response);
+  }
+
+  if (response instanceof DOMException) {
+    return new FetchJsonErrors.FetchJsonRequestNetworkError({}, response);
+  }
+
+  const json = await response
+    .json()
+    .then(value => value as Kit.Json.Value)
+    // @see https://developer.mozilla.org/en-US/docs/Web/API/Response/json#exceptions
+    .catch(
+      Kit.oneOf(
+        error =>
+          error instanceof SyntaxError ||
+          error instanceof TypeError ||
+          error instanceof DOMException,
+      ),
+    );
+
+  if (json instanceof DOMException) {
+    return new FetchJsonErrors.FetchJsonRequestNetworkError({}, json);
+  }
+
+  if (json instanceof TypeError) {
+    return new FetchJsonErrors.FetchJsonResponseTypeError({ response }, json);
+  }
+
+  if (json instanceof SyntaxError) {
+    return new FetchJsonErrors.FetchJsonResponseSyntaxError({ response }, json);
+  }
+
+  if (schema) {
+    const result = schema.safeParse(json);
+    if (!result.success) {
+      return new FetchJsonErrors.FetchJsonResponseSchemaError(
+        { response, json, schema },
+        result.error,
+      );
+    }
+    return result.data as any; // z.infer<Exclude<schema, undefined>>;
+  }
+
+  return json as any; // Kit.Json.Value;
+}
+
+// =================================
+// Error Classes
+// =================================
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
+export namespace FetchJsonErrors {
+  export type FetchJsonErrors = FetchJsonResponseErrors | FetchJsonRequestErrors;
+
+  // --------------------------------
+  // Response Error Classes
+  // --------------------------------
+
+  export type FetchJsonRequestErrors = FetchJsonRequestTypeError | FetchJsonRequestNetworkError;
+
+  export class FetchJsonRequestNetworkError extends Kit.Errors.ContextualError<
+    'FetchJsonRequestNetworkError',
+    {},
+    DOMException
+  > {
+    message = 'Network failure.';
+  }
+
+  export class FetchJsonRequestTypeError extends Kit.Errors.ContextualError<
+    'FetchJsonRequestTypeError',
+    { requestInit?: RequestInit },
+    TypeError
+  > {
+    message = 'Invalid request.';
+  }
+
+  // --------------------------------
+  // Response Error Classes
+  // --------------------------------
+
+  export abstract class FetchJsonResponseError<
+    $Name extends string,
+    $Context extends {
+      response: Response;
+    },
+    $Cause extends z.ZodError | SyntaxError | TypeError | DOMException,
+  > extends Kit.Errors.ContextualError<$Name, $Context, $Cause> {
+    message = 'Invalid response.';
+  }
+
+  export type FetchJsonResponseErrors =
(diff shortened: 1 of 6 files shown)


jasonkuhrt:
I don't know how to test Sentry yet in our test suites, will explore that next.

packages/web/app/src/lib/fetch-json.ts
```diff
@@ -0,0 +1,134 @@
...
+    if (!result.success) {
+      return new FetchJsonErrors.FetchJsonResponseSchemaError(
+        { response, json, schema },
+        result.error,
+      );
+    }
+    return result.data as any; // z.infer<Exclude<schema, undefined>>;
+  }
+
+  return json as any; // Kit.Json.Value;
```
jdolle, line 59: nit: prefer `unknown` to force us to use validation or cast

packages/web/app/src/lib/fetch-json.ts
```diff
@@ -0,0 +1,134 @@
...
+  // --------------------------------
+
+  export abstract class FetchJsonResponseError<
+    $Name extends string,
+    $Context extends {
+      response: Response;
+    },
+    $Cause extends z.ZodError | SyntaxError | TypeError | DOMException,
+  > extends Kit.Errors.ContextualError<$Name, $Context, $Cause> {
+    message = 'Invalid response.';
```
jdolle, line 103: Prefer if we can provide more insight than this. E.g. should we include the cause's error message?

packages/web/app/src/lib/kit/errors.ts
```diff
@@ -0,0 +1,33 @@
...
+    constructor(
+      ...args: undefined extends $Cause
+        ? [context: $Context, cause?: $Cause]
+        : [context: $Context, cause: $Cause]
+    ) {
+      const [context, cause] = args;
+      super('Something went wrong.', { cause });
+      this.name = this.constructor.name as $Name;
+      this.context = context;
+      this.cause = cause as $Cause;
```
jdolle, line 20: should the call to `super` set this?

packages/web/app/src/lib/kit/helpers.ts
```diff
@@ -8,3 +8,20 @@ export const tryOr = <$PrimaryResult, $FallbackResult>(
     return fallback();
   }
 };
+
+export const oneOf = <type extends readonly unknown[]>(
```
jdolle, line 12: nit: Please add a description block to make the behavior more obvious when we use this function.
When seeing it used, I wasnt 100% sure it threw, or what type of error it would throw.

packages/web/app/src/lib/kit/helpers.ts
```diff
@@ -8,3 +8,20 @@ export const tryOr = <$PrimaryResult, $FallbackResult>(
...
+  ...guards: OneOfCheck<type>
+): ((value: unknown) => type[number]) => {
+  return (value: unknown) => {
+    for (const guard of guards) {
+      if (guard(value)) {
+        return value;
+      }
+    }
+    throw new Error(`Unexpected value received by oneOf: ${value}`);
```
jdolle, line 21: should this be a unique error class?

packages/web/app/src/server/slack.ts
```diff
@@ -130,9 +182,65 @@ export function connectSlack(server: FastifyInstance) {
...
 }
+
+// =================================
+// Error Classes
+// =================================
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
+namespace SlackIntegrationErrors {
+  export class DisabledError extends Kit.Errors.ContextualError<'DisabledError'> {
+    message = 'The Slack integration is not enabled.';
```
jdolle, line 203: Are there any recommendations you could provide in this message for how to correct this if it needs correcting?
e.g. "set environment variable X" or "see <url>"

packages/web/app/src/server/slack.ts
```diff
@@ -11,81 +15,112 @@ const SlackIntegration_addSlackIntegration = graphql(/* GraphQL */ `
...
   server.get('/api/slack/callback', async (req, res) => {
+    const sentryTagsRouteLevel = {
+      ...sentryTagsComponentLevel,
+      route: '/api/slack/callback',
+    };
+
     if (env.slack === null) {
-      throw new Error('The Slack integration is not enabled.');
+      const error = new SlackIntegrationErrors.DisabledError({});
+      Sentry.captureException(error, { tags: sentryTagsRouteLevel });
```
jdolle, line 44: Do we have a request ID or any other identifiers that we could use to identify why/for who this is failing?

packages/web/app/src/server/slack.ts
```diff
@@ -11,128 +15,232 @@
...
+      for (const errorCause of error.cause.errors) {
+        req.log.error(errorCause);
       }
-      throw new Error('Failed setting slack token.');
+      throw error;
     }
 
-    void res.redirect(`/${organizationSlug}/view/settings`);
+    void res.redirect(`/${organizationId}/view/settings`);
+  });
```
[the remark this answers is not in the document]
jdolle: This suggestion is interesting. I wonder if we should throttle people eventually..

packages/web/app/src/lib/slack-api.ts
```diff
@@ -0,0 +1,73 @@
+import { stringify } from 'node:querystring';
+import { z } from 'zod';
+import { fetchJson } from './fetch-json';
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
+export namespace SlackAPI {
```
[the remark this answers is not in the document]
jasonkuhrt: I prefer using ESM but I received pushback using that approach on this repo. If the desired style of this repo is ultimately to use types and functions with stringly namespaced identifiers (e.g. `slackAPIauthorize(...)` instead of `SlackAPI.authorize(...)) then I wasn't aware of that, and I think the DX on that is worse. I usually trade producer complexity for consumer simplicity based on my feeling that complexity should go down, not up, the stack.

packages/web/app/src/lib/kit/helpers.ts
jasonkuhrt: Maybe! Its failure represents a defect not a normal error so we could encode that in a special class for sure.

packages/web/app/src/lib/fetch-json.ts
jasonkuhrt: It's all available through the chain data in a structured way. The missing part IMO is error renderers that are up to date with modern error types. Cause is part of the ECMA  standard so it's confusing to me that default renderings are still AFAIK failing to raise good context about them. We would just have to add our own where it is needed such as top level catch all or logging middleware, etc.

packages/web/app/src/lib/fetch-json.ts
jasonkuhrt: Since this is in the return position of an implementation it would not work and it does not matter, TS just doesn't make it possible to make implementations type safe. At best we could cast it to something matching the return type but its more maintenance for little gain and no increased type safety since it is a cast either way.

packages/web/app/src/server/slack.ts
jasonkuhrt: We receive a string from the Slack API, quite lame. I have not attempted to increase the quality here by parsing it, reviewing the Slack docs, if any exist for this string, etc. It would definitely be an improvement! Worth at least a TODO here I agree.

jasonkuhrt: Definitely worth putting a TODO about that here. We could write a catch all documentation here or link to some docs. Since this case is classified, however, upstream error rendering has the potential to switch on this and provide recommendations where it by definition has more context.

Error rendering is a really important thing that isn't addressed in this PR. For example that is where localization would be factored in, I think. Keeping strings out of error classes is generally good. But as you've pointed out in a few places, without good rendering, all that context and structure is just potential energy, not realized.

jasonkuhrt:
@jdolle thanks for the feedback. Feel free to take over this PR if you're interested! I'm currently on other tasks now without any plan to return here currently. If you see value here that's great, hope it helps! 🙏

Issue #124, opened 2022-05-29
Separate tokens permissions and allow to create operations reporting token only


dotansimha:

kamilkisiela:
1. Add usage:write scope to all existing tokens containing registry:write
2. Check for usage:write scope in the Usage service
3. Add usage:write scope to UI
4. Make sure we restart the Tokens service (to purge the cache)

This way we keep backwards-compatibility and still offer more granular access for new tokens

PR #6538, opened 2025-02-20
feat: report usage with organization access tokens

n1ru4l:
### Background

Re-use the new authentication/authorization layer introduced within our graphql API service within the usage service for reporting usage with organization access tokens.

### Description

- Adds redis and postgres connection to usage service
- Adds endpoint `/<target-uuid>` for reporting usage with an organization access token
- Adds endpoint `/<org-slug>/<target-slug>/<target-slug>` for reporting usage with an organization access token

https://linear.app/the-guild/issue/CONSOLE-955/support-organization-access-tokens-in-the-usage-reporting-service

Closes https://github.com/graphql-hive/console/issues/124

### Checklist

- [x] Input validation
- [x] Output encoding
- [x] Session management
- [x] Access control
- [x] Data protection
- [x] System configuration
- [x] Testing


tsconfig.json
```diff
@@ -26,16 +26,7 @@
...
-        "./packages/services/api/src/modules/schema/providers/artifact-storage-reader.ts"
-      ],
-      "@hive/api/src/modules/shared/providers/redis": [
-        "./packages/services/api/src/modules/shared/providers/redis.ts"
-      ],
-      "@hive/api/src/modules/shared/providers/pub-sub": [
-        "./packages/services/api/src/modules/shared/providers/pub-sub.ts"
-      ],
-
+      "@hive/api/*": ["./packages/services/api/src/*"],
```
n1ru4l, line 29: Allowing wildcard imports seemed to be more flexible until we figure out how we want to organize shared dependencies (e.g. break them into their own packages).

n1ru4l:
@coderabbitai review

packages/services/usage/src/environment.ts
```diff
@@ -65,6 +65,22 @@
...
 ]);
 
+const PostgresModel = zod.object({
+  POSTGRES_SSL: emptyString(zod.union([zod.literal('1'), zod.literal('0')]).optional()),
+  POSTGRES_HOST: zod.string(),
+  POSTGRES_PORT: NumberFromString,
+  POSTGRES_DB: zod.string(),
+  POSTGRES_USER: zod.string(),
+  POSTGRES_PASSWORD: emptyString(zod.string().optional()),
+});
```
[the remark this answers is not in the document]
n1ru4l: Some users that self-host this service do not use a postgres password. If we would make this required we would force people to use a password.

packages/services/usage/src/environment.ts
```diff
@@ -65,6 +65,22 @@ const KafkaModel = zod.union([
...
+
+const RedisModel = zod.object({
+  REDIS_HOST: zod.string(),
+  REDIS_PORT: NumberFromString,
+  REDIS_PASSWORD: emptyString(zod.string().optional()),
+  REDIS_TLS_ENABLED: emptyString(zod.union([zod.literal('1'), zod.literal('0')]).optional()),
+});
```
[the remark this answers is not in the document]
n1ru4l: Some users that self-host this service do not use a redis password. If we would make this required we would force people to use a password.

packages/services/usage/src/authn.ts
```diff
@@ -0,0 +1,37 @@
...
+  return new AuthN({
+    strategies: [
+      (logger: Logger) =>
+        new OrganizationAccessTokenStrategy({
+          logger,
+          organizationAccessTokensCache,
+          organizationAccessTokenValidationCache,
+        }),
+    ],
+  });
```
n1ru4l, line 36: I decided to only support the new organization access token strategy for the new endpoints (`/:uuid` / `/:orgSlug/:projectSlug/:targetSlug`).
However, we could also support the target access tokens here if needed.

When we make the documentation and SDK information public, people will be able to already create organization access tokens. Adding the option to still provide target access tokens here seems like something that will result in making their complete deprecation and removal happen later than expected.

deployment/services/usage.ts
```diff
@@ -39,60 +44,78 @@ export function deployUsage({
...
+      .withSecret('REDIS_HOST', redis.secret, 'host')
+      .withSecret('REDIS_PORT', redis.secret, 'port')
+      .withSecret('REDIS_PASSWORD', redis.secret, 'password')
+      // PG
+      .withSecret('POSTGRES_HOST', postgres.pgBouncerSecret, 'host')
+      .withSecret('POSTGRES_PORT', postgres.pgBouncerSecret, 'port')
+      .withSecret('POSTGRES_USER', postgres.pgBouncerSecret, 'user')
+      .withSecret('POSTGRES_PASSWORD', postgres.pgBouncerSecret, 'password')
+      .withSecret('POSTGRES_DB', postgres.pgBouncerSecret, 'database')
+      .withSecret('POSTGRES_SSL', postgres.pgBouncerSecret, 'ssl')
```
n1ru4l, line 112: Diff is a bit shitty here due to reformatting. These are the relevant lines that were added.

deployment/index.ts
```diff
@@ -163,6 +163,8 @@ const usage = deployUsage({
   docker,
   environment,
   tokens,
+  redis,
```
dotansimha, line 166: I wonder about adding these - is there a way we can avoid it? 

Redis should be fine, since it's in-cluster for now and it does not have issues, even if something external fails. 
But, Postgres means that now usage service depends on a new external thing, and I wonder if we should avoid that? (in the past, we did try to avoid introducing more complexity/deps to usage service) 

@n1ru4l @kamilkisiela thoughts?

n1ru4l: The usage service already independently depends upon postgres, but via the tokens service. I could implement this via the tokens service again and add more complexity. We decided to eventually kill the tokens service as it is unnecessary.

kamilkisiela: as long as we deduplicate `getOrSet` bento calls and have `redis` to share tokens between `usage` replicas, we're good

dotansimha (approved):
lgtm. let's wait also for Kamil's feedback before merging?

packages/services/usage/package.json
```diff
@@ -20,10 +22,12 @@
     "dotenv": "16.4.7",
     "got": "14.4.5",
     "graphql": "16.9.0",
+    "ioredis": "5.4.2",
     "kafkajs": "2.2.4",
     "lru-cache": "11.0.2",
     "p-limit": "6.2.0",
     "pino-pretty": "11.3.0",
+    "reflect-metadata": "^0.2.0",
```
jdolle, line 30: the rest are pinned to a specific version, do we want to do the same for this package?

packages/services/usage/src/authn.ts
jdolle: So long as we do not break existing behavior, and have a way for users to upgrade easily, I don't see issue with this.
But if we go that route, I think we will want to determine an appropriate timeline for deprecation and mention that in any error / in our docs where relevant

tsconfig.json
jdolle: I am not super strongly opinionated if this makes this easier for you, but I tend to prefer keeping a single entry point because I think the restrictions force you to make better decisions around package boundaries

jdolle (approved):

packages/services/usage/package.json
n1ru4l: I copied the line from api/package.json which was the only one that did not use a pinned version :D

I took care of it in https://github.com/graphql-hive/console/pull/6538/commits/ef77f6222544207df12d4a43a0dd1bc0431e1a22

n1ru4l:
@dotansimha 
image

PR #6540, opened 2025-02-20
Replaces rate-limit, stripe-billing and usage-estimator with commerce

kamilkisiela:
The `rate-limit`, `stripe-billing` and `usage-estimator` services are specific to Hive Cloud.
I'm merging them into one, called `commerce`.

I also merged `billing`, `rate-limit` and `usage-estimator` modules (in `@hive/api`) into `commerce`.

Everything stayed almost identical.

Closes CONSOLE-1083


packages/services/commerce/src/rate-limit/limiter.ts
```diff
@@ -1,11 +1,10 @@
...
+import type { Storage } from '@hive/api';
 import type { ServiceLogger } from '@hive/service-common';
 import { traceInline } from '@hive/service-common';
-import { createStorage as createPostgreSQLStorage, Interceptor } from '@hive/storage';
-import type { UsageEstimatorApi } from '@hive/usage-estimator';
 import * as Sentry from '@sentry/node';
-import { createTRPCProxyClient, httpLink } from '@trpc/client';
-import type { RateLimitInput } from './api';
+import { UsageEstimator } from '../usage-estimator/estimator';
+// import type { RateLimitInput } from './api';
```
jdolle, line 7: remove comment

packages/services/commerce/src/rate-limit/limiter.ts
```diff
@@ -294,7 +271,6 @@ export function createRateLimiter(config: {
         clearInterval(intervalHandle);
         intervalHandle = null;
       }
-      await (await postgres$).destroy();
```
jdolle, line 297: Why not this? It looks like postgres is still being used

kamilkisiela: it's part of global shutdown in `index.ts`

packages/services/server/README.md
```diff
@@ -10,7 +10,6 @@ The GraphQL API for GraphQL Hive.
 | `ENCRYPTION_SECRET`                                  | **Yes**                                            | Secret for encrypting stuff.                                                                  | `8ebe95cg21c1fee355e9fa32c8c33141`                        |
 | `WEB_APP_URL`                                        | **Yes**                                            | The url of the web app.                                                                       | `http://127.0.0.1:3000`                                   |
 | `GRAPHQL_PUBLIC_ORIGIN`                              | **Yes**                                            | The origin of the GraphQL server.                                                             | `http://127.0.0.1:4013`                                   |
-| `RATE_LIMIT_ENDPOINT`                                | **Yes**                                            | The endpoint of the rate limiting service.                                                    | `http://127.0.0.1:4012`                                   |
```
jdolle, line 13: Should this be replaced by one for the `COMMERCE_ENDPOINT`?

jdolle (commented):
In general this architectural change makes a lot of sense. There are a lot of details in this PR though.
And is there a migration plan or are we comfortable taking downtime?

packages/services/server/README.md
kamilkisiela: It's in the `Hive Cloud Configuration` below

kamilkisiela:
> In general this architectural change makes a lot of sense. There are a lot of details in this PR though. And is there a migration plan or are we comfortable taking downtime?

The only downtime (1) here could be stripe-billing / usage-estimator related. Very unlikely that somebody will make a request that involves one of them.
The rate-limit defaults to `not rate limited` so we're good here.

I think the downtime (1) is fine, otherwise we would have to do a 2 step rollout with 2 services running at the same time, not sure it's worth the hassle.

docker/docker.hcl
```diff
@@ -123,27 +123,6 @@ target "emails" {
   ]
 }
 
-target "rate-limit" {
```
dotansimha, line 126: I think `docker.community` file should also get updated?

docker/docker.hcl
```diff
@@ -123,27 +123,6 @@ target "emails" {
...
-  inherits = ["service-base", get_target()]
```
dotansimha, line 127: How do we deprecate these images after merging this PR?

docker/docker.hcl
dotansimha: ah, actually it was never part of that Compose, right?

dotansimha (approved):
1. changeset missing?
2. what do we do with existing images?
3. product update?

kamilkisiela:
> 1. changeset missing?
> 
> 2. what do we do with existing images?
> 
> 3. product update? 
> 
> 4. I'll need your help with updating our internal architecture diagrams and network diagrams, to match this new setup
> 
> 

None of these are user facing. It's part of Hive Cloud, not self-hosting. Why would we need a changeset here?

dotansimha:
Yeah I wrote that before and then realized it's all internals, good point. resolving

PR #6544, opened 2025-02-21
Docs/Gateway: Logging & Error Handling

ardatan:
Closes https://github.com/graphql-hive/gateway/issues/645
Completes GW-181
Completes GW-191


packages/web/docs/src/app/gateway/gateway-marquee-rows.tsx
```diff
@@ -40,7 +40,7 @@ const terms = new Map<string[], string /* href */>([
   [['Rate Limiting'], '/docs/gateway/other-features/security/rate-limiting'],
   [['Cost Limit'], '/docs/gateway/other-features/security/cost-limit'],
   [['Security'], '/docs/gateway/other-features/security'],
-  [['maskedErrors'], '/docs/gateway/other-features/security/error-masking'],
```
dotansimha, line 43: @ardatan i think this will be 404? can you add a redirect if it's needed?

packages/web/docs/src/content/gateway/logging-and-error-handling.mdx
```diff
@@ -0,0 +1,305 @@
...
+  "data": null
+}
+```
+
+<Callout>
+  When `INTERNAL_SERVER_ERROR` is passed from the subgraph, it is masked by the gateway and sent to
+  the client as `INTERNAL_SERVER_ERROR`. But the gateway will still log the original error.
+</Callout>
+
+#### Disable for development
```
dotansimha, line 252: I think development should actually be the same as prod. This is also what we do in Hive. 
Because of the frontend developer assumes some structure during it, it will be different in prod. 

@n1ru4l @kamilkisiela what do you think?

dotansimha (approved):
Looks great in general, added a few questions

packages/web/docs/src/content/gateway/logging-and-error-handling.mdx
```diff
@@ -0,0 +1,305 @@
+---
+title: Logging & Error Handling
```
dotansimha, line 2: My general take is that logging and error-handling are a bit different, but i do understand why it makes sense when i read the content :) thanks @ardatan

packages/web/docs/src/content/gateway/logging-and-error-handling.mdx
ardatan: It is already same by default in development. This section shows how to disable error masking for debugging purposes.

dotansimha: maybe we should note that this should be disabled only for debugging? and not for actual development? not sure how to phrase it 🤔

ardatan: I copied it from Yoga docs actually :/

ardatan: Improved it!

PR #6553, opened 2025-02-24
Standardize the design and content of all email templates

kamilkisiela:
We had HTML (supertokens related), mjml and very different design in both

I removed the image in the header, because it could affect deliverability of the emails (the email domain is different than the image's hostname).

| Before              | After |
| :---------------- | :------: |
| Screenshot 2025-02-24 at 15 26 38        |   Screenshot 2025-02-24 at 15 18 39   |


saihaj (commented):
can we use SVG for logo in the email?

kamilkisiela:
> can we use SVG for logo in the email?

Yes, for sure

dotansimha (approved):

n1ru4l (approved):

PR #6555, opened 2025-02-25
feat: Add metadata filter to explorer page

jdolle:
### Background

A continuation of https://github.com/graphql-hive/console/pull/6517 for https://github.com/graphql-hive/console/issues/2818

### Description

Adds a way to filter fields by their associated metadata.
Screenshot 2025-02-24 at 9 26 44 PM

The current implementation ORs the metadata filter, but AND may be more useful... I can see both being useful but don't think we should implement both at this time because it would complicate the UI.

### Checklist

- [x] Testing


n1ru4l:
@jdolle I am not sure about the accordion within the dropdown 🤔 

Maybe we should aim for something that looks more like the following?
image

What is the purpose of select all?

jdolle:
The purpose of select all was mainly to allow people to easily select all-but-one, or to reset the selection.
Having to click every option could be quite annoying, and then unselecting them would be similarly annoying.

From what I could find, having a select all which is also capable of deselecting all is a convenient way to help both of these cases.

jdolle:
I like search but now that you bring it up, I wonder if both search and the accordion are overkill at this time.
The goal was to make these groups easier to look through but we only know of a couple of use-cases at this time.

How would you feel about excluding search for now and perhaps in the future add it if there are more than some number of items in the list?

n1ru4l:
@jdolle I also think we do not need the search now. My main point is that the inclusion of collapsible makes the dropdown a bit messy.

What is the use-case of selecting all but one metadata filter? I am fine with having this filter if it is functionally necessary.

I was also thinking that we might have the Switch next to the category/label name, that when toggled checks/selects all values within the category? 🤔 Mainly, because I think it is cleaner.

jdolle:
Selecting all but one is an edge case. I started with a "reset" button but thought the "select all" would be more like other tools out there and would also provide the reset functionality.

Isolating this to a group still makes sense to me. I'm happy to try that.

jdolle:
I've adjusted the UI per your recommendation:
Screenshot 2025-02-26 at 4 37 09 PM

packages/web/app/src/components/target/explorer/object-type.tsx
```diff
@@ -34,7 +34,9 @@ export function GraphQLObjectTypeComponent(props: {
   const router = useRouter();
   const searchObj = router.latestLocation.search;
   const search =
-    'search' in searchObj && typeof searchObj.search === 'string' ? searchObj.search : undefined;
+    'search' in searchObj && typeof searchObj.search === 'string'
+      ? searchObj.search.toLowerCase()
```
jdolle, line 38: I modified the search behavior wherever I saw it, so that it's not case sensitive. Having a case sensitive search is typically very confusing UX

n1ru4l (approved):

PR #6558, opened 2025-02-27
fix: treat all environments without ENVIRONMENT being set to "development" as production

n1ru4l:
### Background

https://linear.app/the-guild/issue/CONSOLE-1086/graphql-introspection-query-enabled

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/server/src/index.ts
+++ b/packages/services/server/src/index.ts
@@ -443,7 +443,7 @@ export async function main() {
         connectionUri: env.supertokens.connectionURI,
         apiKey: env.supertokens.apiKey,
       },
-      isProduction: env.environment === 'prod',
+      isProduction: env.environment !== 'development',
       release: env.release,
       hiveConfig: env.hive,
       hivePersistedDocumentsConfig: env.hivePersistedDocuments,


packages/services/server/src/index.ts
```diff
@@ -443,7 +443,7 @@ export async function main() {
         connectionUri: env.supertokens.connectionURI,
         apiKey: env.supertokens.apiKey,
       },
-      isProduction: env.environment === 'prod',
+      isProduction: env.environment !== 'development',
```
n1ru4l, line 446: We specify `ENVIRONMENT=development` within our .env file for local development, so wont impact us locally.

n1ru4l: The environment values we use during deployment are:
- `prod`
- `dev`
- `staging`

I wonder whether we actually need them aside from analytics values and instead should make this a `GRAPHQL_INTROSPECTION_ENABLED` flag instead...

kamilkisiela (approved):

PR #6561, opened 2025-02-27
fix: members page with member:modify

n1ru4l:
### Background

https://linear.app/the-guild/issue/CONSOLE-1097/dashboard-membermodify-permission-error

### Description

The dashboard shows an error page when having `member:describe` but not `member:modify` permissions.

image

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/organization/providers/organization-manager.ts
+++ b/packages/services/api/src/modules/organization/providers/organization-manager.ts
@@ -239,7 +239,7 @@ export class OrganizationManager {
 
   @cache((selector: OrganizationSelector) => selector.organizationId)
   async getInvitations(selector: OrganizationSelector) {
-    await this.session.assertPerformAction({
+    const canAccessInvitations = await this.session.canPerformAction({
       action: 'member:modify',
       organizationId: selector.organizationId,
       params: {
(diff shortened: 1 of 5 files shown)


packages/services/api/src/modules/organization/providers/organization-manager.ts
```diff
@@ -239,15 +239,19 @@ export class OrganizationManager {
...
         organizationId: selector.organizationId,
       },
     });
 
-    return this.storage.getOrganizationInvitations(selector);
+    if (!canAccessInvitations) {
+      return null;
+    }
+
+    return await this.storage.getOrganizationInvitations(selector);
```
n1ru4l, line 254: This is a pragmatic solution to fix this problem. 
Alternatively, we could ignore the error on the frontend, but that would require more logic there for error handling.

Thinking of the upcoming public API, we might even want to have a "throw" field for the public API, but a "just return null" field for our dashboard API...

n1ru4l: I added a reference for this to https://www.notion.so/theguildoss/Hive-GraphQL-API-Design-Patterns-18ab6b71848a80678939c669699fe8e6?pvs=4#1a7b6b71848a807aa645eee88f98c93a

jdolle (approved):
It'd be nice if we can add a unit test.

n1ru4l:
@jdolle I added integration tests

Issue #6249, opened 2025-01-02
Improve DB index for `getMaybeLatestVersion` database query 


dotansimha:
The following PG query has a missing PG index, and sometimes it's running slow (~6-10s):
https://github.com/graphql-hive/console/blob/5e94f5dee5359b6e2421a2671d09dea5ca2ff72d/packages/services/storage/src/index.ts#L2181-L2199

We should follow the same approach as in https://github.com/graphql-hive/console/pull/6158 and   https://github.com/graphql-hive/console/pull/6154 , and add a new index for this query.

jdolle:
```
sql`/* getMaybeLatestVersion */
          SELECT
            ${schemaVersionSQLFields(sql`sv.`)}
          FROM schema_versions as sv
          LEFT JOIN targets as t ON (t.id = sv.target_id)
          WHERE sv.target_id = ${args.targetId} AND t.project_id = ${args.projectId}
          ORDER BY sv.created_at DESC
          LIMIT 1
        `,
```

If we're already matching on the `target_id`, do we need to match `project_id`?
And if we then are relying solely on SV, it should be able to leverage the `schema_versions_cursor_pagination` index.
```
"schema_versions_cursor_pagination" ON "schema_versions" (
          "target_id" ASC
          , "created_at" DESC
          , "id" DESC
        );
      `,
```

I need to do some testing against some real data though. My local plants aren't helpful.

PR #6562, opened 2025-02-27
Add index get maybe latest version

egoodwinx:
### Background

Addressing: https://github.com/graphql-hive/console/issues/6249
Added index for getMaybeLatestVersion 

### Description

Created postgres migration for schema_versions on target_id sorted by created_id.

Ran `db:migrator` and it ran successfully


n1ru4l:
Running the following query with `SET enable_seqscan = OFF;`

```
explain SELECT 
  sv.*
FROM schema_versions as sv 
LEFT JOIN targets as t ON (t.id = sv.target_id) 
WHERE sv.target_id = gen_random_uuid() AND t.project_id = gen_random_uuid()
ORDER BY sv.created_at DESC 
LIMIT 1 
```

Without this index:

```
Limit  (cost=61.67..61.68 rows=1 width=379)
  ->  Sort  (cost=61.67..61.68 rows=1 width=379)
        Sort Key: sv.created_at DESC
        ->  Nested Loop  (cost=0.29..61.66 rows=1 width=379)
              ->  Index Scan using schema_versions_target_id on schema_versions sv  (cost=0.14..48.15 rows=1 width=379)
                    Filter: (target_id = gen_random_uuid())
              ->  Index Scan using targets_pkey on targets t  (cost=0.15..8.17 rows=1 width=16)
                    Index Cond: (id = sv.target_id)
                    Filter: (project_id = gen_random_uuid())
```

after:

```
Limit  (cost=61.67..61.68 rows=1 width=379)
  ->  Sort  (cost=61.67..61.68 rows=1 width=379)
        Sort Key: sv.created_at DESC
        ->  Nested Loop  (cost=0.29..61.66 rows=1 width=379)
              ->  Index Scan using idx_maybe_latest_version_id on schema_versions sv  (cost=0.14..48.15 rows=1 width=379)
                    Filter: (target_id = gen_random_uuid())
              ->  Index Scan using targets_pkey on targets t  (cost=0.15..8.17 rows=1 width=16)
                    Index Cond: (id = sv.target_id)
                    Filter: (project_id = gen_random_uuid())
```

I am not sure this index actually improves the performance here.

dotansimha:
As @n1ru4l mentioned here, we are not sure this index actually works. I tried to add it in another PR but it failed to improve as well.

jdolle:
I believe this can be removed. We've remediated indices elsewhere and the slowness of maybeLatestVersion doesn't seem related to the index.

PR #6569, opened 2025-02-28
docs: add hive link docs

jdolle:
### Background

Metadata support was recently added via a linked directive, `@meta`.

### Description

This adds a page to our existing documentation that describes our Hive link spec. Rather than create an entirely new set of docs we have to maintain, I opted to use the existing docs. We can and should then create a redirect from `https://specs.graphql-hive.com/hive/*` to this page.


packages/web/docs/src/content/specs/link-specs.mdx
```diff
@@ -0,0 +1,87 @@
...
+## Features
+
+### @meta
+
+<Tabs items={["v1.0"]}>
+<Tabs.Tab>
+
+Use this directive to enhance schema types with additional data, visible via the Hive Console. _This
+intentionally does not appear in the API Schema to avoid bloat. But if you want this directive to be
+returned to your gateway, then you can use `@composeDirective`_
```
n1ru4l, line 56: Should there be an example of using `@composeDirective`?

jdolle: Good question. I could definitely provide one. Based on chatting with Kamil, I should maybe just not mention this in the docs. Or leave it at composeDirective being unnecessary.
The reason I mentioned it was in case teams want to piggyback off this directive to write other custom plugins in the gateway. But it may be too early to say if that would happen.

n1ru4l (approved):

packages/web/docs/src/content/specs/link-specs.mdx
n1ru4l, line 1: Nit: we can use `link-specifications` instead of `link-specs`. I see no need to abbreviate the URL

jdolle: I've replaced all instances of `link-specs` with `link-specifications`

PR #6570, opened 2025-02-28
fix: show meta filter on explorer only if there is meta

jdolle:
### Background

Small bug on the metadata filter where when metadata is all removed from a subgraph, an empty array is returned from the API which causes the filter to still show even though it's empty.

### Description

This correctly hides the metadata filter on the explorer if there are no values in the metadata returned.


dotansimha (approved):

dotansimha:
oopsi, didn't see the auto-merge, was it ready to merge @jdolle ? 

cc @n1ru4l @kamilkisiela

PR #6572, opened 2025-03-03
docs(instrumentation): Add documentation for new Instrumentation API

EmrysMyrddin:


packages/web/docs/src/content/gateway/other-features/custom-plugins.mdx
```diff
@@ -826,6 +826,138 @@ Prefer `onRequestParse` when possible, or wrap the hook code in a `try` block.
 | `serverContext` | The final context object that is shared between all hooks and the GraphQL execution. [Learn more about the context](https://the-guild.dev/graphql/yoga-server/docs/features/context#server-context). |
 | `response`      | The outgoing HTTP response as WHATWG `Response` object. [Learn more about the response interface](https://developer.mozilla.org/en-US/docs/Web/API/Response).                                        |
 
+### `instrumentation`
```
dotansimha, line 829: Maybe add a general overview of when to use instrumentation? (tracing, ...) 
Should we mention that in most cases, this is a very advanced use-case?

packages/web/docs/src/content/gateway/other-features/custom-plugins.mdx
```diff
@@ -826,6 +826,138 @@ Prefer `onRequestParse` when possible, or wrap the hook code in a `try` block.
...
+An optional `instrumentation` instance can be present in the plugin.
+
+This `Instrumentation` instance allows to wrap an entire phase execution (including all plugin
+hooks), meaning running code just before, just after and around the execution of the phase.
+
+Instrumentation doesn't have access to input/output of a phase, use hooks to have access to those
+data. If needed, we recommend to share data between instrumentation and hooks with a `WeakMap` and
+the given `context` as the key.
+
+All instrumentation takes 2 parameters:
```
dotansimha, line 840: Hmm, we only have an example here for composition, and not for basic instrumentation. should we add one?

PR #6573, opened 2025-03-04
feat: organization access token error handling

n1ru4l:
### Background

https://linear.app/the-guild/issue/CONSOLE-993/prepare-cli-for-passing-target-slug-to-server

### Description

Show proper error messages on the CLI level if `--target` is missing for organization access tokens.

### Checklist

- [x] Input validation
- [x] Testing


packages/services/api/src/modules/schema/resolvers/Query/latestValidVersion.ts
```diff
@@ -9,24 +8,23 @@ export const latestValidVersion: NonNullable<QueryResolvers['latestValidVersion'
...
-        throw new InsufficientPermissionError('project:describe');
-      },
-    });
+  const selector = await injector.get(IdTranslator).resolveTargetReference({
+    reference: args.target ?? null,
+  });
+
+  if (!selector) {
+    session.raise('project:describe');
+  }
```
n1ru4l, line 17: For some reason this does not make `selector` non-null below, causing a TS error...

The same code works fine in other parts...

packages/libraries/cli/src/helpers/errors.ts
```diff
@@ -137,6 +137,16 @@ export class MissingCdnEndpointError extends HiveCLIError {
   }
 }
 
+export class MissingTargetError extends HiveCLIError {
+  constructor() {
+    super(
+      ExitCode.ERROR,
+      errorCode(ErrorCategory.GENERIC, 9),
+      `A target is required to perform the action. Please provide the "--target" flag.`,
```
jdolle, line 145: I would recommend the `MissingArgumentsError`.
This is a generic error class that can be passed information about the target argument to better instruct users.

n1ru4l: good catch 🥳 changed in https://github.com/graphql-hive/console/pull/6573/commits/e84c7f387fa6414cc8f840178a1ed2755432f027

PR #6582, opened 2025-03-05
feat: accept url in schema check

### LINKED ISSUE #3979: allow service url changes for schema check
Right now service URL changes are only shown for schema publishes. We should also allow service URL changes for schema checks if users provide a service URL parameter.

jdolle:
### Background

Resolves #3979

### Description

Adds optional url argument to schema check input.
This is backwards compatible so that users can opt to pass in a url or not depending on their wants.

Documentation will follow as a separate PR.

### Checklist

- [x] Testing


n1ru4l (changes requested):
we need a changeset here 😅

n1ru4l (approved):

Issue #6423, opened 2025-01-27
sanitize/validate service name input


n1ru4l:
We should validate and sanitize the service name for schema publishes and schema checks. Right now the service string can be anything, including whitespace and special characters. 

We should limit it more to have:
- maximum length
- disallow spaces
- disallow certain special characters

We need to consider already existing services published to our database and ensure that introducing the limits does not break any existing users.

Characters used within our pg database today:

```
 "-./0123456789_abcdefghijklmnopqrstuvwxyz{}“”
```

<details><summary>how we extracted the list</summary>
<p>

```
select string_agg(c,'')
from (
  select distinct regexp_split_to_table(lower(service_name),'') as c
  from schema_log
) t
```

</p>
</details> 

Some of those look like unintended user mistakes when publishing the schema:
- `{` and `}` seem like wrongly interpolated environment variables on CI systems (projects using this are not active; or all services deleted)
- `”` and `”` also seem like wrongly interpolated environment variables on CI systems (project using this are not active; or all services deleted)

Proposed allowed characters:
- alphanumeric
- dots
- underscore
- slash
- dash

jdolle:
For now I think we should only validate when creating a new service. We can then work to migrate existing services if we need to. But I don't want to break anyone's services if for some reason they are using one of these characters.

PR #6585, opened 2025-03-05
feat: restrict service creation by characters and length

jdolle:
### Background
https://github.com/graphql-hive/console/issues/6423

### Description

Ideally this validation should exist on the model but if we did that, we'd start rejecting services that already exist.
This is at least a start to restricting services to better names.

Modifying documentation to mention this new restriction will come as a followup PR.

### Checklist

- [x] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -1040,6 +1040,23 @@ export class SchemaPublisher {
       this.schemaManager.getMaybeLatestVersion(target),
     ]);
 
+    if (!latestVersion && input.service) {
```
n1ru4l, line 1043: I think we would rather have to check whether the latestVersion has a service that is named like this service? 🤔

n1ru4l: There would also be an edge case, where someone deletes a service (`hive schema:delete`) and tries to republish it, but then it fails, because it would no longer be in the previous version 🤔

jdolle: Youre right about the latestVersion. I'll fix that.

For the edge case, I think we can ignore that. If they are deleting then we can treat like new.

packages/services/api/src/modules/schema/providers/schema-helper.ts
```diff
@@ -37,7 +37,7 @@ export function ensureSingleSchema(schema: Schema | Schema[]): SingleSchema {
   throw new Error('Expected a single schema');
 }
 
-export function ensureCompositeSchemas(schemas: readonly Schema[]): CompositeSchema[] | never {
```
jdolle, line 40: I assumed this `never` was because the function threw an array, but it only filters. So I'm not sure why this never is here... removed.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
jdolle: Added validation to schema check also. This will prevent us from having issues where the check passes but push fails because of the service name.

n1ru4l (changes requested):
Lets add a test here

n1ru4l:
I am not sure whether we want to allow `/`.

My main argument for this is that our resource identifiers use the `/` character to separate the resource partials.

Examples:

```
target/${args.targetId}/appDeployment/${args.appDeploymentName}`
target/${args.targetId}/service/${args.serviceName}
```

I think it is very unlikely that service name or app deployment will have a "sub" resource e.g.

```
target/${args.targetId}/service/${args.serviceName}/foobars
```

But if we would introduce something like this, it would require additional adjustments if the `/` character is allowed in service names (e.g. url encode the service name).

Thoughts?

jdolle:
That makes sense to me. It seems like best practices to keep the character list to a minimum also to avoid edge cases such as this. So I'll adjust that.

n1ru4l (approved):
Nice 🥳

PR #6586, opened 2025-03-06
Ensure type extension fields are properly considered in unused schema detection

kamilkisiela:
The logic that converts SDL into a GraphQL schema representation was ignoring fields from type extensions. Since this representation is what gets fed into the resolvers, those fields were never rendered as unused parts of the schema in the app. It only affected monolithic schemas (`single` projects).

I used `@graphql-tools/merge` to merge type extensions into their base types. Now, type extensions aren't left behind.

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-version-helper.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-version-helper.ts
@@ -135,7 +135,7 @@ export class SchemaVersionHelper {
       'SchemaVersionHelper.getCompositeSchemaAst: Composite',
     );
 
-    return this.autoFixCompositeSchemaAst(compositeSchemaAst);
+    return compositeSchemaAst;
   }
 
   @cache<SchemaVersion>(version => version.id)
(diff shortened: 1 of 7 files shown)


packages/services/api/src/modules/schema/providers/schema-version-helper.ts
```diff
@@ -332,8 +333,29 @@ export class SchemaVersionHelper {
     return schemaVersion.isComposable && schemaVersion.hasContractCompositionErrors === false;
   }
 
+  /**
+   * There's a possibility that the composite schema contains type extensions.
+   * This is a problem, because other parts of the system may expect it to be clean from type extensions.
+   *
+   * This function will also check for type system extensions and merge them into matching definitions.
+   */
+  private autoFixCompositeSchemaAst(schemaAst: DocumentNode): DocumentNode {
```
jdolle, line 342: I think it's a little confusing having `autoFixCompositeSchemaSdl` and `autoFixCompositeSchemaAst` but they do very different things.

kamilkisiela: I only followed the convention there. The original autoFix* is specific to federation, but in future it could do many other things.

kamilkisiela: @jdolle I removed the new method and apply its logic to the original one. This way whenever we access composite schema, we will look for type extensions.

n1ru4l (approved):

PR #6588, opened 2025-03-06
Prevent compositeSchemaSdl from having type extensions

kamilkisiela:
Follow-up https://github.com/graphql-hive/console/pull/6586 to apply the merging logic to the schema that's written to db.

Proof it works when the post-fix (#6586) is disabled - https://github.com/graphql-hive/console/actions/runs/13705781167


jdolle (approved):
Ty :)

PR #6594, opened 2025-03-07
fix: if insights start and end is the same, then set the range to tha…

jdolle:
…t entire day

### Background

https://linear.app/the-guild/issue/CONSOLE-1007/absolute-date-range-issue-in-hive-insights-when-setting-start-and-end

### Description

The most intuitive solution I could come up with is to set the window to the entire day that is selected.
Another option is that we prevent users from selecting the same day, but I think this solution provides the least friction and best experience.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/.changeset/big-cars-tie.md
+++ b/.changeset/big-cars-tie.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+Fix insights range if selecting same start and end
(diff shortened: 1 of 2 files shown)


kamilkisiela (approved):

n1ru4l (approved):

n1ru4l:
@jdolle Changeset!

Issue #6464, opened 2025-02-04
show CLI publish command within dashboard


n1ru4l:
> Also having a set of commands displayed for publishing your first schema would be helpful:
<img width="901" alt="Screenshot 2025-02-03 at 8 40 50 AM" src="https://github.com/user-attachments/assets/ec7c8d05-e0e7-4ec3-8515-5fe6fde7f006" />

_Originally posted by @jdolle in https://github.com/graphql-hive/console/pull/6449#discussion_r1939704323_

https://github.com/graphql-hive/console/pull/6449

PR #6602, opened 2025-03-11
feat: add directions for publishing to no schema component

jdolle:
### Background

Closes https://github.com/graphql-hive/console/issues/6464

### Description

Adds publish instructions, which are specific to the project type.


n1ru4l:
We could use a component that is more similar to our documentation code block.

My main critique here:
- no mono-spaced font
- on schema tab we should show publish command
- on checks tab we should show the check command


image


(ignore content the content of the code block)

jdolle:
I didnt see any existing component that matched this aesthetic. But I can add monospacing to the input.

Regarding check vs publish, I think on the check page I'd be okay showing just the check command, but for the schema or explorer, I'd prefer showing both check and publish. I think we should encourage best practices and checking the schema prior to publishing makes the most sense.

jdolle:
I've adjusted this to use the `InlineCode` component and separated the check and publish steps.

Screenshot 2025-03-11 at 12 52 35 PM
Screenshot 2025-03-11 at 12 52 43 PM

n1ru4l:
`InlineCode` is kind of a relict from the past that we want to stop using (like most components within `packages/web/app/src/components/v2`. 😅

We can create the component with that aesthetic based on https://github.com/search?q=repo%3Ashuding%2Fnextra%20nextra-code&type=code

jdolle:
Screenshot 2025-03-12 at 11 05 39 PM

packages/web/app/src/components/ui/empty-list.tsx
```diff
@@ -43,13 +48,61 @@ export const noSchema = (
...
+    if (recommendedAction === 'check') {
+      children = (
+        <>
+          <div className="flex w-full justify-center py-2 text-xs text-gray-500">
+            It's recommended to check that the schema is valid and compatible with the state of the
+            registry before publishing.
+          </div>
+          <div className="flex w-full justify-center">
+            <Code>
+              {`hive schema:check ${isDistributed ? '--service <service-name> --url <url> ' : ''}<path/schema.graphql>`}
```
n1ru4l, line 72: Now that we have the `--target` parameter mandatory for the new access tokens, we should probably also include it here? 🤔 

```
hive schema:check --target "org/project/target ...."
```

https://the-guild.dev/graphql/hive/docs/migration-guides/organization-access-tokens#performing-actions-with-the-cli

n1ru4l (approved):
looking good!

We could improve a bit here tho
https://github.com/graphql-hive/console/pull/6602#discussion_r2005014515

PR #6609, opened 2025-03-11
Mark usage-ingestor as unhealthy when consumer crashes

kamilkisiela:
Recover when it successfully fetches messages from the broker.

Closes CONSOLE-1129


saihaj (approved):
For the pod to auto restart we can just exit process with -1?

PR #6614, opened 2025-03-12
feat(api): filter private schema into public schema using federation contracts

n1ru4l:
### Background

Implementation for CONSOLE-948 using apollo federation and our existing schema filter logic based on tags.

Completes CONSOLE-948 

### Description

We need to run the composition filtering etc. on start-up since self-hosters can not load the schema from the actual Hive CDN.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -290,7 +290,7 @@ export default gql`
     """
     Whether the CLI supports retrying the schema publish, in case acquiring the schema publish lock fails due to a busy queue.
     """
-    supportsRetry: Boolean = False
+    supportsRetry: Boolean = false
```
n1ru4l, line 293: `False` is not valid according to apollo federation

packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -125,7 +125,7 @@ export default gql`
     period: Int!
     percentage: Float!
     requestCount: Int! = 1
-    breakingChangeFormula: BreakingChangeFormula! = PERCENTAGE
+    breakingChangeFormula: BreakingChangeFormula!
```
n1ru4l, line 128: This caused a composition error.

n1ru4l: fixed via https://github.com/the-guild-org/federation/pull/87

packages/services/api/src/modules/schema/module.graphql.ts
n1ru4l: and also according to the graphql spec :D

packages/services/api/src/modules/shared/module.graphql.ts
```diff
@@ -7,12 +7,22 @@ export default gql`
...
 
+  extend schema
+    @link(url: "https://specs.apollo.dev/link/v1.0")
+    @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@tag"])
+
+  directive @link(url: String!, import: [String!]) repeatable on SCHEMA
+
+  directive @tag(
+    name: String!
+  ) repeatable on FIELD_DEFINITION | INTERFACE | OBJECT | UNION | ARGUMENT_DEFINITION | SCALAR | ENUM | ENUM_VALUE | INPUT_OBJECT | INPUT_FIELD_DEFINITION | SCHEMA
```
n1ru4l, line 18: We turn our graphql schema into an apollo subgraph, so we can run our federation logic for filtering the schema on it :D

packages/services/api/src/modules/shared/module.graphql.ts
```diff
@@ -7,12 +7,22 @@ export default gql`
...
+
   type Query {
-    noop: Boolean
+    _: Boolean @tag(name: "public")
   }
 
   type Mutation {
-    noop(noop: String): Boolean
+    _: Boolean @tag(name: "public")
```
jdolle, line 25: I'm not very familiar with the contract logic.
if the schema is not "public", then is this mutation empty or does the entire Mutation type get removed?
If empty, then this will break. Types must have at least one field.

packages/services/server/src/public-graphql-handler.ts
```diff
@@ -0,0 +1,106 @@
...
+  logger: FastifyBaseLogger;
+  authN: AuthN;
+  hiveUsageConfig: HiveUsageConfig;
+  tracing?: TracingInstance;
+};
+
+export const createPublicGraphQLHandler = (
+  args: CreatePublicGraphQLHandlerArgs,
+): RouteHandlerMethod => {
+  const publicSchema = createPublicGraphQLSchema<Context>(args.registry);
```
jdolle, line 32: I think this flow is a bit confusing when combined with our schema registry.

I saw you upgrading to a federated schema and was thinking we'd use our schema registry to do the filtering and then expose that version of the schema in this service, rather than do the filtering in this service. But this makes more sense with a federated schema where the supergraph does validation to prevent executing private queries in subgraphs...

n1ru4l: If we use the schema registr for filtering, we have a problem for self-hosters as they would have to retrieve the public api schema from somewhere.

We do the filtering on startup, in addition we will later use the contracts on Hive for CI/CD checks.

packages/services/api/src/modules/shared/module.graphql.ts
n1ru4l: It will cause a composition error, that is why I added this placeholder field here. We will add mutation fields to the public api soon, and we can then remove the placeholder fields.

packages/services/api/src/modules/shared/resolvers/Mutation/_.ts
```diff
@@ -0,0 +1,5 @@
+import type { MutationResolvers } from './../../../../__generated__/types';
+
+export const _: NonNullable<MutationResolvers['_']> = async (_parent, _arg, _ctx) => {
```
kamilkisiela, line 3: I love it

kamilkisiela (approved):

Issue #6565, opened 2025-02-28
Flag for treating adding union and enum members as breaking with conditional breaking changes


n1ru4l:
**Report by customer**:
> In our system, some changes that aren't _technically breaking_, like adding to an existing enum or union value, can cause issues in our downstream services.

stephenh:
Hi all! Just adding some rationale for this...

We have several TypeScript frontends that pull in our GraphQL schema, and when we have enums, we prefer to use "exhaustive enum checks", i.e. things like:

```ts
enum Color {
  Red = "red",
  Green = "green",
  Blue = "blue",
}

function assertNever(value: never): never {
  throw new Error(`Unhandled case: ${value}`);
}

function getColorMessage(color: Color): string {
  if (color === Color.Red) {
    return "You selected Red!";
  } else if (color === Color.Green) {
    return "You selected Green!";
  } else if (color === Color.Blue) {
    return "You selected Blue!";
  }

  return assertNever(color); // Ensures exhaustiveness checking
}

// Example usage
const myColor: Color = Color.Red;
console.log(getColorMessage(myColor));
```

And so when our backends add a new enum value, this is not technically breaking from a wire protocol perspective, but it does risk causing both compile time errors (when the frontends pull in the next version of the GQL schema) & and also runtime errors (the `throw new Error(handled)`) in existing clients.

Given the severity of both of these issues, we'd like to use graphql-hive/console to track these as breaking changes. Thanks!

jdolle:
I'll be taking on this work. We have a few options for how to prevent this issue, but the proposed approach of requiring schema approval is the most intuitive for now.

jdolle:
After reviewing the source code and possible conditions that could cause issues, I've found overlap with the "dangerous" changes flagged by the `graphql-inspector` logic.

I think it'd be best to make the option "Require approval for dangerous changes". So it will encompass more than just enums. I think this is the safest approach and lets us avoid duplicating a lot of the logic that's done by the inspector already.

blimmer:
That sounds like an excellent solution to this problem. Offloading the behavior to the "dangerous" rules upstream sounds like a very reasonable option 👍

jdolle:
I've completed work on this feature. It will go out with our next release. Please reach out if this doesn't meet your expectations or if you have other ideas for improvements :)

stephenh:
Amazing! Thank you @jdolle !

blimmer:
[...]

PR #6626, opened 2025-03-18
feat: Add setting to make dangerous changes into breaking

jdolle:
### Background

Resolves https://github.com/graphql-hive/console/issues/6565

### Description

This adds a simple toggle to the target's breaking change settings to make any dangerous change turn into a breaking change -- thus requiring schema approval.

This PR includes updated documentation.

Screenshot 2025-03-18 at 1 03 08 PM

### DIFF
--- a/packages/migrations/src/actions/2025.03.20T00-00-00.dangerous_breaking.ts
+++ b/packages/migrations/src/actions/2025.03.20T00-00-00.dangerous_breaking.ts
@@ -1,0 +1,12 @@
+import type { MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.03.20T00-00-00.dangerous_breaking.ts',
+  noTransaction: true,
+  run: ({ sql }) => sql`
+    ALTER TABLE
+      targets
+    ADD COLUMN
+      consider_dangerous_breaking BOOLEAN NOT NULL DEFAULT FALSE;
+  `,
+} satisfies MigrationExecutor;
--- a/packages/services/api/src/modules/schema/providers/registry-checks.ts
+++ b/packages/services/api/src/modules/schema/providers/registry-checks.ts
@@ -2,7 +2,7 @@ import { URL } from 'node:url';
 import type { GraphQLSchema } from 'graphql';
 import { Injectable, Scope } from 'graphql-modules';
 import hashObject from 'object-hash';
-import { CriticalityLevel } from '@graphql-inspector/core';
+import { CriticalityLevel, Rule } from '@graphql-inspector/core';
 import type { CheckPolicyResponse } from '@hive/policy';
 import type { CompositionFailureError, ContractsInputType } from '@hive/schema';
 import { traceFn } from '@hive/service-common';
@@ -425,6 +425,7 @@ export class RegistryChecks {
     approvedChanges: null | Map<string, SchemaChangeType>;
     /** Settings for fetching conditional breaking changes. */
     conditionalBreakingChangeConfig: null | ConditionalBreakingChangeDiffConfig;
+    considerDangerousToBeBreaking: null | boolean;
   }) {
     let existingSchema: GraphQLSchema | null = null;
     let incomingSchema: GraphQLSchema | null = null;
@@ -462,7 +463,11 @@ export class RegistryChecks {
       } satisfies CheckResult;
     }
 
-    let inspectorChanges = await this.inspector.diff(existingSchema, incomingSchema);
+    let inspectorChanges = await this.inspector.diff(
+      existingSchema,
+      incomingSchema,
+      args.considerDangerousToBeBreaking ? [dangerousBreaking] : undefined,
+    );
 
     // Filter out federation specific changes as they are not relevant for the schema diff and were in previous schema versions by accident.
     if (args.filterOutFederationChanges === true) {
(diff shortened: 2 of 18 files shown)


packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -20,6 +20,10 @@ export default gql`
     updateTargetGraphQLEndpointUrl(
       input: UpdateTargetGraphQLEndpointUrlInput!
     ): UpdateTargetGraphQLEndpointUrlResult!
+    updateTargetDangerousChangeClassification(
+      target: TargetReferenceInput!
```
jdolle, line 24: I recall talk of moving to using the UUIDs, but existing mutations use the selectors.
I assumed the longer goal was to allow both everywhere using this `TargetReferenceInput`?
If not, I can adjust this.

jdolle:
I am considering writing a cypress test for the frontend but i dont see very many cypress tests. Would we prefer something else?

packages/services/api/src/modules/target/module.graphql.ts
n1ru4l: We always use a input type named after the field for mutation fields.

```suggestion
      input: UpdateTargetDangerousChangeClassificationInput!
```

```
input UpdateTargetDangerousChangeClassificationInput {
  target: TargetReferenceInput!
}
```

packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -45,6 +49,23 @@ export default gql`
     graphqlEndpointUrl: String
   }
 
+  input UpdateTargetDangerousChangeClassificationInput {
+    considerDangerousToBeBreaking: Boolean!
+  }
```
n1ru4l, line 54:
```suggestion
  input UpdateTargetDangerousChangeClassificationInput {
    target: TargetReferenceInput!
    considerDangerousToBeBreaking: Boolean!
  }
```

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -854,3 +859,19 @@ function isFederationRelatedChange(change: SchemaChangeType) {
...
+        criticality: {
+          ...change.criticality,
+          level: CriticalityLevel.Breaking,
+        },
+      };
+    }
+
+    return change;
+  });
+};
```
n1ru4l, line 877: Should we really change the criticality of the schema?

Instead we could also adjust the check whether the schema check is a failure? https://github.com/graphql-hive/console/blob/cbf8b935f0e4b1549f27f414999b82c3c9166cfa/packages/services/api/src/modules/schema/providers/registry-checks.ts#L538-L567

Have you thought about it and wighted the pros and cons of these two different approaches?

jdolle: I was initially thinking it would be very similar. But I do have concerns with my approach in that it will make it hard to distinguish them in the UI.
I like this recommendation, I think it will do exactly what we need while not making the change types as ambiguous.

jdolle: I've adjusted the implementation and terminology in our UI and documentation.
Adjusting the variable names and db column now. But I'm considering this resolved for all intensive purposes.

jdolle:
I've adjusted this per your feedback @n1ru4l and opened a PR for the relevant product update https://github.com/graphql-hive/console/pull/6631

packages/migrations/src/actions/2025.03.20T00-00-00.dangerous_breaking.ts
```diff
@@ -0,0 +1,12 @@
+import type { MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.03.20T00-00-00.dangerous_breaking.ts',
+  noTransaction: true,
+  run: ({ sql }) => sql`
+    ALTER TABLE
+      targets
+    ADD COLUMN
+      fail_dangerous_checks BOOLEAN NOT NULL DEFAULT FALSE;
```
n1ru4l, line 10: NIT: `fail_diff_on_dangerous_change` would be a more descriptive naming (would also include renaming the code variables)

n1ru4l (approved):
Approved with small comment:
https://github.com/graphql-hive/console/pull/6626#discussion_r2005027386

packages/services/api/src/modules/target/resolvers/Mutation/updateTargetDangerousChangeClassification.ts
```diff
@@ -0,0 +1,44 @@
...
+  } else if (targetRef.bySelector) {
+    const translator = injector.get(IdTranslator);
+    [organizationId, projectId, targetId] = await Promise.all([
+      translator.translateOrganizationId(targetRef.bySelector),
+      translator.translateProjectId(targetRef.bySelector),
+      translator.translateTargetId(targetRef.bySelector),
+    ]);
+  } else {
+    throw new Error();
+  }
```
n1ru4l, line 24: Shouldn't this use the `resolveTargetReference` method from `packages/services/api/src/modules/shared/providers/id-translator.ts` ?

n1ru4l: Also I would move this to the `target-manager` `updateTargetDangerousChangeClassification` method 🤔

jdolle: `resolveTargetReference` is very handy. I had forgot it existed. Ty.

PR #6630, opened 2025-03-19
Introduce built-in armor security features

enisdenjo:
Merge after https://github.com/graphql-hive/gateway/pull/871.

Contains some additional cleanup, improvements and fixes around gateway security.

### DIFF
--- a/packages/web/docs/src/content/gateway/other-features/security/disable-introspection.mdx
+++ b/packages/web/docs/src/content/gateway/other-features/security/disable-introspection.mdx
@@ -24,9 +24,7 @@ import { defineConfig } from '@graphql-hive/gateway'
 
 export const gatewayConfig = defineConfig({
   disableIntrospection: {
-    disableIntrospection: {
-      disableIf: () => true
-    }
+    disableIf: () => true
   }
 })
 ```
--- a/packages/web/docs/src/content/gateway/other-features/security/disable-introspection.mdx
+++ b/packages/web/docs/src/content/gateway/other-features/security/disable-introspection.mdx
@@ -6,7 +6,7 @@ searchable: false
 
 import { Callout } from '@theguild/components'
 
-# Introspection
+# Disable Introspection
 
 A powerful feature of GraphQL is schema introspection. This feature is used by GraphiQL for
 exploring the schema and also by tooling such as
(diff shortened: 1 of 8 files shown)


dotansimha (approved):

ardatan (approved):

packages/web/docs/src/content/gateway/other-features/security/disable-introspection.mdx
```diff
@@ -44,46 +54,3 @@ export const gatewayConfig = defineConfig({
   }
 })
 ```
-
-## Disabling Field Suggestions
```
ardatan, line 48: Isn't this important to have this here? Because this plugin prevents the leak of the introspection even if it is disabled. So user will know it can still leak.

enisdenjo: It's still there, just on another, dedicated, page: https://pr-6630.hive-landing-page.pages.dev/docs/gateway/other-features/security/block-field-suggestions.

enisdenjo: BTW the "block field suggestions" has nothing to do with introspection. It's the suggestion GQL generates in an error when you mistype a field.

```
Cannot query field "upcie" on type "Product". Did you mean "upc"?
```

->

```
Cannot query field "upcie" on type "Product". [Suggestion hidden]
```

ardatan: `Did you mean X?` exposes the parts of introspection. And theoritically it is possible to get the schema fully even if the introspection is disabled. I'm trying to find that tool that creates the entire schema for you by brute forcing this suggestion thing. So I think it is related. The point of disabling introspection is to prevent your schema to leak. So user should know this is not enough which is a common mistake.

ardatan: See https://github.com/nikitastupin/clairvoyance?ref=escape.tech

enisdenjo: https://github.com/graphql-hive/console/pull/6630/commits/d8ae11210012a8a85a226632391cec2de4ceacdd

PR #6632, opened 2025-03-19
fix: Stripe load error

jdolle:
### Background

This is an uncommon error that gets raised to sentry. https://the-guild-z4.sentry.io/issues/6190340612/?environment=prod&project=5691450&query=error.unhandled%3Atrue%20is%3Aunresolved&referrer=issue-stream&sort=freq&stream_index=3

This fix should capture the error as a warning in sentry. This should lower the priority there but keep it visible if we decide we care.

I tried to look through stripe.js docs and issues to find if others have experienced this, and what I found was that this is potentially because of an adblocker on some browsers, and the recommended fix was to ignore it and possibly display some message asking users to disable any adblocker. I opted just for the first one for now.


n1ru4l (approved):
Let's create a follow up task for the adblock warning? Then we can omit this froms entry completly.

PR #6657, opened 2025-03-26
fix(policy): update graphql-eslint to v4, add more tests

dotansimha:
## Background 

Continue from https://github.com/graphql-hive/console/pull/6101 

This change also introduces fixed for linting process, and upgrades eslint to latest. 

The testing of this PR was a bit complex, as I needed to use actual configuration samples for testing. 

## Changelog 

- [x] latest graphql-eslint (v4)
- [x] latest eslint
- [x] fix local development lint issues 
- [x] verify v4 of graphql-eslint with policy service 
- [x] backward compatible for existing configs 
- [x] migration (`2025.03.26T00-00-00.graphql-eslint.v4.ts`) + migration tests for existing v3 data 
- [x] make sure all existing data can be migrated 
- [x] reproduce issues with `no-unreachable-types`
- [x] verify UI working 

## Migrations 

Some rules needs adjustments to the configuration, leading to changes in the UI and also in the DB. 

### `alphabetize` 

In v3, `values` flags accepted an array, but now it's just a boolean. 

before/after 

image

- [x] migration file
- [x] migration test
- [x] ui check

### `no-case-insensitive-enum-values-duplicates` 

Changed name to `unique-enum-value-names` .

before/after 

image

- [x] migration file
- [x] migration test
- [x] ui check


packages/migrations/src/actions/2025.03.26T00-00-00.graphql-eslint.v4.ts
```diff
@@ -0,0 +1,104 @@
...
+
+          await connection.query(
+            sql`UPDATE schema_policy_config SET config = ${sql.json(v4Config)} WHERE resource_type = ${resource_type} AND resource_id = ${resource_id}`,
+          );
+          return { resource_type, resource_id, config: v4Config };
+        }
+
+        return null;
+      }),
+    );
```
n1ru4l, line 26: I don't like the idea of starting an uncontrolled amount of database queries in parallel.

Can we please buffer the `schema_policy_config`?

packages/services/policy/package.json
```diff
@@ -9,17 +9,18 @@
...
     "@trpc/server": "10.45.2",
-    "@types/eslint": "8.56.12",
+    "@types/eslint": "9.6.1",
     "ajv": "8.17.1",
     "dotenv": "16.4.7",
-    "eslint": "8.57.1",
+    "eslint": "9.23.0",
     "fastify": "4.29.0",
     "graphql": "16.9.0",
+    "json-schema-to-ts": "3.1.1",
```
n1ru4l, line 23: what is this dependency used for?

packages/services/policy/src/api.ts
```diff
@@ -64,10 +64,10 @@ export const schemaPolicyApiRouter = t.router({
...
     return RELEVANT_RULES.map(([name, rule]) => ({
       name,
-      description: rule.meta.docs?.description || '',
-      recommended: rule.meta.docs?.recommended ?? false,
-      url: rule.meta.docs?.url,
-      schema: normalizeAjvSchema(rule.meta.schema) as object | null,
+      description: rule.meta!.docs!.description || '',
+      recommended: rule.meta!.docs!.recommended ?? false,
+      url: rule.meta!.docs!.url,
+      schema: normalizeAjvSchema(rule.meta!.schema),
```
n1ru4l, line 70: Please avoid using `!`. If you need to use it add a comment on why it is necessary or safe to use it here.

packages/services/policy/package.json
dotansimha: It added it because `graphql-eslint` uses it internally to create the programmatic API types for the rules config schema. Without it, it was all `any`, but now the test samples are also fully type-checked.

packages/migrations/src/actions/2025.03.26T00-00-00.graphql-eslint.v4.ts
dotansimha: Yes, working on it. Fully agree

dotansimha: Thanks, refactored it to use Zod and make actual validations on the data. Also it's running it serially now.

packages/services/policy/src/api.ts
dotansimha: fixed

n1ru4l (commented):
So it is not backwards compatible - meaning if the configurations are in a broken state via the migration (e.g. wrongly migrated) - we cannot roll back to the previous version.

I am okay taking that risk if we can verify it works for our existing records.

Otherwise, I propose writing the configuration to a new column and using it from there, so we can immediately rollback this change if any issues occur that disrupt the workflows of the customers.

dotansimha:
> So it is not backwards compatible - meaning if the configurations are in a broken state via the migration (e.g. wrongly migrated) - we cannot roll back to the previous version.

Yeah, agree.

> I am okay taking that risk if we can verify it works for our existing records.

It is tested, but still worth reducing risk. 

> Otherwise, I propose writing the configuration to a new column and using it from there, so we can immediately rollback this change if any issues occur that disrupt the workflows of the customers.

Just done, pushed :) Added a new field `config_v4` and we modify this one in the migration. The service code is using the new field. So this change is not backward compatible (the commit, not the eslint config).

packages/services/storage/src/index.ts
```diff
@@ -153,7 +153,7 @@ export async function createStorage(
   function transformSchemaPolicy(schema_policy: schema_policy_config): SchemaPolicy {
     return {
       id: `${schema_policy.resource_type}_${schema_policy.resource_id}`,
-      config: schema_policy.config,
+      config: schema_policy.config_v4 || schema_policy.config,
```
n1ru4l, line 156: Shouldn't we always use config_v4 here?

n1ru4l: New records should always be written to the config_4 column

dotansimha: crap, great catch, didn't push these :facepalm:  and yeah, fixed to use v4 always ✅

jdolle:
@dotansimha should this be closed?

dotansimha:
We can close this, and re-open/iterate when this becomes relevant again or if it becomes a blocker.

PR #6670, opened 2025-03-28
fix: retain `@oneOf` in public api

n1ru4l:
### Description

The `@oneOf` directive was stripped out of the public API definitions, as it was not retained in the federation composition.

### DIFF
--- a/packages/services/api/src/modules/shared/module.graphql.ts
+++ b/packages/services/api/src/modules/shared/module.graphql.ts
@@ -1,7 +1,19 @@
 import { gql } from 'graphql-modules';
 
 export default gql`
-  scalar DateTime @tag(name: "public")
+  """
+  A date-time string at UTC, such as \`2007-12-03T10:15:30Z\`, is compliant with the date-time format outlined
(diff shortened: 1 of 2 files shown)


kamilkisiela (approved):

packages/services/api/src/modules/shared/module.graphql.ts
```diff
@@ -1,17 +1,34 @@
...
   scalar Date
   scalar JSON
   scalar JSONSchemaObject
   scalar SafeInt
 
   extend schema
     @link(url: "https://specs.apollo.dev/link/v1.0")
-    @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@tag"])
+    @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@tag", "@composeDirective"])
+    @link(url: "https://github.com/graphql/graphql-spec/pull/825/v0.1", import: ["@oneOf"])
```
jdolle, line 25: Minor: I'd prefer a more permanent URL.
But since this isnt a federated graph, it will be easy to adjust in the future so this is a minor complaint.

PR #6678, opened 2025-03-31
feat(api): add `SeverityLevelType`

n1ru4l:
### Description

`CriticalityLevel` does not follow our specified conventions.

- Enum types should end with `Type`
- Enum type fields should be in upper snake case

`SeverityLevelType` replaces `CriticalityLevel` and follows the specified criteria.

This PR also adjusts both CLI and our App to use these new fields.

### DIFF
--- a/packages/libraries/cli/src/commands/schema/fetch.ts
+++ b/packages/libraries/cli/src/commands/schema/fetch.ts
@@ -26,7 +26,7 @@ const SchemaVersionForActionIdQuery = graphql(/* GraphQL */ `
   ) {
     schemaVersionForActionId(actionId: $actionId, target: $target) {
       id
-      valid
+      isValid
       sdl @include(if: $includeSDL)
       supergraph @include(if: $includeSupergraph)
       schemas @include(if: $includeSubgraphs) {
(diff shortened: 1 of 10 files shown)


packages/libraries/cli/src/commands/schema/fetch.ts
```diff
@@ -59,7 +59,7 @@ const LatestSchemaVersionQuery = graphql(/* GraphQL */ `
   ) {
     latestValidVersion(target: $target) {
       id
-      valid
+      isValid
```
kamilkisiela, line 62: I see that `valid` exists still in the schema, my eyes are working right?

kamilkisiela (approved):

packages/libraries/cli/src/commands/schema/fetch.ts
n1ru4l: yup its still used in previous versions of the cli

jdolle (changes requested):
needs a changeset and my preference is providing a date in the `@deprecated` description for when we would like to delete the fields

n1ru4l:
@jdolle We will remove the field when no customer is using a CLI version using that field anymore. So we cannot have a fixed date here. :/ The field is only relevant for our own usage (app/cli).

jdolle (approved):
Thanks, that makes sense.

PR #6679, opened 2025-03-31
chore: make graphql schema ready for hive federation project 

n1ru4l:
### Background

We want to transform the `the-guild/graphql-hive` project on hive into a federation project so we can leverage contracts for getting schema diffs. This requires a few adjustments so the GraphQL schema is accepted by our API.

**Note:** After merging this PR additional database edits are needed. I tested these steps on `dev`.

1. Find the latest schema version in our database
2. Add a service name (`graphql`) and service url on the database record
3. Make sure native composition is enabled for the project

### DIFF
--- a/codegen.mts
+++ b/codegen.mts
@@ -12,7 +12,10 @@ const config: CodegenConfig = {
         typeDefsFilePath: false,
         mergeSchema: {
           path: '../../../../schema.graphql',
-          config: { includeDirectives: true },
+          config: {
+            includeDirectives: true,
+            append: '\n\n directive @oneOf on INPUT_OBJECT',
+          },
         },
         resolverGeneration: 'minimal',
         resolverMainFileMode: 'modules',
--- a/deployment/utils/publish-graphql-schema.ts
+++ b/deployment/utils/publish-graphql-schema.ts
@@ -15,7 +15,7 @@ export function publishGraphQLSchema(args: {
 }) {
   const command = (accessToken: string) =>
     `schema:publish` +
-    ` --registry.endpoint ${args.registry.endpoint} --registry.accessToken ${accessToken} --target ${args.registry.target}` +
+    ` --registry.endpoint ${args.registry.endpoint} --registry.accessToken ${accessToken} --target ${args.registry.target} --service graphql --url https://api.graphql-hive.com/graphql` +
     ` --commit ${args.version.commit} --author "Hive CD" ./schema.graphqls`;
 
   return new local.Command(
--- a/deployment/utils/publish-graphql-schema.ts
+++ b/deployment/utils/publish-graphql-schema.ts
@@ -15,7 +15,7 @@ export function publishGraphQLSchema(args: {
 }) {
   const command = (accessToken: string) =>
     `schema:publish` +
-    ` --registry.endpoint ${args.registry.endpoint} --registry.accessToken ${accessToken} --target ${args.registry.target} --service graphql --url https://api.graphql-hive.com/graphql` +
+    ` --registry.endpoint ${args.registry.endpoint} --registry.accessToken ${accessToken} --target ${args.registry.target} --service graphql` +
     ` --commit ${args.version.commit} --author "Hive CD" ./schema.graphqls`;
 
   return new local.Command(
--- a/deployment/utils/publish-app-deployment.ts
+++ b/deployment/utils/publish-app-deployment.ts
@@ -74,7 +74,7 @@ export function publishAppDeployment(args: {
           `docker run --rm --name "publish-app-deployment-${args.appName}"` +
           ` ${dockerImage}` +
           ` app:publish` +
-          ` --registry.endpoint ${args.registry.endpoint} --registry.accessToken ${accessToken} --target ${args.registry.target} --service graphql --url https://api.graphql-hive.com/graphql` +
+          ` --registry.endpoint ${args.registry.endpoint} --registry.accessToken ${accessToken} --target ${args.registry.target}` +
           ` --name ${args.appName} --version ${args.version.commit}`,
       ),
     },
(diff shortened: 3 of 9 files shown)


deployment/utils/publish-graphql-schema.ts
```diff
@@ -15,7 +15,7 @@ export function publishGraphQLSchema(args: {
 }) {
   const command = (accessToken: string) =>
     `schema:publish` +
-    ` --registry.endpoint ${args.registry.endpoint} --registry.accessToken ${accessToken} --target ${args.registry.target}` +
+    ` --registry.endpoint ${args.registry.endpoint} --registry.accessToken ${accessToken} --target ${args.registry.target} --service graphql` +
```
kamilkisiela, line 18: :D

kamilkisiela (approved):

codegen.mts
```diff
@@ -12,7 +12,10 @@ const config: CodegenConfig = {
         typeDefsFilePath: false,
         mergeSchema: {
           path: '../../../../schema.graphql',
-          config: { includeDirectives: true },
+          config: {
+            includeDirectives: true,
+            append: '\n\n directive @oneOf on INPUT_OBJECT',
```
n1ru4l, line 17: Retaining the `@oneOf` directive (which is not yet part of the graphql specification), which is not yet supported by GraphQL federation, but shipped with `graphql-js` by default is a fever dream. The easiest solution to retain it for the output schema that we publish to the schema registry is to manually append it, which I achieved by patching codegen...

Once we support `@oneOf` within Federation natively, we can remove this part.

deployment/utils/publish-graphql-schema.ts
n1ru4l: --url is now optional :P

deployment/config/public-graphql-api-gateway/gateway.config.ts
```diff
@@ -0,0 +1,15 @@
+// @ts-expect-error not a dependency
```
dotansimha, line 1: I think you can add as a pinned devDependency? The way renovate will update it, and you don't need to have it statically defined in this line: https://github.com/graphql-hive/console/blob/experiment-federation/deployment/services/public-graphql-api-gateway.ts#L15 (you can read `deployment/package.json` in Pulumi code and take the version from it)

deployment/config/public-graphql-api-gateway/gateway.config.ts
```diff
@@ -0,0 +1,15 @@
...
+import { defineConfig } from '@graphql-hive/gateway';
+
+export const gatewayConfig = defineConfig({
```
dotansimha, line 4: Any other plugins we should consider here? should we offload some of the plugins in graphql-api to the gw? (in a separate pr?)

deployment/services/public-graphql-api-gateway.ts
```diff
@@ -0,0 +1,90 @@
...
+import { ServiceDeployment } from '../utils/service-deployment';
+import { type Docker } from './docker';
+import { type Environment } from './environment';
+import { type GraphQL } from './graphql';
+
+/**
+ * Hive Gateway Docker Image Version
+ * Bump this to update the used gateway version.
+ */
+const dockerImage = 'ghcr.io/graphql-hive/gateway:1.13.4';
```
dotansimha, line 15: see previous comment regarding this

deployment/services/public-graphql-api-gateway.ts
```diff
@@ -0,0 +1,90 @@
...
+  const supergraphEndpoint = cdnEndpoint + '/contracts/public';
+
+  // Note: The persisted documents access key is also valid for reading the supergraph
+  const persistedDocumentsSecret = new ServiceSecret('persisted-documents', {
+    cdnAccessKeyId: apiConfig.requireSecret('hivePersistedDocumentsCdnAccessKeyId'),
+  });
+
+  const configMap = new k8s.core.v1.ConfigMap('public-graphql-api-gateway-config', {
+    data: {
+      'gateway.config.ts': fs.readFileSync(gatewayConfigPath, 'utf-8'),
```
dotansimha, line 49: nit: I saw in modern examples of Pulumi that they use this instead of reading the file with `fs`: https://www.pulumi.com/registry/packages/local/api-docs/file/ 

Not sure if it's needed, as `fs` worked for us before.

deployment/config/public-graphql-api-gateway/gateway.config.ts
n1ru4l: no please not there will be a dependency cycle issue since gateway depends on packages in this repo.

deployment/config/public-graphql-api-gateway/gateway.config.ts
```diff
@@ -0,0 +1,15 @@
...
+  transportEntries: {
```
dotansimha, line 5: Ah, @n1ru4l , now that this gw is running, you can hook it with otel and we'll have better insights on how our public API is being used (as separate traces from the internal one). 

So maybe it is worth considering OTEL plugin here, with OTEL context-propagation. 

Should be as simple as: 
```ts
  openTelemetry: {
    exporters: [
      createOtlpHttpExporter({
        url: '<get-this-one-from-Pulumi>'
      })
    ]
  }
```

dotansimha: cc @EmrysMyrddin

codegen.mts
dotansimha: anything @eddeee888 can help with here?

n1ru4l: no really its not worth putting time into

n1ru4l:
I moved everything related to the gateway to a separate PR https://github.com/graphql-hive/console/pull/6703

PR #6683, opened 2025-04-01
Publish url optional

### LINKED ISSUE #6659: Schema Publish: Make `url` optional for subsequent schema publishes
If we already have a subgraph (service) published, we can make the url for subsequent schema publishes optional and re-use the existing one.

jdolle:
### Background

Resolves #6659 , Ref CONSOLE-1169 

### Description

Defaults to previously published url instead of erroring.

### DIFF
--- a/.changeset/tidy-singers-double.md
+++ b/.changeset/tidy-singers-double.md
@@ -1,0 +1,5 @@
+---
+'hive': minor
+---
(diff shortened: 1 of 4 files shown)


n1ru4l (approved):
legit - there is a formatting error tho

PR #6689, opened 2025-04-01
update support ticket to support categories, project, and target options

egoodwinx:
### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/support/module.graphql.ts
+++ b/packages/services/api/src/modules/support/module.graphql.ts
@@ -30,6 +30,9 @@ export default gql`
 
   input SupportTicketCreateInput {
     organizationSlug: String!
+    project: String
+    target: String
+    category: SupportCategory!
     subject: String!
     description: String!
     priority: SupportTicketPriority!
@@ -71,6 +74,9 @@ export default gql`
     id: ID!
     status: SupportTicketStatus!
     priority: SupportTicketPriority!
+    category: SupportCategory!
+    project: String
+    target: String
     createdAt: DateTime!
     updatedAt: DateTime!
     subject: String!
@@ -93,6 +99,13 @@ export default gql`
     createdAt: DateTime!
     body: String!
     fromSupport: Boolean!
+  }
+
+  enum SupportCategory {
+    TECHNICAL_ISSUE
+    BILLING
+    COMPLIANCE
+    OTHER
   }
 
   enum SupportTicketPriority {
--- a/packages/services/api/src/modules/support/providers/support-manager.ts
+++ b/packages/services/api/src/modules/support/providers/support-manager.ts
@@ -10,6 +10,7 @@ import { Logger } from '../../shared/providers/logger';
 import { Storage } from '../../shared/providers/storage';
 import { OrganizationManager } from './../../organization/providers/organization-manager';
 import { SUPPORT_MODULE_CONFIG, type SupportConfig } from './config';
+import { SupportCategory } from 'packages/libraries/core/src/client/__generated__/types';
 
 export const SupportTicketPriorityAPIModel = z.enum(['low', 'normal', 'high', 'urgent']);
 export const SupportTicketStatusAPIModel = z.enum([
@@ -23,6 +24,7 @@ export const SupportTicketStatusAPIModel = z.enum([
 
 export const SupportTicketPriorityModel = z.nativeEnum(SupportTicketPriority);
 export const SupportTicketStatusModel = z.nativeEnum(SupportTicketStatus);
+export const SupportTicketCategoryModel = z.nativeEnum(SupportCategory);
 
 const SupportTicketModel = z.object({
   id: z.number(),
@@ -45,6 +47,9 @@ const SupportTicketModel = z.object({
 
     return SupportTicketStatusModel.parse(value);
   }),
+  category: SupportTicketCategoryModel,
+  project: z.string().optional(),
+  target: z.string().optional(),
   created_at: z.string(),
   updated_at: z.string(),
   subject: z.string(),
@@ -80,6 +85,9 @@ const SupportTicketCommentListModel = z.object({
 
 const SupportTicketCreateRequestModel = z.object({
   organizationId: z.string(),
+  category: SupportTicketCategoryModel,
+  project: z.string().optional(),
+  target: z.string().optional(),
   subject: z.string().min(3),
   description: z.string().min(3),
   priority: SupportTicketPriorityModel,
--- a/packages/services/api/src/modules/support/resolvers/SupportTicket.ts
+++ b/packages/services/api/src/modules/support/resolvers/SupportTicket.ts
@@ -3,6 +3,15 @@ import type { SupportTicketResolvers } from './../../../__generated__/types';
 
 export const SupportTicket: SupportTicketResolvers = {
   comments: async (ticket, _args, { injector }) => {
+    return {
+      edges: [],
+      pageInfo: {
+        endCursor: '',
+        hasNextPage: false,
+        hasPreviousPage: false,
+        startCursor: '',
+      },
+    };
     const response = await injector.get(SupportManager).getTicketComments(ticket.id);
 
     return {
--- a/packages/web/app/src/components/layouts/project-selector.tsx
+++ b/packages/web/app/src/components/layouts/project-selector.tsx
@@ -1,5 +1,6 @@
 import { Select, SelectContent, SelectItem, SelectTrigger } from '@/components/ui/select';
 import { FragmentType, graphql, useFragment } from '@/gql';
+import { SelectValue } from '@radix-ui/react-select';
 import { Link, useRouter } from '@tanstack/react-router';
 
 const ProjectSelector_OrganizationConnectionFragment = graphql(`
@@ -21,7 +22,15 @@ export function ProjectSelector(props: {
   currentOrganizationSlug: string;
   currentProjectSlug: string;
   organizations: FragmentType<typeof ProjectSelector_OrganizationConnectionFragment> | null;
+  onValueChange?: Function;
+  optional?: boolean;
+  showOrganization?: boolean;
 }) {
+  const optional = typeof props.optional !== 'undefined' ? props.optional : false;
+  const showOrganization =
+    typeof props.showOrganization !== 'undefined' ? props.showOrganization : true;
+  const onValueChangeFunc =
+    typeof props.onValueChange !== undefined ? props.onValueChange : undefined;
   const router = useRouter();
 
   const organizations = useFragment(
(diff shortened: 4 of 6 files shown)


packages/services/api/src/modules/support/resolvers/SupportTicket.ts
```diff
@@ -3,6 +3,15 @@ import type { SupportTicketResolvers } from './../../../__generated__/types';
...
   comments: async (ticket, _args, { injector }) => {
+    return {
+      edges: [],
+      pageInfo: {
+        endCursor: '',
+        hasNextPage: false,
+        hasPreviousPage: false,
+        startCursor: '',
+      },
+    };
```
jdolle, line 14: youll want to remove these stubs. This was just for offline development.

packages/services/api/src/modules/support/module.graphql.ts
```diff
@@ -30,6 +30,9 @@ export default gql`
 
   input SupportTicketCreateInput {
     organizationSlug: String!
+    project: String
+    target: String
+    category: SupportCategory!
```
jdolle, line 35: Minor: I defer to my colleagues, but I think we should default this or make it nullable to avoid taking any downtime during the deployment window.

We don't get a ton of support tickets so I doubt it'd cause an issue, but I'm extra cautious.

packages/web/app/src/components/layouts/project-selector.tsx
```diff
@@ -38,47 +47,74 @@ export function ProjectSelector(props: {
...
       ) : (
-        <div className="h-5 w-48 max-w-[200px] animate-pulse rounded-full bg-gray-800" />
+        ''
       )}
-      {projects?.length && currentProject ? (
+
+      {(projects?.length && currentProject) || optional ? (
         <>
-          <div className="italic text-gray-500">/</div>
+          {showOrganization ? <div className="italic text-gray-500">/</div> : <></>}
```
jdolle, line 68: can use `null` instead of an empty element `<></>`

packages/web/app/src/components/layouts/project-selector.tsx
```diff
@@ -38,47 +47,74 @@ export function ProjectSelector(props: {
...
-                params: {
-                  organizationSlug: props.currentOrganizationSlug,
-                  projectSlug: id,
-                },
-              });
-            }}
+            onValueChange={
+              onValueChangeFunc
+                ? id => {
+                    onValueChangeFunc(id);
```
jdolle, line 74: minor: to clean up the jsx, I'd move this defaulting logic to the top of the function alongside the other defaulting logic. And you shouldnt need to wrap the `onValueChangeFunc`

packages/web/app/src/components/layouts/project-selector.tsx
```diff
@@ -38,47 +47,74 @@ export function ProjectSelector(props: {
...
+                  key={'empty'}
+                  value={'empty'}
+                  data-cy={`project-picker-option-Unassigned`}
                 >
-                  {project.slug}
+                  Unassigned
                 </SelectItem>
-              ))}
+              ) : (
+                <></>
```
jdolle, line 103: null here also

packages/web/app/src/components/layouts/project-selector.tsx
```diff
@@ -38,47 +47,74 @@ export function ProjectSelector(props: {
...
+                        projectSlug: id,
+                      },
+                    });
+                  }
+            }
           >
             <SelectTrigger variant="default" data-cy="project-picker-trigger">
               <div className="font-medium" data-cy="project-picker-current">
-                {currentProject.slug}
+                {optional ? <SelectValue placeholder="Pick an option" /> : ''}
```
jdolle, line 89: minor: This can be shortened to
```
{optional ? <SelectValue placeholder="Pick an option" /> : (currentProject?.slug ?? '')}
```

packages/web/app/src/components/layouts/target-selector.tsx
```diff
@@ -43,70 +58,102 @@ export function TargetSelector(props: {
...
+            className="max-w-[200px] shrink-0 truncate font-medium"
+          >
+            {currentOrganization.slug}
+          </Link>
+        ) : (
+          <div className="h-5 w-48 max-w-[200px] animate-pulse rounded-full bg-gray-800" />
+        )
       ) : (
-        <div className="h-5 w-48 max-w-[200px] animate-pulse rounded-full bg-gray-800" />
+        <></>
```
jdolle, line 83: null here also

packages/web/app/src/components/layouts/target-selector.tsx
```diff
@@ -43,70 +58,102 @@ export function TargetSelector(props: {
...
+            className="max-w-[200px] shrink-0 truncate font-medium"
+          >
+            {currentProject.slug}
+          </Link>
+        ) : (
+          <div className="h-5 w-48 max-w-[200px] animate-pulse rounded-full bg-gray-800" />
+        )
       ) : (
-        <div className="h-5 w-48 max-w-[200px] animate-pulse rounded-full bg-gray-800" />
+        <></>
```
jdolle, line 102: null here also

packages/web/app/src/components/layouts/target-selector.tsx
```diff
@@ -43,70 +58,102 @@ export function TargetSelector(props: {
...
       )}
-      <div className="italic text-gray-500">/</div>
-      {targets?.length && currentOrganization && currentProject && currentTarget ? (
+      {showProject ? <div className="italic text-gray-500">/</div> : <></>}
```
jdolle, line 104: null here also

packages/web/app/src/components/layouts/target-selector.tsx
```diff
@@ -43,70 +58,102 @@ export function TargetSelector(props: {
...
+                      },
+                    });
+                  }
+            }
           >
             <SelectTrigger variant="default" data-cy="target-picker-trigger">
               <div className="font-medium" data-cy="target-picker-current">
-                {currentTarget.slug}
+                {isOptional ? <SelectValue placeholder="Pick an option" /> : ''}
+                {currentTarget && !isOptional ? currentTarget.slug : ''}{' '}
```
jdolle, line 129: ^ simplify like the prior comment

packages/web/app/src/pages/organization-support.tsx
```diff
@@ -199,6 +228,91 @@ function NewTicketForm(props: {
...
+                  control={form.control}
+                  name="project"
+                  render={({ field }) => (
+                    <FormItem>
+                      <FormLabel>Project (Optional)</FormLabel>
+                      <FormControl>
+                        <ProjectSelector
+                          currentOrganizationSlug={props.organizationSlug}
+                          currentProjectSlug={field.value ? field.value : ''}
+                          onValueChange={(value: any) => {
```
jdolle, line 277: avoid using `any` if possible.

packages/services/api/src/modules/support/providers/support-manager.ts
```diff
@@ -568,6 +578,12 @@ export class SupportManager {
     });
     const customerType = this.resolveCustomerType(organization);
 
+    const formattedBody = ` "Category: " + ${request.data.category ? request.data.category : null}\n\n
+        "Project: " + ${request.data.project ? request.data.project : null}\n\n
+        "Target: " + ${request.data.target ? request.data.target : null}\n\n
```
jdolle, line 583: minor: since this is just text and not jsx, I'd prefer "not selected" as the default over `null`

packages/services/api/src/modules/support/providers/support-manager.ts
```diff
@@ -606,6 +622,9 @@ export class SupportManager {
       metadata: {
         ticketDescription: input.description,
         ticketPriority: input.priority,
+        ticketCategory: input.category ?? '',
+        ticketProject: input.project ?? '',
+        ticketTarget: input.target ?? '',
```
jdolle, line 627: prefer not including the metadata if it isnt present here.
I.e. `...(input.category ? { ticketCategory: input.category } : {}),`

jdolle (changes requested):
Great work on this overall. These are pretty minor details that I'd prefer be slightly different.

packages/services/api/src/modules/support/module.graphql.ts
```diff
@@ -71,6 +74,9 @@ export default gql`
     id: ID!
     status: SupportTicketStatus!
     priority: SupportTicketPriority!
+    category: SupportCategory!
+    project: String
+    target: String
```
n1ru4l, line 79: Do these fields actually work? They are not used on the frontend.

n1ru4l (changes requested):
I added a few comments

packages/services/api/src/modules/support/module.graphql.ts
egoodwinx: Is it not set here?
https://github.com/graphql-hive/console/pull/6689/files/edf992ca77a945a9a219ecd27b8acc96b2e8eedc#diff-5035f21b1e4e2a9955ac5a1334e68dcce4ed8cea2b181addefe7a11c8d7401d1R133-R135

packages/services/api/src/modules/support/module.graphql.ts
```diff
@@ -30,6 +30,9 @@ export default gql`
 
   input SupportTicketCreateInput {
     organizationSlug: String!
+    project: String
+    target: String
```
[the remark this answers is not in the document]
egoodwinx: Is it not easier to read the ticket if we use the names?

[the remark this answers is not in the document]
jdolle: Somewhat recently we added support for UUIDs in addition to paths (combination of slugs).
The reason for this change was because these UUIDs don't change if people rename things.

For a support ticket, we typically get to them quick enough that I dont think it would make a huge difference, but on the off chance that they change the name of a resource, a UUID would be better because then we don't need to ask again.

Readability isn't a concern -- it's all about being able to look up information in logs etc.

packages/services/api/src/modules/support/module.graphql.ts
jdolle: They are sent in the input but they may not be getting resolved.
For simplicty, I think these fields should be removed from the SupportTicket type. The requirement was to add this information to the support ticket body. I dont think we need it both in the body and in other fields.

n1ru4l:
@egoodwinx can you please take care of the eslint and type errors?

n1ru4l (changes requested):
See https://github.com/graphql-hive/console/pull/6689#issuecomment-2875485552

jdolle (approved):

packages/web/app/src/components/layouts/project-selector.tsx
```diff
@@ -23,9 +24,25 @@ export function ProjectSelector(props: {
   currentOrganizationSlug: string;
   currentProjectSlug: string;
   organizations: FragmentType<typeof ProjectSelector_OrganizationConnectionFragment> | null;
+  onValueChange?: ((value: string) => void) | undefined;
+  optional?: boolean;
+  showOrganization?: boolean;
 }) {
   const router = useRouter();
-
+  const optional = typeof props.optional !== 'undefined' ? props.optional : false;
```
n1ru4l: This would be shorted when using nullish coalescing

```suggestion
  const optional = props.optional ?? false;
```

packages/web/app/src/components/layouts/project-selector.tsx
```diff
@@ -23,9 +24,25 @@ export function ProjectSelector(props: {
...
+  const showOrganization =
+    typeof props.showOrganization !== 'undefined' ? props.showOrganization : true;
```
n1ru4l:
```suggestion
  const showOrganization = props.showOrganization ?? true
```

packages/web/app/src/components/layouts/target-selector.tsx
```diff
@@ -52,65 +74,73 @@ export function TargetSelector(props: {
...
-                >
-                  {edge.node.slug}
+              {isOptional ? (
+                <SelectItem key="empty" value="empty" data-cy="project-picker-option-Unassigned">
+                  Unassigned
                 </SelectItem>
-              ))}
+              ) : (
+                <></>
+              )}
```
n1ru4l, line 130: This can be `isOptional  && ()`, to avoid the ternary. TBH I don't care too much about using a ternary here, but rather having the else condition being a Fragment instead of simply `null`.

This expression can just be shorter.

n1ru4l:
Since you mentioned that you did not check the network calls, we first need to verify this is working either on staging or our development environment. Neither of those currently have a zendesk setup, so we first need to improve that.

PR #6703, opened 2025-04-04
chore: hive gateway as entrypoint for the public api

n1ru4l:
### Background

Use Hive Gateway as the entry-point for the private API endpoint.

### DIFF
--- a/deployment/config/public-graphql-api-gateway/gateway.config.ts
+++ b/deployment/config/public-graphql-api-gateway/gateway.config.ts
@@ -41,10 +41,11 @@ export const gatewayConfig = defineConfig({
         ],
       }
     : false,
-  demandControl: {
-    maxCost: 1000,
-    includeExtensionMetadata: true,
-  },
+  // TODO: this causes runtime errors
+  // demandControl: {
+  //   maxCost: 1000,
+  //   includeExtensionMetadata: true,
+  // },
   maxTokens: 1_000,
   maxDepth: 20,
 });
--- a/deployment/config/public-graphql-api-gateway/gateway.config.ts
+++ b/deployment/config/public-graphql-api-gateway/gateway.config.ts
@@ -41,11 +41,10 @@ export const gatewayConfig = defineConfig({
         ],
       }
     : false,
-  // TODO: this causes runtime errors
-  // demandControl: {
-  //   maxCost: 1000,
-  //   includeExtensionMetadata: true,
-  // },
+  demandControl: {
+    maxCost: 1000,
+    includeExtensionMetadata: true,
+  },
   maxTokens: 1_000,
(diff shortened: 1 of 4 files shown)


deployment/config/public-graphql-api-gateway/gateway.config.ts
```diff
@@ -0,0 +1,44 @@
...
+    key: process.env['HIVE_CDN_ACCESS_TOKEN'],
+  },
+  graphiql: {
+    title: 'Hive Console - GraphQL API',
+    defaultQuery,
+  },
+  propagateHeaders: {
+    fromClientToSubgraphs({ request }) {
+      return {
+        'x-request-id': request.headers.get('x-request-id'),
```
ardatan, line 27: GW should do this automatically actually 😅

deployment/config/public-graphql-api-gateway/gateway.config.ts
```diff
@@ -0,0 +1,44 @@
+// @ts-expect-error not a dependency
+import { createOtlpHttpExporter, defineConfig } from '@graphql-hive/gateway';
+
+const defaultQuery = `#
+# Welcome to the Hive Console GraphQL API.
```
dotansimha, line 5: add link to docs maybe? or TODO to update it when public API docs are live?

deployment/services/public-graphql-api-gateway.ts
```diff
@@ -0,0 +1,105 @@
...
+import { type Docker } from './docker';
+import { type Environment } from './environment';
+import { type GraphQL } from './graphql';
+import { type Observability } from './observability';
+
+/**
+ * Hive Gateway Docker Image Version
+ * Bump this to update the used gateway version.
+ */
+const dockerImage = 'ghcr.io/graphql-hive/gateway:1.13.4';
```
n1ru4l, line 17:
```suggestion
const dockerImage = 'ghcr.io/graphql-hive/gateway:1.13.5';
```

deployment/config/public-graphql-api-gateway/gateway.config.ts
ardatan: Request id header is passed to the subgraph http requests automatically.

dotansimha: yeah, didn't find it in the docs, but it it forwarding for sure, see https://github.com/graphql-hive/gateway/blob/e888ff1c4755056d5c45ff0b7593b092e1dc909c/packages/runtime/src/plugins/useRequestId.ts#L21 + https://github.com/graphql-hive/gateway/blob/main/packages/runtime/src/createGatewayRuntime.ts#L995

kamilkisiela (approved):

PR #6706, opened 2025-04-04
fix(schema): building contracts with extend keyword

n1ru4l:
### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/schema/src/lib/federation-tag-extraction.spec.ts
+++ b/packages/services/schema/src/lib/federation-tag-extraction.spec.ts
@@ -1818,7 +1818,7 @@ describe('applyTagFilterOnSubgraphs', () => {
         field1: String! @inaccessible
       }
 
-      extend type Type1 @inaccessible {
+      extend type Type1 {
         field2: String! @inaccessible
       }
     `);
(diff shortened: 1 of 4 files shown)


packages/services/schema/src/lib/federation-tag-extraction.spec.ts
```diff
@@ -1818,7 +1818,7 @@ describe('applyTagFilterOnSubgraphs', () => {
         field1: String! @inaccessible
       }
 
-      extend type Type1 @inaccessible {
+      extend type Type1 {
```
n1ru4l, line 1821: This was invalid output before.

packages/services/schema/src/lib/federation-tag-extraction.ts
```diff
@@ -231,11 +231,7 @@ export function applyTagFilterToInaccessibleTransformOnSubgraphSchema(
         }),
       } as ObjectLikeNode;
 
-      if (
-        !rootTypeNames.has(node.name.value) &&
```
n1ru4l: The root type name check here is unnecessary. If someone defines a contract that excludes the root type that is not our problem to ignore...

PR #6718, opened 2025-04-08
upgrade got to fix timeouterror

jdolle:
### Background

https://the-guild-z4.sentry.io/issues/6328589755/?query=is%3Aunresolved%20timeout&referrer=issue-stream&sort=date&stream_index=16

https://github.com/sindresorhus/got/issues/2248

### Description

Upgrades the "got" package to fix a TimeoutError edge case that can happen when the event loop is busy or blocked


dotansimha (approved):
LGTM. Let's make sure to test it in staging before proceeding to prod. I remember last `got` upgrades were a pain because of some unexpected errors (that's also why we have some weird patches...)

PR #6722, opened 2025-04-10
feat: user management public api

n1ru4l:
### Background

Adds member/invitations management fields to the public graphql api.

### Description

Adds the following to the public API schema.
- `Mutation.assignMemberRole`
- `Mutation.createMemberRole`
- `Mutation.deleteMemberRole`
- `Mutation.deleteOrganizationInvitation`
- `Mutation.inviteToOrganizationByEmail`
- `Mutation.updateMemberRole`

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -285,12 +303,13 @@ export default gql`
...
     cleanId: ID! @deprecated(reason: "Use the 'slug' field instead.")
     name: String! @deprecated(reason: "Use the 'slug' field instead.")
-    owner: Member!
+    owner: Member! @tag(name: "public")
     me: Member!
-    members: MemberConnection
-    invitations: OrganizationInvitationConnection
+    members(first: Int @tag(name: "public"), after: String @tag(name: "public")): MemberConnection!
+      @tag(name: "public")
+    invitations: OrganizationInvitationConnection @tag(name: "public")
```
n1ru4l, line 310: TODO: add pagination arguments

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -285,12 +303,13 @@ export default gql`
...
     getStarted: OrganizationGetStarted!
-    memberRoles: [MemberRole!]
+    memberRoles: MemberRoleConnection @tag(name: "public")
```
n1ru4l, line 312: TODO: add pagination arguments

packages/migrations/src/actions/2025.04.29T00-00-00.organization-member-pagination.ts
```diff
@@ -0,0 +1,55 @@
...
+export default {
+  name: '2025.04.29T00-00-00.organization-member-pagination.ts',
+  noTransaction: true,
+  // Adds a default role to OIDC integration and set index on "oidc_integrations"."default_role_id"
+  run: ({ sql }) => [
+    {
+      name: 'Add "organization_member"."created_at" column',
+      query: sql`
+        ALTER TABLE "organization_member"
+          ADD COLUMN IF NOT EXISTS "created_at" timestamptz NOT NULL DEFAULT ${sql.literalValue(createdAt)}::timestamp
```
jdolle, line 20: Does `::timestamp` auto format the createdAt literal? Because `date.getDate()` for example, will return `1` and not `01`

packages/services/api/src/modules/auth/module.graphql.ts
```diff
@@ -53,17 +53,17 @@ export default gql`
     total: Int!
   }
 
-  enum AuthProvider {
-    GOOGLE
-    GITHUB
+  enum AuthProviderType {
+    GOOGLE @tag(name: "public")
```
jdolle, line 57: I couldnt quickly find docs on this... but could you add the tag to the enum type to apply it to all its values? I don't see a world where we'd want to only expose some of these values.

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -240,40 +244,54 @@ export default gql`
...
+    email: String @tag(name: "public") @tag(name: "public")
   }
 
   """
   @oneOf
   """
   type InviteToOrganizationByEmailResult {
-    ok: OrganizationInvitation
-    error: InviteToOrganizationByEmailError
+    ok: InviteToOrganizationByEmailResultOk @tag(name: "public")
```
jdolle, line 277: This is a breaking change. I doubt these are used by anyone else, but we should check before deploying.

packages/services/api/src/modules/organization/providers/organization-manager.ts
```diff
@@ -471,26 +478,72 @@ export class OrganizationManager {
...
+  async inviteByEmail(input: {
+    organization: GraphQLSchema.OrganizationReferenceInput;
+    email: string;
+    role?: string | null;
+  }) {
+    await this.inMemoryRateLimiter.check(
+      'inviteToOrganizationByEmail',
+      5_000, // 5 seconds
+      6, // 6 invites
+      `Exceeded rate limit for inviting to organization by email.`,
```
jdolle, line 513: Minor: I'd prefer this be in redis since it's possible for a user to be routed to a different instance. But I see this is copying the existing behavior.

packages/services/api/src/modules/organization/providers/organization-member-roles.ts
```diff
@@ -83,19 +88,74 @@ export class OrganizationMemberRoles {
...
+    return {
+      edges,
+      pageInfo: {
+        hasNextPage,
+        hasPreviousPage: cursor !== null,
+        get endCursor() {
+          return edges[edges.length - 1]?.cursor ?? '';
+        },
+        get startCursor() {
+          return edges[0]?.cursor ?? '';
```
jdolle, line 155: In case you have a preference: Per the spec these cursors can be null. But it seems like they also accept non nulls.

https://relay.dev/graphql/connections.htm#sec-undefined.PageInfo

packages/services/storage/src/index.ts
```diff
@@ -222,11 +222,16 @@ export async function createStorage(
...
+        ).toString('hex');
+      },
       email: invitation.email,
-      organization_id: invitation.organization_id,
+      organizationId: invitation.organization_id,
       code: invitation.code,
-      created_at: invitation.created_at as any,
-      expires_at: invitation.expires_at as any,
+      createdAt: invitation.created_at as any,
+      expiresAt: invitation.expires_at as any,
```
jdolle, line 234: is it possible to get rid of the cast `as any`?

jdolle (commented):
The biggest concern is that these are breaking changes to our API. This can create a poor experience for active users. (They have to refresh  the page.)
And any external parties making API calls to these paths would also need to update their client.

But in the past this hasnt stopped these sorts of changes.
Should we consider this a major change at least?

packages/migrations/src/actions/2025.04.29T00-00-00.organization-member-pagination.ts
n1ru4l: Good question, I just double checked and it seems fine.

image

packages/services/api/src/modules/auth/module.graphql.ts
n1ru4l: Our implementation of contract filtering does not support it - I think it should though.

packages/services/api/src/modules/organization/providers/organization-member-roles.ts
n1ru4l: I don't really have a preference here tbh. 🤔

packages/services/api/src/modules/organization/providers/organization-manager.ts
n1ru4l: If I remember correctly this was a quick an dirty implementation for a security audit.

jdolle: I'd be in favor of creating a ticket in the backlog to improve this if we dont already have one.

jdolle (approved):

PR #6729, opened 2025-04-14
disable redis snapshots

jdolle:
### Background

Our redis image enables this feature by default.

### Description

We don't benefit from redis snapshots at all and the disadvantages can be relatively major:
https://redis.io/docs/latest/operate/oss_and_stack/management/persistence/#rdb-disadvantages

This disables snapshots to remove this unused functionality and to prevent issues with latency due to this process.

### DIFF
--- a/docker/docker-compose.community.yml
+++ b/docker/docker-compose.community.yml
@@ -89,6 +89,8 @@ services:
       REDIS_DISABLE_COMMANDS: 'FLUSHDB,FLUSHALL'
     volumes:
       - './.hive/redis/db:/bitnami/redis/data'
+    command: |
+      /opt/bitnami/scripts/redis/run.sh --maxmemory 1300mb --maxmemory-policy allkeys-lru
 
   supertokens:
     image: registry.supertokens.io/supertokens/supertokens-postgresql:9.3
--- a/docker/docker-compose.community.yml
+++ b/docker/docker-compose.community.yml
@@ -89,8 +89,6 @@ services:
       REDIS_DISABLE_COMMANDS: 'FLUSHDB,FLUSHALL'
     volumes:
       - './.hive/redis/db:/bitnami/redis/data'
-    command: |
-      /opt/bitnami/scripts/redis/run.sh --maxmemory 1300mb --maxmemory-policy volatile-ttl
 
   supertokens:
     image: registry.supertokens.io/supertokens/supertokens-postgresql:9.3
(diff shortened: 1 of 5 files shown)


docker/docker-compose.community.yml
```diff
@@ -89,6 +89,8 @@ services:
       REDIS_DISABLE_COMMANDS: 'FLUSHDB,FLUSHALL'
     volumes:
       - './.hive/redis/db:/bitnami/redis/data'
+    command: |
+      /opt/bitnami/scripts/redis/run.sh --maxmemory 1300mb --maxmemory-policy allkeys-lru
```
jdolle, line 93: I suspect that this `maxmemory` should actually be slightly lower than the pod's `mem_limit` because the OS needs some to run. But I'm not super familiar with kubernetes to know.

dotansimha: Yeah, agree. Let's do here `1800mb` and in the deployment config limit it to `2000mb`? 
We have plenty of mem to use in prod, it's an adjustment we can make.

dotansimha: what are the other flags? anything we should think about before?

n1ru4l:
How will this change affect bulljs worker queues?

dotansimha (changes requested):
Please see my comments, also you need to adjust the prod config of the Redis container, not just the Docker ones. 
See https://github.com/graphql-hive/console/blob/main/deployment/services/redis.ts

docker/docker-compose.community.yml
jdolle: Other flags are listed and described [here](https://raw.githubusercontent.com/redis/redis/unstable/redis.conf). These are less critical than the two I set, but there are other optimizations we can make:
- `lazyfree-lazy-eviction yes` lets us have constant eviction time regardless of the size of the record. This runs the risk of delaying eviction and therefore can cause us to run out of memory. But if we set a reasonable limit then this risk is minimized. (I.e. 60-70% of total available memory)
- `save ""` can be used to disable snapshotting. If we use this redis strictly as a cache, then we can set `save ""` to disable snapshotting, which can otherwise cause latency spikes and elevated CPU during the snapshot window.

There are also other lazy free options we could enable, but I propose we not worry about setting expire manually and we avoid manually deleting from the cache also:

```
lazyfree-lazy-expire <yes|no>
lazyfree-lazy-server-del <yes|no>
replica-lazy-flush <yes|no>
lazyfree-lazy-user-del <yes|no>
lazyfree-lazy-user-flush <yes|no>
```

jdolle:
In order for this to work, we need to move existing redis queue implementations, like bullmq, to a different queue mechanism, and then dedicate redis to caching.

PR #6738, opened 2025-04-16
feat(api): store access token information in audit log

n1ru4l:
### Background

GraphQL API actions can now be performed with both organization access tokens/ dashboard access tokens linked to an user account.
The former is not yet tracked within our access log.

### Description

This PR introduces storing basic information about the organization access token within clickhouse.

Adds a `access_token_id` column to our clickhouse table that includes the access token id (if the action is performed with such). In addition to that, we also store more detailed information about the access token within the metadata json column.

The change is tested with an integration test.

### DIFF
--- a/integration-tests/tests/api/audit-logs/audit-log-record.spec.ts
+++ b/integration-tests/tests/api/audit-logs/audit-log-record.spec.ts
@@ -185,7 +185,7 @@ test.concurrent('export audit log for schema policy', async () => {
 
 test.concurrent('access token actions are stored within the audit log', async () => {
   const { createOrg, ownerToken } = await initSeed().createOwner();
-  const { createProject, organization, createOrganizationAccessToken } = await createOrg();
+  const { organization, createOrganizationAccessToken } = await createOrg();
   // First we create an access token with the organization owner token
   const accessToken = await createOrganizationAccessToken({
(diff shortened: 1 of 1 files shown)


packages/migrations/src/clickhouse-actions/014-audit-logs-access-token.ts
```diff
@@ -0,0 +1,10 @@
+import type { Action } from '../clickhouse';
+
+export const action: Action = async exec => {
+  await exec(`
+    ALTER TABLE "audit_logs"
+      ADD COLUMN IF NOT EXISTS
+        "access_token_id" String CODEC(ZSTD(1))
```
dotansimha, line 7: nit(?): if we only store id, it might not have enough info when the token is deleted? should it be also a name?

n1ru4l: We store more information about the token within the metadata column. See https://github.com/graphql-hive/console/pull/6738/files#diff-2ef2e8ba27a98af0972ca875f10b24fe905bcc7d31bdb1791df098ff42fa0872R47-R53

n1ru4l:
I added a integration test in 174b0bd86b1f23453811650493687f785a02f315

dotansimha (approved):

PR #6743, opened 2025-04-16
Extend timeout for schema check using cli from 20s to 65s

jdolle:
### Background

We had an issue where a schema check was being aborted after 20s and not hitting our service. Hive's global timeout is 60s. Therefore, it's possible for these schema checks to succeed, but be aborted by clients after 20s. Instead, we should allow the service to process the schema fully before giving up.

### Description

Sets the client timeout to slightly longer than the server in order to minimize the possibility of the schema check being aborted early while it's still being processed.

### DIFF
--- a/.changeset/twelve-cycles-lick.md
+++ b/.changeset/twelve-cycles-lick.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/cli': patch
+---
+
+Increase schema check client timeout from 20s to 65s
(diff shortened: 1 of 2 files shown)


n1ru4l (changes requested):
This value can not be higher than the proxy timeout https://github.com/graphql-hive/console/blob/8e3873fa466302e324e1ffbb964c2fb05413f7b2/deployment/services/proxy.ts#L75

jdolle:
> This value can not be higher than the proxy timeout
> 
> https://github.com/graphql-hive/console/blob/8e3873fa466302e324e1ffbb964c2fb05413f7b2/deployment/services/proxy.ts#L75

why? Don't we want the server to timeout the request rather than the client?

PR #6754, opened 2025-04-24
Exit the worker thread correctly instead of hanging

jdolle:
### Background

If the worker thread throws an uncaught exception, the worker thread stops receiving messages and `postMessage` from the main thread will hang.

This is because when the worker thread calls process.exit, it does not report the error to the main process

### Description

To resolve this, we should not capture this error and should instead allow the main thread to capture and handle the error.

### DIFF
--- a/packages/services/schema/src/composition-scheduler.ts
+++ b/packages/services/schema/src/composition-scheduler.ts
@@ -102,8 +102,8 @@ export class CompositionScheduler {
       });
     };
 
+    // catch uncaught exception from worker thread. Worker thread gets terminated.
     worker.on('error', error => {
-      console.error(error);
       this.logger.error('Worker error %s', error);
       Sentry.captureException(error, {
         extra: {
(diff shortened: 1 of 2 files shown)


packages/services/schema/src/composition-scheduler.ts
```diff
@@ -102,8 +102,8 @@ export class CompositionScheduler {
       });
     };
 
+    // catch uncaught exception from worker thread. Worker thread gets terminated.
     worker.on('error', error => {
-      console.error(error);
```
jdolle, line 106: already logged in ithe next line

packages/services/schema/src/composition-worker.ts
```diff
@@ -22,21 +22,7 @@ export function createCompositionWorker(args: {
...
-  });
-
-  process.on('uncaughtException', function (err) {
-    console.error('uncaughtException', err);
-    args.port.postMessage({
-      code: 'ERROR',
-      err,
-    });
-    process.exit(1);
+    throw err;
```
jdolle, line 25: `unhandledRejection` should throw to have the main thread kill the process. We shouldnt have unhandled rejections.

PR #6755, opened 2025-04-25
Reduce retry counts; adjust usage producer fallback logic

jdolle:
### Background

The short form is that when we push the the fallback producer queue, our buffer is never set back to the `Ready` status.

And the pod is killed after the grace period -- `terminationGracePeriodSeconds: 60,`

The most recent case when this happend looked like this:
Screenshot 2025-04-25 at 1 15 34 PM
Screenshot 2025-04-25 at 1 15 37 PM
which supports this theory.

### Description

This change addresses the readiness logic to make sure after the fallback queue is processed, that the state is changed back to Ready.

The one thing I am not entirely sure about is why the memory shows such a dramatic spike. All other metrics make sense (503s, kafka latency ... ).

### DIFF
--- a/packages/services/usage/src/usage.ts
+++ b/packages/services/usage/src/usage.ts
@@ -35,7 +35,7 @@ const retryOptions = {
   initialRetryTime: 500,
   factor: 0.2,
   multiplier: 2,
-  retries: 10,
+  retries: 5,
 } satisfies RetryOptions; // why satisfies? To be able to use `retryOptions.retries` and get `number` instead of `number | undefined`
 
 export function splitReport(report: RawReport, numOfChunks: number) {
@@ -165,7 +165,7 @@ export function createUsage(config: {
       return Object.keys(report.map).length;
     },
     split(report, numOfChunks) {
-      logger.info('Splitting into %s', numOfChunks);
+      logger.info('Splitting report into %s (id=%s)', numOfChunks, report.id);
       return splitReport(report, numOfChunks);
     },
     onRetry(reports) {
@@ -186,6 +186,7 @@ export function createUsage(config: {
       try {
         bufferFlushes.inc();
         const stopTimer = kafkaDuration.startTimer();
+
         const meta = await producer
           .send({
             topic: config.kafka.topic,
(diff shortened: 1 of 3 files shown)


packages/services/usage/src/usage.ts
```diff
@@ -212,21 +213,11 @@ export function createUsage(config: {
...
         rawOperationFailures.inc(numOfOperations);
 
         changeStatus(Status.Unhealthy);
-        logger.error(`Failed to flush (id=%s, error=%s)`, batchId, error.message);
-        Sentry.setTags({
-          batchId,
-          message: error.message,
-          numOfOperations,
-        });
-        Sentry.captureException(error);
```
jdolle, line 227: Moved this to the fallback queue since this is being handled.
I dont believe we should send handled errors to sentry.

packages/services/usage/src/usage.ts
```diff
@@ -306,7 +302,7 @@ export function createUsage(config: {
       },
     ),
     readiness() {
-      return status === Status.Ready && fallback.size() === 0;
+      return status === Status.Ready;
```
jdolle, line 305: checking fallback size here is redundant

packages/services/usage/src/usage.ts
```diff
@@ -35,7 +35,7 @@ const retryOptions = {
   initialRetryTime: 500,
   factor: 0.2,
   multiplier: 2,
-  retries: 10,
+  retries: 5,
```
jdolle, line 38: 10 retries is a lot... I'd rather fallback faster

packages/services/usage/src/usage.ts
```diff
@@ -255,6 +245,11 @@ export function createUsage(config: {
       } finally {
         stopTimer();
       }
+
+      if (fallback.size() === 0) {
+        logger.info('Fallback queue flushed')
+        changeStatus(Status.Ready);
```
jdolle, line 251: this is the important part

saihaj (approved):

n1ru4l (approved):

PR #6758, opened 2025-04-28
feat(api): target crud operations

n1ru4l:
### Background

Adds crud operations to the public graphql api.

### Description

- Merge `Mutation.setTargetValidation` and `Mutation.updateTargetValidationSettings` into `Mutation.updateTargetConditionalBreakingChangeConfiguration`.
- Move business logic from resolvers to manager/business logic classes
- Follow our practices for resolving entities by using `*ReferenceInput` types for mutation inputs
- Add following to the public graphql api schema
   - `Mutation.updateTargetConditionalBreakingChangeConfiguration`
   - `Mutation.updateTargetSlug`
   - `Mutation.updateTargetDangerousChangeClassification`
   - `Mutation.createTarget`
   - `Mutation.deleteTarget`
   - `Mutation.createCdnAccessToken`
   - `Mutation.deleteCdnAccessToken`
   - `Target.conditionalBreakingChangeConfiguration`
   - `Target.failDiffOnDangerousChange`
   - `Target.cdnAccessTokens`
   - `BreakingChangeFormulaType`
   - `Organization.availableOrganizationAccessTokenPermissionGroups`


packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -139,43 +172,71 @@ export default gql`
...
+    requestCount: Int @tag(name: "public")
+    """
+    List of target ids within the same project, whose operations are used for the breaking change detection.
+    """
+    targetIds: [ID!] @tag(name: "public")
+    """
+    List of client names that are excluded from the breaking change detection.
+    """
+    excludedClients: [String!] @tag(name: "public")
+  }
```
n1ru4l, line 208: I did not really see any benefit of having the `isEnabled` and other configuration options separated. Thus I merged the two mutation fields into a single one that can handle both cases.

packages/services/api/src/modules/target/providers/target-manager.ts
n1ru4l, line 1: Changes here are mostly moving code from resolvers here.

PR #6759, opened 2025-04-28
Reduce usage service readiness threshold

jdolle:
### Background

We've been investigating an issue with the usage producer where the memory spikes and the service may or may not crash or remain with high memory. 

The fallback/retry logic was adjusted here https://github.com/graphql-hive/console/pull/6755

### Description

This PR reduces the failure threshold for readiness with usage service.

We know that there's a window where requests can continue to be made to the usage service after it's flagged as unhealthy. I had incorrectly thought previously, that these messages were being send to kafka via our producer in the normal (non-fallback) flow, even though the service was unhealthy. _However_, inside the endpoint's handler, we check readiness of the usage client and if it's unhealthy, return a 503.

To reduce the number of 503s, we should be more aggressive with the readiness check for usage. Even if it triggers more false positives (which I doubt), moving a pod from ready to not ready isn't a big deal and having fewer unnecessary 503s is a positive thing for our consumers.

### DIFF
--- a/deployment/services/usage.ts
+++ b/deployment/services/usage.ts
@@ -54,7 +54,7 @@ export function deployUsage({
         readinessProbe: {
           initialDelaySeconds: 10,
           periodSeconds: 5,
-          failureThreshold: 2,
+          failureThreshold: 1,
           timeoutSeconds: 5,
           endpoint: '/_readiness',
         },
--- a/packages/services/usage/src/buffer.ts
+++ b/packages/services/usage/src/buffer.ts
@@ -322,6 +322,7 @@ export function createKVBuffer<T>(config: {
       logger.info('Stopping buffer');
       if (timeoutId) {
         clearTimeout(timeoutId);
+        timeoutId = null;
       }
       await send({
         scheduleNextSend: false,
--- a/packages/services/usage/src/usage.ts
+++ b/packages/services/usage/src/usage.ts
@@ -268,9 +268,17 @@ export function createUsage(config: {
     status = newStatus;
   }
 
+  producer.on(producer.events.CONNECT, () => {
+    logger.info(`Kafka producer: connected`);
+  })
+
   producer.on(producer.events.REQUEST_TIMEOUT, () => {
     logger.info('Kafka producer: request timeout');
   });
+
+  producer.on(producer.events.DISCONNECT, () => {
+    logger.info(`Kafka producer: disconnected`);
+  })
 
   async function stop() {
     logger.info('Started Usage shutdown...');
@@ -281,7 +289,6 @@ export function createUsage(config: {
     await fallback.stop();
     logger.info(`Fallback stopped`);
     await producer.disconnect();
-    logger.info(`Producer disconnected`);
 
     logger.info('Usage stopped');
   }
--- a/packages/services/usage/src/usage.ts
+++ b/packages/services/usage/src/usage.ts
@@ -270,7 +270,7 @@ export function createUsage(config: {
 
   producer.on(producer.events.CONNECT, () => {
     logger.info(`Kafka producer: connected`);
-  })
+  });
 
   producer.on(producer.events.REQUEST_TIMEOUT, () => {
     logger.info('Kafka producer: request timeout');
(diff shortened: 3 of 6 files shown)


packages/services/usage/src/buffer.ts
```diff
@@ -322,6 +322,7 @@ export function createKVBuffer<T>(config: {
       logger.info('Stopping buffer');
       if (timeoutId) {
         clearTimeout(timeoutId);
+        timeoutId = null;
```
jdolle, line 325: this isn't necessary based on the current logic, but it's best practice being a good steward of our variables to avoid confusing logic potentially later.

packages/services/usage/src/usage.ts
```diff
@@ -268,10 +268,18 @@ export function createUsage(config: {
...
+  producer.on(producer.events.CONNECT, () => {
+    logger.info(`Kafka producer: connected`);
+  });
+
   producer.on(producer.events.REQUEST_TIMEOUT, () => {
     logger.info('Kafka producer: request timeout');
   });
 
+  producer.on(producer.events.DISCONNECT, () => {
+    logger.info(`Kafka producer: disconnected`);
```
jdolle, line 280: I want to know whenever these things happen to better understand what our kafka producer is doing when we see the spikes in memory and in general.

deployment/services/usage.ts
```diff
@@ -54,7 +54,7 @@ export function deployUsage({
         readinessProbe: {
           initialDelaySeconds: 10,
           periodSeconds: 5,
-          failureThreshold: 2,
+          failureThreshold: 1,
```
jdolle, line 57: Fail fast. Avoid 503s

packages/services/usage/src/usage.ts
```diff
@@ -32,7 +41,7 @@ const levelMap = {
 
 const retryOptions = {
   maxRetryTime: 30_000,
-  initialRetryTime: 500,
+  initialRetryTime: 300,
```
jdolle, line 44: A continuation of more aggressive times in general. 300ms is the default for kafkajs and I don't see a reason to wait longer.

packages/services/usage/src/usage.ts
```diff
@@ -103,6 +112,25 @@ export function createUsage(config: {
...
+  const socketFactory: ISocketFactory = ({ host, port, ssl, onConnect }) => {
+    const socket = ssl
+      ? tls.connect(
+          Object.assign({ host, port }, !net.isIP(host) ? { servername: host } : {}, ssl),
+          onConnect,
+        )
+      : net.connect({ host, port }, onConnect);
+
+    // This is equivalent to kafka's "connections.max.idle.ms"
+    socket.setKeepAlive(true, 180_000);
```
jdolle, line 126: default is 60s in kafkajs, but event hub recommends 180s https://github.com/Azure/azure-event-hubs-for-kafka/blob/master/CONFIGURATION.md#producer-configurations-only

packages/services/usage/src/usage.ts
```diff
@@ -140,10 +168,11 @@ export function createUsage(config: {
       };
     },
     // settings recommended by Azure EventHub https://docs.microsoft.com/en-us/azure/event-hubs/apache-kafka-configurations
-    requestTimeout: 60_000, //
+    requestTimeout: 30_000,
```
jdolle, line 171: recommendation is 30-60s.
Our average duration is 190ms, so 30s is more than enough.

packages/services/usage/src/usage.ts
```diff
@@ -268,10 +297,25 @@ export function createUsage(config: {
     status = newStatus;
   }
 
+  producer.on(producer.events.CONNECT, () => {
+    logger.info(`Kafka producer: connected`);
+
+    if (status === Status.Unhealthy) {
+      changeStatus(Status.Ready);
```
jdolle, line 304: My thought here is that we should be more proactive with setting the usage service to unhealthy and ready, instead of waiting for messages to fail.

n1ru4l (approved):

Issue #6649, opened 2025-03-25
Advanced Breaking Change Detection


n1ru4l:
## Background

> Currently, we visit fields, variables, arguments, and object fields in DocumentNode, collect their return types and parent types, and turn them into schema coordinates. When a schema is modified, Hive creates a list of touched schema coordinates and checks in ClickHouse if a schema coordinate was registered or not and if it’s a breaking change or not.
>
> This is too safe 🙂 

There are scenarios where a change detected as a breaking change might not really be a breaking change. See https://github.com/kamilkisiela/graphql-breaking-changes

Some examples:

- Changing type `String` to `ID` does not break clients (unless the scalar is used in a operation arguments definition
- Changing `Int` to `Int!` for an argument is not breaking if all operation argument usages already define it as `Int!`
- Renaming an Object type is not breaking if the `__typename` is not used.

We want to be able to cover these kind of "safe" breaking changes with our breaking change detection.

n1ru4l:
This is now implemented and shipped.

PR #6764, opened 2025-05-02
feat: Advanced breaking change detection for inputs and arguments

jdolle:
### Background

Related to https://github.com/graphql-hive/console/issues/6649

### Description

To accurately determine whether a change is breaking or not based on traffic, we must track whether or not an argument or input's value has actually been provided or not.

I also considered how we should handle default values, and after some thought determined that those can be detected during the schema check since they are in the SDL. This change is all about the runtime values.

#### How are values provided?

- If the value is a literal for the coordinate
- If the value is a variable, and that variable is not null and not undefined for the coordinate

#### Limitations

- If `processVariables` is not enabled, then nested variables cannot be correctly identified as provided or not.

### Checklist

- [x] Testing


packages/libraries/core/tests/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "include": ["."],
   "compilerOptions": {
     "baseUrl": ".",
-    "rootDir": ".",
+    "rootDir": "..",
```
jdolle, line 6: if rootdir doesnt include imported files, then tsc complains. This is doesn't do anything in reality, but it's annoying in the IDE

packages/libraries/core/tests/usage.spec.ts
```diff
@@ -368,7 +369,6 @@ test('should not leak the exception', { retry: 3 }, async () => {
     [INF] [hive][usage] Sending report (queue 1)
     [INF] [hive][usage] POST http://404.localhost.noop (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) Attempt (1/2)
     [ERR] [hive][usage] Error: getaddrinfo ENOTFOUND 404.localhost.noop
-    [ERR] [hive][usage]     at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:666:666)
```
jdolle, line 371: removed stack from `fastFetchError` error result.

packages/libraries/yoga/tests/yoga.spec.ts
```diff
@@ -420,36 +420,30 @@ test('reports usage with response cache', async ({ expect }) => {
       }),
     ],
   });
-  await new Promise<void>((resolve, reject) => {
```
jdolle, line 423: Previous implementation was not consistently resolving in time. I simplified it by making it serial awaits, and added a wait after them to trigger the event loop. This seems much more consistent.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -489,10 +489,27 @@ export class RegistryChecks {
...
+              });
+
+              if (
+                advancedNullabilityTotalRequests[`${change.breakingChangeSchemaCoordinate}!`] >=
+                totalRequestCounts[change.breakingChangeSchemaCoordinate]
+              ) {
+                // All requests for this coordinate provide the value for the coordinate. So moving to non-null is allowed.
+                return;
+              }
+            }
```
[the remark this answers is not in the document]
jdolle: Added

integration-tests/tests/usage-reporting/advanced-breaking-changes.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    });
+    await userResult.expectNoGraphQLErrors();
+
+    const usageReport = await collectUsage({
+      size: 2,
+      map: {
+        '4de79319': {
+          operationName: 'user',
+          operation: 'query user { user(id: "1234") { id } }',
+          fields: ['Query', 'Query.user', 'Query.user.id', 'User', 'User.id', 'ID'],
```
n1ru4l, line 32: Shouldn't there be a `Query.user.id!` being reported here? Can you elaborate on how this works?

jdolle: Turns out I was not waiting long enough after collecting usage.
I've fixed that and fixed a case with arguments that exists in the current version.
This issue is that if an argument is not provided, then its coordinate is not collected. And then the total count is off for that coordinate for the nullability case. I added a special condition for this to count the number of times the parent field is used when the change is from nullable to required.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -483,19 +483,47 @@ export class RegistryChecks {
             return;
           }
 
+          let checkCoordinate = change.breakingChangeSchemaCoordinate;
+          if (isNullToRequiredArgumentChange(change)) {
+            // If this is a nullable argument changing to required, then we must count all calls -- even those not passing an argument.
+            // Otherwise, if those operations are not considered, they could still be getting sent. And since they are not passing the
+            // argument, they would be broken... This could likely be addressed by adjusting the coordinates collection logic.
```
n1ru4l, line 490: > This could likely be addressed by adjusting the coordinates collection logic.

I think we can not easily change the coordinate collection logic as we cannot control what version people are using. 🤔 

We can only add new features on top of the existing coordinate collection that people can use if they upgrade.

If we fundamentally change how it works, we would need to introduce versioning.

But I am not sure if this is relevant here.

jdolle: It is. This is exactly why i ultimately went thr route that i did

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -483,19 +483,47 @@ export class RegistryChecks {
             return;
           }
 
+          let checkCoordinate = change.breakingChangeSchemaCoordinate;
+          if (isNullToRequiredArgumentChange(change)) {
```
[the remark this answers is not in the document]
jdolle: An argument is included in the usage report's list of coordinates only if it's in the operation. E.g.
```
query Foo { foo(a: "b") }
```
Would report `Query.foo, Query.foo.a, Query.foo.a!`

```
query Foo2 { foo }
```
Would only report `Query.foo`.

And so when changing an argument from nullable to non-nullable, we need to check the parent field could rather than the argument count. Otherwise, the second example wouldn't trigger the change as "breaking".

integration-tests/tests/usage-reporting/advanced-breaking-changes.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+          },
+        },
+      ],
+    });
+    expect(usageReport.status).toBe(200);
+    expect(typeof usageReport.body).not.toBeInstanceOf(String);
+    const body = usageReport.body as Exclude<typeof usageReport.body, string>;
+    expect(body.operations.accepted).toBe(2);
+
+    await waitFor(8_000);
```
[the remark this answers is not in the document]
jdolle: Tweaked to use polling and it's sped up significantly.
```
 RERUN  testkit/flow.ts x1 
        Filename pattern: tests/usage-reporting/advanced-breaking-changes.spec.ts

 ✓ tests/usage-reporting/advanced-breaking-changes.spec.ts (4 tests) 4177ms
   ✓ advanced breaking changes > an argument can safely migrate from nullable to non nullable if all usages provide that argument  1098ms
   ✓ advanced breaking changes > an argument can NOT migrate from nullable to non nullable if NOT all usages provide that argument  1010ms
   ✓ advanced breaking changes > an input field can migrate from nullable to non nullable if all usages provide that argument  1040ms
   ✓ advanced breaking changes > an input field can NOT migrate from nullable to non nullable if NOT all usages provide that argument  1027ms

 Test Files  1 passed (1)
      Tests  4 passed (4)
   Start at  14:13:51
   Duration  4.54s
```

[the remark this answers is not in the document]
jdolle: We should do the same for `integration-tests/tests/api/target/usage.spec.ts`, but I would like to do that as a followup to avoid code creep

packages/services/api/src/modules/schema/providers/registry-checks.ts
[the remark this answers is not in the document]
n1ru4l: Now I get it thank you ❤️  Let's add this example as a comment in code for future reference?

integration-tests/tests/api/target/usage.spec.ts
```diff
@@ -275,6 +275,7 @@ test.concurrent(
       );
     }
 
+    // @todo
```
n1ru4l, line 278: TODO!

n1ru4l (approved):
Looks good 😍 

We still need to add:
- Changeset for SDKs (core, apollo, yoga, envelop)
- Changeset for Hive Docker images

integration-tests/tests/usage-reporting/advanced-breaking-changes.spec.ts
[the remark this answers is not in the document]
n1ru4l: @jdolle Nice 👌

integration-tests/tests/api/target/usage.spec.ts
jdolle: oops. I was going to make this poll but then opted to move it to its own ticket. I'll remove the comment since it's captured in a ticket already.

PR #6765, opened 2025-05-06
Improve unit test stability

jdolle:
### Background

Unit tests fail semi-regularly due to issues with async and inconsistent dns settings across environments (local vs remote/docker)

### Description

This PR fixes these cases so that the async process is guaranteed to run consistently and the the dns issue is avoided by having a separate implementation of fetch provided.

### Checklist

- [x] Testing

### DIFF
--- a/packages/libraries/core/tests/tsconfig.json
+++ b/packages/libraries/core/tests/tsconfig.json
@@ -3,7 +3,7 @@
   "include": ["."],
   "compilerOptions": {
     "baseUrl": ".",
-    "rootDir": ".",
+    "rootDir": "..",
     "target": "es2017",
     "module": "esnext",
     "skipLibCheck": true,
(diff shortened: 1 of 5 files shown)


packages/libraries/core/tests/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "include": ["."],
   "compilerOptions": {
     "baseUrl": ".",
-    "rootDir": ".",
+    "rootDir": "..",
```
jdolle, line 6: This is to make the local IDE happy since within the tests we are importing from the `../src`, and Typescript complains when importing files outside the root directory.

packages/libraries/core/tests/usage.spec.ts
```diff
@@ -368,7 +369,6 @@ test('should not leak the exception', { retry: 3 }, async () => {
     [INF] [hive][usage] Sending report (queue 1)
     [INF] [hive][usage] POST http://404.localhost.noop (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) Attempt (1/2)
     [ERR] [hive][usage] Error: getaddrinfo ENOTFOUND 404.localhost.noop
-    [ERR] [hive][usage]     at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:666:666)
```
jdolle, line 371: I removed the stack from the error intentionally.

n1ru4l (approved):

PR #6770, opened 2025-05-07
Contract root types

jdolle:
### Background

https://linear.app/the-guild/issue/CONSOLE-1239/adjust-contract-logic-to-allow-empty-mutations

Contract behavior is to remove types that have all inaccessible fields. However, currently all root types are kept. This causes issues because per graphql, only the Query type needs to be defined in a schema.

### Description

This change adjusts the contracts logic so that only the root query type is kept no matter what.

### Checklist

- [x] Testing

### DIFF
--- a/packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
+++ b/packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
@@ -392,7 +392,9 @@ describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
 
       const output = applyTagFilterToInaccessibleTransformOnSubgraphSchema(sdl, filter);
 
-      expect(output.typesWithAllFieldsInaccessible.entries().toArray()).toEqual([['Mutation', true]]);
+      expect(output.typesWithAllFieldsInaccessible.entries().toArray()).toEqual([
+        ['Mutation', true],
+      ]);
 
       expect(print(output.typeDefs)).toMatchInlineSnapshot(`
         schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@tag"]) {
--- a/packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
+++ b/packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
@@ -3,7 +3,7 @@ import {
   applyTagFilterOnSubgraphs,
   applyTagFilterToInaccessibleTransformOnSubgraphSchema,
   type Federation2SubgraphDocumentNodeByTagsFilter,
-} from './federation-tag-extraction';
+} from '../federation-tag-extraction';
 
 describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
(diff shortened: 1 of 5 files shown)


packages/services/schema/src/lib/__tests__/reachable-type-filter.spec.ts
jdolle, line 1: Moved into the test directory to match other parts of our code base

packages/services/schema/src/lib/federation-tag-extraction.ts
```diff
@@ -63,34 +63,20 @@ function hasIntersection<T>(a: Set<T>, b: Set<T>): boolean {
   return false;
 }
 
-function getRootTypeNamesFromDocumentNode(document: DocumentNode) {
+function getRootQueryTypeNameFromDocumentNode(document: DocumentNode) {
```
jdolle, line 66: This is specifically what changed.

packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
```diff
@@ -3,7 +3,7 @@ import {
   applyTagFilterOnSubgraphs,
   applyTagFilterToInaccessibleTransformOnSubgraphSchema,
   type Federation2SubgraphDocumentNodeByTagsFilter,
-} from './federation-tag-extraction';
+} from '../federation-tag-extraction';
```
jdolle, line 6: moved this test file into a test directory to match other parts of our code base

packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
```diff
@@ -370,6 +370,46 @@ describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
...
+
+    test('mutation object type is returned in "typesWithAllFieldsInaccessible" if all its fields are excluded', () => {
+      const filter: Federation2SubgraphDocumentNodeByTagsFilter = {
+        include: new Set(),
+        exclude: new Set(['exclude']),
+      };
+      const sdl = parse(/* GraphQL */ `
+        schema @link(url: "https://specs.apollo.dev/federation/v2.0", import: ["@tag"]) {
+          query: Query
+        }
```
n1ru4l, line 382: Isn't `mutation: Mutation` missing here? By explicitly declaring `schema` and omitting `mutation` we are indicating that this subgraph has no Mutation fields and `Mutation` is just a "normal" object type without any special semantic meaning. 🤔

n1ru4l (approved):
I created additional test cases for reviewing and verifying these changes are working "as a whole" in https://github.com/graphql-hive/console/pull/6773. @jdolle Let me know what you think and then also merge these here before merging this pr.

n1ru4l:
Nevermind... Auto-merge was enabled

packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
jdolle: Yeah it was... 😨

PR #6772, opened 2025-05-09
fix: support ticket user lookup

jdolle:
### Background

https://linear.app/the-guild/issue/CONSOLE-1232/support-ticket-failed-with-internal-server-error

A customer was unable to submit a support ticket and retried 3x. Each time failing the user lookup.

### Description

The current URL is to get a list of users, but is not searching that list. We should be using: https://developer.zendesk.com/api-reference/ticketing/users/users/#search-users

And this returns a list of users that match. And on the off chance that users have an email that is a partial match to another email in our system, we should find the email in the returned list instead of checking the first.


n1ru4l (approved):
legit!

PR #6779, opened 2025-05-13
fix(schema): always treat network errors as schema publish rejections

n1ru4l:
### Checklist

- [x] Input validation
- [x] Output encoding
- [ ] Session management
- [x] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [x] Memory management
- [x] Testing


packages/services/schema/src/api.ts
```diff
@@ -100,7 +100,6 @@ export const schemaBuilderApiRouter = t.router({
                 data: {
                   type: 'federation',
                   args,
-                  requestTimeoutMs: ctx.cache.timeoutMs,
```
n1ru4l, line 103: This is now defined via a environment variable

packages/services/schema/src/lib/compose.ts
```diff
@@ -25,14 +25,13 @@ interface BrokerPayload {
...
 
-const EXTERNAL_COMPOSITION_RESULT = z.union([
+const ExternalCompositionResultModel = z.union([
   z.object({
     type: z.literal('success'),
     result: z.object({
       supergraph: z.string(),
       sdl: z.string(),
     }),
-    includesNetworkError: z.boolean().default(false),
```
n1ru4l, line 35: We never want to receive this from the external composition service. It i state that only exists on our end.

packages/services/schema/src/lib/compose.ts
```diff
@@ -46,11 +45,15 @@ const EXTERNAL_COMPOSITION_RESULT = z.union([
         }),
       ),
     }),
-    includesNetworkError: z.boolean().default(false),
```
n1ru4l, line 49: We never want to receive this from the external composition service. It i state that only exists on our end.

packages/services/schema/src/lib/compose.ts
```diff
@@ -315,57 +338,78 @@ async function callExternalService(
...
       },
       timeout: {
         request: timeoutMs,
         // connecting should be quick
-        lookup: 10_000,
-        connect: 10_000,
-        secureConnect: 10_000,
+        lookup: 5_000,
+        connect: 5_000,
+        secureConnect: 5_000,
```
n1ru4l, line 363: We have
`SCHEMA_COMPOSITION_TIMEOUT_MS` - Timeout for the whole composition process (main process kills worker)
`SCHEMA_EXTERNAL_COMPOSITION_TIMEOUT_MS` - Timeout for a external composition call

`SCHEMA_COMPOSITION_TIMEOUT_MS` must be higher than `SCHEMA_EXTERNAL_COMPOSITION_TIMEOUT_MS * EXTERNAL_COMPOSITION_RETRY_COUNT`.

```
EXTERNAL_COMPOSITION_RETRY_COUNT -> 9_000
EXTERNAL_COMPOSITION_RETRY_COUNT -> 3
TOTAL = 27_000
```

That means `SCHEMA_COMPOSITION_TIMEOUT_MS` set to `30_000` is okay. That way we get the actual HTTP/Socket error we can show to the end user (for their own external composition service). Instead of a generic schema composition timed out or unexpected error.

packages/services/schema/src/api.ts
```diff
@@ -160,23 +159,25 @@ export const schemaBuilderApiRouter = t.router({
         assertAllCasesExhausted(input);
       } catch (error) {
         // Treat timeouts caused by external composition as "expected errors"
-        if (ctx.cache.isTimeoutError(error) && input.type === 'federation' && input.external) {
+        if (ctx.cache.isTimeoutError(error)) {
           return {
             errors: [
               {
-                message: error.message,
-                source: 'graphql',
```
n1ru4l, line 168: No idea why this was 'graphql' before. With the new guard for rejecting within `packages/services/api/src/modules/schema/providers/models/composite.ts` it makes more sense to have this be `composite`.

packages/services/schema/src/api.ts
```diff
@@ -160,23 +159,25 @@ export const schemaBuilderApiRouter = t.router({
...
+                message: 'The composition timed out. Please try again later.',
+                source: 'composition',
               },
             ],
             sdl: null,
             supergraph: null,
-            includesNetworkError: true,
+            includesNetworkError: false,
+            includesException: true,
```
n1ru4l, line 173: This is not a network error but an exception because this basically means that we either
1. have an infinite loop in our composition logic
2. have external composition timeouts configured so they do not time out before external composition calls time out

kamilkisiela (approved):

packages/services/schema/src/lib/compose.ts
```diff
@@ -315,57 +346,74 @@ async function callExternalService(
...
-    return JSON.parse(response.body) as unknown;
+    return { type: 'success', data: JSON.parse(response.body) as unknown };
   } catch (error) {
     if (error instanceof RequestError) {
       if (!error.response) {
-        span.setAttribute('error.message', error.message);
+        let message = error.message;
+
+        if (!message && error.cause instanceof AggregateError) {
+          message = error.cause.message;
```
jdolle, line 382: Should we collect the `error.errors` messages instead? For an aggregate error, there may be multiple messages that would be helpful.

jdolle (approved):

packages/services/schema/src/lib/compose.ts
n1ru4l: Usually the error messages are the same for each attempt. I already have some notes to further improve the whole flow. We can do this in a follow up pr.

PR #6780, opened 2025-05-13
Add pg indexes to help with org delete

jdolle:
### Background

Large organizations are hitting timeouts when trying to delete their organization.
One such call ran for over 10min.

### Description

To address this performance issue and to allow organizations to self-serve, I've added an index for every foreign key that uses cascading deletes.

This should significantly improve the speed of this operation because when a delete is cascading, it must find those records in the table to perform the delete. Without an index, it must scan through all records.

### DIFF
--- a/packages/migrations/src/actions/2025.05.14T00-00-00.cascade-deletion-indices-2.ts
+++ b/packages/migrations/src/actions/2025.05.14T00-00-00.cascade-deletion-indices-2.ts
@@ -1,0 +1,151 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+/**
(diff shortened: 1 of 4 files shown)


packages/migrations/src/actions/2025.05.14T00-00-00.cascade-deletion-indices-2.ts
n1ru4l, line 1: Just out of curiosity, was there some "smart" way to get a list of all the columns that need indices or did it require going through all of the foreign keys?

n1ru4l (approved):

PR #6781, opened 2025-05-14
refactor(api): remove abstraction of schema service caller

n1ru4l:
### Background

We have 3 class implementations that are ~~almost~~ the same, with redundant logic that needs to be updated every single time we adjust how the schema service is called. We can replace it with a single class instead, reducing the need to touch 3 files every time an adjustment is made. Also some object no longer need to be passed through classes. We have dependency injection for that purpose.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/schema/providers/schema-manager.ts
```diff
@@ -189,8 +184,6 @@ export class SchemaManager {
       };
     }
 
-    const orchestrator = this.matchOrchestrator(project.type);
```
jdolle, line 192: Should we offload the `'Only Federation projects are supported'` error logic to the `compositionOrchestrator`?
It will error if the number of schemas is > 1 and the project type is Single. So it already has some logic around what each projectType supports.
Right now all this feels a little confusing where we're mostly handling project type inside the orchestrator but still have these edge cases.

jdolle (approved):
Overall I like the change. I love removing duplicate code.
I dont think it was a priority though.

n1ru4l:
@jdolle Yeah true, but it was annoying me too much because I had to touch these files a lot in the past few weeks :D I wanted to do this since at least 1 year 😆

PR #6782, opened 2025-05-14
chore: fix docker image publish (?)

n1ru4l:
### Background

The workflow tag push trigger was [not invoked since the push is made with the github token](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow), leading to no docker images being tagged with the version number

### Description

I changed the workflow for pushing the docker images to be invoked by the `release-stable.yml` workflow instead.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/.github/workflows/build-and-dockerize.yaml
+++ b/.github/workflows/build-and-dockerize.yaml
@@ -1,6 +1,10 @@
 on:
   workflow_call:
     inputs:
+      # Reference to checkout
+      ref:
+        default: ''
+        type: string
       dockerize:
         default: true
         type: boolean
@@ -54,6 +58,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
         with:
           fetch-depth: 2
+          ref: ${{ inputs.ref }}
 
       - name: setup environment
         if: ${{ inputs.build }}
@@ -185,7 +190,7 @@ jobs:
 
             <details>
               <summary>Docker Bake metadata</summary>
-              
+
               ```json
               ${{ steps.docker-bake.outputs.metadata }}
               ```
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-dockerize.yaml
```diff
@@ -1,6 +1,10 @@
 on:
   workflow_call:
     inputs:
+      # Reference to checkout
+      ref:
+        default: ''
+        type: string
```
n1ru4l, line 7: I am not really sure whether this is needed. I don't really understand how the context is propagated for a workflow call.

.github/workflows/release-docker.yaml
```diff
@@ -31,6 +32,7 @@ jobs:
     if: ${{ contains(needs.version.outputs.number, '.') }}
     uses: ./.github/workflows/build-and-dockerize.yaml
     with:
+      ref: refs/tags/${{ inputs.tag }}
```
n1ru4l, line 35: I think this is not necessary, since `main.yml` invokes `release-stable.yml` which invokes `release-docker.yml` - which means the ref should be the same as the tagged version?

.github/workflows/build-and-dockerize.yaml
dotansimha: This should work just fine, you can also set `default: ${{ github.ref_name }}` here as far as I know.

.github/workflows/build-and-dockerize.yaml
```diff
@@ -1,6 +1,10 @@
...
       dockerize:
         default: true
         type: boolean
```
dotansimha, line 10: I guess `publishLatest:` also needs to be configured externally using `inputs` when you use workflow call? otherwise it won't never tag as `lastest`

Issue #6742, opened 2025-04-16
Cannot delete target with contract


jdolle:
Recent changes to contracts may have caused an issue with deleting targets:

> ForeignKeyIntegrityConstraintViolationError
> Query violates a foreign key integrity constraint. update or delete on table "targets" violates foreign key constraint "contracts_target_id_fkey" on table "contracts"

[Found via sentry](http://the-guild-z4.sentry.io/issues/6545357643/?environment=prod&project=5691450&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&sort=date&stream_index=4)

We should possibly be using CASCADE deletes.

PR #6792, opened 2025-05-14
Adjust contracts' foreign key to delete when the target is deleted

jdolle:
### Background

Resolves https://github.com/graphql-hive/console/issues/6742

https://linear.app/the-guild/issue/CONSOLE-1206/cannot-delete-target-with-contract

Contracts should delete when the target is deleted

### Description

You cannot alter a constraint to add the cascade deletes, so you must first remove the constraint and then re-add it.

After migration:

```
registry=# \d+ contracts
                                                                           Table "public.contracts"
                     Column                      |           Type           | Collation | Nullable |      Default       | Storage  | Compression | Stats target | Description 
-------------------------------------------------+--------------------------+-----------+----------+--------------------+----------+-------------+--------------+-------------
 id                                              | uuid                     |           | not null | uuid_generate_v4() | plain    |             |              | 
 target_id                                       | uuid                     |           | not null |                    | plain    |             |              | 
 contract_name                                   | text                     |           | not null |                    | extended |             |              | 
 include_tags                                    | text[]                   |           |          |                    | extended |             |              | 
 exclude_tags                                    | text[]                   |           |          |                    | extended |             |              | 
 remove_unreachable_types_from_public_api_schema | boolean                  |           | not null |                    | plain    |             |              | 
 created_at                                      | timestamp with time zone |           | not null | now()              | plain    |             |              | 
 is_disabled                                     | boolean                  |           | not null | false              | plain    |             |              | 
Indexes:
    "contracts_pkey" PRIMARY KEY, btree (id)
    "contracts_connection_pagination" btree (target_id, created_at DESC, id DESC)
    "contracts_target_id_contract_name_key" UNIQUE CONSTRAINT, btree (target_id, contract_name)
Foreign-key constraints:
    "contracts_target_id_fkey" FOREIGN KEY (target_id) REFERENCES targets(id) ON DELETE CASCADE
Referenced by:
    TABLE "contract_checks" CONSTRAINT "contract_checks_contract_id_fkey" FOREIGN KEY (contract_id) REFERENCES contracts(id) ON DELETE CASCADE
    TABLE "contract_schema_change_approvals" CONSTRAINT "contract_schema_change_approvals_contract_id_fkey" FOREIGN KEY (contract_id) REFERENCES contracts(id) ON DELETE CASCADE
    TABLE "contract_versions" CONSTRAINT "contract_versions_contract_id_fkey" FOREIGN KEY (contract_id) REFERENCES contracts(id) ON DELETE SET NULL

```


packages/migrations/src/actions/2025.05.15T00-00-00.contracts-foreign-key-constraint-fix.ts
```diff
@@ -0,0 +1,32 @@
...
+    // add back but this time with cascade delete
+    {
+      name: 'add constraint contracts_target_id_fkey',
+      query: sql`
+        ALTER TABLE "contracts" ADD CONSTRAINT "contracts_target_id_fkey"
+          FOREIGN KEY ("target_id")
+          REFERENCES "targets"("id")
+          ON DELETE CASCADE
+      `,
+    },
```
n1ru4l, line 30:
```suggestion
    {
      name: 'drop constraint contracts_target_id_fkey',
      query: sql`
        ALTER TABLE "contracts"
          DROP CONSTRAINT IF EXISTS "contracts_target_id_fkey"
          , ADD CONSTRAINT "contracts_target_id_fkey"
            FOREIGN KEY ("target_id")
            REFERENCES "targets"("id")
            ON DELETE CASCADE
      `,
    },
```

I think this can be a single statement. 🤔

packages/migrations/src/actions/2025.05.15T00-00-00.contracts-foreign-key-constraint-fix.ts
n1ru4l, line 1: Even though adding a foreign key constraint will block the table - I think it is okay for the following reasons:
- Relatively low amount of records
- Very infrequent writes to the table

Did a bit of research and other ways of using `NOT VALID` etc. seem to not be worth the effort given the table size.

n1ru4l (approved):
Approved with comment

PR #6799, opened 2025-05-15
feat: Add a cancellable connection util; cancel org delete on request abort

jdolle:
### Background

We recently had an org delete call get timed out yet still go through. At least for expensive operations, we should cancel the operation if the request is cancelled. This should provide the best experience to users also who maybe reconsider at the very last minute.

### Description

Adds a utility wrapper to cancel pg connections.

I tried several times to write a test to cover this, but figure out a good way to verify that the db call was cancelled. Integration tests are too far outside the execution to capture the abort, and unit tests can't rely on the necessary pg commands.

I did a local integration test by adjusting the sql statement to be a long sleep, adding some logs, and refreshing the page after the call. And it did cancel the statement.


n1ru4l:
I don't think introducing this complexity is necessary if we instead can make the deletion time sub seconds or async/lazy.

jdolle:
I think you're right that this is excessive in many cases, but on critical operations, this provides a better experience to users and avoids long running queries if we miss future indices.

The complexity is hidding inside the utility function. And the actual usage is pretty simple:
https://github.com/graphql-hive/console/pull/6799/files?diff=split&w=1#diff-9f85c4b4385684df371c05ce89163d5efd753804848e497c6ffc77915e614b37R756

But I don't think this should be used in every call. Maintaining the extra connection to kill the task would be excessive. (Although I think this can be optimized to reuse a single connection for all calls.) But for delete org, and probably delete project, I think it makes sense to use this safeguard.

n1ru4l:
> provides a better experience to users and avoids long running queries if we miss future indices.

What is the user experience improvement here? If people are annoyed by the delete not going through reloading the page and the operation being aborted and them retrying will not really make a difference here. 😓 

I see this as an attempted workaround for actually fixing the core issue, that is not addressed here, which is:
Deletes should be near instant

jdolle:
The delete latency was addressed in a prior PR by adding indices. Although we are unlikely to have another large delete call in production soon, so it's difficult to say definitively that this matter is resolved.

The user experience improvement is that if in the future something is changed to where it once again is missing an index, or if there's a huge org that deletes, that they won't encounter this massive delay in deletion. It's confusing that the request times out and shows an error to the user and then completes in the background.

I agree that this is likely temporary though. In thr long run we should move this to async and tell users it will complete in the background. But i don't think we need that yet.

And there's value in guaranteeing these calls don't run for 10min+ because long running queries can add up to severely impact the system

jdolle:
We're going in another direction for this.

PR #6803, opened 2025-05-16
tests: Use pollFor logic instead of static wait times in integration tests

jdolle:
### Background

Integration tests currently contain a lot of `waitFor` calls which are static wait periods to ensure certain async operations are processed. To avoid excessively flaky tests, the wait period must exceed what is typically necessary. This time adds up.

### Description

A `pollFor` testkit function was recently added. This PR makes use of, and enhances that function, to improve test consistency and speed.

`tests/api/target/usage.spec.ts` was especially impacted by this change. Locally, the time to run these tests went from 87s to 26s.

### Checklist

- [x] Testing

### DIFF
--- a/integration-tests/tests/api/target/usage.spec.ts
+++ b/integration-tests/tests/api/target/usage.spec.ts
@@ -8,13 +8,19 @@ import { graphql } from 'testkit/gql';
 import { BreakingChangeFormulaType, ProjectType } from 'testkit/gql/graphql';
 import { execute } from 'testkit/graphql';
 import { getServiceHost } from 'testkit/utils';
+import { T } from 'vitest/dist/chunks/reporters.d.CfRkRKN2';
 import { UTCDate } from '@date-fns/utc';
 // eslint-disable-next-line hive/enforce-deps-in-dev
 import { normalizeOperation } from '@graphql-hive/core';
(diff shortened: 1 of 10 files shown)


integration-tests/tests/api/target/usage.spec.ts
```diff
@@ -2776,36 +2829,41 @@ test.concurrent(
       },
     });
 
-    await waitFor(10000);
+    let firstSchemaCheckId: string | undefined;
+    await pollFor(async () => {
+      try {
+        const used = await token
```
jdolle, line 2835: I dont love having to run these checks inside the pollFor to determine when something has cycled through the system, but I couldn't figure out a better way for a number of these cases.

jdolle:
I see a similar 60s decrease for the usage spec in github actions, but the total duration of the integration tests wasnt impacted much.
I believe some files get ran in parallel and since there's a couple test files that are significantly longer running than the rest, it slows down the entire workflow.
If we want to improve the speed of our github actions, we can split these large test files into several smaller ones, or work to improve the performance of what they call.
I see that the `publish` command is particularly slow running in our tests, and this is an important operation for us, so I'd start there.

n1ru4l (approved):

PR #6814, opened 2025-05-20
Fix infinite loop on schema checks route

jdolle:
### Background

Schema checks is maintaining filter data as both the url search param and as state and there is a race condition where this can cause an infinite render loop. This is amplified by a small re-render bug in @tanstack/react-router which could cause it to render routes multiple times unnecessarily. (Upgrading tanstack will be in a separate PR)

### Description

This addresses the infinite loop bug and simplifies the route to only use the state from the URL. Lastly, this adjusts some of the render logic to better indicate loading times and to avoid elements popping in and out frequently while loading.


packages/web/app/package.json
```diff
@@ -62,9 +62,10 @@
     "@stripe/react-stripe-js": "3.1.1",
     "@stripe/stripe-js": "5.5.0",
     "@tanstack/react-query": "5.63.0",
-    "@tanstack/react-router": "1.34.9",
+    "@tanstack/react-router": "1.120.5",
     "@tanstack/react-table": "8.20.6",
-    "@tanstack/router-devtools": "1.34.9",
+    "@tanstack/router-devtools": "1.120.6",
+    "@tanstack/zod-adapter": "1.120.5",
```
jdolle, line 68: adds a tanstack -> zod integration to validate search params.

dotansimha (approved):
lgtm, let's wait for Laurin to confirm. CI needs a fix (typecheck) and please let's also validate that the upgrade doesn't break anything else.

jdolle:
I double checked and the infinite loop is fixed without upgrading tanstack. We can save that for a future PR so this can stay small and focused on the real issue.

packages/web/app/src/components/ui/empty-list.tsx
```diff
@@ -22,7 +22,10 @@ export const EmptyList = ({
 }): ReactElement => {
   return (
     <Card
-      className={cn('flex grow cursor-default flex-col items-center gap-y-2 p-4', className)}
+      className={cn(
+        'flex max-h-screen min-h-[400px] grow cursor-default flex-col items-center gap-y-2 p-4',
```
jdolle, line 26: Addresses when the list of checks is very long. It causes this element to grow to where the text is all off screen.

Issue #6819, opened 2025-05-23
invalid error names used for missing author and commit for schema publish


n1ru4l:
The name `GithubCommitRequiredError` and `GithubAuthorRequiredError` is misleading, as the commit and author is nothing that is related to Github. 

https://github.com/graphql-hive/console/blob/d919e2db6444caba25ad25d2e5bff3cd1b6faa14/packages/libraries/cli/src/commands/schema/publish.ts#L241-L248

PR #6820, opened 2025-05-23
fix(cli): use correct error codes and add more context

n1ru4l:
### Background

Closes https://github.com/graphql-hive/console/issues/6819


packages/libraries/cli/src/commands/schema/check.ts
```diff
@@ -227,7 +227,7 @@ export default class SchemaCheck extends Command<typeof SchemaCheck> {
 
       if (usesGitHubApp) {
         if (!commit) {
-          throw new GithubCommitRequiredError();
+          throw new CommitRequiredError();
```
jdolle, line 230: These could probably be generic bad input errors instead of special cases

jdolle (approved):

PR #6836, opened 2025-06-03
Initial schema and pg tables for schema proposals feature

jdolle:
### Background

https://linear.app/the-guild/issue/CONSOLE-1245/collaborative-schema-designerproposer

Our goal is to build a space where developers/teams can come together and collaborate on schema. "Schema Proposals", as it's currently named, is a feature that would do just this. By allowing users to submit a proposed change and others to review and comment.

### Description

This PR lays some of the groundwork for this feature. I have a proposed database structure and initial graphql API.
This very realistically may change even if approved and merged if other requirements arise. But I wanted to get this in front of people in the early stages to avoid as many issues as possible.

### DIFF
--- a/packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
+++ b/packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
@@ -1,0 +1,173 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+/**
+ * This migration establishes the schema proposal tables.
+ */
+export default {
+  name: '2025.05.29T00-00-00.schema-proposals.ts',
+  run: ({ sql }) => [
+    {
+      name: 'create schema_proposal tables',
+      query: sql`
+        CREATE TYPE IF NOT EXISTS
+          schema_proposal_stage AS ENUM('DRAFT', 'OPEN', 'APPROVED', 'IMPLEMENTED', 'CLOSED')
+        ;
+        /**
+         * Request patterns include:
+         * - Get by ID
+         * - List target's proposals by date
+         * - List target's proposals by date, filtered by author/user_id and/or stage (for now)
+         */
+        CREATE TABLE IF NOT EXISTS "schema_proposals"
+        (
+          id UUID PRIMARY KEY DEFAULT uuid_generate_v4 ()
+          , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          , updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          , title VARCHAR(72) NOT NULL
+          , stage schema_proposal_stage NOT NULL
+          , target_id UUID NOT NULL REFERENCES targets (id) ON DELETE CASCADE
+          -- ID for the user that opened the proposal
+          , user_id UUID REFERENCES users (id) ON DELETE SET NULL
+          -- schema version that is used to calculate the diff. In case the version is deleted,
+          -- set this to null to avoid completely erasing the change... This should never happen.
+          , diff_schema_version_id UUID NOT NULL REFERENCES schema_version (id) ON DELETE SET NULL
+        )
+        ;
+        CREATE INDEX IF NOT EXISTS schema_proposals_list ON schema_proposals (
+          target_id
+          , created_at DESC
+        )
+        ;
+        CREATE INDEX IF NOT EXISTS schema_proposals_list_by_user_id ON schema_proposals (
+          target_id
+          , user_id
+          , created_at DESC
+        )
+        ;
+        CREATE INDEX IF NOT EXISTS schema_proposals_list_by_stage ON schema_proposals (
+          target_id
+          , stage
+          , created_at DESC
+        )
+        ;
+        CREATE INDEX IF NOT EXISTS schema_proposals_list_by_user_id_stage ON schema_proposals (
+          target_id,
+          , user_id
+          , stage
+          , created_at DESC
+        )
+        ;
+        -- For performance during schema_version delete
+        CREATE INDEX IF NOT EXISTS schema_proposals_diff_schema_version_id on schema_proposals (
+          diff_schema_version_id
+        )
+        ;
+        -- For performance during user delete
+        CREATE INDEX IF NOT EXISTS schema_proposals_diff_user_id on schema_proposals (
+          user_id
+        )
+        ;
+        /**
+         * Request patterns include:
+         * - Get by ID
+         * - List proposal's latest versions for each service
+         * - List all proposal's versions ordered by date
+         */
+        CREATE TABLE IF NOT EXISTS "schema_proposal_versions"
+        (
+          id UUID PRIMARY KEY DEFAULT uuid_generate_v4 ()
+          , user_id UUID REFERENCES users (id) ON DELETE SET NULL
+          , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          , schema_proposal_id UUID NOT NULL REFERENCES schema_proposals (id) ON DELETE CASCADE
+          , service_name text
+          , schema_sdl text NOT NULL
+        )
+        ;
+        CREATE INDEX IF NOT EXISTS schema_proposal_versions_list_latest_by_distinct_service ON schema_proposal_versions(
+          schema_proposal_id
+          , service_name
+          , created_at DESC
+        )
+        ;
+        CREATE INDEX IF NOT EXISTS schema_proposal_versions_schema_proposal_id_created_at ON schema_proposal_versions(
+          schema_proposal_id
+          , created_at DESC
+        )
+        ;
+        /**
+         * Request patterns include:
+         * - Get by ID
+         * - List proposal's latest versions for each service
+         * - List all proposal's versions ordered by date
+         */
+         /**
+          SELECT * FROM schema_proposal_comments as c JOIN schema_proposal_reviews as r
+            ON r.schema_proposal_review_id = c.id
+            WHERE schema_proposal_id = $1
+            ORDER BY created_at
+            LIMIT 10
+            ;
+          */
+        CREATE TABLE IF NOT EXISTS "schema_proposal_reviews"
+        (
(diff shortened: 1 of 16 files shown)


jasonkuhrt:
Will look tomorrow!

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
```diff
@@ -0,0 +1,173 @@
...
+        CREATE TABLE IF NOT EXISTS "schema_proposals"
+        (
+          id UUID PRIMARY KEY DEFAULT uuid_generate_v4 ()
+          , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          , updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          , title VARCHAR(72) NOT NULL
+          , stage schema_proposal_stage NOT NULL
+          , target_id UUID NOT NULL REFERENCES targets (id) ON DELETE CASCADE
+          -- ID for the user that opened the proposal
+          , user_id UUID REFERENCES users (id) ON DELETE SET NULL
```
jasonkuhrt:
```suggestion
          , user_id UUID REFERENCES users (id) ON DELETE SET NULL
```

nit, I tend to try to bake the semantic of the relation into the column name. Based on the comment above I imagine `author_id` being more descriptive.

More wild thought is relations eliding _id suffixes so you get relation columns like "target" / "author" / ... etc. -- not sure its a net good or evil.

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
```diff
@@ -0,0 +1,173 @@
...
+         * - List proposal's latest versions for each service
+         * - List all proposal's versions ordered by date
+         */
+        CREATE TABLE IF NOT EXISTS "schema_proposal_versions"
+        (
+          id UUID PRIMARY KEY DEFAULT uuid_generate_v4 ()
+          , user_id UUID REFERENCES users (id) ON DELETE SET NULL
+          , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          , schema_proposal_id UUID NOT NULL REFERENCES schema_proposals (id) ON DELETE CASCADE
+          , service_name text
```
jasonkuhrt: What is "service_name"?

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
```diff
@@ -0,0 +1,173 @@
...
+          , stage_transition schema_proposal_stage NOT NULL
+          , user_id UUID REFERENCES users (id) ON DELETE SET NULL
+          , schema_proposal_id UUID NOT NULL REFERENCES schema_proposals (id) ON DELETE CASCADE
+          -- store the originally proposed version to be able to reference back as outdated if unable to attribute
+          -- the review to another version.
+          , original_schema_proposal_version_id UUID NOT NULL REFERENCES schema_proposal_versions (id) ON DELETE SET NULL
+          -- store the original text of the line that is being reviewed. If the base schema version changes, then this is
+          -- used to determine which line this review falls on. If no line matches in the current version, then
+          -- show as outdated and attribute to the original line.
+          , line_text text
```
jasonkuhrt: Could another way be to store a relation to the base schema version?

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
```diff
@@ -0,0 +1,173 @@
...
+            ORDER BY created_at
+            LIMIT 10
+            ;
+          */
+        CREATE TABLE IF NOT EXISTS "schema_proposal_reviews"
+        (
+          id UUID PRIMARY KEY DEFAULT uuid_generate_v4 ()
+          , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          -- null if just a comment
+          , stage_transition schema_proposal_stage NOT NULL
```
jasonkuhrt: For me this wasn't clear. Is it just me or worth expanding a bit?

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
```diff
@@ -0,0 +1,173 @@
...
+         */
+        CREATE TABLE IF NOT EXISTS "schema_proposal_comments"
+        (
+          id UUID PRIMARY KEY DEFAULT uuid_generate_v4 ()
+          , user_id UUID REFERENCES users (id) ON DELETE SET NULL
+          , body TEXT NOT NULL
+          , created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          , updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
+          , schema_proposal_review_id UUID REFERENCES schema_proposal_reviews (id) ON DELETE CASCADE
+        )
```
jasonkuhrt: Could we bake in more semantics here rather than just free form? For example maybe a category enum.

type: kudos | confused | disagree | idea 


The thinking is that structured semantics here would allow some interesting UX downstream, such as us being able to present the user with triaged comments e.g. "blocking" | "optional" etc. to reduce review process overhead.

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
jdolle: i considered `author_user_id`. But I like knowing, based on the column name, what it's referencing. I'm open to removing the ID but I wanted to follow the design of our other tables, which include `_id`

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
jdolle: For federated graphs, it's the name of the subgraph. It gets passed in on schema check/publish etc https://the-guild.dev/graphql/hive/docs/api-reference/cli#publish-a-schema

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
jdolle: Yes, and of course I'm open to adjusting this, but then youd need to calculate this text at runtime by diffing that version and the "original_schema_proposal_version". Otherwise you wouldn't know what the review is referencing and when the diff changes, the line number would likely reference a totally different line.
String matching is unfortunately the only way I could come up with that made sense.

E.g.
on a diff
```
line 10: - type FooUser {
line 11: + type FooUser implements User {
```
If I comment on line 11, and in the next version several lines get added above this, then I want my comment to still reference `+ type FooUser implements User {`

And if this line is modified in later version and no longer has a match, then it the review will revert to being at its original line (11), be flagged as stale, and show a preview of what the line used to be -- which is "line_text".
So I believe this column is essential if we want to be performant and accurate.

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
jasonkuhrt: Yeah it's not very ambiguous here!

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
jdolle: I think this is a great idea but it might belong on the `review` object and is better suited to a v2.

For more context -- One constraint of the current table design is that every comment must be tied to a review. This is because I opted to use sql and I wanted to avoid complicated merging + pagination logic in the UI to display multiple sets or chronological data at the same time.
If using nosql then this could all be in the same table and have different structures.

packages/migrations/src/actions/2025.05.29T00.00.00.schema-proposals.ts
jdolle: Every action on a proposal is a "review". E.g. move back to draft, move to open, approve, close...
I agree this is slightly odd conceptually but I explained the reason in a separate comment below. (TLDR; to keep chronological order of changes clean)

packages/migrations/src/actions/2025.12.30T00-00-00.schema-proposals-part-2.ts
n1ru4l, line 1: this is not used?

packages/services/api/src/modules/auth/lib/authz.ts
```diff
@@ -357,6 +357,18 @@ function defaultAppDeploymentIdentity(
...
+  args: { schemaProposalId: string | null } & Parameters<typeof defaultTargetIdentity>[0],
+) {
+  const ids = defaultTargetIdentity(args);
+
+  if (args.schemaProposalId !== null) {
+    ids.push(`target/${args.targetId}/schemaProposal/${args.schemaProposalId}`);
+  }
+
+  return ids;
+}
```
n1ru4l: Why do we introduce a new permission level here? This implies that we allow users to modify and view individual schema proposals within a target. 🤔 

E.g. he that someone can see one specific schema proposal, isn't this permission on the target good enough for now?

packages/migrations/src/actions/2025.12.30T00-00-00.schema-proposals-part-2.ts
jdolle: I was in the midst of factoring it out for a followup PR, but based on our discussions we want to move forward as is and adjust.
I'll add the import back

packages/services/api/src/modules/auth/lib/authz.ts
jdolle: 💯
I was following what app deployments did without thinking. I've adjusted this.

packages/services/api/src/modules/organization/resolvers/Organization.ts
```diff
@@ -206,8 +205,8 @@ export const Organization: Pick<
       },
     });
   },
-  availableMemberPermissionGroups: () => {
-    return OrganizationMemberPermissions.permissionGroups;
```
jdolle, line 210: This didn't account for feature flags.

PR #6842, opened 2025-06-11
Schema proposals UI list

jdolle:
### Background

The schema proposals feature is a large undertaking with many subsections. This is one such subsection.

### Description

This adds a proposals list UI. This view is intended to be consistent with others while adjusting the UX to better handle a variety of window sizes and content. The side nav can be expanded, contracted, hidden, and it tries to account for the size of content in these various states.

### DIFF
--- a/packages/web/app/src/components/target/proposals/user-filter.tsx
+++ b/packages/web/app/src/components/target/proposals/user-filter.tsx
@@ -1,0 +1,116 @@
+import { useRouter, useSearch } from "@tanstack/react-router";
+import { useMemo, useState } from "react";
+import { Command, CommandEmpty, CommandGroup, CommandInput, CommandItem } from "@/components/ui/command";
+import { Popover, PopoverContent, PopoverTrigger } from "@/components/ui/popover";
+import { ScrollArea } from "@/components/ui/scroll-area";
+import { ChevronsUpDown } from 'lucide-react';
+import { Checkbox } from "@/components/v2";
+import { Button } from "@/components/ui/button";
+import { useQuery } from "urql";
+import { graphql } from '@/gql';
+
+const UsersSearchQuery = graphql(`
+  query UsersSearch($organizationSlug: String!, $after: String, $first: Int) {
+    organization(reference: { bySelector: { organizationSlug: $organizationSlug }}) {
+      viewerCanSeeMembers
+      members(first: $first, after: $after) {
+        edges {
+          node {
+            user {
+              id
+              displayName
+              fullName
+            }
+          }
+        }
+        pageInfo {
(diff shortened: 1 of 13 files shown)


packages/web/app/src/components/target/proposals/user-filter.tsx
```diff
@@ -0,0 +1,127 @@
...
+    query: UsersSearchQuery,
+    variables: {
+      after: pages[pages.length - 1]?.after,
+      first: pages[pages.length - 1]?.first,
+      organizationSlug,
+    },
+  });
+  const search = useSearch({ strict: false });
+  const users = query.data?.organization?.members.edges.map(e => e.node.user) ?? [];
+  // @todo handle preloading selected users to populate on refresh.... And only search on open.
```
jdolle, line 65: This will not be part of this PR, since this is a prototype and not yet dealing with real data.

packages/web/app/src/components/target/proposals/user-filter.tsx
```diff
@@ -0,0 +1,127 @@
...
+      <PopoverContent align="end" className="p-0">
+        <Command>
+          <CommandInput placeholder="Search org members..." />
+          <CommandEmpty>No results.</CommandEmpty>
+          <CommandGroup>
+            <ScrollArea className="relative max-h-screen">
+              {users?.map(user => (
+                <CommandItem
+                  key={user.id}
+                  // @todo upgrade cmdk and use "keywords" attribute
```
jdolle, line 95: this will be done in a separate PR.
Recent versions of `cmdk` have an additional attribute called "keywords" that lets you search by multiple terms. This would allow us to keep the value as `user.id` but let it search by displayName etc.

packages/web/app/src/components/v2/checkbox.tsx
```diff
@@ -1,12 +1,16 @@
...
 import { CheckIcon } from '@radix-ui/react-icons';
 
 export const Checkbox = (props: CheckboxProps): ReactElement => {
   return (
     <Root
-      className="flex size-5 shrink-0 items-center justify-center rounded-sm border border-orange-500 bg-gray-800 text-orange-500 hover:border-orange-700 disabled:cursor-not-allowed disabled:border-gray-900"
       {...props}
+      className={cn(
+        'flex size-5 shrink-0 items-center justify-center rounded-sm border border-orange-500 bg-gray-800 text-orange-500 hover:border-orange-700 disabled:cursor-not-allowed disabled:border-gray-900',
+        props.className,
```
jdolle, line 12: This is a bug with the current checkbox. It shouldn't override the className in its entirety if defined.

PR #6847, opened 2025-06-18
Update resources-requirements.mdx

comatory:
Mention `DEBUG` variable in context of performance.

### Background

During rollout of the gateway, we noticed increased CPU usage when compared to existing gateway service. We investigated that debug logs affect this.

### Description

Mention `DEBUG` env variable in the documentation.

### Checklist

- [X] System configuration

### DIFF
--- a/packages/web/docs/src/content/gateway/deployment/resources-requirements.mdx
+++ b/packages/web/docs/src/content/gateway/deployment/resources-requirements.mdx
@@ -36,3 +36,5 @@ By default, the gateway uses the available parallelism (number of CPUs, based on
 
 You can set the `FORK` environment variable to specify the number of workers you want to spawn. For
 example, to spawn 12 workers, set `FORK=12`.
+
+Consider setting `DEBUG=0` if you don't need debug logs. This can have significant impact on CPU usage.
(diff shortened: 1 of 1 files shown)


packages/web/docs/src/content/gateway/deployment/resources-requirements.mdx
```diff
@@ -36,3 +36,5 @@ By default, the gateway uses the available parallelism (number of CPUs, based on
 
 You can set the `FORK` environment variable to specify the number of workers you want to spawn. For
 example, to spawn 12 workers, set `FORK=12`.
+
+Consider setting `DEBUG=0` if you don't need debug logs. This can have significant impact on CPU usage.
```
dotansimha: @enisdenjo is it enabled by default? I think it shouldn't be the case?

comatory: It's not enabled by default. I just thought it is important to note when rolling out the gateway to production environment. We enabled it to have better visibility into potential problem but did not realize it could've had such a big impact.

enisdenjo:
```suggestion
Consider unsetting the `DEBUG` env variable if you don't need debug logs. This can have significant impact on CPU usage.
```
Because at certain legacy codepoints we do a very naive `if (!process.env.DEBUG)`. 😓

comatory: ok makes sense, this is better

PR #6850, opened 2025-06-20
fix(schema): invalid schema contract composition for shared scalars/types 

n1ru4l:
### Background

We created an invalid supergraph. In order to correctly determind whether a type/field is tagged or not we first need to visit all types, fields, and arguments directives. Since a tag defined on one type/field propagates to all subgraphs.

Related https://github.com/graphql-hive/federation-composition/pull/152

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/schema/src/lib/federation-tag-extraction.ts
+++ b/packages/services/schema/src/lib/federation-tag-extraction.ts
@@ -421,7 +421,14 @@ function collectTagsBySchemaCoordinateFromSubgraph(
     }
   }
 
-  function addTagsPerSchemaCoordinate(schemaCoordinate: string, tagValues: Set<string>) {
+  function addTagsPerSchemaCoordinate(
+    schemaCoordinate: string,
+    tagValues: Set<string> | undefined,
+  ) {
+    if (tagValues === undefined) {
+      return;
+    }
+
     let values = map.get(schemaCoordinate);
     if (values === undefined) {
       values = new Set();
@@ -432,7 +439,9 @@ function collectTagsBySchemaCoordinateFromSubgraph(
     }
   }
 
-  function getTagsForNode(node: { directives?: readonly ConstDirectiveNode[] }): Set<string> {
+  function getTagsForNode(node: {
+    directives?: readonly ConstDirectiveNode[];
+  }): Set<string> | undefined {
     const tags = new Set<string>();
     node.directives?.forEach(directiveNode => {
       const tagValue = extractTag(directiveNode);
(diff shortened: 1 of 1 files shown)


packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
```diff
@@ -366,7 +399,7 @@ describe('applyTagFilterToInaccessibleTransformOnSubgraphSchema', () => {
       }
 
       type Type2 @federation__inaccessible {
-        field1: ID!
+        field1: ID! @federation__inaccessible
```
n1ru4l, line 402: `Type2` has is tagged with `tag2`, so the tagging propagates down to `Type2.field1`, making it also excluded, since excluded takes precedence.

See https://www.apollographql.com/docs/graphos/platform/schema-management/delivery/contracts/reference#dependent-tags

packages/services/schema/src/lib/__tests__/federation-tag-extraction.spec.ts
```diff
@@ -1813,7 +1974,7 @@ describe('applyTagFilterOnSubgraphs', () => {
 
     expect(print(result[0].typeDefs)).toMatchInlineSnapshot(`
       type Query {
-        field1: String! @inaccessible
+        field1: String!
```
n1ru4l, line 1977: Since we now create type register before, this is no longer marked as inaccessible.

packages/services/schema/src/composition/__tests__/federation.spec.ts
```diff
@@ -519,3 +519,78 @@ test('contract: mutation type is part of the public schema if at least one field
...
+
+  expect(contractResult?.id).toEqual('1');
+  expect(contractResult?.result.type).toEqual('success');
+
+  const line = contractResult?.result.result.supergraph
+    ?.split('\n')
+    .find(line => line.includes('scalar Brr'));
+  expect(line).toEqual(
+    `scalar Brr @join__type(graph: FOO_GRAPHQL)  @join__type(graph: FOO2_GRAPHQL) `,
+  );
```
n1ru4l, line 595: Before the changes this line would also include `@inaccessible`.

jdolle (approved):
Looks great. I love that we're gradually shifting more and more to extracting import names using links instead of hardcoding directive names.

When released, let's monitor composition times just to make sure this doesn't increase latency by much. I don't think it will, but composition is already a relatively high latency call.

We may want to take a pass at optimizing any low hanging fruit to get that latency down.

dotansimha (approved):

PR #6863, opened 2025-06-26
Add logs for active subscriptions that are missing payment methods

jdolle:
### Background

We recently had an issue where a customer was unable to update their payment information even though they have an active subscription.

### Description

This adds logs to help us better understand if the missing payment method case is happening. And it updates the payment method call to use stripe's recommended "List Customer Payment Methods" api

### DIFF
--- a/packages/services/commerce/src/stripe-billing/index.ts
+++ b/packages/services/commerce/src/stripe-billing/index.ts
@@ -92,10 +92,13 @@ export const stripeBillingRouter = router({
 
       const actualSubscription = subscriptions[0] || null;
 
-      const paymentMethod = await ctx.stripeBilling.stripe.customers.listPaymentMethods(customer.id, {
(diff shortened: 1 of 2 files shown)


packages/services/commerce/src/stripe-billing/index.ts
```diff
@@ -92,9 +92,9 @@ export const stripeBillingRouter = router({
 
       const actualSubscription = subscriptions[0] || null;
 
-      const paymentMethod = await ctx.stripeBilling.stripe.paymentMethods.list({
-        customer: customer.id,
+      const paymentMethod = await ctx.stripeBilling.stripe.customers.listPaymentMethods(customer.id, {
```
jdolle: if preferred, we can keep this call the same and assess before making this change. LMK

n1ru4l (approved):

PR #6888, opened 2025-07-10
fix(commerce): avoid overriding email when updating subscription

dotansimha:
The condition doesn't make any sense, it's overriding the email to the one we have on record. Since subscription's main email is managed externally on Stripe, we don't really need to hold it on our end, or update it at all. A follow up PR can even remove this column from the database, as I don't this it's being used. 

Users are able to update their billing email using the Stripe magic link. 

Fixes CONSOLE-1308


n1ru4l (approved):
LGTM

kamilkisiela:
> LGTM

metoo

PR #6889, opened 2025-07-10
docs: links to features in pricing table

n1ru4l:


Urigo (approved):
awesome work!
for gateway, should we add Defer and Role-based Access Control (RBAC)?
RBAC mentioned for Console but not for Gateway

PR #6897, opened 2025-07-21
docs: update to logger-winston

PlayAnyData:
If I see your published packages correctly, @graphql-hive/winston has only been published as an alpha version and was succeeded by @graphql-hive/logger-winston


enisdenjo (approved):
Thanks!

PR #6898, opened 2025-07-21
feat(deployment): upgrade contour to 1.31.1 (chart 20.0.3)

dotansimha:
- [x] upgrade chart (https://github.com/projectcontour/contour/releases/tag/v1.32.0) 
```
bitnami/contour         	20.0.3       	1.31.1    
```
- [x] fix generate script 
- [x] update types 
- [x] pass typecheck 
- [x] experiment on `dev`


deployment/generate.ts
```diff
@@ -64,37 +61,8 @@ async function generateVectorDevTypes() {
...
-
-  const valuesFile = await fetch(url)
-    .then(r => r.arrayBuffer())
-    .then(r => Buffer.from(r))
-    .then(r => gunzipSync(r))
-    .then(r => Readable.from(r))
-    .then(r => getFileFromTar(r, 'contour/values.yaml'))
-    .then((r: any) => r.toString());
+  const helmValuesFileUrl = `https://raw.githubusercontent.com/bitnami/charts/contour/${CONTOUR_CHART.version}/bitnami/contour/values.yaml`;
+  const valuesFile = await fetch(helmValuesFileUrl).then(r => r.text());
```
jdolle, line 36: Does this file need validated at all in case it's ever not the format we expect?

dotansimha: It's always YAML so all good here ✅

dotansimha: And if it's not, it will fail later when we try to generate TS types from that YAML 😆

Issue #6816, opened 2025-05-21
support advanced nullability reporting in apollo-router hive plugin


n1ru4l:
The changes introduced to the Hive JS SDK in https://github.com/graphql-hive/console/pull/6764 must be ported to the apollo-router hive plugin.

Whether an input field is used as nullable or non-nullable must be included within the reported schema coordinates using the `!` character.

dotansimha:
This one is done. Thank you @egoodwinx ! 

Will be available in next release :)

PR #6906, opened 2025-07-30
feat: Advanced breaking change detection for inputs and arguments for apollo-router-hive-fork

egoodwinx:
### Background
Addressing: https://github.com/graphql-hive/console/issues/6816

Whether an input field is used as nullable or non-nullable must be included within the reported schema coordinates using the ! character.

### Description
To accurately determine whether a change is breaking or not based on traffic, we must track whether or not an argument or input's value has actually been provided or not.

Track used variables, if input variables were provided, non null variables, and variables with defaults.

Add unit tests based on https://github.com/graphql-hive/console/issues/6816

**Didn't E2E test - just added unit tests.**

**Also not sure how to check passed in variables if anyone could help with that.**

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/.changeset/sixty-guests-fall.md
+++ b/.changeset/sixty-guests-fall.md
@@ -2,4 +2,4 @@
 'hive-apollo-router-plugin': patch
 ---
 
-Update Rust version to 1.89
+Update Rust version to 1.90
--- a/.github/workflows/apollo-router-updater.yaml
+++ b/.github/workflows/apollo-router-updater.yaml
@@ -28,7 +28,7 @@ jobs:
       - name: Install Rust
         uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1
         with:
-          toolchain: '1.89.0'
+          toolchain: '1.90.0'
           default: true
           override: true
 
--- a/.github/workflows/publish-rust.yaml
+++ b/.github/workflows/publish-rust.yaml
@@ -84,7 +84,7 @@ jobs:
       - name: Install Rust
         uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # v1
         with:
-          toolchain: '1.89.0'
+          toolchain: '1.90.0'
           target: ${{ env.RUST_TARGET }}
           default: true
           override: true
--- a/docker/router.dockerfile
+++ b/docker/router.dockerfile
@@ -2,7 +2,7 @@
 FROM scratch AS pkg
 FROM scratch AS config
 
-FROM rust:1.89 AS build
+FROM rust:1.90-slim-bookworm AS build
 
 # Required by Apollo Router
 RUN apt-get update
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.89.0"
+channel = "1.90.0"
 components = ["rustfmt", "clippy"]
 profile = "minimal"


packages/libraries/router/src/graphql.rs
```diff
@@ -41,29 +46,34 @@ pub fn collect_schema_coordinates(
...
-            match named_type {
-                Some(named_type) => match named_type {
+    } else {        
+        for type_name in ctx.used_input_fields {
+            if is_builtin_scalar(&type_name) {
+                ctx.schema_coordinates.insert(type_name);
+            } else if let Some(type_def) = schema.type_by_name(&type_name) {
+                match type_def {
+                    TypeDefinition::Scalar(scalar_def) => {
+                        ctx.schema_coordinates.insert(scalar_def.name.clone());
```
jdolle, line 70: Curious -- why clone scalar name but not type_name above?

packages/libraries/router/src/graphql.rs
```diff
@@ -178,25 +329,29 @@ impl<'a> OperationVisitor<'a, SchemaCoordinatesContext> for SchemaCoordinatesVis
...
+        ctx.used_input_fields.insert(type_name.clone());
 
-        ctx.input_types_to_collect.insert(type_name);
+        if let Some(default_value) = &var.default_value {
+            self.process_default_value(info, ctx, &type_name, default_value);
+        }
     }
-
-    fn enter_argument(
+        fn enter_argument(
```
jdolle: minor: Formatting^

packages/libraries/router/src/graphql.rs
```diff
@@ -214,60 +369,68 @@ impl<'a> OperationVisitor<'a, SchemaCoordinatesContext> for SchemaCoordinatesVis
...
+                    let arg_type_name = self.resolve_type_name(arg_def.value_type.clone());
+                    
+                    match arg_value {
+                        Value::Enum(value) => {
+                            let value_str: String = value.to_string();
+                            ctx.schema_coordinates
+                                .insert(format!("{arg_type_name}.{value_str}").to_string());
+                        }
+                        Value::List(_) => {
+                            // handled by enter_list_value
```
jdolle, line 415: 👍

packages/libraries/router/src/graphql.rs
```diff
@@ -214,60 +369,68 @@ impl<'a> OperationVisitor<'a, SchemaCoordinatesContext> for SchemaCoordinatesVis
...
-
-            let arg_value = arg.1.clone();
+            let (arg_name, arg_value) = arg;
+
+            let coordinate = format!("{type_name}.{field_name}.{arg_name}");
+            
+            let has_value = match arg_value {
+                Value::Null => false,
+                Value::Variable(var_name) => {
+                    ctx.variables_with_defaults.contains(var_name)
```
jdolle, line 393: should you check `ctx.non_null_variables` here also?

packages/libraries/router/src/graphql.rs
egoodwinx: scalar_def here is a shared reference so it needs to be a copy to insert into schema coordinates, type_name is in a loop so its owned by the loop and can be directly inserted from my understanding

packages/libraries/router/src/graphql.rs
egoodwinx: yes I think so

packages/libraries/router/src/graphql.rs
```diff
@@ -382,7 +383,7 @@ impl<'a> OperationVisitor<'a, SchemaCoordinatesContext> for SchemaCoordinatesVis
             let has_value = match arg_value {
                 Value::Null => false,
                 Value::Variable(var_name) => {
-                    ctx.variables_with_defaults.contains(var_name)
+                    ctx.variables_with_defaults.contains(var_name) && ctx.non_null_variables.contains(var_name)
```
jdolle: I think this should be an `or`.
Please add a test specific to this case so we can confirm this is behaving as expected.

jdolle (approved):
This looks good to me.
The output matches what our js library produces.

@kamilkisiela could you please review the GoLang specifics?

dotansimha:
Thanks @egoodwinx , this is great! 

Can you please elaborate on this? I can try to help 

> Also not sure how to check passed in variables if anyone could help with that.

.changeset/breezy-forks-press.md
```diff
@@ -0,0 +1,5 @@
+---
+'hive-apollo-router-plugin': major
```
dotansimha: I think this could be `minor` as it adds new functionality, and doesn't really break anything?

packages/libraries/router/src/graphql.rs
```diff
@@ -703,7 +909,7 @@ mod tests {
         type Query {
             project(selector: ProjectSelectorInput!): Project
             projectsByType(type: ProjectType!): [Project!]!
-            projectsByTypes(types: [ProjectType!]!): [Project!]!
+            projectsByTypes(types: [ ProjectType!]!): [Project!]!
```
dotansimha, line 940: oops, space can be removed

dotansimha (commented):
lgtm. Added a few minor notes. 

@egoodwinx i think it worth confirming that we didn't miss any test case, based on the JS tests here? https://github.com/graphql-hive/console/blob/main/packages/libraries/core/tests/collect-schema-coordinates.spec.ts

dotansimha:
Pushed the fixes for changeset, and also addressed unrelated issues with Rust version.

dotansimha:
@n1ru4l I pushed some changes and fixed the Rust compile issue. As far as I can see, we can proceed once you approve :)

PR #6907, opened 2025-07-30
fix: do not apply contract logic on external object's fields

jdolle:
### Background

Requires federation-composition upgrade https://github.com/graphql-hive/federation-composition/pull/167

### Description

A recent change fixed a bug with `@external` fields for contracts. This extends that fix to objects.


jdolle:
This code was migrated in https://github.com/graphql-hive/federation-composition/pull/180

PR #6909, opened 2025-07-31
fix: correctly trim operation name on insights and add popover title

jdolle:
### Background

https://linear.app/the-guild/issue/CONSOLE-1328/fix-cutoff-long-operation-names-in-insights-table
The operation name is being truncated but is center aligned and is not using ellipsis to indicate there is more to the name.

### Description

This change fixes the truncation and adds an overflow-x scroll effect to the operation stats table to prevent the contents from spilling outside the table and to better indicate that there is more content to this table.
Screenshot 2025-07-31 at 2 39 45 PM


n1ru4l (approved):
Looking good, can you also add a changeset?

PR #6918, opened 2025-08-21
fix: Grant access when permissions don't have a resource set

jdolle:
### Background

When a permission is not scoped to a specific resource, it fails the resource check.

### Description

This adjust the assertion logic to pass the resource check and then run the action check.

### DIFF
--- a/packages/web/docs/src/content/gateway/other-features/security/demand-control.mdx
+++ b/packages/web/docs/src/content/gateway/other-features/security/demand-control.mdx
@@ -62,8 +62,8 @@ export const gatewayConfig = defineConfig({
 
 When a query is planned by the gateway,
 
-- For each operation, the `operationTypeCost` function is called with the operation type. (By default
-  only mutations have a cost `10`)
+- For each operation, the `operationTypeCost` function is called with the operation type. (By
+  default only mutations have a cost `10`)
 - For each field in the operation, the `fieldCost` function is called with the field node. (By
   default fields have a cost of `0`)
 - For each type in the operation, the `typeCost` function is called with the type. (By default


packages/services/api/src/modules/auth/lib/authz.spec.ts
```diff
@@ -213,4 +213,32 @@ describe('Session.assertPerformAction', () => {
...
+    const bId = 'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb';
+    const cId = 'cccccccc-cccc-cccc-cccc-cccccccccccc';
+
+    const session = new TestSession([
+      {
+        effect: 'allow',
+        action: 'appDeployment:create',
+        resource: [],
+      },
+    ]);
```
n1ru4l, line 228: If you want to allow creation on app deployments on any target the resource must be

```json5
// within any target in the org
{
  effect: 'allow',
  action: 'appDeployment:create',
  resource: [`hrn:aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa:organization/*`],
},
```

or

```json5
// within specific target
{
  effect: 'allow',
  action: 'appDeployment:create',
  resource: [`hrn:aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa:target/${cId}`],
},
```

or

```json5
// within any target in the project
{
  effect: 'allow',
  action: 'appDeployment:create',
  resource: [`hrn:aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa:project/${bId}`],
},
```

packages/services/api/src/modules/auth/lib/authz.spec.ts
```diff
@@ -213,4 +213,32 @@ describe('Session.assertPerformAction', () => {
...
+  test('Allow if no resource was specified on the permission', async () => {
+    const orgId = 'aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa';
+    const bId = 'bbbbbbbb-bbbb-bbbb-bbbb-bbbbbbbbbbbb';
+    const cId = 'cccccccc-cccc-cccc-cccc-cccccccccccc';
+
+    const session = new TestSession([
+      {
+        effect: 'allow',
+        action: 'appDeployment:create',
+        resource: [],
```
n1ru4l, line 226: Supporting this allow an to create app deployments within any organization, this also breaks an existing integration test for permissions see [`prevent access to projects with assigned resources on member` ](https://github.com/graphql-hive/console/actions/runs/17136185798/job/48613650721?pr=6918)

PR #6919, opened 2025-08-21
Hive json default

jdolle:
### Background

If the hive.json is used but a specific key is not provided, then the Config.get logic returns null, not undefined.
`packages/libraries/cli/src/helpers/config.ts`

This causes the fallback logic to not run and the value to be set to null in this case.

### Description

This change checks for null and undefined before falling back to the defaultValue.

### DIFF
--- a/packages/libraries/cli/src/base-command.ts
+++ b/packages/libraries/cli/src/base-command.ts
@@ -153,7 +153,7 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
     } else {
       const configValue = this._userConfig!.get(key) as GetConfigurationValueType<TKey>;
 
-      if (configValue !== undefined && configValue !== null) {
+      if (configValue != null) {
         value = configValue;
       } else if (defaultValue) {
         value = defaultValue;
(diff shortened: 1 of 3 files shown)


packages/libraries/cli/src/base-command.ts
```diff
@@ -151,11 +151,9 @@ export default abstract class BaseCommand<T extends typeof Command> extends Comm
     } else if (envName && env[envName] !== undefined) {
       value = env[envName] as TArgs[keyof TArgs] as NonNullable<GetConfigurationValueType<TKey>>;
     } else {
-      const configValue = this._userConfig!.get(key) as NonNullable<
-        GetConfigurationValueType<TKey>
-      >;
+      const configValue = this._userConfig!.get(key) as GetConfigurationValueType<TKey>;
 
-      if (configValue !== undefined) {
+      if (configValue !== undefined && configValue !== null) {
```
n1ru4l: nit: this could be

```suggestion
      if (configValue != null) {
```

n1ru4l (approved):

PR #6929, opened 2025-08-27
fix appDeployment access token permissions presentation and name filter

jdolle:
### Background

Selecting granular appDeployments vs "ALL" was addressed by https://github.com/graphql-hive/console/pull/6924 , but there are additional bugs around appDeployment permissions. This addresses those.

### Description

Fixes appDeployment names being filtered out because the name was not being referenced correctly, and fixes the local appDeployment mode referencing the service's mode.

Tested locally -- this shot shows I created and can see an appDeployment with granular and all modes.
Screenshot 2025-08-27 at 2 32 36 PM


packages/services/api/src/modules/organization/providers/resource-assignments.ts
```diff
@@ -235,7 +235,9 @@ export class ResourceAssignments {
                   mode: 'granular',
                   appDeployments:
                     targetRecord.appDeployments.appDeployments
-                      ?.filter(name => AppDeploymentNameModel.safeParse(name).success)
+                      ?.filter(
+                        ({ appDeployment: name }) => AppDeploymentNameModel.safeParse(name).success,
+                      )
```
n1ru4l, line 240: 👀

n1ru4l (approved):
Good catch, thank you!

Issue #6920, opened 2025-08-22
Cannot select "All" app deployments in target when creating an access token


n1ru4l:
In the access token creation form it is currently impossible to grant access on all app deployments within a single target.

When the "Apps" tab is active, toggling between "Select / All" does nothing.

When you select "All" within the "Services" tab, switching to the "Apps" tab shows the state of the "Service" tab.

<img width="668" height="462" alt="Image" src="https://github.com/user-attachments/assets/e8ba0e01-17f4-4a28-bcff-75a64f68b592" />

jdolle:
It's now possible to select all since #6924 
And UX improvements that make this much more clear are in #6935 

There's still opportunity to improve the UX but for now this ticket is resolved.

PR #6935, opened 2025-09-01
fix: AppDeployment permissions granted to all if none selected by def…

jdolle:
…ault in access token UI. Fix app deployments feature flag check

### Background

Relates to https://github.com/graphql-hive/console/issues/6920

Since appdeploys are hidden behind a toggle atm, the granular permission is a bit confusing. The user doesnt automatically see that they need to select an appDeployment.

ALL is the most reasonable default for these permissions until the UI can be adjusted to potentially better indicate the need to select an app deployment.

### Description

Fixes some confusion around app deployment permissions when creating an access token.
Fixes environment variable reference in Organization.availableOrganizationAccessTokenPermissionGroups

Screenshot 2025-09-01 at 12 04 09 PM


packages/services/api/src/modules/app-deployments/resolvers/Organization.ts
```diff
@@ -13,8 +13,6 @@ import type { OrganizationResolvers } from './../../../__generated__/types';
 export const Organization: Pick<OrganizationResolvers, 'isAppDeploymentsEnabled' | '__isTypeOf'> = {
   /* Implement Organization resolver logic here */
   isAppDeploymentsEnabled(organization, _, { injector }) {
-    return (
-      injector.get<boolean>(APP_DEPLOYMENTS_ENABLED) || organization.featureFlags.appDeployments
-    );
+    return injector.get<boolean>(APP_DEPLOYMENTS_ENABLED, organization.featureFlags.appDeployments);
```
n1ru4l: @jdolle Previously the logic was check whether they are enabled globally for everyone, otherwise check if it is enabled for this organization.

Now if it is is not enabled by default for everyone, we cannot enable it for a single organization. I think we need to revert these lines

n1ru4l (approved):
thanks! CI failed because of missing docker images, I merged main in here so now everything should hopefully pass 🤞

Issue #6950, opened 2025-09-09
Make deprecatedSchema & unusedSchema part of the public API


martyganz:
Like shortly discussed at GQLConf. 

I would like to query deprecatedSchema & unusedSchema from Hive's public API so I can spin up an mcp server locally and see how good co-pilot is at cleaning up the unused/deprecated bits which are also not used by us on production (your usage statistics will come in handy there).

PR #6951, opened 2025-09-09
tag unusedSchema & deprecatedSchema as public API

martyganz:
### Background

#6950 

### Description

API is affected as it makes the private used API for deprecatedSchema & unusedSchema public

We could also consider adding a new field which does not ask for an input or period and just gives back everything it can find, but that will be a very long list in our case atleast :)

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/schema/module.graphql.ts
+++ b/packages/services/api/src/modules/schema/module.graphql.ts
@@ -796,8 +796,8 @@ export default gql`
     Experimental: This field is not stable and may change in the future.
     """
     explorer(usage: SchemaExplorerUsageInput): SchemaExplorer
-    unusedSchema(usage: UnusedSchemaExplorerUsageInput): UnusedSchemaExplorer
-    deprecatedSchema(usage: DeprecatedSchemaExplorerUsageInput): DeprecatedSchemaExplorer
+    unusedSchema(usage: UnusedSchemaExplorerUsageInput): UnusedSchemaExplorer @tag(name: "public")
+    deprecatedSchema(usage: DeprecatedSchemaExplorerUsageInput): DeprecatedSchemaExplorer @tag(name: "public")
 
     schemaCompositionErrors: SchemaErrorConnection @tag(name: "public")
 


n1ru4l:
image

You unfortunately also need to tag all the referenced types. 😄

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -433,11 +433,11 @@ export default gql`
       )
     """
     The severity level of this schema change.
-    Note: A schema change with the impact \`SeverityLevelType.BREAKING\` can still be safe based on the usage (\`SchemaChange.isSafeBasedOnUsage\`).
+    Note: A schema change with the impact 'SeverityLevelType.BREAKING' can still be safe based on the usage ('SchemaChange.isSafeBasedOnUsage').
     """
     severityLevel: SeverityLevelType! @tag(name: "public")
     """
-    The reason for the schema changes severity level (\`SchemaChange.severityLevel\`)
+    The reason for the schema changes severity level ('SchemaChange.severityLevel')
```
n1ru4l, line 440: Just some cosmetic changes, it was driving me nuts that syntax highlighting in this file was broken.

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -925,15 +974,27 @@ export default gql`
...
+    """
+    The usage of the type within the specified period.
+    """
+    usage: SchemaCoordinateUsage!
+    """
+    Metadata specific to Apollo Federation Projects.
+    Is null if no meta information is available (e.g. this is not an apollo federation project).
+    """
+    supergraphMetadata: SupergraphMetadata
+  }
```
n1ru4l, line 1002: Having this as an interface type makes it slightly less painful to query.

PR #6993, opened 2025-09-15
fix(deps): cleanup some noisy cves, latest pnpm, update deps, fix rust build

dotansimha:
- [x] align nodejs version in ci pipelines
- [x] upgrade pnpm to latest
- [x] fix lockfile
- [x] address some recent cves  
- [x] upgrade vite to latest 
- [x] upgrade rust version to latest and fix rust deps issues


packages/web/docs/next.config.js
```diff
@@ -10,6 +10,7 @@ export default withGuildDocs({
       treeShaking: true,
     },
   },
+
```
n1ru4l, line 13:
```
👁️ 👓 👁️ 
   👃
```

PR #6995, opened 2025-09-15
docs(pricing): clarify high volume discounts

n1ru4l:
People ask me a lot what is the actual pricing and the answer is it depends. But we have this here because we want to give you a price that is lower than pro.

Screenshot 2025-09-15 at 12 51 55


hasparus:
I'll take the liberty of merging it without further reviews because it's a trivial change (and I missclicked the button).

PR #7007, opened 2025-09-16
feat: native composition report

n1ru4l:
### Background

This should make it easier for users to see the differences in the composition results and share them with us, so we can fix them within our composition library.

This also changes our authorization layer to automatically grant the `describe` action on all resources if they are an admin account, making it easier to access the data of projects from customers having trouble.

### DIFF
--- a/packages/services/api/src/modules/auth/lib/authz.ts
+++ b/packages/services/api/src/modules/auth/lib/authz.ts
@@ -281,16 +281,22 @@ function isActionMatch(actionContainingWildcard: string, action: string) {
   if (actionContainingWildcard === '*') {
     return true;
   }
+
   // exact match
   if (actionContainingWildcard === action) {
     return true;
   }
 
-  const [actionScope] = action.split(':');
+  const [actionScope, actionId] = action.split(':');
   const [userSpecifiedActionScope, userSpecifiedActionId] = actionContainingWildcard.split(':');
 
   // wildcard match "scope:*"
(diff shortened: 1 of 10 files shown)


packages/services/api/src/modules/auth/lib/authz.ts
```diff
@@ -514,7 +520,7 @@ const actionDefinitions = {
 
 type Actions = keyof typeof actionDefinitions;
 
-type ActionStrings = Actions | '*';
+type ActionStrings = Actions | '*' | '*:describe';
```
n1ru4l, line 523: I listed this as a special action string in order to avoid having to get a major in typescript and further cooking the ide performance.

n1ru4l: Aside from `describe`, I see no other wildcard usage right now.

packages/services/api/src/modules/auth/lib/supertokens-strategy.ts
```diff
@@ -67,6 +67,18 @@ export class SuperTokensCookieBasedSession extends Session {
...
+      // This makes it much more pleasant to debug.
+      if (user.isAdmin) {
+        return [
+          {
+            action: '*:describe',
+            effect: 'allow',
+            resource: `hrn:${organizationId}:organization/${organizationId}`,
+          },
+        ];
+      }
```
n1ru4l, line 80: This allows admin accounts to debug customer issues

PR #7030, opened 2025-09-19
pass selected operations filter to query for pagination; add operation counts by selected client

jdolle:
### Background

https://linear.app/the-guild/issue/CONSOLE-1359/investigate-case-sensitivity-bug-in-hive-insights-tab

We received a report of filtering not behaving correctly on the operations list view.

### Description

This passes the list of operations to the query so that the requested list includes only those operations. This is to avoid over fetching and having to filter client side.

And I've adjusted the dropdown for the operation list on insights so that it shows the number of requests for the selected clients filter, and the total count.

Screenshot 2025-09-19 at 5 40 49 PM

If we like this, then we can apply the same logic to the client dropdown so that each accurately reflects what will be shown once filtered.

### DIFF
--- a/packages/web/app/src/components/target/insights/List.tsx
+++ b/packages/web/app/src/components/target/insights/List.tsx
@@ -241,7 +241,7 @@ function OperationsTable({
       )}
     >
       <Section.Title>Operations</Section.Title>
-      <Section.Subtitle>List of all operations with their statistics</Section.Subtitle>
+      <Section.Subtitle>List of all operations with their statistics, filtered by selected clients.</Section.Subtitle>
 
       <Table>
         <THead>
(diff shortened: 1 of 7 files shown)


scripts/seed-local-env.ts
```diff
@@ -20,6 +20,7 @@ const usageReportingEndpoint =
     : envName === 'dev'
       ? 'https://app.dev.graphql-hive.com/usage'
       : 'http://localhost:4001';
+const target = process.env.TARGET;
```
jdolle, line 23: I did this for testing because I was using a new access token and they require passing in the target.

packages/web/app/src/components/target/insights/List.tsx
```diff
@@ -410,13 +412,6 @@ function OperationsTableContainer({
     const records: Operation[] = [];
     if (operationStats) {
       for (const { node: op } of operationStats.operations.edges) {
-        if (
```
jdolle, line 413: this is filtered in our resolver now.

GabrielDelepine:
The UI could be confusing since the client filter didn’t affect this list. This update clears that up nicely.

PR #7032, opened 2025-09-22
Edit diff for logger

PlayAnyData:
Edit diff for loggers to make visible, that Logger is also newly imported


PlayAnyData:
Done. pls recheck :-)

PR #7035, opened 2025-09-22
feat: Add selected counts to insights client filter

jdolle:
### Background

Same as https://github.com/graphql-hive/console/pull/7030 but for the client filter component

### Description
Screenshot 2025-09-22 at 9 50 26 AM


n1ru4l (approved):
nice addition 🥳

PR #7045, opened 2025-09-24
fix: adjust token creation ui to make toggling all/granular and services/apps more intuitive

jdolle:
### Background

Existing UI is unclear if the Services / Apps toggle is clickable.

### Description

- Auto selects newly added resources
- Replaced "Select / All" with a checkbox
- Split "Services / Apps" to separate toggleable sections
- Added counts for services/apps when section is hidden

Screenshot 2025-09-25 at 5 37 57 PM


jdolle:
Moved to draft because I think more substantive changes are warranted.

n1ru4l:
image


@jdolle Can we make a smaller checkbox variant for this use-case, i feels a bit oversized here?

n1ru4l:
Screenshot 2025-09-26 at 09 58 05
224910

Maybe we can completely remove the space between the Service and App header if the Service is collapsed?

jdolle:
Screenshot 2025-09-26 at 11 01 58 AM

PR #7046, opened 2025-09-24
add progress log to cli app:create command

jdolle:
### Background

https://linear.app/the-guild/issue/CONSOLE-1378/hive-cli-appcreate-retries-and-never-exits-on-its-own

Investigating this shows that perhaps app:create is working but is taking a long time due to a large number of documents.

### Description

This change adds a progress log so that in these cases, users can see why this command is taking so long and roughly estimate the remaining time.


n1ru4l (approved):
nice addition 🥳

Issue #7044, opened 2025-09-24
App Deployments list needs to be searchable


jdolle:
Currently the app deployments list is an infinite list and there's no way of finding a specific element. We should allow users to more easily navigate to specific versions. This could be done via another input on the page or using Hive CLI or a multitude of other options.

PR #7060, opened 2025-09-29
feat: search app deployment version by operation name

jdolle:
### Background

Relates to https://github.com/graphql-hive/console/issues/7044

### Description

Adds search to an app deployment by operation name. I did not modify the indices because the number of operations per app deployment version isnt expected to be outside of a thousand or so. Even for large apps.

Screenshot 2025-09-29 at 12 43 14 PM
Screenshot 2025-09-29 at 12 43 04 PM
Screenshot 2025-09-29 at 12 43 49 PM

### DIFF
--- a/.changeset/thirty-laws-hang.md
+++ b/.changeset/thirty-laws-hang.md
@@ -1,0 +1,5 @@
+---
+'hive': minor
+---
+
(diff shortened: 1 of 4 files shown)


n1ru4l (approved):
nice 🥳

PR #7063, opened 2025-09-30
Integrate Server Preset v0.13

eddeee888:
### Description

Server Preset v0.13 comes with better integration with the base `typescript-resolvers` types. Most notably, meta resolvers like `__isTypeOf` will only be picked from the base type if it is available.

Previously, `__isTypeOf` is generated on every object type, and is always picked, which can be misleading as this is not the default GraphQL behaviour. 

This is just a type change so should have no runtime impact. Note that in the future, Server Preset may export and enforce consistent type to avoid having to update resolver files like this on version bumps e.g.

```ts
import type { ServerPreset } from './../../../__generated__/types';

export const Project: ServerPreset['resolvers']['Project'] = { // alias of `Pick<ProjectResolvers, 'alertChannels' | 'alerts'>`
```


packages/services/api/src/modules/organization/resolvers/Organization.ts
```diff
@@ -34,11 +34,7 @@ export const Organization: Pick<
   | 'viewerCanModifySlug'
   | 'viewerCanSeeMembers'
   | 'viewerCanTransferOwnership'
-  | '__isTypeOf'
 > = {
-  __isTypeOf: organization => {
-    return !!organization.id;
-  },
```
eddeee888, line 41: `Organization` shouldn't be able to implement `__isTypeOf`:
- it is not an implementing type of an Interface
- it is not a Union member

The new codegen version checks the above to correctly add `__isTypeOf` to generated type

So, this function has been here but it hasn't been called.

n1ru4l (approved):
looks good 🥳

PR #7066, opened 2025-09-30
Fix first schema check landing width & sidenav loading flicker

jdolle:
### Background

I know there was a ticket or issue somewhere for this width but I can't find it...
Also took the opportunity to improve the ux for checks by tweaking the loading logic

Current:
Screenshot 2025-09-30 at 10 16 28 AM

### Description
Screenshot 2025-09-30 at 9 07 18 AM

and when adjusting the filter on the check sidebar nav, currently it flickers a loading indicator. This is because of the placement of the spinner and that the loading is not debounced/delayed.

This logic adds a delay to showing the loader, so that it only shows after 500ms which should cause it not to show at all on fast connections, and adjusts the location and other logic around loading/staleness so that the transitions are much more fluid.

### DIFF
--- a/.changeset/stale-mice-push.md
+++ b/.changeset/stale-mice-push.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+fix explorer field input cursor reset
--- a/.changeset/thirty-laws-hang.md
+++ b/.changeset/thirty-laws-hang.md
@@ -1,0 +1,5 @@
+---
+'hive': minor
+---
+
+add search to app deployment version by operation name
--- a/.changeset/wicked-rooms-mate.md
+++ b/.changeset/wicked-rooms-mate.md
@@ -1,0 +1,5 @@
+---
+'hive': minor
+---
+
+Refine schema composition settings UI
--- a/deployment/config/public-graphql-api-gateway/gateway.config.ts
+++ b/deployment/config/public-graphql-api-gateway/gateway.config.ts
@@ -1,5 +1,24 @@
 // @ts-expect-error not a dependency
-import { createOtlpHttpExporter, defineConfig } from '@graphql-hive/gateway';
+import { defineConfig } from '@graphql-hive/gateway';
+// @ts-expect-error not a dependency
+import { openTelemetrySetup } from '@graphql-hive/gateway/opentelemetry/setup';
+import { AsyncLocalStorageContextManager } from '@opentelemetry/context-async-hooks';
+import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-http';
+
+openTelemetrySetup({
+  // Mandatory: It depends on the available API in your runtime.
+  // We recommend AsyncLocalStorage based manager when possible.
+  // `@opentelemetry/context-zone` is also available for other runtimes.
+  // Pass `false` to disable context manager usage.
+  contextManager: new AsyncLocalStorageContextManager(),
+
+  traces: {
+    // Define your exporter, most of the time the OTLP HTTP one. Traces are batched by default.
+    exporter: new OTLPTraceExporter({ url: process.env['OPENTELEMETRY_COLLECTOR_ENDPOINT']! }),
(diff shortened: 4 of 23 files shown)


packages/web/app/src/pages/target-checks-single.tsx
```diff
@@ -281,20 +281,20 @@ function ConditionalBreakingChangesMetadataSection(props: {
         {numberOfTargets <= 3 && (
           <>
             {allTargets.map((target, index) => (
-              <>
+              <Fragment key={target.slug}>
```
jdolle, line 284: fixes a react `key` warning

packages/web/app/src/components/ui/empty-list.tsx
```diff
@@ -100,7 +100,9 @@ export const NoSchemaVersion = ({
     <EmptyList
       title="Hive is waiting for your first schema"
       description="You can publish a schema with Hive CLI and Hive Client"
-      docsUrl="/features/schema-registry#publish-a-schema"
+      docsUrl={
+        recommendedAction === 'publish' ? '/features/schema-registry#publish-a-schema' : undefined
```
jdolle, line 104: opted to hide for checks since there is a separate link to docs for them

packages/web/app/src/pages/target-checks.tsx
```diff
@@ -337,40 +346,46 @@ function ChecksPageContent(props: {
...
-                    onLoadMore={cursor => setPaginationVariables(cursors => [...cursors, cursor])}
+                    onLoadMore={loadMore}
                     key={cursor ?? 'first'}
                     showOnlyChanged={showOnlyChanged}
                     showOnlyFailed={showOnlyFailed}
                   />
                 ))}
               </div>
-            ) : query.fetching || query.stale ? (
-              <Spinner />
```
jdolle, line 348: moved outside this ternary so I could better control its rendering with the other elements.

packages/web/app/src/pages/target-checks.tsx
```diff
@@ -410,3 +425,28 @@ export function TargetChecksPage(props: {
     </>
   );
 }
+
+const useDebouncedLoader = (isLoading: boolean, delay = 500) => {
```
jdolle, line 429: I also looked at using the `use-debounce` package's useDebounce helper but it debounces in both directions... I only want to delay the loading icon showing. Once loaded, it should hide immediately to keep the rendering as quick as possible.

packages/web/app/src/pages/target-checks.tsx
```diff
@@ -410,3 +425,28 @@ export function TargetChecksPage(props: {
...
+      setShowLoadingIcon(false);
+    }
+
+    // Cleanup function to clear the timer on unmount or if isLoading changes
+    return () => {
+      clearTimeout(timerRef.current);
+    };
+  }, [isLoading, delay]);
+
+  return [showLoadingIcon];
```
XiNiHa, line 451: Doesn't seem to have to be a tuple 😅

packages/web/app/src/pages/target-checks.tsx
XiNiHa: Haven't deeply looked into, but looks like this lib offers a variety of options for scheduling 👀 Not sure whether it's worth adding a dep though
https://tanstack.com/pacer/latest/docs/framework/react/reference/index

packages/web/app/src/pages/target-checks.tsx
XiNiHa: Actually the whole ternaries look too bloated; maybe we should consider constructing the JSX parts outside of the return statement then combine 🫠

XiNiHa (approved):
The functional change itself looks good! Left a few nits around the code but maybe we can just merge this regardless 🤷

XiNiHa:
Oops it had auto merge enabled lol

jdolle:
hehe
no worries, I'll adjust in a separate PR. Ty for the feedback

PR #7050, opened 2025-09-25
feat(app): refine schema composition settings UI

XiNiHa:
### Background

Closes #7020

The schema composition settings UI didn't reflect the mutual exclusiveness of native/external/Apollo composition modes. This was both hard to use and understand.

### Description

<details>
  <summary>Screenshots</summary>

  스크린샷 2025-09-25 오후 11 00 05
  스크린샷 2025-09-25 오후 11 00 16
  스크린샷 2025-09-25 오후 11 00 26
</details>

Combines the composition settings as a single card, separated into tabs. Enabling one mode automatically disables others.

The current implementation involves client-side waterfalls: one mutation for updating the native composition status and another for updating the external composition status. In the long run, we could unify these as a single mutation, possibly using `@oneOf` to express the mutual exclusiveness of each mode.

### Checklist

- [ ] Input validation
- [x] Output encoding
- [ ] Authentication management
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [x] Error handling and logging
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

PR #7069, opened 2025-10-01
feat(app): unify schema composition settings mutations

XiNiHa:
### Background

Follow-up of #7050

### Description

This eliminates the client-side waterfall introduced in #7050, by unifying the process of updating native composition and external composition into a single mutation.
This also allowed refactoring the frontend code to be more top-down.

### Checklist

- [x] Input validation
- [x] Output encoding
- [ ] Session management
- [x] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/schema/providers/schema-manager.ts
```diff
@@ -661,14 +598,61 @@ export class SchemaManager {
     });
 
     if (project.type !== ProjectType.FEDERATION) {
-      throw new HiveError(`Native schema composition is supported only by Federation projects`);
+      throw new HiveError(`Schema composition is supported only by Federation projects`);
```
jdolle: I think we had this wrong before.
This should also be a typed error rather than a graphql error.

packages/services/storage/src/index.ts
```diff
@@ -1482,7 +1482,10 @@ export async function createStorage(
         await pool.one<projects>(sql`/* updateNativeSchemaComposition */
           UPDATE projects
           SET
-            native_federation = ${enabled}
+            native_federation = ${enabled},
+            external_composition_enabled = FALSE,
+            external_composition_endpoint = NULL,
+            external_composition_secret = NULL
```
jdolle: As a convenience to users, we _could_ opt to keep the endpoint and secret so that we can prepopulate if composition is toggled back to external. I could see someone accidentally switching to another composition type and losing their secret.

packages/web/app/src/components/project/settings/legacy-composition.tsx
```diff
@@ -52,71 +36,54 @@ export function LegacyCompositionSettings(props: {
...
-          return toast({
+      } else {
+        const updateResult = useFragment(LegacyCompositionSettings_UpdateResultFragment, result);
+        if (updateResult.ok) {
+          toast({
+            title: 'Successfully enabled legacy composition',
+            description: `Your project is no longer using ${
+              previousCompositionMode === 'external'
+                ? 'external schema composition.'
+                : 'our Open Source composition library for GraphQL Federation.'
```
jdolle:
```suggestion
                : 'native schema composition.'
```
I'd prefer we're consistent about the name of things. We have an explanation already for the "Native Composition" tab, so the extra details of this message shouldn't be necessary here.

n1ru4l: I agree with this!

packages/services/api/src/modules/schema/providers/schema-manager.ts
n1ru4l: Yeah we can return this as a typed error instead.

packages/services/storage/src/index.ts
n1ru4l: I also think it would be convenient to keep these.

n1ru4l (approved):
Looking good! After addressing the feedback from @jdolle, feel free to merge this!

dotansimha (approved):

XiNiHa:
@jdolle I'll wait for your approval to make sure things are done correctly 😋

PR #7070, opened 2025-10-01
Kube log labels

jdolle:
### Background

A recent change to our kubernetes setup caused logs to lose labels we used

### Description

Adds back the labels that were removed


deployment/utils/observability.ts
```diff
@@ -95,6 +95,8 @@ export class Observability {
                 namespace: '{{`{{ kubernetes.pod_namespace }}`}}',
                 pod_name: '{{`{{ kubernetes.pod_name }}`}}',
                 node: '{{`{{ kubernetes.pod_node_name }}`}}',
+                container_name: '{{`{{ kubernetes.container_name }}`}}',
+                env: this.environment.envName,
```
n1ru4l:
```suggestion
```

n1ru4l:
I removed the changeset and the env line, as they are not really relevant to self-hosters, since this is for our own infrastructure.

PR #7071, opened 2025-10-01
fix: error logs missing error text

jdolle:
### Background

Some of our error logs are not printing the error objects.

### Description

This change adjust a number of our logs to add error text

### DIFF
--- a/.changeset/tangy-ravens-love.md
+++ b/.changeset/tangy-ravens-love.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+fix error printing in logs


packages/services/api/src/modules/alerts/providers/adapters/msteams.ts
```diff
@@ -72,7 +72,13 @@ export class TeamsCommunicationAdapter implements CommunicationAdapter {
         }),
       );
     } catch (error) {
-      this.logger.error(`Failed to send Microsoft Teams notification`, error);
+      const errorText =
+        error instanceof Error
+          ? error.toString()
+          : typeof error === 'string'
+            ? error
+            : JSON.stringify(error);
```
jdolle, line 80: honestly im not 100% sure this conversion logic is necessary.
I couldn't figure out well our logger handles converting unknown types to strings.

PR #7073, opened 2025-10-03
fix: reorder render to fix project error --  Rendered fewer hooks tha…

jdolle:
…n expected

### Background

Recently adjusted project and org pages which added more hooks.
We've seen a new error since: http://the-guild-z4.sentry.io/issues/6918059571/?alert_rule_id=5155401&alert_type=issue&notification_uuid=072e0ad1-27b7-4fd3-b98b-7484df1ca8f7&project=5691450&referrer=slack

### Description

This PR puts the error render block after the hooks, so that the hooks run every time.

### DIFF
--- a/packages/web/app/src/pages/organization.tsx
+++ b/packages/web/app/src/pages/organization.tsx
@@ -348,10 +348,6 @@ function OrganizationPageContent(
       });
   }, [projectsConnection, props.search, sortKey, sortOrder]);
 
-  if (query.error) {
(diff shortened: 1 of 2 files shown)


jdolle:
didnt add a changeset bc this issue doesn't matter much. It's an edge case for when the page errors out which would make the page non-functioning regardless of the error log from react.

n1ru4l (approved):
Legit 🥳

PR #7074, opened 2025-10-06
fix: oom for deprecated schema explorer

n1ru4l:
### Background

Some big GraphQL schemas exhaust resources via the deprecated schema explorer GraphQL types and fields.

### Description

Add memoization for the type stats lookup, to prevent allocating a lot of objects.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/operations/providers/operations-manager.ts
+++ b/packages/services/api/src/modules/operations/providers/operations-manager.ts
@@ -907,15 +907,6 @@ export class OperationsManager {
       typename: string;
     } & TargetSelector,
   ) {
-    await this.session.assertPerformAction({
-      action: 'project:describe',
-      organizationId: args.organizationId,
-      params: {
-        organizationId: args.organizationId,
-        projectId: args.projectId,
-      },
-    });
-
     const loader = this.getClientNamesPerCoordinateOfTypeLoader({
       target: args.targetId,
       period: args.period,
--- a/packages/services/api/src/modules/operations/providers/operations-manager.ts
+++ b/packages/services/api/src/modules/operations/providers/operations-manager.ts
@@ -1118,6 +1118,14 @@ export class OperationsManager {
   }: {
     period: DateRange;
   } & TargetSelector) {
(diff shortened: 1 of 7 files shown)


packages/services/api/src/modules/operations/providers/operations-manager.ts
```diff
@@ -907,15 +907,6 @@ export class OperationsManager {
...
     } & TargetSelector,
   ) {
-    await this.session.assertPerformAction({
-      action: 'project:describe',
-      organizationId: args.organizationId,
-      params: {
-        organizationId: args.organizationId,
-        projectId: args.projectId,
-      },
-    });
```
n1ru4l, line 917: Not really needed as the context around already asserts this permission.

packages/services/api/src/modules/schema/resolvers/DeprecatedSchemaExplorer.ts
```diff
@@ -6,21 +7,25 @@ export const DeprecatedSchemaExplorer: DeprecatedSchemaExplorerResolvers = {
...
-        operationsManager,
-        typename,
-      });
-    }
+        return withUsedByClients(stats, {
+          selector: usage,
+          period: usage.period,
+          operationsManager,
+          typename,
+        });
```
n1ru4l, line 24: `withUsedByClients` is hella expensive and was called multiple per times for each typename. By only calling it once we can skip a lot of iterations (for each type usage within the schema for the deprecated schema view).

packages/services/api/src/modules/operations/providers/operations-manager.ts
jdolle: it looks like `getClientNamesPerCoordinateOfType` is calling this also.

Issue #7055, opened 2025-09-26
body/html has different shades of black


n1ru4l:
It should only be one (the one that appears first)

<img width="811" height="323" alt="Image" src="https://github.com/user-attachments/assets/9fe12d97-352d-4fad-9ea5-eb0cad8af326" />

<img width="2202" height="1068" alt="Image" src="https://github.com/user-attachments/assets/ab060f4c-83bc-481b-8fe3-f1de36206643" />

PR #7079, opened 2025-10-07
fix(app): background color

n1ru4l:
### Description

Closes https://github.com/graphql-hive/console/issues/7055

I couldn't really find a better way to fix this...

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/web/app/src/index.css
+++ b/packages/web/app/src/index.css
@@ -204,6 +204,10 @@
   @apply absolute right-2 size-1.5 rounded-full border border-orange-600 bg-orange-400 content-[''];
 }
 
+#root {
+  background-color: rgb(3 7 17 / var(--tw-bg-opacity, 1)) !important;
+}
+
 #root > * {
   background-color: rgb(3 7 17 / var(--tw-bg-opacity, 1)) !important;
 }


packages/web/app/src/index.css
```diff
@@ -203,3 +203,7 @@
 .hive-badge-is-changed:after {
   @apply absolute right-2 size-1.5 rounded-full border border-orange-600 bg-orange-400 content-[''];
 }
+
+#root > * {
+  background-color: rgb(3 7 17 / var(--tw-bg-opacity, 1)) !important;
+}
```
n1ru4l, line 213: The background color is a global style that is applied to body, but somehow does not cascade to the react app root containers children...

packages/web/app/src/components/not-found.tsx
```diff
@@ -1,5 +1,4 @@
 import ghost from '../../public/images/figures/ghost.svg?url';
-import { Helmet } from 'react-helmet-async';
```
jdolle, line 2: Why do we no longer want the helmet functionality?

n1ru4l: I couldn't see a difference with and without it, so I removed it 😄

Issue #7078, opened 2025-10-07
schema explorer: batch client usage lookup


n1ru4l:
Visiting the (deprecated) schema explorer triggers a huge amount of Clickhouse queries for looking up the top clients that use a schema coordinate

See

https://github.com/graphql-hive/console/blob/05dfe0839be50495c7f7cb78367a0ecdfb21931a/packages/services/api/src/modules/operations/providers/operations-manager.ts#L883-L891

and

https://github.com/graphql-hive/console/blob/05dfe0839be50495c7f7cb78367a0ecdfb21931a/packages/services/api/src/modules/operations/providers/operations-reader.ts#L1438

We should look into batching this into a single ClickHouse query.

PR #7090, opened 2025-10-09
fix: batch client name coordinate lookups

n1ru4l:
### Background

The deprecated schema explorer is slow and uses a lot of memory resources on the api.

### Description

Closes https://github.com/graphql-hive/console/issues/7078

Instead of looking up schema coordinate client usages via the typename matching, we can speed up things a lot by looking up exact schema coordinates. The page now almost loads instantly for big schemas compared to previously, where it would allocate huge amounts of memory.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/.changeset/free-worms-melt.md
+++ b/.changeset/free-worms-melt.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+Further improve the performance of loading the deprecated schema view for large schemas.


packages/services/api/src/modules/operations/providers/operations-reader.ts
```diff
@@ -1483,13 +1472,9 @@ export class OperationsReader {
...
-              target: args.targetId,
-              period: args.period,
-              extra: [
-                sql`( co.coordinate = ${args.typename} OR co.coordinate LIKE ${
-                  args.typename + '.%'
-                } )`,
-              ],
+              target: targetId,
+              period: period,
+              extra: [sql`co.coordinate IN (${sql.array(coordinates, 'String')})`],
```
n1ru4l, line 1477: Instead of looking up coordinates per typename, just looking up exact coordinates seems a lot more efficient and makes the deprecated schema view load much faster for bigger schemas.

Issue #6851, opened 2025-06-20
[Hive Explorer] incorrectly displays subgraph owner of a field when a type is used with @external


rickbijkerk:
we have a setup similar to this where there's a subgraph A which defines the **Product** and **ProductWeight** type

```
#Subgraph A
type Product @key(fields: "id") {
    id: ID!
    weight: ProductWeight!
}

type ProductWeight  {
    weight: Int!
    // .. 
}
```

and a subgraph B which uses the **ProductWeight** data to enrich the **Product** type 

```
#Subgraph B
type Product @key (fields: "id"){
    id: ID!
    weight: ProductWeight! @external
    shippingCosts: Int! @requires(fields: "weight { id }")
}


type ProductWeight @external {
    id: ID!
}
```
**Problem**:
based on the schema above i would expect hive to display Subgraph A to be the owner of the _weight_ field on **Product** and all of **ProductWeight** fields, however in reality Hive displays Subgraph B to be the owner of the _weight_ field and of the id field on the **ProductWeight** type

**Expected outcome**:
Hive should exclude subgraphs that have a field marked as @external when looking for the owner.

(Note:  currently hive does correctly mark both subgraph A and B as owner of the **ProductWeight** Type which i think makes sense as it makes it easy to find usages of certain types)

If you need more info let me know!

n1ru4l:
I just tried to reproduce it based on your example using the following schemas:

**Subgraph A**
```graphql
extend schema @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@key"])

type Product @key(fields: "id") {
  id: ID!
  weight: ProductWeight!
}

type ProductWeight {
  id: ID!
  weight: Int!
}

type Query {
  a: Product!
  g: ProductWeight!
}

```


**Subgraph B**
```graphql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@key", "@external", "@requires"])

type Product @key(fields: "id") {
  id: ID!
  weight: ProductWeight! @external
  shippingCosts: Int! @requires(fields: "weight { id }")
}

type ProductWeight @external {
  id: ID!
}

type Query {
  c: Product!
}
```


As you can see in this explorer screenshot, this is not really showing the outcome you described. Can you share an example that showcases the described issue?


<img width="1192" height="347" alt="Image" src="https://github.com/user-attachments/assets/32755a5d-bd80-4324-8d39-1279e81e67ed" />

n1ru4l:
[...]

PR #7097, opened 2025-10-10
test: ensure correct ownership of external field

n1ru4l:
### Background

Trying to reproduce https://github.com/graphql-hive/console/issues/6851#issuecomment-3389791517

### DIFF
--- a/packages/services/api/src/modules/schema/resolvers/SchemaCoordinateStats.ts
+++ b/packages/services/api/src/modules/schema/resolvers/SchemaCoordinateStats.ts
@@ -18,7 +18,6 @@ export const SchemaCoordinateStats: Pick<SchemaCoordinateStatsResolvers, 'superg
     // @note: `SchemaVersionHelper.getSupergraphAst` is cached
     const supergraphAst = await injector.get(SchemaVersionHelper).getSupergraphAst(latestVersion);
     const supergraph = supergraphAst ? extractSuperGraphInformation(supergraphAst) : null;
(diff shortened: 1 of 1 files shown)


packages/services/api/src/modules/schema/lib/__tests__/federation-super-graph.spec.ts
```diff
@@ -480,3 +481,65 @@ describe('extractSuperGraphInformation', () => {
...
+    {
+      name: 'b',
+      typeDefs: b,
+    },
+  ]);
+
+  const supergraphInfo = extractSuperGraphInformation(parse(result.supergraphSdl!));
+  expect(supergraphInfo.schemaCoordinateServicesMappings.get('Product.weight')).toEqual(['a']);
+  expect(supergraphInfo.schemaCoordinateServicesMappings.get('ProductWeight.id')).toEqual(['a']);
+  expect(supergraphInfo.schemaCoordinateServicesMappings.get('ProductWeight')).toEqual(['a', 'b']);
```
XiNiHa, line 541: Do we expect to also have `b` included here?

n1ru4l: As per https://github.com/graphql-hive/console/issues/6851#issue-3163316649 I guess it might make sense?

> (Note: currently hive does correctly mark both subgraph A and B as owner of the ProductWeight Type which i think makes sense as it makes it easy to find usages of certain types)

😁

PR #7099, opened 2025-10-13
fix: improve query performance

n1ru4l:
### Background

Removes unnecessary join condition and redundant method only used in tests.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/integration-tests/tests/api/schema/delete.spec.ts
+++ b/integration-tests/tests/api/schema/delete.spec.ts
@@ -183,7 +183,7 @@ test.concurrent(
           ping: String
         }
       `);
-      expect(latestVersion!.schemaCompositionErrors).toEqual(null);
+      expect(latestVersion.schemaCompositionErrors).toEqual(null);
       expect(latestVersion.hasPersistedSchemaChanges).toEqual(true);
       expect(latestVersion.isComposable).toEqual(true);
 


packages/services/storage/src/index.ts
```diff
@@ -2034,31 +2034,17 @@ export async function createStorage(
...
-    },
-
     async getMaybeLatestVersion(args) {
       const version = await pool.maybeOne<unknown>(
         sql`/* getMaybeLatestVersion */
           SELECT
             ${schemaVersionSQLFields(sql`sv.`)}
-          FROM schema_versions as sv
-          LEFT JOIN targets as t ON (t.id = sv.target_id)
-          WHERE sv.target_id = ${args.targetId} AND t.project_id = ${args.projectId}
```
n1ru4l, line 2049: It is safe to remove this join, we do not need to make this assertion here, as any code that calls this method already asserts whether the project belongs to the target.

integration-tests/tests/api/schema/delete.spec.ts
```diff
@@ -171,18 +171,19 @@ test.concurrent(
...
       });
+      assertNonNull(latestVersion);
 
       expect(latestVersion.compositeSchemaSDL).toMatchInlineSnapshot(`
         type Query {
           ping: String
         }
       `);
-      expect(latestVersion.schemaCompositionErrors).toEqual(null);
+      expect(latestVersion!.schemaCompositionErrors).toEqual(null);
```
XiNiHa: Why do we need `!` here? Looks like the others are without `!`

n1ru4l: good catch, i first did it like this, then I decided to do it the clean way :D

n1ru4l:
```suggestion
      expect(latestVersion.schemaCompositionErrors).toEqual(null);
```

PR #7101, opened 2025-10-13
fix: make query work on clickhouse cloud

n1ru4l:
### Background

```
Duplicate alias in ARRAY JOIN: value: While processing SELECT 'errorCode' AS key, toString(value) AS value, count() AS count FROM otel_traces_normalized ARRAY JOIN graphql_error_codes AS value WHERE target_id = ???
```

The solution is to simply give the `ARRAY JOIN` column different name.

The annoying thing is that this issue does not occur with the local clickhouse docker container. 😢 

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/operations/providers/traces.ts
+++ b/packages/services/api/src/modules/operations/providers/traces.ts
@@ -343,7 +343,7 @@ export class TraceBreakdownLoader {
       const statements: SqlValue[] = [];
       const arrJoinColumnAlias = 'arr_join_column_value';
 
-      const (let { key, columnExpression, limit, arrayJoinColumn } of inputs) {
+      for (const { key, columnExpression, limit, arrayJoinColumn } of inputs) {
         statements.push(sql`
           SELECT
             '${sql.raw(key)}' AS "key"


packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -325,39 +325,54 @@ export class TraceBreakdownLoader {
...
+    ),
     { value: string; count: number }[],
     string
   >(
     async inputs => {
       const statements: SqlValue[] = [];
+      const arrJoinColumnAlias = 'arr_join_column_value';
 
-      for (const { key, columnExpression, limit, arrayJoinColumn } of inputs) {
+      for (let { key, columnExpression, limit, arrayJoinColumn } of inputs) {
```
n1ru4l:
```suggestion
      const (let { key, columnExpression, limit, arrayJoinColumn } of inputs) {
```

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -325,39 +325,54 @@ export class TraceBreakdownLoader {
...
+            "count" DESC
+          ${sql.raw(limit ? `LIMIT ${limit}` : '')}
+        `);
       }
 
+      const query = sql`
+        ${sql.join(statements, ' UNION ALL ')}
+      `;
+
+      console.log(query.sql);
```
n1ru4l:
```suggestion
```

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -325,39 +325,52 @@ export class TraceBreakdownLoader {
...
-        GROUP BY value
-        ORDER BY count DESC
-        ${sql.raw(limit ? `LIMIT ${limit}` : '')}
-      `);
+          SELECT
+            '${sql.raw(key)}' AS "key"
+            , toString(${sql.raw(columnExpression ?? arrJoinColumnAlias)}) AS "value"
+            , count(*) AS "count"
+          FROM "otel_traces_normalized"
+            ${sql.raw(arrayJoinColumn ? `ARRAY JOIN ${arrayJoinColumn} AS "${arrJoinColumnAlias}"` : '')}
```
n1ru4l, line 353: Basically the alias here conflicting with the alias in Line 350 would cause the exception. By simply renaming the alias to something different we fix this.

The diff is so big because of reformatting the sql query 😢

PR #7116, opened 2025-10-14
Use newer PNPM and link to newest GraphQLConf

hasparus:
### Description

Updated Components to have the new conference card (now it accepts props, so we don't have to update it for this reason anymore).

I also let PNPM bump itself because it kept pestering me. LMK if you want me to tidy it up into separate PRs.


package.json
```diff
@@ -12,7 +12,7 @@
   },
   "license": "MIT",
   "private": true,
-  "packageManager": "pnpm@10.16.1",
+  "packageManager": "pnpm@10.18.3+sha512.bbd16e6d7286fd7e01f6b3c0b3c932cda2965c06a908328f74663f10a9aea51f1129eea615134bf992831b009eabe167ecb7008b597f40ff9bc75946aadfb08d",
```
jdolle, line 15: If you use corepack it should automatically use what's defined in here so you dont need to worry much about it.

jdolle (approved):

Issue #7117, opened 2025-10-14
Hive CLI schema:fetch actionId is not exposed anywhere


jdolle:
We should either expose the actionId via the UI, or allow fetching by another identifier such as the contextId or uuid.

PR #7124, opened 2025-10-15
select latest log in version by commit; add version details to history

jdolle:
### Background

Related to https://github.com/graphql-hive/console/issues/7117

Using Hive CLI to fetch a schema by action ID can result in unpredictable behavior because the multiple logs may share the same commit.

### Description

This change makes sure to take the latest log, and clarifies some of the fields on the history page by adding a header.
Screenshot 2025-10-15 at 4 22 34 PM

### DIFF
--- a/integration-tests/tests/api/schema/schema-version.spec.ts
+++ b/integration-tests/tests/api/schema/schema-version.spec.ts
@@ -5,8 +5,8 @@ import { execute } from '../../../testkit/graphql';
 import { initSeed } from '../../../testkit/seed';
 
 const SchemaByActionIdQuery = graphql(/* GraphQL */ `
-  query SchemaByActionIdQuery($actionId: ID!) {
-    schemaVersionForActionId(actionId: $actionId) {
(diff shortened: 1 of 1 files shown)


n1ru4l (commented):
Let's add a small integration test to confirm this behaviour and prevent regressions?

kamilkisiela:
Let's maybe set a maximum with on the commit div, truncate with `...` and add a copy button (simple icon that is wrapped with an invisible button and displayed on hover?)

https://github.com/user-attachments/assets/155f05ae-ed83-443e-a699-b50355f38629

https://v0.app/chat/dA3ddnhqCOj

jdolle:
Added an integration test, copy button, and made the content dynamic/truncated. Also applied these changes to the checks page.

At a small enough width, there's still some resizing issues, but this is because of the navigation bar, and the multiple tabs on the checks + history page for schema, supergraph, etc.
These fall outside the current task and can be addressed if we opt to make all pages even more dynamic.
Screenshot 2025-10-20 at 2 18 58 PM

PR #7128, opened 2025-10-16
Member invites clarification, ui improvements, and invite link toast

jdolle:
### Background

I'm working on some other invite related issues and when creating and accepting invites locally, it took me a second to understand what was happening...

### Description

1. Clarifies that invites are intended for non-OIDC accounts. This was not very clear initially since it may all be tied to your email.
2. When going to the join URL (invitation link), if the user was not logged in, they would be redirected to signin without any additional information... So I added a toast to notify them that they need to log in before accepting the invite. I think we can still do a lot to improve this flow, possibly with a custom flow, but this met the minimum requirement.
3. Adjusts sizes of things at different resolutions to improve ux.

**Before**
Screenshot 2025-10-16 at 3 43 32 PM

**After**
Screenshot 2025-10-16 at 3 29 11 PM
Screenshot 2025-10-16 at 3 27 23 PM

### DIFF
--- a/packages/web/app/src/components/organization/members/invitations.tsx
+++ b/packages/web/app/src/components/organization/members/invitations.tsx
@@ -398,8 +398,12 @@ function Invitation(props: {
         ) : null}
       </AlertDialog>
       <tr>
-        <td className="py-3 text-sm font-medium">{invitation.email}</td>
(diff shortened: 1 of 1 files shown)


n1ru4l (approved):
This is great ❤️

PR #7131, opened 2025-10-17
Add supergraph generation to federation demo on workers

kamilkisiela:
Adds a script to generate the supergraph schema and configures the wrangler to deploy it under /supergraph.graphql

### DIFF
--- a/packages/services/demo/federation/src/generate-supergraph.ts
+++ b/packages/services/demo/federation/src/generate-supergraph.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
 // @ts-nocheck
 // Reason: this script is executed directly via `node --experimental-strip-types`
 // which allows running `.ts` files without transpilation. TypeScript doesn’t


packages/services/demo/federation/tsconfig.json
```diff
@@ -5,5 +5,5 @@
     "module": "esnext",
     "rootDir": "../../.."
   },
-  "include": ["./src/**/*.ts"]
+  "include": ["./src/main.ts"]
```
kamilkisiela:
```suggestion
  "include": ["./src/**/*.ts"]
```

packages/services/demo/federation/tsconfig.json
kamilkisiela, line 4:
```suggestion
    "target": "ES2020",
    "allowImportingTsExtensions": true,
```

packages/services/demo/federation/src/generate-supergraph.ts
```diff
@@ -0,0 +1,35 @@
+import { writeFileSync } from 'node:fs';
```
kamilkisiela, line 8:
```suggestion
// @ts-nocheck
// Reason: this script is executed directly via `node --experimental-strip-types`
// which allows running `.ts` files without transpilation. TypeScript doesn’t
// allow `.ts` import specifiers unless `allowImportingTsExtensions` is enabled,
// which we don’t want globally. Disabling type checking here avoids TS5097
// while keeping the rest of the project strict.
import { writeFileSync } from 'node:fs';
```

packages/services/demo/federation/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "compilerOptions": {
     "target": "ES2020",
     "module": "esnext",
+    "allowImportingTsExtensions": true,
```
kamilkisiela:
```suggestion
```

packages/services/demo/federation/public/supergraph.graphql
n1ru4l, line 1: did u want to commit this badboy

kamilkisiela: yes, but it can mess the prettier check, up to you if to ignore file or not

PR #7135, opened 2025-10-21
OIDC default resource assignments

jdolle:
### Background

OIDC can have member roles assigned by default, but those roles have access to every resource.

### Description

This PR adds a resource assignment configuration to the OIDC settings, and adds an API to support that.

### DIFF
--- a/integration-tests/tests/api/oidc-integrations/assigned-resources.spec.ts
+++ b/integration-tests/tests/api/oidc-integrations/assigned-resources.spec.ts
@@ -1,0 +1,266 @@
+import { graphql } from 'testkit/gql';
+import { ResourceAssignmentModeType } from 'testkit/gql/graphql';
+import { execute } from 'testkit/graphql';
+import { initSeed } from 'testkit/seed';
+
+const AssignedResourcesSpec_CreateOIDCIntegrationMutation = graphql(`
+  mutation AssignedResourcesSpec_CreateOIDCIntegrationMutation(
+    $input: CreateOIDCIntegrationInput!
+  ) {
+    createOIDCIntegration(input: $input) {
+      ok {
+        createdOIDCIntegration {
+          id
+          defaultResourceAssignment {
+            mode
+          }
+        }
+      }
+    }
+  }
+`);
+
+const AssignedResourcesSpec_ReadDefaultTest = graphql(`
+  query AssignedResourcesSpec_ReadDefaultTest($organizationSlug: String!) {
+    organization(reference: { bySelector: { organizationSlug: $organizationSlug } }) {
+      id
+      oidcIntegration {
+        defaultResourceAssignment {
+          mode
+          projects {
+            project {
+              id
+              slug
+            }
+            targets {
+              mode
+              targets {
+                target {
+                  id
+                  slug
+                }
+                services {
+                  mode
+                  services
+                }
+                appDeployments {
+                  mode
+                  appDeployments
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`);
+
+const AssignedResourcesSpec_UpdateDefaultMutation = graphql(`
+  mutation AssignedResourcesSpec_UpdateDefaultMutation(
+    $input: UpdateOIDCDefaultResourceAssignmentInput!
+  ) {
+    updateOIDCDefaultResourceAssignment(input: $input) {
+      ok {
+        updatedOIDCIntegration {
+          id
+          defaultResourceAssignment {
+            mode
+            projects {
+              project {
+                id
+                slug
+              }
+              targets {
+                mode
+                targets {
+                  target {
+                    id
+                    slug
+                  }
+                  services {
+                    mode
+                    services
+                  }
+                  appDeployments {
+                    mode
+                    appDeployments
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+      error {
+        message
+      }
+    }
+  }
+`);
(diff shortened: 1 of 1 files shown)


packages/web/app/src/components/organization/members/resource-selector.tsx
```diff
@@ -152,7 +152,8 @@ export function ResourceSelector(props: {
   // whether we show the service or apps in the last tab
   const [serviceAppsState, setServiceAppsState] = useState(ServicesAppsState.service);
 
-  const toggleServiceAppsState = () => {
+  const toggleServiceAppsState = (e: MouseEvent<HTMLButtonElement>) => {
```
jdolle, line 155: This helps in case this is within a `<form>` element -- which I experimented with briefly.

packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
```diff
@@ -702,6 +741,118 @@ function OIDCDefaultRoleSelector(props: {
...
+          resources: resourceSlectionToGraphQLSchemaResourceAssignmentInput(resources),
+        },
+      });
+    },
+    [debouncedMutate, setSelection, props.oidcIntegrationId],
+  );
+
+  return (
+    <div className="relative">
+      {(debouncedMutate.isPending() || isMutating) && (
```
jdolle, line 844: `debouncedMutate.isPending()` is true if the debounce is waiting.
This way the user gets feedback more quickly that something is happening, even though it doesn't represent the exact response times.

packages/services/api/src/modules/oidc-integrations/resolvers/OIDCIntegration.ts
```diff
@@ -40,4 +41,12 @@ export const OIDCIntegration: OidcIntegrationResolvers = {
...
     return role;
   },
+  defaultResourceAssignment: async (oidcIntegration, _, { injector }) => {
+    if (oidcIntegration.defaultResourceAssignment) {
+      return await injector.get(ResourceAssignments).resolveGraphQLMemberResourceAssignment({
+        organizationId: oidcIntegration.linkedOrganizationId,
+        resources: oidcIntegration.defaultResourceAssignment,
+      });
+    }
+  },
```
n1ru4l, line 53: nit: this could be an early return `null`

packages/services/storage/src/db/types.ts
```diff
@@ -11,6 +11,7 @@ export type alert_channel_type = 'MSTEAMS_WEBHOOK' | 'SLACK' | 'WEBHOOK';
 export type alert_type = 'SCHEMA_CHANGE_NOTIFICATIONS';
 export type breaking_change_formula = 'PERCENTAGE' | 'REQUEST_COUNT';
 export type schema_policy_resource = 'ORGANIZATION' | 'PROJECT';
+export type schema_proposal_stage = 'APPROVED' | 'CLOSED' | 'DRAFT' | 'IMPLEMENTED' | 'OPEN';
```
n1ru4l: 🤔  I think this leaked 😄

packages/services/storage/src/index.ts
```diff
@@ -519,6 +519,11 @@ export async function createStorage(
                     WHERE organization_id = ${linkedOrganizationId} AND name = 'Viewer')
                 )
               ),
+              (
+                SELECT default_assigned_resources FROM oidc_integrations
+                WHERE id = ${args.oidcIntegrationId})
+                )
+              ),
```
n1ru4l, line 525: Looks like too many `)` 🤔

packages/services/storage/src/db/types.ts
jdolle: 😛 Yeah...

packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
n1ru4l: Minor note here, even though I don't think its that critical, is that we are mixing "auto-save" and "click-to-save" elements here. Maybe sticking to click to save for everything might be less confusing - but not 100% sure

packages/web/app/src/lib/urql.ts
```diff
@@ -82,6 +82,13 @@ export const urqlClient = createClient({
         FilterStringOption: noKey,
         FilterBooleanOption: noKey,
         TracesFilterOptions: noKey,
+        ResourceAssignment: noKey,
+        TargetServicesResourceAssignment: noKey,
+        TargetAppDeploymentsResourceAssignment: noKey,
+        TargetResouceAssignment: noKey,
+        ProjectTargetsResourceAssignment: noKey,
+        ProjectResourceAssignment: noKey,
+        BillingConfiguration: noKey,
```
dotansimha, line 91: I assume you added `BillingConfiguration` because it emitted a warning as well? just confirming

packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
jdolle: I completely agree, but I think then we'd need to do the same for the existing role selector inputs, and also potentially modify the API to avoid making 3 mutations all together. I think this is better done in a future revamp of this flow. Currently, I think it's way too much information in a single modal window.

packages/web/app/src/lib/urql.ts
jdolle: yeah

Issue #6699, opened 2025-04-03
Publish initial contract with composition error shows wrong UI state on dashboard


n1ru4l:
When you attempt to publish an invalid contract for the first time, it shows both the `First composable version` component and `Composition Errors` component. The `First composable version` should not be shown in that scenario.

<img width="1088" alt="Image" src="https://github.com/user-attachments/assets/c786919b-f3ee-4857-a0dd-87e94da38061" />

PR #7140, opened 2025-10-21
Fix: Schema contract intial schema composition error to not show Firs…

mskorokhodov:
…t composable schema state

### Background

Fix for #6699 

### Description

ContractVersion.isFirstComposableVersion resolver fix to return proper value


packages/services/api/src/modules/schema/providers/contracts-manager.ts
```diff
@@ -217,7 +217,7 @@ export class ContractsManager {
   public async getIsFirstComposableVersionForContractVersion(contractVersion: ContractVersion) {
     const diffableContractVersion =
       await this.getDiffableContractVersionForContractVersion(contractVersion);
-    return !diffableContractVersion;
+    return !!diffableContractVersion;
```
n1ru4l: I think this must return `true` if there is no previously diffable (sucessful) contract version and this version has no errors.

```suggestion
    return !diffableContractVersion && args.contractVersion.schemaCompositionErrors === null
```

Issue #7091, opened 2025-10-09
loading state for default/deprecated/unused schema explorer


n1ru4l:
The deprecated schema explorer can take some time to load. We shoud at least show a loading spinner or skeleton. In the best case scenario we can also defer the expensive part of the graphql query to show UI early and make the full experience more pleasant.

PR #7141, opened 2025-10-21
Add: skeletons for schema explorer

mskorokhodov:
### Background

#7091 

### Description

Add skeleton for loading/stale states of schema explorer


n1ru4l (approved):
Really nice!

image

PR #7143, opened 2025-10-21
New shared cargo package: `hive-console-sdk`

ardatan:
- Create a new package to be shared by Apollo Router fork and Hive Router
- Avoid Mutex as much as we can. And avoid having Mutexes in multiple places nestedly. I removed all other mutexes except one for the queue.
- Use thread-safe moka instead of the combination of `lru` and Mutex
- Single agent to support reports from multiple schemas on the queue to handle the cases while the schema is being updated by the router.
- Fix the default "enabled" flag for the persisted operations plugin. It is now "false" previously "true" causing arbitrary operations fail by default

Related https://github.com/graphql-hive/router/pull/499


kamilkisiela:
> Use thread-safe moka instead of the combination of lru and Mutex
Moka can also deadlock if used incorrectly, just so you know. DashMap too.

---

Can you rename it to `hive-console-sdk`? @n1ru4l wdyt?

ardatan:
Renamed to SDK @kamilkisiela

.changeset/funny-rockets-speak.md
```diff
@@ -0,0 +1,7 @@
+---
+'hive-console-sdk-rs': patch
+'hive-apollo-router-plugin': patch
+---
+
+Extract Hive Console integration implementation into a new package `hive-console-sdk` which can
```
dotansimha, line 6: Please also document here the internal changes to the LRU cache implementation, and also the refactor to reduce the Mutex count.

ardatan: Added 👍

PR #7150, opened 2025-10-22
adjust local clickhouse configuration

jdolle:
### Background

When running clickhouse through docker locally, I frequently encounter out of memory errors. 

### Description

This is an attempt to reduce the memory usage as well as increase the available memory.

Looking at the way we've set up our docker images, I believe this should only impact the dev instance.


n1ru4l (approved):
Lets see how that goes 🔮 , I also had it crashing a lot 😢

PR #7152, opened 2025-10-22
[ignore] advanced breaking changes PR - local branch

dotansimha:
a local branch of https://github.com/graphql-hive/console/pull/6906 so I can get a Docker image for testing

### DIFF
--- a/docker/router.dockerfile
+++ b/docker/router.dockerfile
@@ -2,7 +2,7 @@
 FROM scratch AS pkg
 FROM scratch AS config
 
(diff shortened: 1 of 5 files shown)


docker/router.dockerfile
ardatan, line 37:
```suggestion
FROM debian:trixie-slim AS runtime
```

Latest Rust needs GLIBC_2.39 while bookworm has 2.36. So we need to update to Debian 13 / trixie

PR #7157, opened 2025-10-23
router-docs: added supergraph, probes pages

dotansimha:
- [x] Added a new page for Probes
- [x] Adde a new page explaining the process of loading/reloading of the Supergraph
- [x] Updated `supergraph` api ref page 
- [x] Slight updates to `Getting Started` page


kamilkisiela:
Let's maybe add "Observability and Logging" group or something similar and put probes under that? This way when we add Prom and OTel, we can  put it there too

PR #7160, opened 2025-10-23
OIDC assigned resources

jdolle:
> Previous OIDC PR was auto-merged and reverted. This is reopening further review.

Background
OIDC can have member roles assigned by default, but those roles have access to every resource.

Description
This PR adds a resource assignment configuration to the OIDC settings, and adds an API to support that.

### DIFF
--- a/packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
+++ b/packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
@@ -851,11 +851,14 @@ function OIDCDefaultResourceSelector(props: {
   function MutateState() {
     if (debouncedMutate.isPending() || mutateState === 'loading') {
       return <Spinner className="absolute right-0 top-0" />;
-    } else if (mutateState === 'error') {
+    }
+    if (mutateState === 'error') {
       return <XIcon className="absolute right-0 top-0 text-red-500" />;
-    } else if (mutateState === 'success') {
+    }
+    if (mutateState === 'success') {
       return <CheckIcon className="absolute right-0 top-0 text-emerald-500" />;
     }
+    return null;
   }
 
   return (
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17626,8 +17626,8 @@ snapshots:
     dependencies:
       '@aws-crypto/sha256-browser': 3.0.0
       '@aws-crypto/sha256-js': 3.0.0
-      '@aws-sdk/client-sso-oidc': 3.596.0
-      '@aws-sdk/client-sts': 3.596.0(@aws-sdk/client-sso-oidc@3.596.0)
+      '@aws-sdk/client-sso-oidc': 3.596.0(@aws-sdk/client-sts@3.596.0)
+      '@aws-sdk/client-sts': 3.596.0
       '@aws-sdk/core': 3.592.0
       '@aws-sdk/credential-provider-node': 3.596.0(@aws-sdk/client-sso-oidc@3.596.0)(@aws-sdk/client-sts@3.596.0)
       '@aws-sdk/middleware-host-header': 3.577.0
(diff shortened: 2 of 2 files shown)


packages/web/app/src/lib/urql.ts
```diff
@@ -82,6 +82,13 @@ export const urqlClient = createClient({
         FilterStringOption: noKey,
         FilterBooleanOption: noKey,
         TracesFilterOptions: noKey,
+        ResourceAssignment: noKey,
+        TargetServicesResourceAssignment: noKey,
+        TargetAppDeploymentsResourceAssignment: noKey,
+        TargetResouceAssignment: noKey,
```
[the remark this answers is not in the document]
jdolle: There is a typo, but it's in the name of an existing type and we don't want to change that name.

packages/services/storage/src/index.ts
```diff
@@ -4633,6 +4669,7 @@ const OktaIntegrationBaseModel = zod.object({
   client_secret: zod.string(),
   oidc_user_access_only: zod.boolean(),
   default_role_id: zod.string().nullable(),
+  default_assigned_resources: zod.any().nullable(),
```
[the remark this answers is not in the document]
n1ru4l: Instead of any you can use the zod model here:

https://github.com/graphql-hive/console/blob/3dbb04533e55cb4c8fc277fe597221753bdd59f4/packages/services/api/src/modules/organization/lib/resource-assignment-model.ts#L80

packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
```diff
@@ -1042,15 +1203,30 @@ function UpdateOIDCIntegrationForm(props: {
...
+            </p>
+          </div>
+          <div className="pb-4">
+            <OIDCDefaultResourceSelector
+              oidcIntegrationId={props.oidcIntegration.id}
+              disabled={!props.isAdmin}
+              organization={props.organization}
+              resourceAssignment={props.oidcIntegration.defaultResourceAssignment ?? {}}
+            />
+          </div>
```
n1ru4l, line 1249: `Organization.projects` is filtered based on what the viewer can see. In order to work around this we could do the following:

Add `Organization.projectsForResourceSelector` -> Returns the Projects in a minimal form based on what is needed for the resource selector to work.

```graphql
type ProjectForResourceSelector {
  id: ID!
  slug: String!
  targets: [TargetForResourceSelector!]!
}

type TargetForResourceSelector {
  id: ID!
  slug: String!
  services: [String!]!
  appDeployments: [String!]!
}
```

n1ru4l: If we don't do this, we might have a weird scenario where assigned resources "vanish", after someone without all project access edits this list.

n1ru4l: It could then simply check if the viewer has the permissions for the contexts where it is used e.g.`oidc:modify` and `member:xxxx`

packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
```diff
@@ -702,6 +741,118 @@ function OIDCDefaultRoleSelector(props: {
...
+      });
+    },
+    [debouncedMutate, setSelection, props.oidcIntegrationId],
+  );
+
+  return (
+    <div className="relative">
+      {(debouncedMutate.isPending() || isMutating) && (
+        <Spinner className="absolute right-0 top-0" />
+      )}
```
n1ru4l: Can we somehow show "Saved :checkmark:" after the save has finished? I think that way it is more obvious that an auto-safe has happened.

packages/services/api/src/modules/audit-logs/providers/audit-logs-types.ts
```diff
@@ -1,5 +1,5 @@
 import { z } from 'zod';
-import { ResourceAssignmentModel } from '../../organization/lib/resource-assignment-model';
+import { ResourceAssignmentModel } from '@hive/storage/resource-assignment-model';
```
jdolle, line 2: had to move because now storage uses the model and i didnt want to make services a dependency of storage.

jdolle:
@n1ru4l will be taking over the remainder of this task. Currently we believe this means graying out projects/targets that the user does not have access to, and restricting the ability to add permissions to those resources.
They should still be able to view and remove permissions though.

n1ru4l:
I cherry-picked some commits from here into https://github.com/graphql-hive/console/pull/7252, and will continue there.

PR #7161, opened 2025-10-23
Link to Wealthsimple case study

hasparus:
### Background

image

Kamil raised that case studies are not linked to from the landing page and proposed adding a section similar to testimonials.

The initial design of testimonials section assumed they'll link to case studies.

### Description

As a first step, I added a link to the only case study that overlaps with testimonials.

@saihaj how hard would you expect it to be to get testimonial-like quotes from all the companies we have case studies for? Then we could use the same section, it's already scrollable and has these little dots, so we could feet like ~7 of them without any additional design work.

image

alternatively, @kamilkisiela if you think it deserves an entirely new section, how do you imagine it?

### DIFF
--- a/packages/web/docs/src/components/company-testimonials/index.tsx
+++ b/packages/web/docs/src/components/company-testimonials/index.tsx
@@ -62,6 +62,7 @@ const testimonials: Testimonial[] = [
     picture: {
       img: wealthsimplePicture,
     },
+    caseStudyHref: '/case-studies/wealthsimple',
   },
   {
     company: 'Prodigy',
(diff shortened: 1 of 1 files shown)


packages/web/docs/src/components/company-testimonials/index.tsx
```diff
@@ -206,7 +207,7 @@ export function CompanyTestimonialsSection({ className }: { className?: string }
                           href={caseStudyHref}
                           className="absolute bottom-0 w-full md:w-fit"
                         >
-                          Read Case Study
+                          Read the case study
```
hasparus, line 210: After the initial designs we settled on sentence case.

PR #7162, opened 2025-10-24
improve development environment seeding

jdolle:
### Background

Our seeding script was outdated and didnt produce a very realistic schema. It also seeded usage and schemas in the same script, which really only is useful on the first run since the schemas never change.

### Description

Rewrites to split usage and schema seeding. Creates a better monorepo example and a more flexible file structure so we can continue to add operations or subgraphs as we want to test more cases.


scripts/seed-schemas.ts
```diff
@@ -0,0 +1,148 @@
...
+    loaders: [new GraphQLFileLoader()],
+  });
+  const sdl = printSchema(schema);
+  return publishSchema({
+    sdl,
+    target,
+  });
+}
+
+async function federation() {
```
n1ru4l, line 133: Here we could also use the schemas from here: `packages/services/demo/federation`

jdolle: Added some of the definitions found in the demo to the seed.
We could use the same schema for both but I think unifying the two is a separate task and probably isnt really necessary. The point of these seeds is to setup the conditions we want to manually verify or develop against. I think the demo benefits from being a bit simpler.

Issue #6946, opened 2025-09-07
hive-apollo-router plugin reports `local` as version instead of crate version


dotansimha:
This seems to be depending on some runtime env var: https://github.com/graphql-hive/console/blob/b97cf7774ea8e94fa81dd74db7846e2244849928/packages/libraries/router/src/registry.rs#L27 that 100% of the times is not set at runtime, so it unwrap into `local`.

Instead, we should use the plugin version as it set during compilation.

PR #7173, opened 2025-10-26
fix(apollo-router-fork): use correct apollo-router-plugin version instead of commit

dotansimha:
Closes https://github.com/graphql-hive/console/issues/6946 

`env!` resolved the env var at built time, and Cargo is in charge of setting it during `cargo build`, so the version will be "burned" as static into the compiled service.

### DIFF
--- a/packages/libraries/router/src/agent.rs
+++ b/packages/libraries/router/src/agent.rs
@@ -1,3 +1,5 @@
+use crate::consts::PLUGIN_VERSION;
+
 use super::graphql::OperationProcessor;
 use graphql_parser::schema::{parse_schema, Document};
 use reqwest::Client;
@@ -9,8 +11,6 @@ use std::{
 };
 use thiserror::Error;
 use tokio::sync::Mutex as AsyncMutex;
-
-static COMMIT: Option<&'static str> = option_env!("GITHUB_SHA");
 
 #[derive(Serialize, Debug)]
 pub struct Report {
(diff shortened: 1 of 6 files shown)


packages/libraries/router/src/agent.rs
```diff
@@ -299,7 +299,7 @@ impl UsageAgent {
                 )
                 .header(
                     reqwest::header::USER_AGENT,
-                    format!("hive-apollo-router/{}", COMMIT.unwrap_or("local")),
+                    format!("hive-apollo-router/{}", ROUTER_VERSION),
```
n1ru4l: Please rename this const to `PLUGIN_VERSION`

n1ru4l (approved):
Minor comment: https://github.com/graphql-hive/console/pull/7173/files#r2464894375

I coudln't find a integration test for this, so I had to set up my own fake usage reporting server offline to tets the functionality.

The headers seem okay:

```
{
  'x-usage-api-version': '2',
  authorization: 'Bearer torlololol',
  'user-agent': 'hive-apollo-router/2.1.3',
  'content-type': 'application/json',
  accept: '*/*',
  'accept-encoding': 'gzip',
  host: 'host.docker.internal:8811',
  'content-length': '271'
}
```

packages/libraries/router/src/agent.rs
dotansimha: done!

dotansimha:
> I coudln't find a integration test for this, so I had to set up my own fake usage reporting server offline to tets the functionality.

Yeah, no tests for this at the moment. We can create an follow up issues to really test the functionality. It might be worth investing because we are separating this into it's own pkg and going to use this in hive-router too.

PR #7193, opened 2025-10-29
feat: complete --forceSafe implementation to auto-approve breaking changes

### LINKED ISSUE #6902: CLI Schema Check --forceSafe argument should automatically approve the change
--forceSafe should auto-approve breaking changes (assuming the token has permissions), probably with a custom approval message like "Check force approved automatically by ___".

Current behavior is that the CLI ignores the error and therefore does not return an exit code (1). But the hive check is still considered breaking and there is no indication that it passed the check via --forceSafe.

adambenhassen:
Closes #6902

Completes the implementation of the `--forceSafe` CLI flag to actually auto-approve breaking schema changes in Hive, rather than just suppressing the error exit code.

### DIFF
--- a/packages/services/api/src/modules/schema/resolvers/SchemaChangeApproval.ts
+++ b/packages/services/api/src/modules/schema/resolvers/SchemaChangeApproval.ts
@@ -1,8 +1,22 @@
+import { parseCliAuthor } from '../lib/parse-cli-author';
 import { SchemaManager } from '../providers/schema-manager';
 import type { SchemaChangeApprovalResolvers } from './../../../__generated__/types';
 
 export const SchemaChangeApproval: SchemaChangeApprovalResolvers = {
   approvedBy: (approval, _, { injector }) =>
-    injector.get(SchemaManager).getUserForSchemaChangeById({ userId: approval.userId }),
+    approval.userId
+      ? injector.get(SchemaManager).getUserForSchemaChangeById({ userId: approval.userId })
+      : null,
+  cliApprovalMetadata: approval => {
+    if (!approval.author) {
+      return null;
+    }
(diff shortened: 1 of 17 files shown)


packages/services/api/src/modules/schema/resolvers/SchemaChangeApproval.ts
```diff
@@ -1,8 +1,33 @@
...
+        email,
+        displayName,
+        fullName: displayName,
+        provider: 'CLI' as any,
+        externalAuthUserId: null,
+        createdAt: approval.date,
+        isAdmin: false,
+        canSwitchOrganization: false,
+      } as any;
+    }
```
n1ru4l: I would rather prefer adding a new field to the `SchemaChangeApprovel` type for exposing the CLI approval metadata and consume that from the CLI, instead of having this "fake" user object here, even if it means upgrading the CLI version for users.

packages/services/api/src/modules/schema/resolvers/SuccessfulSchemaCheck.ts
```diff
@@ -31,13 +32,38 @@ export const SuccessfulSchemaCheck: SuccessfulSchemaCheckResolvers = {
...
+    }
+
+    if (schemaCheck.manualApprovalUserId) {
+      return injector.get(SchemaManager).getApprovedByUser({
+        organizationId: schemaCheck.selector.organizationId,
+        userId: schemaCheck.manualApprovalUserId,
+      });
+    }
+
+    if (schemaCheck.breakingSchemaChanges) {
```
n1ru4l, line 47: Same as https://github.com/graphql-hive/console/pull/7193/files#r2473979057

PR #7200, opened 2025-10-30
Smaller refinements for Progressive Override

enisdenjo:
Draft because maybe there's more improvements coming in.


enisdenjo:
Landed this, no new comments.

PR #7203, opened 2025-10-31
fix(ui): handle escaped single-quoted strings in schema changes

### LINKED ISSUE #6692: Deprecation changes are incorrectly formatted on schema version view
**SDL**:
```graphql
enum CriticalityLevel {
  Breaking
    @deprecated(
      reason: "Use 'SeverityLevelType' instead. This field will be removed once it is no longer in use by a client."
    )
  Dangerous
    @deprecated(
      reason: "Use 'SeverityLevelType' instead. This field will be removed once it is no longer in use by a client."
    )
  Safe
    @deprecated(
      reason: "Use 'SeverityLevelType' instead. This field will be removed once it is no longer in use by a client."
    )
}
```

Rendered UI:

Image

Example on staging: https://app.staging.graphql-hive.com/the-guild/graphql-hive/staging/history/c4bd48a7-5bbc-493b-9d8a-eb6848894479

adambenhassen:
Closes #6692

Fixes the formatting of deprecation changes on the schema version view by improving the labelize function to properly handle both single and double-quoted identifiers in schema change messages.

Screenshot 2025-11-10 at 19 14 07

Note: this issue is still persistent in CLI output. 
the root cause of this issue lies in [graphql-inspector](https://github.com/graphql-hive/graphql-inspector), which a follow up PR will address.

PR fixes/feat for graphql-inspector: https://github.com/graphql-hive/graphql-inspector/pull/2904

### DIFF
--- a/.changeset/beige-chicken-wonder.md
+++ b/.changeset/beige-chicken-wonder.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+handle escaped single-quoted strings in schema changes
--- a/packages/web/app/src/components/target/history/errors-and-changes.tsx
+++ b/packages/web/app/src/components/target/history/errors-and-changes.tsx
@@ -31,11 +31,10 @@ import { CheckCircledIcon, InfoCircledIcon } from '@radix-ui/react-icons';
 import { Link } from '@tanstack/react-router';
 
 export function labelize(message: string) {
-  // Turn " into '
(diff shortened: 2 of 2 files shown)


n1ru4l:
I don't think we need to labelize/parse and style the deprecation reason that originates from the input schema. Instead we could just escape the string.

RatStar811 (approved):

PR #7133, opened 2025-10-17
feat(api): personal access tokens

n1ru4l:
### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Authentication management
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Error handling and logging
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

PR #7206, opened 2025-11-03
feat(api): project and personal scoped access tokens

n1ru4l:
### Background

Alternative approach for https://github.com/graphql-hive/console/pull/7133

Introduces
- Project scoped access tokens
- User (organization membership) scoped access tokens

Screenshot 2025-11-07 at 13 50 44

Screenshot 2025-11-07 at 13 54 32

Screenshot 2025-11-07 at 13 54 49

### Checklist

- [x] Input validation
- [ ] Output encoding
- [x] Session management
- [x] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [x] Database security
- [ ] File management
- [x] Memory management
- [x] Testing

### DIFF
--- a/packages/libraries/cli/src/commands/whoami.ts
+++ b/packages/libraries/cli/src/commands/whoami.ts
@@ -8,6 +8,7 @@ import {
   MissingEndpointError,
   MissingRegistryTokenError,
 } from '../helpers/errors';
+import { colors } from '../helpers/texture/texture';
 
 const myTokenInfoQuery = graphql(/* GraphQL */ `
   query myTokenInfo($showAll: Boolean!) {
@@ -113,25 +114,52 @@ export default class WhoAmI extends Command<typeof WhoAmI> {
       this.log(`Level: ${permLevel.level}`);
       this.log(`Resources: ${permLevel.resolvedResourceIds?.join(', ') ?? '<none>'}`);
 
-      const table = new Table({
-        head: ['Group', 'Permission ID', 'Title', 'Granted', 'Description'],
-        wordWrap: true,
-        style: { head: ['cyan'] },
-      });
+      const data = [['Group', 'Permission ID', 'Title', 'Granted', 'Description']];
 
       for (const group of permLevel.resolvedPermissionGroups) {
         for (const perm of group.permissions) {
-          table.push([
+          data.push([
             group.title,
             perm.permission.id,
             perm.permission.title,
-            perm.isGranted ? '✓' : '✗',
+            perm.isGranted ? colors.green('✓') : colors.red('✗'),
             perm.permission.description,
           ]);
         }
       }
 
+      const colWidths = calculateColWidths(data);
+
+      const table = new Table({
+        head: data[0],
+        wordWrap: true,
+        style: { head: ['cyan'] },
+        colWidths,
+      });
+
+      data.slice(1).forEach(row => table.push(row));
+
       this.log(table.toString());
     }
   }
+}
+
+function calculateColWidths(data: Array<Array<string>>) {
+  const colCount = data[0].length;
+  const maxColWidths = new Array(colCount).fill(0);
+
+  data.forEach(row => {
+    row.forEach((cell, i) => {
+      maxColWidths[i] = Math.max(maxColWidths[i], cell.length);
+    });
+  });
+
+  const terminalWidth = process.stdout.columns || 80;
+
+  // Account for borders: numCols + 1
+  const totalBorder = colCount + 1;
+  const totalContentWidth = maxColWidths.reduce((a, b) => a + b, 0);
+
+  // Calculate scaled width for each column
+  return maxColWidths.map(w => Math.floor((w / totalContentWidth) * (terminalWidth - totalBorder)));
 }


packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export default {
+  name: '2025.10.17T00-00-00.organization-access-tokens-project-scope.ts',
+  noTransaction: true,
+  run: ({ sql }) => [
+    {
+      name: 'add new columns to "organization_access_tokens"',
+      query: sql`
+        ALTER TABLE "organization_access_tokens"
+          ADD COLUMN IF NOT EXISTS "project_id" UUID REFERENCES "projects" ("id") ON DELETE CASCADE
```
n1ru4l, line 11: If this column is populated, the access token is a access token bound to a project scope.

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
```diff
@@ -0,0 +1,37 @@
...
+          , ADD COLUMN IF NOT EXISTS "user_id" UUID REFERENCES "users" ("id") ON DELETE CASCADE
```
n1ru4l, line 12: If this column is populated, the access token is a access token bound to a specific user.

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
```diff
@@ -0,0 +1,37 @@
...
+          , ALTER COLUMN "permissions" DROP NOT NULL;
```
n1ru4l, line 13: For users it is possible that this column has the value `null`. In such a scenario, the access token will inherit all the permissions defined by the users organization membership.

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
```diff
@@ -0,0 +1,37 @@
...
+      `,
+    },
+    {
+      name: 'add index "organization_access_tokens_pagination_target"',
+      query: sql`
+        CREATE INDEX CONCURRENTLY IF NOT EXISTS "organization_access_tokens_pagination_target" ON "organization_access_tokens" (
```
n1ru4l:
```suggestion
        CREATE INDEX CONCURRENTLY IF NOT EXISTS "organization_access_tokens_pagination_project" ON "organization_access_tokens" (
```

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
```diff
@@ -0,0 +1,37 @@
...
+      name: 'add new columns to "organization_access_tokens"',
+      query: sql`
+        ALTER TABLE "organization_access_tokens"
+          ADD COLUMN IF NOT EXISTS "project_id" UUID REFERENCES "projects" ("id") ON DELETE CASCADE
+          , ADD COLUMN IF NOT EXISTS "user_id" UUID REFERENCES "users" ("id") ON DELETE CASCADE
+          , ALTER COLUMN "permissions" DROP NOT NULL;
+      `,
+    },
+    {
+      name: 'add index "organization_access_tokens_pagination_target"',
```
n1ru4l:
```suggestion
      name: 'add index "organization_access_tokens_pagination_project"',
```

packages/services/api/src/modules/auth/lib/authz.ts
```diff
@@ -393,6 +393,7 @@ const permissionsByLevel = {
     z.literal('schemaLinting:modifyOrganizationRules'),
     z.literal('auditLog:export'),
     z.literal('accessToken:modify'),
+    z.literal('personalAccessToken:modify'),
```
n1ru4l, line 396: This new permission allows users to create, delete and use their own access tokens. If the `personalAccessToken:modify` permission is revoked from a user, all his access tokens are invalid/revoked, until the permission is re-added to the user.

These personal access tokens (PAT), are special compared to project and organization scope access tokens, as the user can only authorise actions on permissions and resouces they have access on.

If a permission or resource access is revoked from the users organization membership, this is reflected in the access token (until the permissions/resources are again authorized for the user).

PATs are crucial for developers, so they can compose their local subgraphs with the subgraph in the schema registry, without having to ask a user with a project or organization access to issue them an access token.

packages/services/api/src/modules/auth/lib/authz.ts
```diff
@@ -401,6 +402,7 @@ const permissionsByLevel = {
     z.literal('alert:modify'),
     z.literal('schemaLinting:modifyProjectRules'),
     z.literal('target:create'),
+    z.literal('projectAccessToken:modify'),
```
n1ru4l, line 405: This new permission allows users to create and delete access tokens on a specific project.

These access tokens are similar to organization access tokens, with the difference of being bound to a single project. The user that has access to these access tokens could abuse it to perform actions that their user role might not be granted with.

The purpose of this access token is to have a way for project owners to issue CI/CD access tokens, without requiring full "admin" access in the whole organization, which is a problem in bigger organizations, where the organization admin is a complete IT department that just bootstraps the projects and project members, and leaves all other tasks to the project members.

packages/services/api/src/modules/organization/lib/organization-access-token-permissions.ts
```diff
@@ -147,6 +152,11 @@ export const permissionGroups: Array<PermissionGroup> = [
         title: 'Publish app deployment',
         description: 'Grant access to publishing app deployments.',
       },
+      {
+        id: 'appDeployment:retire',
```
n1ru4l, line 156: this one was missing 😓

packages/services/api/src/modules/organization/lib/organization-member-permissions.ts
```diff
@@ -237,6 +244,71 @@ export const permissionGroups: Array<PermissionGroup> = [
...
+        dependsOn: 'personalAccessToken:modify',
+      },
+      {
+        id: 'traces:report',
+        title: 'Report OTEL traces',
+        description: 'Grant access to reporting traces.',
+        dependsOn: 'personalAccessToken:modify',
+      },
+    ],
+  },
```
n1ru4l, line 311: Since personal access token inherit their permissions from the organization members role, the admin MUST assign these permissions to the organization member role, so they can grant themselves these permissions.

IMHO, most subgraph developers will only ever need `schema:compose`. However, this is super flexible, and will allow all kind of constructs...

Documentation will clarify this and also quadruple highlight that you should use project access tokens for schema publish, usage reprting and other CI/CD/prod stuff.

packages/services/api/src/modules/organization/lib/resource-assignment-model.ts
```diff
@@ -79,3 +87,111 @@ export const ResourceAssignmentModel = z.union([
  */
 export type ResourceAssignmentGroup = z.TypeOf<typeof ResourceAssignmentModel>;
 export type GranularAssignedProjects = z.TypeOf<typeof GranularAssignedProjectsModel>;
+
+/**
+ * Get the intersection of two resource assignments
+ */
+export function intersectResourceAssignments(
```
n1ru4l, line 94: Helper function for narrowing down access token resources based on the personal access tokens owners permissions.

packages/services/api/src/modules/organization/providers/organization-access-tokens-cache.ts
```diff
@@ -5,11 +5,29 @@ import { Inject, Injectable, Scope } from 'graphql-modules';
...
+  id: string;
+  organizationId: string;
+  /** undefined, since older cache records might not contain this property. */
+  projectId?: string | null;
+  /** undefined, since older cache records might not contain this property. */
+  userId?: string | null;
+  authorizationPolicyStatements: Array<AuthorizationPolicyStatement>;
+  hash: string;
+  firstCharacters: string;
+};
```
n1ru4l, line 30: We are now only storing a subset of the access token record in the cache, it did not make sense to also store the "raw" permissions and resources, even though in reality it would probably not make a big performance difference.

packages/services/api/src/modules/organization/providers/organization-access-tokens-cache.ts
```diff
@@ -50,29 +68,76 @@ export class OrganizationAccessTokensCache {
...
 
+  private async makeCacheRecord(
+    logger: Logger,
+    accessToken: OrganizationAccessToken,
+  ): Promise<CachedAccessToken | null> {
+    logger.debug('create cache record for access token');
+    let authorizationPolicyStatements: Array<AuthorizationPolicyStatement> | null = null;
+
+    // in this case, we need to load the membership and filter down permissions based on the viewer.
+    if (accessToken.userId) {
```
n1ru4l, line 79: as mentioned before, every time we resolve a personal access token, we need to load the membership and narrow down the permissions/resources based on that.

We cache access tokens for 5 minutes. In the worst case a user will have elevated rights for that period. We could further invalidate access tokens when a users role/resources are updated etc - but it is right now not worth the hassle.

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -315,31 +567,533 @@ export class OrganizationAccessTokens {
...
+      .map(group => ({
+        ...group,
+        permissions: group.permissions.filter(permission =>
+          validProjectResourceLevels.has(getPermissionGroup(permission.id)),
+        ),
+      }))
+      .filter(group => group.permissions.length !== 0);
+  }
+
+  private createFeatureFlagPermissionFilter(organization: Organization) {
```
n1ru4l, line 891: One unified place to apply permission filters based on feature flags (at least for access tokens)? 🤞

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -315,31 +567,533 @@ export class OrganizationAccessTokens {
...
+          .filter(permission => filter(permission.id))
+          .map(permission => ({
+            ...permission,
+            isAssignableByViewer: true,
+          })),
+      }))
+      .filter(group => group.permissions.length !== 0);
+  }
+
+  async getAvailablePermissionGroupsForMembership(membership: OrganizationMembership) {
```
n1ru4l, line 920: `getAvailablePermissionGroupsForMembership`, `getAvailablePermissionGroupsForOrganization` and `getAvailablePermissionsGroupsForProject` are now the "source" of truth for the assignable permissions on the frontend.

All the permissions are filtered on:
- feature flag
- resource level (org or project)
- membership permissions (only for PAT)

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -315,31 +567,533 @@ export class OrganizationAccessTokens {
...
+        ...group,
+        permissions: group.permissions.map(permission => ({
+          ...permission,
+          isAssignableByViewer: memberPermissions.has(permission.id),
+        })),
+      }))
+      .filter(group => group.permissions.length !== 0);
+  }
+
+  async getPermissionsForAccessToken(accessToken: OrganizationAccessToken) {
```
n1ru4l, line 938: Resolves the permissions for an access token, takes into account all constraints such as for PATs

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -315,31 +567,533 @@ export class OrganizationAccessTokens {
...
+    return this.resourceAssignments.resolveGraphQLMemberResourceAssignment({
+      organizationId: accessToken.organizationId,
+      resources: intersectResourceAssignments(
+        accessToken.assignedResources,
+        membership?.assignedRole.resources,
+      ),
+    });
+  }
+
+  getGraphQLResolvedResourcePermissionGroupForAccessToken =
```
n1ru4l, line 993: This one is a bit tricky... 😅 

This "resolves" all the permissions and resources from a access token (or access token like for compatibility with legacy access tokens), and exposes them in a easy to consume format for both the CLI and Web App Access Token Detail View.

It is not perfect, as it could give more insights for the following:
- PATs: show which permissions where granted, but got revoked because of membership permission changes (same for resources)
- deleted resources: show deleted resources, that no longer exists using their UUID instead

However, the latter is something for a follow up pull request to reduce the already high complexity in this PR.

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -315,31 +567,533 @@ export class OrganizationAccessTokens {
...
+          includeAll
+            ? true
+            : group.resolvedPermissionGroups.length !== 0 && group.resolvedResourceIds?.length,
+        );
+    };
+
+  static computeAuthorizationStatements(
+    accessToken: OrganizationAccessToken,
+    membership: OrganizationMembership,
+  ) {
```
n1ru4l, line 1140: Function used by access token cache for personal access tokens. I made it static to avoid dependency injection where we do not use graphql-modules...

packages/services/server/src/index.ts
```diff
@@ -604,6 +605,15 @@ export async function main() {
       targetsByIdCache: registry.injector.get(TargetsByIdCache),
     });
 
+    server.post('/cache/organization-access-token-cache/delete/:id', async (req, res) => {
```
n1ru4l, line 608: very useful for integration tests to check if membership permission/resource updates propagate

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -769,4 +892,81 @@ export default gql`
...
+
+  type ResolvedResourcePermissionGroup {
+    """
+    On what resoucre level this permission applies.
+    """
+    level: PermissionLevelType!
+    """
+    Resource ids that are currently valid for this permission group.
+    """
+    resolvedResourceIds: [String!]
```
n1ru4l, line 1348: These are the "current" resolvable resource IDs.

We could add another field here that exposes more detailed information, as I mentioned before. But we should put that into another PR :)

```graphql

type ResouceAssignment {
  # e.g. "the-guild/graphql-hive/staging"
  # if the resource no longer exists we cannot resolve this anymore, as we internally only store the UUID.
  humanReadableName: String
  # e.g. "<UUID>/<UUID>/<UUID>"
  machineName: String!
}

type ResolvedResouceAssignment {
  # specified upon creation of the access token
  specified: ResouceAssignment!
  # the ones that are currently active
  resolved: [ResouceAssignment!]
  # indicates whether the assignment was "downgraded" due to member role permission changes
  hasChanged: Boolean
}

extend type ResolvedResourcePermissionGroup {
  resourceAssignments: ResolvedResouceAssignment!
}
```

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -769,4 +892,81 @@ export default gql`
...
+    resolvedPermissions(
+      """
+      Whether an overview of all permissions should be included in the output, even if not granted.
+      """
+      includeAll: Boolean = false
+    ): [ResolvedResourcePermissionGroup!]!
+  }
+
+  extend type Query {
+    whoAmI: WhoAmI
```
n1ru4l, line 1387: I wanted a simple field that can expose the permissions of any API consumer. Whether user, org access token or legacy access token.

It is now used in the CLI, to re-introduce the `whoami` command, which was broken after we shipped organization access tokens.

It is also useful for integration tests...

packages/web/app/src/components/organization/members/permission-selector.tsx
```diff
@@ -33,6 +33,7 @@ export const PermissionSelector_PermissionGroupsFragment = graphql(`
       title
       isReadOnly
       warning
+      isAssignableByViewer
```
n1ru4l, line 36: This is new field indicates whether a user can assign the permission, it is only relevant in the context of PATs tho.

packages/web/app/src/components/organization/members/resource-selector.tsx
```diff
@@ -141,14 +141,16 @@ export function ResourceSelector(props: {
   organization: FragmentType<typeof ResourceSelector_OrganizationFragment>;
   selection: ResourceSelection;
   onSelectionChange: (selection: ResourceSelection) => void;
+  /**
+   * Scope the resource selector to a specific project.
+   * If this property is provided, please make sure that the `selection` property contains the project.
+   * */
+  forProjectId?: string;
```
n1ru4l, line 148: For personal access tokens, it does not make sense to select any other project, so I introduced this scoped "mode".

packages/web/app/src/components/organization/members/selected-permission-overview.tsx
```diff
@@ -35,6 +35,8 @@ export type SelectedPermissionOverviewProps = {
   isExpanded?: boolean;
   /** option for injecting additional content within a permission group. */
   additionalGroupContent?: (group: { level: PermissionLevelType }) => React.ReactNode;
+  /** Exclude organization for project settings. */
+  excludeOrganization?: boolean;
```
n1ru4l, line 39: sam as in the resource selector. Does not make sense in project context.

packages/web/app/src/components/organization/settings/access-tokens/permission-detail-view.tsx
n1ru4l, line 1: This new component takes care of showing the permissions of any access token and replaces processing and building the data on the frontend via `OrganizationAccessToken.permissions` and `OrganizationAccessToken.resources`.

packages/web/app/src/components/organization/settings/access-tokens/access-token-detail-view-sheet.tsx
```diff
@@ -18,37 +14,11 @@ const AccessTokenDetailViewSheet_OrganizationQuery = graphql(`
...
-                  mode
-                  services
-                }
-                appDeployments {
-                  mode
-                  appDeployments
-                }
-              }
-            }
-          }
```
n1ru4l, line 46: no longer needed with `packages/web/app/src/components/organization/settings/access-tokens/permission-detail-view.tsx` :)

packages/web/app/src/components/organization/settings/personal-access-tokens/create-personal-access-token-sheet-content.tsx
n1ru4l, line 1: Note here for:
- `organization/settings/access-tokens`
- `organization/settings/personal-access-tokens`
- `project/settings/personal-access-tokens`

The components are almost the same everywhere with slight differences, however, I did not really find an abstraction that made much sense... I avoided trying to make things here as re-usable as possible for now.

.changeset/five-hoops-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-hive/cli': minor
+---
+
+Improve output of the `hive whoami` command. It now also handles the new access token format.
```
jdolle, line 5: Add a change for `hive` also for the token scoping.

integration-tests/tests/api/access-tokens/organization-level.spec.ts
```diff
@@ -0,0 +1,413 @@
...
+  const org = await createOrg();
+
+  const result = await execute({
+    document: CreateOrganizationAccessTokenMutation,
+    variables: {
+      input: {
+        organization: {
+          byId: org.organization.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

integration-tests/tests/api/access-tokens/organization-level.spec.ts
```diff
@@ -0,0 +1,413 @@
...
+      },
+    },
+    authToken: ownerToken,
+  }).then(e => e.expectNoGraphQLErrors());
+  expect(result.createOrganizationAccessToken.error).toEqual(null);
+  expect(result.createOrganizationAccessToken.ok).toEqual({
+    privateAccessKey: expect.any(String),
+    createdOrganizationAccessToken: {
+      id: expect.any(String),
+      title: 'a access token',
```
jdolle:
```suggestion
      title: 'an access token',
```

integration-tests/tests/api/access-tokens/organization-level.spec.ts
```diff
@@ -0,0 +1,413 @@
...
+      },
+    },
+    authToken: ownerToken,
+  }).then(e => e.expectNoGraphQLErrors());
+  expect(result.createOrganizationAccessToken.ok).toEqual(null);
+  expect(result.createOrganizationAccessToken.error).toMatchInlineSnapshot(`
+    {
+      details: {
+        description: null,
+        title: Can only contain letters, numbers, " ", "_", and "-".,
```
jdolle, line 103: technically `'    '` meets this criteria.

integration-tests/tests/api/access-tokens/organization-level.spec.ts
```diff
@@ -0,0 +1,413 @@
...
+  const org = await actor1.createOrg();
+
+  const errors = await execute({
+    document: CreateOrganizationAccessTokenMutation,
+    variables: {
+      input: {
+        organization: {
+          byId: org.organization.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

integration-tests/tests/api/access-tokens/organization-level.spec.ts
```diff
@@ -0,0 +1,413 @@
...
+  const project = await org.createProject(GraphQLSchema.ProjectType.Federation);
+
+  const result = await execute({
+    document: CreateOrganizationAccessTokenMutation,
+    variables: {
+      input: {
+        organization: {
+          byId: org.organization.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

integration-tests/tests/api/access-tokens/organization-level.spec.ts
```diff
@@ -0,0 +1,413 @@
...
+  const project2 = await org.createProject(GraphQLSchema.ProjectType.Federation);
+
+  const result = await execute({
+    document: CreateOrganizationAccessTokenMutation,
+    variables: {
+      input: {
+        organization: {
+          byId: org.organization.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

integration-tests/tests/api/access-tokens/personal-level.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+          description
+        }
+      }
+    }
+  }
+`);
+
+const DeletePersonalAccessTokenMutation = graphql(`
+  mutation DeletePersonalAccessTokenMutation {
+    #   ) #     #$input: DeletePersonalAccessTokenInput! # (
```
jdolle: delete?

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+      },
+    },
+    authToken: ownerToken,
+  }).then(e => e.expectNoGraphQLErrors());
+  expect(result.createPersonalAccessToken.error).toEqual(null);
+  expect(result.createPersonalAccessToken.ok).toEqual({
+    privateAccessKey: expect.any(String),
+    createdPersonalAccessToken: {
+      id: expect.any(String),
+      title: 'a access token',
```
jdolle:
```suggestion
      title: 'an access token',
```

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+
+  const result = await execute({
+    document: CreatePersonalAccessTokenMutation,
+    variables: {
+      input: {
+        organization: {
+          byId: org.organization.id,
+        },
+        title: 'a access token',
+        description: new Array(300).fill('A').join(''),
```
jdolle:
```suggestion
        description: new Array(249).fill('A').join(''),
```

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+  const org = await actor1.createOrg();
+
+  const errors = await execute({
+    document: CreatePersonalAccessTokenMutation,
+    variables: {
+      input: {
+        organization: {
+          byId: org.organization.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

integration-tests/tests/api/access-tokens/personal-level.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+        code: 'UNAUTHORISED',
+      },
+
+      message: `No access (reason: "Missing permission for performing 'personalAccessToken:modify' on resource")`,
+      path: ['createPersonalAccessToken'],
+    },
+  ]);
+});
+
+// test.concurrent('delete: successfuly delete own access token', async ({ expect }) => {
```
jdolle: will adding this be a followup PR or is this not yet done?

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+  const project = await org.createProject(GraphQLSchema.ProjectType.Federation);
+
+  const result = await execute({
+    document: CreatePersonalAccessTokenMutation,
+    variables: {
+      input: {
+        organization: {
+          byId: org.organization.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+  const project2 = await org.createProject(GraphQLSchema.ProjectType.Federation);
+
+  const result = await execute({
+    document: CreatePersonalAccessTokenMutation,
+    variables: {
+      input: {
+        organization: {
+          byId: org.organization.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+    organization: {
+      id: expect.any(String),
+      project: null,
+      slug: org.organization.slug,
+    },
+  });
+});
+
+test.concurrent(
+  'query GraphQL API after membership resources have been downgraded',
```
jdolle, line 371: I love that you're testing this case

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
+    });
+
+    const result = await execute({
+      document: CreatePersonalAccessTokenMutation,
+      variables: {
+        input: {
+          organization: {
+            byId: org.organization.id,
+          },
+          title: 'a access token',
```
jdolle:
```suggestion
          title: 'an access token',
```

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
```
jdolle:
```suggestion
          title: 'an access token',
```

integration-tests/tests/api/access-tokens/personal.spec.ts
```diff
@@ -0,0 +1,762 @@
...
```
jdolle:
```suggestion
          title: 'an access token',
```

integration-tests/tests/api/access-tokens/project-level.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  const { project } = await org.createProject(GraphQLSchema.ProjectType.Federation);
+
+  const result = await execute({
+    document: CreateProjectAccessTokenMutation,
+    variables: {
+      input: {
+        project: {
+          byId: project.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

integration-tests/tests/api/access-tokens/project-level.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+      },
+    },
+    authToken: ownerToken,
+  }).then(e => e.expectNoGraphQLErrors());
+  expect(result.createProjectAccessToken.error).toEqual(null);
+  expect(result.createProjectAccessToken.ok).toEqual({
+    privateAccessKey: expect.any(String),
+    createdProjectAccessToken: {
+      id: expect.any(String),
+      title: 'a access token',
```
jdolle:
```suggestion
      title: 'an access token',
```

integration-tests/tests/api/access-tokens/project-level.spec.ts
```diff
@@ -0,0 +1,160 @@
...
+  const { project } = await org.createProject(GraphQLSchema.ProjectType.Federation);
+
+  const errors = await execute({
+    document: CreateProjectAccessTokenMutation,
+    variables: {
+      input: {
+        project: {
+          byId: project.id,
+        },
+        title: 'a access token',
```
jdolle:
```suggestion
        title: 'an access token',
```

packages/libraries/cli/src/commands/whoami.ts
```diff
@@ -95,62 +94,44 @@ export default class WhoAmI extends Command<typeof WhoAmI> {
...
 
-    return lines.join('\n');
-  };
+      for (const group of permLevel.resolvedPermissionGroups) {
+        for (const perm of group.permissions) {
+          table.push([
+            group.title,
+            perm.permission.id,
+            perm.permission.title,
+            perm.isGranted ? '✓' : '✗',
```
jdolle: It could be nice to do color output also. I find it helps a lot with quickly scanning things.

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
jdolle: in a previous version of this feature, you created a separate table. Why opt to do a new column now?

It seems like tokens are bound to either the org, the user, or the project, but never project and user. I think for this reason these should be separate tables.

packages/services/api/src/modules/audit-logs/providers/audit-logs-types.ts
```diff
@@ -332,8 +332,20 @@ export const AuditLogModel = z.union([
     eventType: z.literal('ORGANIZATION_ACCESS_TOKEN_CREATED'),
     metadata: z.object({
       organizationAccessTokenId: z.string().uuid(),
-      permissions: z.array(z.string()),
+      permissions: z.array(z.string()).nullable(),
       assignedResources: ResourceAssignmentModel,
+      userId: z
+        .string()
+        // optional, because it was introduced later on.
```
jdolle, line 339: using separate tables would simplify this also

packages/services/api/src/modules/auth/lib/authz.ts
jdolle: Was being able to control this requested by anyone? I'm trying to think of a use case where youd want to disable someone's PAT but keep them as a member and I can't really think of anything.
Maybe we could avoid the additional permission?

packages/services/api/src/modules/organization/lib/organization-access-token-permissions.ts
jdolle: good catch

packages/services/api/src/modules/organization/providers/organization-access-tokens-cache.ts
jdolle: I think it's worth doing. But not because of the elevated permissions.
If someone updates their PAT to add additional permissions then it would be best to update that immediately so they can use that token. I can foresee lots of confusion if they add a permission but then can't perform the action.

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -62,10 +78,31 @@ const OrganizationAccessTokenModel = z
     // only used in the context of authorization, we do not need
     // to compute when querying a list of organization access tokens via the GraphQL API.
     get authorizationPolicyStatements() {
-      const permissions = permissionsToPermissionsPerResourceLevelAssignment(record.permissions);
+      if (record.permissions === null && record.userId === null) {
+        throw new Error(
```
jdolle, line 94: should this be a HiveError?

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -122,6 +161,95 @@ export class OrganizationAccessTokens {
...
+      action: 'projectAccessToken:modify',
+    });
+    const canModifyOrganizationAccessTokens = await this.session.canPerformAction({
+      organizationId,
+      params: { organizationId },
+      action: 'accessToken:modify',
+    });
+
+    if (!canModifyProjectAccessTokens && !canModifyOrganizationAccessTokens) {
+      this.session.raise('projectAccessToken:modify');
```
jdolle: i wonder if these should be considered separate rather than hierarchical?

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
```diff
@@ -315,31 +569,533 @@ export class OrganizationAccessTokens {
...
+
+    const organization = await this.storage.getOrganization({
+      organizationId: accessToken.organizationId,
+    });
+    const membership = await this.members.findOrganizationMembership({
+      organization,
+      userId: accessToken.userId,
+    });
+
+    // TODO: we could handle this different TBH
```
jdolle: should we?
Prefer creating a ticket and deleting the line if it matters.

jdolle (commented):
Reviewed half the files so far. I'll get the rest tomorrow

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
n1ru4l: A benefit of this approach is that on the organization level, the admin can still see and revoke all kind of PATs and project scoped access token.

n1ru4l: image

n1ru4l: We could split it into two tables, which would then duplicate a lot of code (see previous PR), and also complicate pagination.

n1ru4l: This kind of visibility for organization admins is sometihng @dotansimha, wished a lot of other tools had 😆

packages/services/api/src/modules/auth/lib/authz.ts
n1ru4l: Nobody explicitly asked for this, to me it felt like it makes sense, since we have permissions for every kind of action.
Maybe in case of a security breach it would make sense to disable these?

I don't see a need for someone that is not a developer needing to create personal access tokens. I don't think it is a big deal "maintaining" this additional permission string here.

packages/services/api/src/modules/organization/providers/organization-access-tokens-cache.ts
n1ru4l: You cannot update access tokens right now, they are create and delete only. The only change happens when a users role assignment changes. In that case we would need to invalidate all cached access tokens by that user. `bentocache` added a [tagging](https://bentocache.dev/docs/tagging) feature, that could help with this. However, I wanted to avoid using an experimental feature for now.

I agree though that we should make the update as immediate as possible in a follow-up task.

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
dotansimha: Yeah, if a token is created by a user and granted access to more than the user itself (project/target), it might end up with dangling tokens that cannot be managed by admin. 

I suggested to allow admin/owner to manage all tokens that were created for the org, as it's something that was missing in many services, and impacting IT teams. 

For example, if a user is diabled in SSO provider, it might not get reflected to our service, so token might still live. Or, in cases where tokens leak, admin needs controll over tokens. 

On the technical aspect, I think we are stepping towards a direction where tokens are a more broad thing (and not really just org-based), so it's a good thing to allow owner accounts to mange these.

dotansimha: ( I don't mind the amount of tables here to ne honest, i mostly want the ability to manage it as an admin. seems like using the same table works well for this)

packages/services/api/src/modules/audit-logs/providers/audit-logs-types.ts
```diff
@@ -332,8 +332,20 @@ export const AuditLogModel = z.union([
...
+        .optional()
+        .nullable()
+        .transform(value => value ?? null),
+      projectId: z
```
dotansimha, line 343: (not really on this line)

I wanted to confirm that we have audit log event for deletion of tokens, because now an owner/admin account could delete tokens that are created by other users, and this needs to be logged in the audit log.

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
jdolle: I think the shared screenshot raises concerns for me because searching for organization access tokens becomes more difficult. I'd rather they be separate lists, and the personal tokens grouped/searchable by user.

packages/services/api/src/modules/auth/lib/authz.ts
jdolle: I agree it isnt a bit deal, but I am thinking as we continue to add features, maybe the list of permissions will become a nuisance. I'm fine either way until that day comes.

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -50,9 +50,145 @@ export default gql`
...
     ): CreateOrganizationAccessTokenResult! @tag(name: "public")
+    createProjectAccessToken(input: CreateProjectAccessTokenInput!): CreateProjectAccessTokenResult!
+    createPersonalAccessToken(
+      input: CreatePersonalAccessTokenInput!
+    ): CreatePersonalAccessTokenResult!
     deleteOrganizationAccessToken(
-      input: DeleteOrganizationAccessTokenInput! @tag(name: "public")
-    ): DeleteOrganizationAccessTokenResult! @tag(name: "public")
+      input: DeleteOrganizationAccessTokenInput!
+    ): DeleteOrganizationAccessTokenResult!
```
jdolle, line 74: did you mean to make this private?

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -50,9 +50,145 @@ export default gql`
     createOrganizationAccessToken(
       input: CreateOrganizationAccessTokenInput! @tag(name: "public")
     ): CreateOrganizationAccessTokenResult! @tag(name: "public")
+    createProjectAccessToken(input: CreateProjectAccessTokenInput!): CreateProjectAccessTokenResult!
+    createPersonalAccessToken(
```
jdolle, line 65: I like that you made these separate mutations :)

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -111,37 +247,113 @@ export default gql`
     description: String @tag(name: "public")
   }
 
-  type OrganizationAccessToken {
+  enum AccessTokenScopeType {
```
jdolle: Why do we need this when we have `__typename`s?

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -111,37 +247,113 @@ export default gql`
...
     id: ID! @tag(name: "public")
+    scope: AccessTokenScopeType!
     title: String! @tag(name: "public")
     description: String @tag(name: "public")
     permissions: [String!]! @tag(name: "public")
     resources: ResourceAssignment! @tag(name: "public")
     firstCharacters: String! @tag(name: "public")
     createdAt: DateTime! @tag(name: "public")
+    resolvedResourcePermissionGroups(
+      includeAll: Boolean = false
```
jdolle:
```suggestion
      includeAll: Boolean! = false
```
This could be non-nullable since it's being defaulted.

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -111,37 +247,113 @@ export default gql`
...
+
+  type ProjectAccessToken implements AccessToken {
+    id: ID! @tag(name: "public")
+    scope: AccessTokenScopeType!
+    title: String! @tag(name: "public")
+    description: String @tag(name: "public")
+    firstCharacters: String! @tag(name: "public")
+    createdAt: DateTime! @tag(name: "public")
+    resolvedResourcePermissionGroups(
+      includeAll: Boolean = false
```
jdolle:
```suggestion
      includeAll: Boolean! = false
```

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -111,37 +247,113 @@ export default gql`
...
+
+  type PersonalAccessToken implements AccessToken {
+    id: ID! @tag(name: "public")
+    scope: AccessTokenScopeType!
+    title: String! @tag(name: "public")
+    description: String @tag(name: "public")
+    firstCharacters: String! @tag(name: "public")
+    createdAt: DateTime! @tag(name: "public")
+    resolvedResourcePermissionGroups(
+      includeAll: Boolean = false
```
jdolle:
```suggestion
      includeAll: Boolean! = false
```

packages/services/api/src/modules/organization/module.graphql.ts
jdolle: Simplicity is good for now.

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -769,4 +995,81 @@ export default gql`
...
+    resolvedPermissionGroups: [ResolvedPermissionsGroup!]!
+  }
+
+  type WhoAmI {
+    title: String!
+    resolvedPermissions(
+      """
+      Whether an overview of all permissions should be included in the output, even if not granted.
+      """
+      includeAll: Boolean = false
```
jdolle:
```suggestion
      includeAll: Boolean! = false
```

packages/services/api/src/modules/organization/providers/organization-member-roles.ts
```diff
@@ -278,6 +287,34 @@ export class OrganizationMemberRoles {
...
+      const query = sql`
+        SELECT
+          ${organizationMemberRoleFields}
+        FROM
+          "organization_member_roles"
+        WHERE
+          "id" = ${roleId}
+      `;
+
+      const result = await deps.pool.maybeOne<unknown>(query);
```
jdolle, line 306: should we wrap this in try/catch to log if it errors?

packages/services/api/src/modules/organization/providers/resource-assignments.ts
```diff
@@ -249,6 +251,106 @@ export class ResourceAssignments {
 
     return resourceAssignmentGroup;
   }
+
+  /**
+   * Translate a resource assignment into resource ids grouped by resource level.
```
jdolle, line 256: I forget if we have this already -- but documentation showing the resource ID structures would be helpful in understanding this. This may be useful for self-hosters, but I think it would especially be useful to our team

packages/services/api/src/modules/organization/resolvers/Query/whoAmI.ts
```diff
@@ -0,0 +1,63 @@
...
+      });
+
+    return {
+      title: `Legacy Target Access Token - ${session.id}`,
+      resolvedPermissions,
+    };
+  }
+
+  if (session instanceof UnauthenticatedSession) {
+    throw new HiveError('Not authneticated.');
```
jdolle:
```suggestion
    throw new HiveError('Not authenticated.');
```
I'm surprised we don't have a special error instance for unauthenticated calls though

packages/web/app/src/components/organization/members/permission-selector.tsx
```diff
@@ -139,11 +140,32 @@ export function PermissionSelector(props: PermissionSelectorProps) {
                         }
                       }}
                     >
-                      <div className={cn(needsDependency && 'opacity-30')}>
+                      <div
+                        className={cn(
+                          (needsDependency ||
+                            !permission.isAssignableByViewer ||
```
jdolle:
```suggestion

```
redundant with the next line

packages/web/app/src/components/project/settings/access-tokens/project-access-tokens-table.tsx
```diff
@@ -0,0 +1,174 @@
...
+            <>
+              <LoaderCircleIcon className="mr-2 inline size-4 animate-spin" /> Loading
+            </>
+          ) : (
+            'Load more'
+          )}
+        </Button>
+      </Table.TableCaption>
+      <Table.TableHeader>
+        <Table.TableRow>
```
jdolle, line 110: I dont see any dynamic resizing based on screen size.
This is one thing I've been more attentive to lately and would like to see continued. How does this UI handle small screens etc?

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
dotansimha: A filter should work in that case?

n1ru4l: Yeah adding a filter is on my todo list

packages/services/api/src/modules/audit-logs/providers/audit-logs-types.ts
n1ru4l: its on the follow up line :D

packages/services/api/src/modules/organization/module.graphql.ts
n1ru4l: this was an accident, i removed it but re-added it :D

packages/migrations/src/actions/2025.10.17T00-00-00.project-access-tokens.ts
n1ru4l: Adding the filter will be a follow up task

integration-tests/tests/api/access-tokens/personal-level.spec.ts
n1ru4l: i readded it, i was doing some stuff before :D

integration-tests/tests/api/access-tokens/personal-level.spec.ts
n1ru4l: re-added it, was just commented out temporarily

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
n1ru4l: It is more of a developer did something wrong error.

packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
n1ru4l: removed

packages/web/app/src/components/project/settings/access-tokens/project-access-tokens-table.tsx
n1ru4l: no idea, but something I will not address in this PR.

.changeset/five-hoops-punch.md
n1ru4l: I added precise changesets, more details is covered in https://github.com/graphql-hive/console/pull/7231

packages/libraries/cli/src/commands/whoami.ts
n1ru4l: done!

packages/services/api/src/modules/organization/providers/organization-member-roles.ts
n1ru4l: If this throws an exception it will be logged, and we will notice. I don't see a need for explicit error handling here.

packages/services/api/src/modules/organization/providers/resource-assignments.ts
n1ru4l: I added some more documentation within the GraphQL schema, authz.ts file also covers a lot of this.

PR #7225, opened 2025-11-05
fix: disable generate token button while submitting on create access token form

### LINKED ISSUE #7211: No operation idempotency check while creating new registry tokens
When you're creating a specific access token for the given registry, it doesn't provide any kind of idempotency per operation. 

An example below, which shows the behavior:

https://github.com/user-attachments/assets/25c4b1c5-34da-4f24-a777-b94602058cfe

---

I do believe that it should have some logic to guarantee that it won't create more than one registry token. I think that the easiest solution would be adding a disabled state into the button while waiting the operation to be concluded.

noghartt:
### Background

Fix #7211 
Closes https://github.com/graphql-hive/console/pull/7226

### Description

It guarantees that each operation, while creating the registry token, won't "n-plicate" the tokens while clicking in the button. As an example of the current behavior, there's a video in the original issue showing it.

### Checklist

- [x] Input validation

### DIFF
--- a/packages/web/app/src/components/target/settings/registry-access-token.tsx
+++ b/packages/web/app/src/components/target/settings/registry-access-token.tsx
@@ -267,7 +267,11 @@ export function GenerateTokenContent(props: {
             <Button
               type="submit"
               data-cy="submit"
-              disabled={!props.form.formState.isValid || props.noPermissionsSelected}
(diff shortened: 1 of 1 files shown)


n1ru4l:
@noghartt thanks!

PR #7228, opened 2025-11-05
development docs updates

jonathanawesome:
### Description

- updates listed versions of node and pnpm in DEVELOPMENT.md
- removes deprecated `version` in docker compose files

### DIFF
--- a/packages/web/app/src/components/target/settings/registry-access-token.tsx
+++ b/packages/web/app/src/components/target/settings/registry-access-token.tsx
@@ -267,7 +267,11 @@ export function GenerateTokenContent(props: {
             <Button
               type="submit"
               data-cy="submit"
-              disabled={!props.form.formState.isValid || props.noPermissionsSelected}
+              disabled={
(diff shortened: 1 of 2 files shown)


docs/DEVELOPMENT.md
```diff
@@ -4,16 +4,16 @@
 
 Developing Hive locally requires you to have the following software installed locally:
 
-- Node.js 21 (or `nvm` or `fnm`)
-- pnpm v9
+- Node.js >=22 (or `nvm` or `fnm`)
+- pnpm >=10.16.0
```
jdolle, line 8: should we mention using corepack? Especially since now it's not installed automatically with npm

jdolle (approved):

docs/DEVELOPMENT.md
jonathanawesome: alternatively, we could remove the mention of any node version manager with the assumption that they're all valid and we don't have a recommendation.

PR #7233, opened 2025-11-06
improvement: documentation for seed script

comatory:
### Background

It was not clear to me what the `TARGET` env variable actually stood for. I also expanded what `access token` is (so people would not confuse them with CDN access tokens).

### Description

I kept hitting a wall with trying to seed the data, always using `<org_name>/<project_name>/<target_name>` since that's what feels natural, as that's the format used with Hive CLI. But the seed script actually expects "Resource ID" passed as `TARGET` variable. I added some docs to make it more clear.


jdolle (approved):
Thanks for clarifying this.

comatory:
I changed formatting a bit, it looked weird when I previewed it 😅

comatory:
Fixed formatting (prettier).

n1ru4l (approved):

PR #7236, opened 2025-11-08
Tailwind v3 clean up

jonathanawesome:
### Background

Newly cloned console and had a number of warnings/errors in VSCode.

### Description

- Updates `.vscode/settings.json` to associate `.css` files with TailwindCSS, avoiding `Unknown @ rule` warnings.
- Converts `tailwind.config.ts` to ESM + typesafe config

### DIFF
--- a/packages/web/app/tailwind.config.ts
+++ b/packages/web/app/tailwind.config.ts
@@ -147,7 +147,7 @@ const config: Config = {
         sm: 'calc(var(--radius) - 4px)',
         xs: 'calc(var(--radius) - 6px)',
       },
-      ringColor: (theme: any) => ({
+      ringColor: ({ theme }) => ({
         DEFAULT: theme('colors.orange.500/75'),
         ...theme('colors'),
       }),


n1ru4l (approved):
Looks good with me, @hasparus please confirm, I am a noob with tailwind :D

packages/web/app/tailwind.config.ts
```diff
@@ -145,78 +147,78 @@ module.exports = {
         sm: 'calc(var(--radius) - 4px)',
         xs: 'calc(var(--radius) - 6px)',
       },
-      ringColor: theme => ({
+      ringColor: (theme: any) => ({
```
hasparus: We should be able to destructure this instead of using `as any`

```suggestion
      ringColor: ({ theme }) => ({
```

packages/web/app/tailwind.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'tailwindcss/lib/util/flattenColorPalette' {
```
hasparus, line 1: We already have this

```ts
declare module 'tailwindcss/lib/util/flattenColorPalette' {
  export default function flattenColorPalette(
    pallette: Record<string, string>,
  ): Record<string, string>;
}
```

in `packages/web/docs/src/deps.d.ts`

Maybe we could have one file with Tailwind type module augmentation included in both projects in `web`?

packages/web/app/tailwind.config.ts
hasparus: The previous syntax worked because it's a circular `theme === theme.theme` for backcompat.

Image

PR #7249, opened 2025-11-10
Console 1509 change close to discard in create token flow

jonathanawesome:
### Background

Clean up copy in the alert that presents when a user is creating an access token and attempts to close/abandon the shelf/sheet.

### Description

This PR not only updates the copy in this alert, but also DRYs up its usage. 

In addition, this PR prunes through our `ui` components to improve tree-shaking (imports clean up) and readability (adds a type inference helper to clean up forwardRef typings). _Exports in these ui component files are left untouched, so consuming component behavior does not change_.

### DIFF
--- a/.changeset/fifty-shirts-strive.md
+++ b/.changeset/fifty-shirts-strive.md
@@ -1,0 +1,7 @@
+---
+'@graphql-hive/envelop': patch
+'@graphql-hive/yoga': patch
+'@graphql-hive/apollo': patch
+---
+
+Send correct version of the client packages, previously they were sending the version of `@graphql-hive/core` package.
--- a/.changeset/five-spiders-serve.md
+++ b/.changeset/five-spiders-serve.md
@@ -1,0 +1,17 @@
+---
+'@graphql-hive/core': patch
+---
+
+Allow to provide `version` to `AgentOptions` in Hive Client integrations.
+
+```ts
+createHive({
+    agent: {
+        name: 'my-custom-agent',
+        version: '1.2.3', // new field
+    },
+})
+```
+
+Currently you can provide `name` but not `version`. This change allows to provide both `name` and `version` to better identify the clients connecting to Hive Console.
+Previously the dependent libraries like Yoga, Envelop and Hive Gateway integrations were incorrectly sending their names with the version of `@graphql-hive/core` package. Now they will be able to send their own versions.
--- a/.changeset/spotty-beans-report.md
+++ b/.changeset/spotty-beans-report.md
@@ -1,0 +1,6 @@
+---
+'@graphql-hive/core': patch
+---
+
+Accept `name` and `version` to `createSupergraphSDLFetcher` to build a more accurate user agent
+instead of passing `hive-client` with `@graphql-hive/core`'s version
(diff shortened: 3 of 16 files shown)


packages/web/app/src/lib/utils.ts
```diff
@@ -18,3 +24,10 @@ export function exhaustiveGuard(_value: never): never {
     `Reached forbidden guard function with unexpected value: ${JSON.stringify(_value)}`,
   );
 }
+
+// typescript utils
+
+/** this type inference helper cleans up some of our ui component code */
+export function createForwardRefComponent<T extends ElementType>(_Component: T) {
+  return forwardRef<ElementRef<T>, ComponentPropsWithoutRef<T>>;
```
jdolle: this being runtime is messing with my head.
Why not use a generic type definition like:
```
type ForwardRefComponent<T> = typeof forwardRef<ElementRef<T>, ComponentPropsWithoutRef<T>>
```
^ assuming i didnt mess up the typing... but you get what i mean

packages/web/app/src/components/organization/settings/access-tokens/access-tokens-sub-page.tsx
```diff
@@ -99,32 +99,12 @@ export function AccessTokensSubPage(props: AccessTokensSubPageProps): React.Reac
...
-                  onClick={() => setCreateAccessTokenState(CreateAccessTokenState.closed)}
-                >
-                  Close
-                </AlertDialog.AlertDialogAction>
-              </AlertDialog.AlertDialogFooter>
-            </AlertDialog.AlertDialogContent>
-          </AlertDialog.AlertDialog>
+          <DiscardAccessTokenDraft
+            continueFn={() => setCreateAccessTokenState(CreateAccessTokenState.open)}
+            discardFn={() => setCreateAccessTokenState(CreateAccessTokenState.closed)}
```
jdolle: minor: I think our convention would be `onDiscard` and `onContinue`

packages/web/app/src/components/ui/dropdown-menu.tsx
```diff
@@ -1,23 +1,44 @@
...
+  Trigger,
+} from '@radix-ui/react-dropdown-menu';
 
-const DropdownMenu = DropdownMenuPrimitive.Root;
-const DropdownMenuTrigger = DropdownMenuPrimitive.Trigger;
-const DropdownMenuGroup = DropdownMenuPrimitive.Group;
-const DropdownMenuPortal = DropdownMenuPrimitive.Portal;
-const DropdownMenuSub = DropdownMenuPrimitive.Sub;
-const DropdownMenuRadioGroup = DropdownMenuPrimitive.RadioGroup;
+const DropdownMenu = Root;
```
jdolle: could technically use `import as` to remove these lines

jdolle (commented):
I like the component reuse and type cleanup, but i'm not totally sure about it being a function rather than a type.

n1ru4l:
I would propose to first do a pull request for deduplicating the modal only, then we can discuss other changes introduced here.

packages/web/app/src/lib/utils.ts
jonathanawesome: great note, unfortunately typescript needs the value to infer from. removing this anyway and reverting DRY cleanup.

PR #7259, opened 2025-11-12
feat(js-sdk): circuit breaker for usage reporting

n1ru4l:
### Description

Support circuit breaking for usage reporting.

Circuit breaking is a fault-tolerance pattern that prevents a system from repeatedly calling a failing service. When errors or timeouts exceed a set threshold, the circuit “opens,” blocking further requests until the service recovers.

This ensures that during a network issue or outage, the service using the Hive SDK remains healthy and is not overwhelmed by failed usage reports or repeated retries.

```ts
import { createClient } from "@graphql-hive/core"

const client = createClient({
  agent: {
    circuitBreaker: {
      /**
       * Count of requests before starting evaluating.
       * Default: 5
       */
      volumeThreshold: 5,
      /**
       * After which time a request should be treated as a timeout in milleseconds
       * Default: 5_000
       */
      timeout: 5_000,
      /**
       * Percentage of requests failing before the circuit breaker kicks in.
       * Default: 50
       */
      errorThresholdPercentage: 1,
      /**
       * After what time the circuit breaker is attempting to retry sending requests in milliseconds
       * Default: 30_000
       */
      resetTimeout: 10_000,
    },
  }
})
```

Feedback I am looking for:
- How do you feel about the default values; are they too aggressive or to loose?
- Is there a need to provide hooks for hooking into the breaker event phases? E.g. so this info can be used for prometheus, otel metrics etc.

### DIFF
--- a/packages/libraries/core/src/client/agent.ts
+++ b/packages/libraries/core/src/client/agent.ts
@@ -320,15 +320,9 @@ export function createAgent<TEvent>(
     try {
       return await breaker.fire(...args);
     } catch (err: unknown) {
-      if (err instanceof Error && 'code' in err) {
-        if (err.code === 'EOPENBREAKER') {
-          breakerLogger.info('circuit open - sending report skipped');
-          return null;
-        }
-        if (err.code === 'ETIMEDOUT') {
-          breakerLogger.info('circuit open - sending report aborted - timed out');
-          return null;
-        }
+      if (err instanceof Error && 'code' in err && err.code === 'EOPENBREAKER') {
+        breakerLogger.info('circuit open - sending report skipped');
+        return null;
       }
 
       throw err;


packages/libraries/core/playground/agent-circuit-breaker.ts
n1ru4l, line 1: It is a bit hard to unit test this functionality, so I added this file that I can play around with the values for the circuit breaker.

I encourage reviewers to do the same to get an understanding on how this works.

packages/libraries/core/src/client/agent.ts
```diff
@@ -143,6 +181,28 @@ export function createAgent<TEvent>(
...
+        retries: options.maxRetries,
+        factor: 2,
+      },
+      logger: options.logger,
+      fetchImplementation: pluginOptions.fetch,
+      signal,
+    });
+  }
+
   async function send(sendOptions?: {
```
n1ru4l, line 208: We are basically doing the following:

```
circuit breaker ( # shuts down if underlying fails or times out aggressively
  http client ( # calls fetch and retries if non okay response
      fetch( # good old fetch
          # the payload we want to send
       )
   )
)
```

packages/libraries/core/src/client/http-client.ts
```diff
@@ -135,6 +142,13 @@ export async function makeFetchCall(
         throw new Error(`Unexpected HTTP error. (x-request-id=${requestId})`, { cause: error });
       });
 
+      if (config.signal?.aborted === true) {
+        const error = config.signal.reason ?? new Error('Request aborted externally.');
+        // TODO: maybe log some message?
```
n1ru4l: not needed, since the other layers already log it.
```suggestion
```

packages/libraries/core/src/client/http-client.ts
```diff
@@ -117,6 +120,8 @@ export async function makeFetchCall(
         body: config.body,
         headers: {
           'x-request-id': requestId,
+          'x-client-timestamp': new Date().toISOString(),
```
ardatan: You can use this maybe?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Date

n1ru4l: Decided to remove it for now after more discussion.

PR #7264, opened 2025-11-13
feat(hive-sdk): reduce http retry noise by introducing debug log level

n1ru4l:
### Background

We currently log all kind of things related to retries as errors. This can be perceived as noise, and we should only log these errors when all the retries fail.

### Description

Logging is still sub-optimal. I also considered introducing `@graphql-hive/logger`, but I would rather do that in a major release.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/libraries/core/src/client/http-client.ts
+++ b/packages/libraries/core/src/client/http-client.ts
@@ -104,6 +104,7 @@ export async function makeFetchCall(
 
   return await asyncRetry(
     async (bail, attempt) => {
+      const isFinalAttempt = attempt > retries;
       const requestId = crypto.randomUUID();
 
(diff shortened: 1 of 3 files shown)


packages/libraries/core/src/client/http-client.ts
n1ru4l, line 148: We should probably also log these as debug, unless it is the "last" attempt.

PR #7266, opened 2025-11-13
Hive Gateway Router Runtime renamed to Rust Query Planner

enisdenjo:
### TODO
- [ ] https://github.com/graphql-hive/federation-gateway-audit/pull/225

### DIFF
--- a/packages/web/docs/src/content/gateway/other-features/rust-query-planner.mdx
+++ b/packages/web/docs/src/content/gateway/other-features/rust-query-planner.mdx
@@ -26,7 +26,7 @@ npm install @graphql-hive/router-runtime
 The package is named `@graphql-hive/router-runtime` because it is the runtime integrating into Hive
 Gateway using the Rust Query Planner and a purpose-built executor for Hive Gateway.
 
-<Callout>
+</Callout>
 
 Then, configure your Hive Gateway to use the Rust Query Planner by updating your gateway's
 configuration:


packages/web/docs/src/content/gateway/other-features/rust-query-planner.mdx
```diff
@@ -0,0 +1,153 @@
...
+```sh npm2yarn
+npm install @graphql-hive/router-runtime
+```
+
+<Callout type="info" emoji="ℹ️">
+
+The package is named `@graphql-hive/router-runtime` because it is the runtime integrating into Hive
+Gateway using the Rust Query Planner and a purpose-built executor for Hive Gateway.
+
+<Callout>
```
kamilkisiela:
```suggestion
</Callout>
```

packages/web/docs/src/content/gateway/other-features/rust-query-planner.mdx
```diff
@@ -0,0 +1,153 @@
...
+| Subscriptions                | ✅         | ⚠️ Limited support    | Cannot populate fields from other subgraphs when resolving the subscription                             |
+| Schema extensions            | ✅         | ⚠️ Limited support    | Schema-level modifications may be limited because hive router does not use an executable schema         |
+| Custom plugins               | ✅         | ⚠️ No stitching hooks | All plugins, except those using [stitching hooks](#no-javascript-query-planner-plugin-hooks), will work |
+| Envelop plugins              | ✅         | ✅                    | All of envelop plugins will work                                                                        |
+| Yoga plugins                 | ✅         | ✅                    | All of Yoga plugins will work                                                                           |
+| Gateway plugins              | ✅         | ✅                    | All gateway plugins will work                                                                           |
```
EmrysMyrddin, line 102: Not entirely compatible. As you pointed out a few lines ago, plugins relying on stitching hooks will not entirely work.

But not sure how much it is worth mentioning. Most plugins don't use them. When they use them it's for monitoring purposes (except if I'm missing a use case?)

packages/web/docs/src/content/gateway/other-features/rust-query-planner.mdx
```diff
@@ -0,0 +1,153 @@
...
+| Persisted documents          | ✅         | ✅                    | Operation allow-listing and security                                                                    |
+| Request plugins              | ✅         | ✅                    | Request-level processing and modification                                                               |
+| Response plugins             | ✅         | ✅                    | Response-level processing and modification                                                              |
+
+## No JavaScript Query Planner Plugin Hooks
+
+All plugin hooks will work with the Rust Query Planner except for those specific to stitching that
+will never work because the runtime is different (Rust Query Planner vs. JavaScript Query Planner).
+
+Those hooks are only the following two hooks:
```
EmrysMyrddin, line 126: Perhaps we can also list the plugins we maintain that needs those hooks? And for each explain how it will affect its behavior?

enisdenjo: There are no plugins, aside from debugging, that use those hooks afaik. @ardatan correct me if I am wrong?

packages/web/docs/src/content/gateway/other-features/rust-query-planner.mdx
enisdenjo: It's good being transparent, given how many things are supported and that it's only a warning stating exactly that stitching hooks dont work - I'd keep them tbh.

packages/web/docs/src/content/gateway/other-features/rust-query-planner.mdx
ardatan: You are right! @enisdenjo Currently we don't have a plugin using that except debugging since they are expensive.

PR #7267, opened 2025-11-14
Upgrade graphql-inspector to v7; update change models

jdolle:
### Background

We recently added a bunch of functionality to graphql-inspector and did a major release.

- [X] Requires https://github.com/graphql-hive/graphql-inspector/pull/2916 first.

### Description

This adds the new change models and upgrades the inspector package.


packages/services/storage/src/schema-change-model.ts
```diff
@@ -354,7 +383,12 @@ export const DirectiveArgumentAddedModel = implement<DirectiveArgumentAddedChang
   meta: z.object({
     directiveName: z.string(),
     addedDirectiveArgumentName: z.string(),
-    addedDirectiveArgumentTypeIsNonNull: z.boolean(),
+    /** Default for backwards compatibility, but it breaks the typing. @todo fix typing */
+    addedDirectiveArgumentTypeIsNonNull: z.boolean().default(false) as any as ZodBoolean,
+    addedDirectiveArgumentType: z.string().default('') as any as ZodString,
+    addedToNewDirective: z.boolean().default(false) as any as ZodBoolean,
```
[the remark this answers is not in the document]
jdolle: This is a good idea. Can you submit these as suggested changes?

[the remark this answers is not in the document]
jdolle: Actually, I tested this and it still had type errors.

packages/services/storage/src/schema-change-model.ts
```diff
@@ -506,17 +563,9 @@ export const DirectiveUsageArgumentDefinitionAddedModel =
       fieldName: z.string(),
       argumentName: z.string(),
       addedDirectiveName: z.string(),
-    }),
-  });
-
-export const DirectiveUsageArgumentDefinitionModel =
```
jdolle, line 512: i couldnt find that this was ever emitted from inspector.

.changeset/loose-dots-learn.md
```diff
@@ -0,0 +1,7 @@
+---
+'@graphql-hive/cli': minor
+'hive': minor
+---
+
+Upgrade graphql-inspector/core to v7 and update the models to be able to handle the new change
+objects
```
n1ru4l: maybe worth mentioning that this now supports showing directive changes?

packages/services/storage/src/schema-change-meta.ts
```diff
@@ -195,6 +199,10 @@ export function schemaChangeFromSerializableChange(
       return directiveUsageInputFieldDefinitionAddedFromMeta(change);
     case ChangeType.DirectiveUsageInputFieldDefinitionRemoved:
       return directiveUsageInputFieldDefinitionRemovedFromMeta(change);
+    case ChangeType.DirectiveUsageArgumentAdded:
+      return directiveUsageArgumentAddedFromMeta(change);
+    case ChangeType.DirectiveUsageArgumentRemoved:
+      return directiveUsageArgumentRemovedFromMeta(change);
```
n1ru4l, line 205: Let's add integration tests for these new added types?

packages/services/storage/src/schema-change-model.ts
```diff
@@ -315,10 +326,31 @@ export const DirectiveRemovedModel = implement<DirectiveRemovedChange>().with({
...
+    directiveName: z.string(),
+  }),
+});
+
 export const DirectiveAddedModel = implement<DirectiveAddedChange>().with({
   type: DirectiveAddedLiteral,
   meta: z.object({
     addedDirectiveName: z.string(),
+    // for backwards compatibility
+    addedDirectiveRepeatable: z.boolean().default(false) as unknown as ZodBoolean, // boolean;
```
n1ru4l: Why is the cast here needed?

jdolle: The typing for our Zod schema defaults is strange.

I would expect a `ZodBoolean` for `z.boolean().default(false)` but it outputs a `ZodOptional<ZodBoolean>` which types the result as `boolean | undefined` -- which isn't how defaults work!

Using `.default`, `.transform`, or even `.precast` modifies the input so that we're guaranteed that value in the parsed result. I haven't been able to find a better way to fix this.

packages/services/storage/src/schema-change-meta.ts
jdolle: added

packages/services/storage/src/schema-change-model.ts
```diff
@@ -1230,9 +1366,9 @@ export const HiveSchemaChangeModel = z
 
         if (
           isInputFieldAddedChange(rawChange) &&
-          rawChange.meta.isAddedInputFieldTypeNullable === false
+          change.meta.isAddedInputFieldTypeNullable === false
         ) {
-          breakingChangeSchemaCoordinate = rawChange.meta.inputName;
```
jdolle, line 1235: I still need to check if we need to adjust the breaking change coordinate for any other change because of the adjustments to `path` in v7

jdolle: I reviewed the changes to inspector and the only paths that got changed were for the deprecation changes so nothing needs changed in relation to the `breakingChangeSchemaCoordinate`

integration-tests/tests/api/policy/policy-check.spec.ts
```diff
@@ -323,13 +323,12 @@ describe('Schema policy checks', () => {
...
         '   - Deprecation reason is required for field foo in type Query. (source: policy-require-deprecation-reason)',
         '',
-        'ℹ Detected 4 changes',
+        'ℹ Detected 3 changes',
         '',
         '   Safe changes:',
         '   - Type User was added',
         '   - Field user was added to object type Query',
         '   - Field Query.foo is deprecated',
-        '   - Directive deprecated was added to field Query.foo',
```
jdolle, line 332: Similarly to new types excluding their new fields on changes -- a new deprecation now excludes this redundant change.

packages/services/storage/src/schema-change-model.ts
```diff
@@ -354,7 +386,12 @@ export const DirectiveArgumentAddedModel = implement<DirectiveArgumentAddedChang
   meta: z.object({
     directiveName: z.string(),
     addedDirectiveArgumentName: z.string(),
-    addedDirectiveArgumentTypeIsNonNull: z.boolean(),
+    /** Default for backwards compatibility, but it breaks the typing. @todo fix typing */
+    addedDirectiveArgumentTypeIsNonNull: z.boolean().default(false) as any as ZodBoolean,
+    addedDirectiveArgumentType: z.string().default('') as any as ZodString,
+    addedToNewDirective: z.boolean().default(false) as any as ZodBoolean,
```
n1ru4l: Lets be consistent with other usages and use `unknown` instead of `any`?

packages/services/storage/src/schema-change-model.ts
n1ru4l: Try this type, instead of the other one, I think it will remove the need for the casts

```
type Implements<Model> = {
  [K in keyof Model]-?: z.ZodTypeAny & { _output: Model[K] };
};
```

jdolle: worked like a charm

n1ru4l:
@jdolle This already introduces several improvements to our current change detection, do you think it is worth to have a follow-up PR with a product update highlighting these improvements?

PR #7269, opened 2025-11-14
docs(js-sdk): document agent and logger configuration options

n1ru4l:
### Description

People asked for more information on configuring the usage reporting agent.

Related:
- https://github.com/graphql-hive/console/pull/7264
- https://github.com/graphql-hive/console/pull/7259

### DIFF
--- a/packages/web/docs/src/content/api-reference/client.mdx
+++ b/packages/web/docs/src/content/api-reference/client.mdx
@@ -465,7 +465,7 @@ useHive({
 
 You can customize the logging level or provide your own (JSON) logger implementation.
 
-#### `logger`
+##### `logger`
 
 - Type: [`Logger`](/docs/logger) or log level string
 - Default: `info`


packages/libraries/core/src/client/agent.ts
```diff
@@ -56,6 +56,7 @@ export interface AgentOptions {
   maxRetries?: number;
   /**
    * 200 by default
```
EmrysMyrddin: I would even remove the description if it has no effect at all today.

EmrysMyrddin:
Gemini's comments are pretty good for once!

packages/libraries/core/src/client/agent.ts
```diff
@@ -56,6 +56,7 @@ export interface AgentOptions {
...
+   * @deprecated This is no longer used.
```
dotansimha: nit: Do we have an alternative field or something related? (if so, worth documenting it)

PR #7275, opened 2025-11-17
[OBSOLETE] feat: add redpanda buffering layer with split ingress/egress otel collectors

adambenhassen:
### Description

Implements a resilient OTLP trace pipeline architecture by splitting the monolithic OTEL collector into separate ingress and egress components with Redpanda as a buffering layer. This provides better scalability, fault tolerance, and performance isolation.

### Architecture Changes
- **Before:** Client -> OTEL Collector -> ClickHouse
- **After:** Client -> Ingress Collectors -> Redpanda -> Egress Collectors -> ClickHouse

#### Split Collector Architecture
  - Ingress Collectors: Handle authentication, receive OTLP data, forward to Redpanda
    - Includes HiveAuth extension for authentication
    - Processes attributes and applies memory limits
    - Exports to Kafka with Snappy compression
  - Egress Collectors: Consume from Redpanda, batch, send to ClickHouse
    - No authentication needed (internal only)
    - Batch processing for efficient ClickHouse inserts
    - Dedicated to ClickHouse export operations

#### Redpanda Message Buffer
  - Acts as durable buffer between ingress and egress
  - Provides backpressure handling and fault tolerance
  - 10 partitions for parallel processing
  - 30-day retention for trace replay capability
  - 10MB max message size for large traces

### DIFF
--- a/deployment/services/otel-collector.ts
+++ b/deployment/services/otel-collector.ts
@@ -5,6 +5,7 @@ import { DbMigrations } from './db-migrations';
 import { Docker } from './docker';
 import { Environment } from './environment';
 import { GraphQL } from './graphql';
+import { Redpanda } from './redpanda';
 
 export type OTELCollector = ReturnType<typeof deployOTELCollector>;
 
@@ -15,9 +16,12 @@ export function deployOTELCollector(args: {
   clickhouse: Clickhouse;
   dbMigrations: DbMigrations;
   graphql: GraphQL;
+  redpanda: Redpanda;
 }) {
-  return new ServiceDeployment(
-    'otel-collector',
+  const kafkaBroker = args.redpanda.brokerEndpoint;
+
+  const ingress = new ServiceDeployment(
+    'otel-collector-ingress',
     {
       image: args.image,
       imagePullSecret: args.docker.secret,
@@ -26,6 +30,7 @@ export function deployOTELCollector(args: {
         HIVE_OTEL_AUTH_ENDPOINT: serviceLocalEndpoint(args.graphql.service).apply(
           value => value + '/otel-auth',
         ),
+        KAFKA_BROKER: kafkaBroker,
       },
       /**
        * We are using the healthcheck extension.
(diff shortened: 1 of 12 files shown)


deployment/services/otel-collector.ts
```diff
@@ -40,11 +46,40 @@ export function deployOTELCollector(args: {
       pdb: true,
       availabilityOnEveryNode: true,
       port: 4318,
-      memoryLimit: args.environment.podsConfig.tracingCollector.memoryLimit,
+      memoryLimit: '512Mi',
```
jdolle: should we keep this configurable by env?

deployment/services/redpanda.ts
```diff
@@ -0,0 +1,198 @@
...
+              resources: {
+                limits: {
+                  cpu: cpuLimit,
+                  memory: memoryLimit,
+                },
+              },
+              args: [
+                'redpanda',
+                'start',
+                '--overprovisioned',
```
jdolle: > Redpanda expects a dev system to be an overprovisioned environment. Based on a [Seastar option](https://docs.seastar.io/master/structseastar_1_1reactor%5F%5Foptions.html#a0caf6c2ad579b8c22e1352d796ec3c1d), setting overprovisioned disables thread affinity, zeros idle polling time, and disables busy-poll for disk I/O.

https://docs.redpanda.com/current/reference/rpk/rpk-redpanda/rpk-redpanda-mode/#production-mode

We want to make sure this is disabled in production

deployment/services/redpanda.ts
```diff
@@ -0,0 +1,198 @@
...
+              ],
+              volumeMounts: [
+                {
+                  name: 'datadir',
+                  mountPath: '/var/lib/redpanda/data',
+                },
+              ],
+              livenessProbe: {
+                httpGet: {
+                  path: '/v1/status/ready',
```
jdolle: for most of our services we use `_health`. I'm not sure if we want to keep this consistent or if we don't mind...

deployment/services/redpanda.ts
```diff
@@ -0,0 +1,198 @@
...
+  // Headless Service for StatefulSet (used for internal cluster communication)
+  const headlessService = new k8s.core.v1.Service('redpanda-headless', {
+    metadata: {
+      name: 'redpanda',
+    },
+    spec: {
+      clusterIP: 'None',
+      selector: labels,
+      ports: [
+        { name: 'kafka', port: 9092, targetPort: 9092 as any },
```
jdolle: why do we need the any casts ?

docker/configs/otel-collector/config-ingress.yaml
```diff
@@ -61,7 +52,7 @@ service:
     - health_check
   telemetry:
     logs:
-      level: DEBUG
+      level: INFO
```
jdolle, line 85: Is this just for testing or is there a reason to use info logs?

adambenhassen: The kafka exporter prints the whole string to console when level is set to DEBUG, during my load tests, it bogged down the process, since the payloads weighed 3.6mb each request.

deployment/services/redpanda.ts
adambenhassen: Right. I remember having some TS errors during the configuration, so I may have cleared the buggy intellisense cache with the any cast 😄

deployment/services/otel-collector.ts
adambenhassen: Right. I have to review the configurable environments.

deployment/services/redpanda.ts
adambenhassen: Redpanda does not expose  `/_health` endpoint

PR #7280, opened 2025-11-17
feat(js-sdk): circuit breaker for usage reporting (attempt 2)

n1ru4l:
### Background

Another attempt for the circuit breaker (originally https://github.com/graphql-hive/console/pull/7259), after I reverted it due to horrible developer experience and not wanting to maintain a page towards manually bundling `opossum` (circuit breaker) into your server that uses the hive-sdk.

I basically inlined `opposum` source code in the SDK, added `events` npm package, so nobody dares to import the forbidden `node:events`, and also converted everything to ESM, so TypeScript does not complain.

I did not bother converting `opposum` to TypeScript, I am fine with it being a follow up PR or starting a revolution in the `opposum` library to use a cross platform `EventEmitter` implementation.

For now I want to as make it as easy and fast as possible help developers to circuit break usage reports when needed.

### Description

Support circuit breaking for usage reporting.

Circuit breaking is a fault-tolerance pattern that prevents a system from repeatedly calling a failing service. When errors or timeouts exceed a set threshold, the circuit “opens,” blocking further requests until the service recovers.

This ensures that during a network issue or outage, the service using the Hive SDK remains healthy and is not overwhelmed by failed usage reports or repeated retries.

```ts
import { createClient } from "@graphql-hive/core"

const client = createClient({
  agent: {
    circuitBreaker: {
      /**
       * Count of requests before starting evaluating.
       * Default: 5
       */
      volumeThreshold: 5,
      /**
       * After which time a request should be treated as a timeout in milleseconds
       * Default: 5_000
       */
      timeout: 5_000,
      /**
       * Percentage of requests failing before the circuit breaker kicks in.
       * Default: 50
       */
      errorThresholdPercentage: 1,
      /**
       * After what time the circuit breaker is attempting to retry sending requests in milliseconds
       * Default: 30_000
       */
      resetTimeout: 10_000,
    },
  }
})
```

### DIFF
--- a/packages/libraries/core/src/circuit-breaker/README.md
+++ b/packages/libraries/core/src/circuit-breaker/README.md
@@ -1,4 +1,4 @@
 # Readme
 
-This is https://github.com/nodeshift/opossum, but we use `events` from `npm`, to make it work on any
-platform :)
+This is https://github.com/nodeshift/opossum, but we use `events` from `npm`, and with esm exports
+to make it work on any platform :)


packages/libraries/core/playground/agent-circuit-breaker.ts
```diff
@@ -0,0 +1,53 @@
...
+      set(d) {
+        data.push(d);
+      },
+    },
+  },
+);
+
+setInterval(() => {
+  agent.capture({});
+  console.log('BURR BURT');
```
jdolle: ?

ardatan (approved):

PR #7284, opened 2025-11-18
fix pagination and add filter to Organization.members

jdolle:
### Background

Ref [CONSOLE-1521](https://linear.app/the-guild/issue/CONSOLE-1521/member-pagination-and-search-api)

### Description

Fixes `Organization.members` pagination by adding an order by `created_at` so that the last record in the array is the oldest, and adds a `searchTerm` argument, which filters the list by matching on the username and email.

### Checklist

- [ ] Testing


packages/services/api/src/modules/organization/providers/organization-members.ts
```diff
@@ -178,11 +188,12 @@ export class OrganizationMembers {
               `
             : sql``
         }
+        ${searching ? sql`AND to_tsvector("u"."display_name" || ' ' || "u"."email") @@ to_tsquery('simple', ${searchTerm + ':*'})` : sql``}
       ORDER BY
         "om"."organization_id" DESC
+        , "om"."created_at" DESC
```
jdolle, line 194: necessary or the order doesn't match the cursor's WHERE clause. Basically, without this the last item in the array may have a cursor that is less than another element's based on:
```sql
AND (
                  "om"."created_at" < ${cursor.createdAt}
                  OR (
                    "om"."created_at" = ${cursor.createdAt}
                    AND "om"."user_id" < ${cursor.id}
                  )
                )
```

packages/services/api/src/modules/organization/providers/organization-members.ts
```diff
@@ -178,11 +188,12 @@ export class OrganizationMembers {
               `
             : sql``
         }
+        ${searching ? sql`AND to_tsvector("u"."display_name" || ' ' || "u"."email") @@ to_tsquery('simple', ${searchTerm + ':*'})` : sql``}
```
jdolle, line 191: this happens after the pagination. So even though it isn't indexed and so it isnt very performant -- running on <= 50 elements is fine.

packages/services/api/src/modules/organization/providers/organization-members.ts
```diff
@@ -178,11 +188,12 @@ export class OrganizationMembers {
...
       ORDER BY
         "om"."organization_id" DESC
+        , "om"."created_at" DESC
         , "om"."user_id" DESC
-        , "om"."user_id" DESC
-      ${first ? sql`LIMIT ${first + 1}` : sql``}
+      LIMIT ${first + 1}
```
jdolle, line 196: This enforces the pagination. Whereas before it was optional. This is critical to avoid performance issues from the filter.

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -881,8 +889,11 @@ export default gql`
     name: String! @deprecated(reason: "Use the 'slug' field instead.")
     owner: Member! @tag(name: "public")
     me: Member!
-    members(first: Int @tag(name: "public"), after: String @tag(name: "public")): MemberConnection!
-      @tag(name: "public")
+    members(
+      first: Int @tag(name: "public")
+      after: String @tag(name: "public")
+      filters: MembersFilter
```
jdolle, line 895: left out of public API until we solidify this pattern

jdolle:
Closing. This has been merged into the frontend branch

PR #7287, opened 2025-11-18
feat(js-sdk): circuit breaker and cdn mirror for CDN apis

n1ru4l:
### Description

- Supports providing a secondary endpoint for persisted documents and supergraph polling
- Introduces circuit breaker for persisted documents and supergraph polling
- introduces new cdn artifact polling implementation, that comes with circuit breaker and cleanup (and is not only for polling the supergraph but all kinds of artifacts like the services list)

__

- See changesets (and typedefs) for API changes
- Documentation will be shipped as part of https://github.com/graphql-hive/console/pull/7269

### Checklist

- [x] Testing

### DIFF
--- a/.changeset/calm-mails-sneeze.md
+++ b/.changeset/calm-mails-sneeze.md
@@ -1,9 +1,7 @@
 ---
 '@graphql-hive/core': minor
 '@graphql-hive/apollo': minor
-'@graphql-hive/apollo': minor
 '@graphql-hive/yoga': minor
-
 ---
 
 **Persisted Documents Improvements**


packages/libraries/apollo/src/index.ts
```diff
@@ -17,34 +19,83 @@ export {
...
+
+/** @deprecated Use {CreateSupergraphManagerArgs} instead */
 export type { SupergraphSDLFetcherOptions } from '@graphql-hive/core';
 
-export function createSupergraphManager({
-  pollIntervalInMs,
-  ...superGraphFetcherOptions
-}: { pollIntervalInMs?: number } & SupergraphSDLFetcherOptions) {
-  pollIntervalInMs = pollIntervalInMs ?? 30_000;
-  const fetchSupergraph = createSupergraphSDLFetcher(superGraphFetcherOptions);
```
n1ru4l, line 27: Note: I introduced a new `CDNArtifactFetcher` that also has a cleanup method, which was needed due to the circuit breaker.

packages/libraries/core/src/client/cdn-artifact-fetcher.ts
```diff
@@ -0,0 +1,171 @@
...
+   * Example:
+   *
+   * ```
+   * [
+   *          "https://cdn.graphql-hive.com/artifacts/v1/9fb37bc4-e520-4019-843a-0c8698c25688/supergraph",
+   *   "https://cdn-mirror.graphql-hive.com/artifacts/v1/9fb37bc4-e520-4019-843a-0c8698c25688/supergraph"
+   * ]
+   * ```
+   */
+  endpoint: string | [string, string];
```
EmrysMyrddin, line 27: Not sure it is necessary to be this strict with the type :-) The code is already working for whatever array length :-) And it avoids casting when using `.map` or `.filter`!
```suggestion
  endpoint: string | string[];
```

.changeset/calm-mails-sneeze.md
```diff
@@ -0,0 +1,55 @@
...
+      accessToken: ''
+    },
+    circuitBreaker: {
+      // open circuit if 50 percent of request result in an error
+      errorThresholdPercentage: 50,
+      // start monitoring the circuit after 10 requests
+      volumeThreshold: 10,
+      // time before the backend is tried again after the circuit is open
+      resetTimeout: 30_000
+    }
```
EmrysMyrddin, line 52: Shouldn't circuit breaker options be under the `cdn` key ? Since it's the CDN's circuit breaker ?

packages/libraries/core/src/client/cdn-artifact-fetcher.ts
n1ru4l: Changed to `[string, ...string[]]`, which seems to make more sense

.changeset/calm-mails-sneeze.md
n1ru4l: I remember I introduced `accessToken` and `endpoint` under `cdn`, so it is obvious you need to provide credentials for the CDN.

Should `fetch` then also be under `cdn` then? I don't feel the strong need to move it there or not. For me it is just fine if it is a name space for credentials.

EmrysMyrddin: As you whish, I don't have strong opinion on it, I just found it surprising by reading the changelog :-)

Anyway, it is typed and documented, so not a big deal.

packages/libraries/core/src/client/cdn-artifact-fetcher.ts
EmrysMyrddin: Which still probably disallow the usage of `.map` and `.filter` ^^'

n1ru4l: https://www.typescriptlang.org/play/?jsx=0#code/MYewdgzgLgBApmAJgBxASzFAXDA2tAJwwHMAaGAOisJNwF06YBePAIgENW6BYAKD9CRYBOBACuAG1gsEKdJgoBbdsgAUAMxAhmAPhisIidodYBKIA

Must be missing something here?

PR #7288, opened 2025-11-18
Add bits to federated seed schema

jonathanawesome:
### Background
This PR updates the federated seed schema with some common bits that will help with local UI development.

### Description

- Notification union type
- various input types
- mutation fields
- descriptions galore
- multiple argument examples

### DIFF
--- a/scripts/seed-schemas/federated/inventory.graphql
+++ b/scripts/seed-schemas/federated/inventory.graphql
@@ -4,10 +4,7 @@ extend schema
     import: ["@key", "@shareable", "@external", "@requires"]
   )
 
-"""
-Product extension adding inventory and delivery information
-"""
-type Product @key(fields: "upc dimensions") {
+extend type Product @key(fields: "upc dimensions") {
   """
   Universal Product Code
   """
--- a/scripts/seed-schemas/federated/notifications.graphql
+++ b/scripts/seed-schemas/federated/notifications.graphql
@@ -131,10 +131,7 @@ extend type User @key(fields: "id") {
   notifications: [Notification!]!
 }
 
-"""
-Product stub reference (full definition in products subgraph)
-"""
-type Product @key(fields: "upc") {
+extend type Product @key(fields: "upc") {
   """
   Universal Product Code
   """


scripts/seed-schemas/federated/products.graphql
```diff
@@ -37,27 +40,141 @@ type Product implements ProductItf & SkuItf @key(fields: "upc") @key(fields: "sk
...
+  """
+  searchProducts(filter: ProductFilterInput!): [ProductItf!]!
+  """
+  Get products filtered by category and price range
+  """
+  products(category: String, minPrice: Float, maxPrice: Float): [ProductItf]
+  """
+  Get products filtered by physical dimensions and shipping class
+  """
+  productsByDimensions(minWeight: Float, maxWeight: Float, size: String, shippingClass: ShippingClass): [ProductItf]
```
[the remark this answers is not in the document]
jonathanawesome: we'll keep this as-is so we have multiple examples.

scripts/seed-schemas/federated/reviews.graphql
```diff
@@ -4,14 +4,47 @@ extend schema
...
+"""
+input CreateReviewInput {
+  """
+  Universal Product Code of the product being reviewed
+  """
+  productUpc: String!
+  """
+  ID of the user creating the review
+  """
+  userId: ID!
```
jdolle: typically this is gathered from the session.

scripts/seed-schemas/federated/reviews.graphql
```diff
@@ -26,3 +59,21 @@ extend type Product @key(fields: "upc") {
...
+
+type Query {
+  """
+  Get a specific review by ID
+  """
+  review(id: Int!): Review
+  """
+  Get reviews filtered by product and rating range
+  """
+  reviews(productUpc: String, minRating: Int, maxRating: Int): [Review]
```
jdolle, line 67: could technically have min and max rating be a filter input

scripts/seed-schemas/federated/users.graphql
```diff
@@ -2,19 +2,234 @@ extend schema
...
 
+"""
+ISO-8601 formatted date and time string
+"""
 scalar DateTime
 
+"""
+Types of notifications that can be sent to users
+"""
+enum NotificationType {
```
jdolle: notifications could be yet another subgraph?

jonathanawesome: great idea...incoming!

scripts/seed-schemas/federated/reviews.graphql
jonathanawesome: Still needs to be passed along with the mutation, no? Am I missing something?

***also, not a real api***

scripts/seed-schemas/federated/reviews.graphql
jonathanawesome: I have a few of these operations accepting multiple arguments even though it's not entirely sensical. We need examples of fields with many arguments so we can build UI around them.

scripts/seed-schemas/federated/reviews.graphql
jdolle: Definitely not a real API but i'm treating it as if it were.
Typically it would be passed via a header to the subgraph, or something like `@fromContext` might be used to populate it. I'm fine with it either way

jonathanawesome: understood...incoming!

PR #7290, opened 2025-11-19
feat: move logger property to top level instead of having it in agent object; use hive-logger

n1ru4l:
### Background

The logger is not only used in agent context, we can lift it UP!

### Description

- Deprecates `debug` boolean flag in favor of using a `logger` object
- Move `logger` option to top level instead of having it in the `agent` sub level (which is a bit confusing)
- Adds test to ensure backwards compatibility for people using the `debug` option

More information in the changeset!

### Checklist

- [x] Testing

### DIFF
--- a/packages/libraries/core/tests/usage.spec.ts
+++ b/packages/libraries/core/tests/usage.spec.ts
@@ -1,10 +1,16 @@
 import { buildSchema, parse } from 'graphql';
 import nock from 'nock';
+import { Logger, MemoryLogWriter } from '@graphql-hive/logger';
 import { createHive } from '../src/client/client';
 import { atLeastOnceSampler } from '../src/client/samplers';
 import type { Report } from '../src/client/usage';
 import { version } from '../src/version';
-import { createHiveTestingLogger, fastFetchError, waitFor } from './test-utils';
+import {
+  createHiveTestingLogger,
+  fastFetchError,
+  normalizeLogMessage,
+  waitFor,
+} from './test-utils';
 
 const headers = {
   'Content-Type': 'application/json',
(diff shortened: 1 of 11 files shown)


packages/libraries/core/tests/http-client.spec.ts
```diff
@@ -36,10 +35,8 @@ test('HTTP with retries and system', async () => {
 
   expect(logger.getLogs()).toMatchInlineSnapshot(`
     [DBG] GET https://ap.localhost.noop (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) Attempt (1/2)
-    [ERR] Error: getaddrinfo ENOTFOUND ap.localhost.noop
     [DBG] GET https://ap.localhost.noop (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) failed (666ms). getaddrinfo ENOTFOUND ap.localhost.noop
     [DBG] GET https://ap.localhost.noop (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) Attempt (2/2)
-    [ERR] Error: getaddrinfo ENOTFOUND ap.localhost.noop
```
jdolle, line 42: Is removing this intentional? I think it's useful to see the raw http client error?

n1ru4l: It is also within the line above, if you provide a hive logger you will also get the error as attributes.

packages/libraries/core/tests/usage.spec.ts
```diff
@@ -640,11 +640,11 @@ test('should not send excluded operation name data to Hive', async () => {
   http.done();
 
   expect(logger.getLogs()).toMatchInlineSnapshot(`
-    [DBG] [hive][usage][agent] Disposing
-    [DBG] [hive][usage][agent] Sending report (queue 2)
-    [DBG] [hive][usage][agent] POST http://localhost/200 (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx)
-    [DBG] [hive][usage][agent] POST http://localhost/200 (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) succeeded with status 200 (666ms).
-    [DBG] [hive][usage][agent] Report sent!
```
jdolle, line 647: minor: i prefer having a space after the tags for readability

packages/libraries/core/tests/http-client.spec.ts
jdolle: I completely missed that. Ignore this then.

packages/libraries/core/tests/usage.spec.ts
n1ru4l: I moved them to the attributes instead, since log prefixing is considered an anti-pattern anyways.

PR #7291, opened 2025-11-19
fix(traces): eliminate clickhouse query timeouts and improve read times

adambenhassen:
### Description
Fixes critical performance issues causing ClickHouse query timeouts on the traces dashboard when handling large datasets. Users with high-traffic targets were experiencing UI hangs and failed queries when viewing trace data.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/operations/providers/traces.ts
+++ b/packages/services/api/src/modules/operations/providers/traces.ts
@@ -45,8 +45,9 @@ export class Traces {
         WHERE
           "trace_id" IN (${sql.array(traceIds, 'String')})
         LIMIT 1 BY "trace_id"
+        SETTINGS max_threads = 8
       `,
-      timeout: 10_000,
+      timeout: 30_000,
       queryId: 'Traces.findTraceByTraceId',
     });
 
@@ -99,11 +100,13 @@ export class Traces {
           ${spanFields}
         FROM
           "otel_traces"
-        WHERE
+        PREWHERE
           "TraceId" = ${traceId}
-          AND "SpanAttributes"['hive.target_id'] = ${targetId}
+        WHERE
+          "SpanAttributes"['hive.target_id'] = ${targetId}
+        SETTINGS max_threads = 8
       `,
-      timeout: 10_000,
+      timeout: 30_000,
       queryId: 'Traces.findSpansForTraceId',
     });
 
@@ -174,11 +177,11 @@ export class Traces {
         const operator = sort?.direction === 'ASC' ? sql`>` : sql`<`;
         paginationSQLFragmentPart = sql`
           AND (
-            (
-               "timestamp" = ${cursor.timestamp}
-               AND "trace_id" < ${cursor.traceId}
+            "timestamp" ${operator} ${cursor.timestamp}
+            OR (
+              "timestamp" = ${cursor.timestamp}
+              AND "trace_id" < ${cursor.traceId}
             )
-            OR "timestamp" ${operator} ${cursor.timestamp}
           )
         `;
       }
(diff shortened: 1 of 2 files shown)


packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -163,49 +165,59 @@ export class Traces {
               "duration" = ${durationStr}
               AND "timestamp" < ${cursor.timestamp}
             )
-            OR (
-              "duration" = ${durationStr}
-              AND "timestamp" = ${cursor.timestamp}
-              AND "trace_id" < ${cursor.traceId}
```
jdolle, line 169: I know the odds of the duration and timestamp being the same seems minuscule, but  don't we need the traceId check to address this edge case?

n1ru4l: In theory two traces can have the same timestamp and could then be be skipped by pagination. 🤔

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -571,7 +604,6 @@ const traceFields = sql`
   , "http_url" AS "httpUrl"
   , "duration"
   , "graphql_operation_name" AS "graphqlOperationName"
-  , "graphql_operation_document" AS "graphqlOperationDocument"
```
n1ru4l, line 574: What is the reason for removing this field? Probably the size?

n1ru4l: Seems it was also unused 😓

packages/services/api/src/modules/operations/providers/traces.ts
adambenhassen: That's true, even if the odds are very small, the case of multiple traces having identical timestamps (and durations when sorting by duration) is very realistic. Good catch! 🚀

packages/services/api/src/modules/operations/providers/traces.ts
adambenhassen: Yes, it was large and most importantly - unused.

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -99,11 +100,13 @@ export class Traces {
           ${spanFields}
         FROM
           "otel_traces"
-        WHERE
+        PREWHERE
```
jdolle, line 103: nice. Clickhouse says this should be done automatically though. Do we know why we have to specify this?

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -99,11 +100,13 @@ export class Traces {
...
           "TraceId" = ${traceId}
-          AND "SpanAttributes"['hive.target_id'] = ${targetId}
+        WHERE
+          "SpanAttributes"['hive.target_id'] = ${targetId}
+        SETTINGS max_threads = 8
       `,
-      timeout: 10_000,
+      timeout: 30_000,
```
jdolle, line 109: Why increase?
Realistically we dont want to wait 30s. Most people would refresh by then and we'd just be wasting time fetching.

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -174,38 +177,59 @@ export class Traces {
...
+      PREWHERE
+        target_id = ${targetId}
+        ${prewhereTimestampFragment}
+      WHERE
+        true
+        ${paginationSQLFragmentPart}
+        ${filterSQLFragment}
+      ORDER BY
+        ${orderByFragment}
+      LIMIT ${sql.raw(String(limit + 1))}
```
jdolle, line 225: Does
```
sql`${limit+1}`
```
not work?

packages/services/api/src/modules/operations/providers/traces.ts
adambenhassen: ClickHouse's automatic optimization is conservative and skips PREWHERE when it sees complex expressions like map accesses (`SpanAttributes['hive.target_id']`) or compound pagination conditions. The explicit PREWHERE ensures the deterministic optimization behavior we expect.

packages/services/api/src/modules/operations/providers/traces.ts
adambenhassen: Good point. The timeout increase was added as a safety margin, but you're right that 30s is too long for a query. It's only there to make sure we render something at some point, regardless if the user focus is lost. I think it's harmless to have it increased.

packages/services/api/src/modules/operations/providers/traces.ts
adambenhassen:
```
Argument of type 'number' is not assignable to parameter of type 'ValueExpression'
```

Edit: Ah, I see. ```${sql`limit + 1`}``` did it. cool gimmick!

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -372,7 +409,7 @@ export class TraceBreakdownLoader {
       }>({
         query,
         queryId: 'traces_filter_options',
-        timeout: 10_000,
+        timeout: 60_000,
```
n1ru4l: 60 seconds is too much, as our api gateway will time out by then.

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -363,6 +399,7 @@ export class TraceBreakdownLoader {
 
       const query = sql`
         ${sql.join(statements, ' UNION ALL ')}
+        SETTINGS max_threads = 8
```
n1ru4l, line 402: What would be the behaviour without this statement? Does adding this limit the amount of allocated threads or increase the amount of allocated threads used for this query?

adambenhassen: Without max_threads = 8, ClickHouse may execute UNION ALL branches sequentially. This query can take up to 60 seconds or more because each branch (there's about 8 of them) waits for others to complete with huge datasets. From my testing, this has massively improved query time. So with this setting set to 8, Clickhouse allocates 8 threads to read columns in parallel across the UNION branch. This reduced read times down to single digit numbers.

packages/services/api/src/modules/operations/providers/traces.ts
adambenhassen: I can suggest lowering this down to 30 seconds, what do you think? Keep in mind that this query takes a lot of time depending on the dataset

adambenhassen:
> The changes look good to me!
> 
> Unfortunately I don't see a good way to "test" this, aside from running it against massive data. 🤔 Maybe another thing we could do as a follow up task is to ensure we always have a project on staging/testing that has massive amounts of data, where we can easily detect regressions?

Good idea! I can suggest adding a command to seed traces (pnpm seed-traces) to duplicate rows in clickhouse

packages/services/api/src/modules/operations/providers/traces.ts
adambenhassen: I had to revert this change. It breaks the code.

PR #7292, opened 2025-11-19
feat: aws lambda cdn handler

n1ru4l:
### Background

Anyone wants to run the CDN handler on AWS?

### Description

- Adds the AWS CDN handler 
- Adds pulumni configuration for deploying the AWS CDN handler with a invoke function

### DIFF
--- a/deployment/services/aws-artifacts-lambda-function.ts
+++ b/deployment/services/aws-artifacts-lambda-function.ts
@@ -47,7 +47,8 @@ export function deployAWSArtifactsLambdaFunction(args: {
     region: 'us-east-2',
     environment: {
       variables: {
-        // yeah this is illegal but I frist need to figure out how to use secret store or some stuff
+        // This could be done better with secrets manager etc.
+        // But it adds a lot of complexity and overhead and runtime logic
         AWS_S3_ENDPOINT: args.s3Mirror.secret.raw.endpoint,
         AWS_S3_BUCKET_NAME: args.s3Mirror.secret.raw.bucket,
         AWS_S3_ACCESS_KEY_ID: args.s3Mirror.secret.raw.accessKeyId,


deployment/services/aws-artifacts-lambda-function.ts
```diff
@@ -0,0 +1,74 @@
...
+    role: lambdaRole.arn,
+    region: 'us-east-2',
+    environment: {
+      variables: {
+        // This could be done better with secrets manager etc.
+        // But it adds a lot of complexity and overhead and runtime logic
+        AWS_S3_ENDPOINT: args.s3Mirror.secret.raw.endpoint,
+        AWS_S3_BUCKET_NAME: args.s3Mirror.secret.raw.bucket,
+        AWS_S3_ACCESS_KEY_ID: args.s3Mirror.secret.raw.accessKeyId,
+        AWS_S3_ACCESSS_KEY_SECRET: args.s3Mirror.secret.raw.secretAccessKey,
```
n1ru4l, line 55: If anyone insists I can do the extra step of storing these in the AWS Secret Manager and reading it within the lambda during function startup from there instead.

jdolle: this should be completely fine. And secret manager is expensive

deployment/services/aws-artifacts-lambda-function.ts
```diff
@@ -0,0 +1,74 @@
...
+      },
+    },
+    // 448mb
+    memorySize: 448,
```
jdolle, line 59: lambda memory impacts CPU also.
It doesn't seem like your lambda is doing anything too CPU intensive but it may be worth fine-tuning this in the future.

packages/services/cdn-worker/src/index-lambda.ts
```diff
@@ -0,0 +1,102 @@
...
+const s3Mirror = null;
+
+const artifactStorageReader = new ArtifactStorageReader(s3, s3Mirror, null, null);
+
+const artifactHandler = createArtifactRequestHandler({
+  isKeyValid: createIsKeyValid({
+    artifactStorageReader,
+    analytics: null,
+    breadcrumb(message: string) {
+      console.log(message);
```
jdolle, line 38: what is this logging?

jdolle (approved):

deployment/services/aws-artifacts-lambda-function.ts
n1ru4l: I did some different tests, and 448mb had the best trade-off in terms of cost and runtime. See https://www.notion.so/theguildoss/AWS-Lambda-CDN-Handler-Fallback-2b0b6b71848a80968ce8ff8d756adf89?source=copy_link#2b0b6b71848a8001be26f7cf15136032 for more context

packages/services/cdn-worker/src/index-lambda.ts
n1ru4l: it gives some context in case there is an exception being thrown in the cloudwatch logs (e.g. what step are we in). breadcrumb is basically a `logger.debug`. 😄 And we introduced it at some point to figure out what is going wrong in Cloudflare Workers. We could refactor the whole code to actually use more "clean" logging, but that is definetly out of scope here.

PR #7299, opened 2025-11-21
Console 1382 member pagination and search

jonathanawesome:
### Background

This PR improves the `org/members` route by adding pagination and filtering based on username or emaill

### Description

- merges @jdolle PR with the API bits
- strips sorting from the `members` and `assigned role` ui
- paginating 20 items per page
- scrollTo window top when clicking "prev" or "next"
- adds a null state: 
Screenshot 2025-11-24 at 4 56 38 PM

### DIFF
--- a/packages/migrations/src/actions/2025.11.25T00-00-00.members-search.ts
+++ b/packages/migrations/src/actions/2025.11.25T00-00-00.members-search.ts
@@ -3,7 +3,15 @@ import { type MigrationExecutor } from '../pg-migrator';
 export default {
   name: '2025.11.25T00-00-00.members-search.ts',
   run: ({ sql }) => sql`
-    CREATE EXTENSION IF NOT EXISTS pg_trgm;
-    CREATE INDEX CONCURRENTLY IF NOT EXISTS "users_search_by_email_and_display_name" on users using gin(LOWER(email|| ' ' || display_name) gin_trgm_ops);
+    -- The order was wrong. This was sorting by org_id, user_id, then created_at...
+    DROP INDEX IF EXISTS "organization_member_pagination_idx";
+
+    -- Replace "organization_member_pagination_idx" with a new index in the correct order
+    CREATE INDEX CONCURRENTLY IF NOT EXISTS "organization_member_pagination"
+    ON "organization_member" (
+      "organization_id" DESC
+      , "created_at" DESC
+      , "user_id" DESC
+    );
   `,
 } satisfies MigrationExecutor;
--- a/packages/services/api/src/modules/organization/providers/organization-members.ts
+++ b/packages/services/api/src/modules/organization/providers/organization-members.ts
@@ -157,7 +157,7 @@ export class OrganizationMembers {
 
     const first = args.first ? Math.min(args.first, 50) : 50;
     const cursor = args.after ? decodeCreatedAtAndUUIDIdBasedCursor(args.after) : null;
-    const searchTerm = args.searchTerm?.trim().toLowerCase() ?? '';
+    const searchTerm = args.searchTerm?.trim() ?? '';
     const searching = searchTerm.length > 0;
 
     const query = sql`
@@ -188,7 +188,7 @@ export class OrganizationMembers {
               `
             : sql``
         }
-        ${searching ? sql`AND (LOWER("u"."display_name" || ' ' || "u"."email") LIKE ${'%' + searchTerm + '%'})` : sql``}
+        ${searching ? sql`AND "u"."display_name" || ' ' || "u"."email" ILIKE ${'%' + searchTerm + '%'}` : sql``}
       ORDER BY
         "om"."organization_id" DESC
         , "om"."created_at" DESC


packages/web/app/src/pages/organization-members.tsx
```diff
@@ -119,17 +142,59 @@ function OrganizationMembersPageContent(props: {
...
+  const [cursorHistory, setCursorHistory] = useState<Array<string | null>>([null]);
+  const [currentPage, setCurrentPage] = useState(0);
+
+  // Reset pagination when search changes
+  useEffect(() => {
+    setCursorHistory([null]);
+    setCurrentPage(0);
+  }, [search.search]);
+
+  const queryVariables = useMemo(
```
jdolle: minor: This memo is probably unnecessary

packages/web/app/src/components/organization/members/list.tsx
```diff
@@ -310,111 +319,112 @@ const OrganizationMembers_OrganizationFragment = graphql(`
...
+          {search && ` - search results for "${search}"`}
+        </div>
+        <div className="flex items-center gap-2">
+          <Button
+            variant="outline"
+            size="sm"
+            onClick={() => {
+              props.onPreviousPage();
+              setTimeout(() => {
+                window.scrollTo({ top: 0, behavior: 'smooth' });
```
jdolle, line 429: should this scroll to the top of the list instead of all the way back to 0?
I'm not sure which is better UX given our design.

packages/web/app/src/components/organization/members/list.tsx
```diff
@@ -1,5 +1,6 @@
-import { useCallback, useMemo, useState } from 'react';
-import { MoreHorizontalIcon, MoveDownIcon, MoveUpIcon } from 'lucide-react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
+import debounce from 'lodash.debounce';
```
jdolle: We have a debounce helper you can use:
`import { useDebouncedCallback } from 'use-debounce';`

packages/web/app/src/components/organization/members/list.tsx
jonathanawesome: I thought about this while adding the `scrollTo`. We have a large number of screens that potentially/probably overflow the viewable space, and this works fine unless you have controls at the bottom of the view that necessitate a `scrollTo`.  I couldn't find an existing pattern in the codebase, so went with what is most common...scrolling to the top of the page after paginating.

There are more interesting ways to handle this, but that's a design exercise.

packages/services/api/src/modules/organization/providers/organization-manager.ts
```diff
@@ -1221,7 +1221,7 @@ export class OrganizationManager {
 
   async getPaginatedOrganizationMembersForOrganization(
     organization: Organization,
-    args: { first: number | null; after: string | null },
+    args: { first: number | null; after: string | null; searchTerm?: string | null },
```
n1ru4l: I prefer being explicit over accidentally forgetting to add a property.

```suggestion
    args: { first: number | null; after: string | null; searchTerm: string | null },
```

packages/services/api/src/modules/organization/providers/organization-members.ts
```diff
@@ -178,11 +188,12 @@ export class OrganizationMembers {
               `
             : sql``
         }
+        ${searching ? sql`AND ("u"."display_name" ILIKE ${'%' + searchTerm + '%'} OR "u"."email" ILIKE ${'%' + searchTerm + '%'})` : sql``}
       ORDER BY
         "om"."organization_id" DESC
+        , "om"."created_at" DESC
         , "om"."user_id" DESC
```
n1ru4l, line 195: How does this affect usage of existing indices?

packages/services/api/src/modules/organization/resolvers/Organization.ts
```diff
@@ -67,6 +67,7 @@ export const Organization: Pick<
       .getPaginatedOrganizationMembersForOrganization(organization, {
         first: args.first ?? null,
         after: args.after ?? null,
+        searchTerm: args.filters?.searchTerm,
```
n1ru4l: See https://github.com/graphql-hive/console/pull/7299/files#r2555587837
```suggestion
        searchTerm: args.filters?.searchTerm ?? null,
```

packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -881,8 +889,11 @@ export default gql`
     name: String! @deprecated(reason: "Use the 'slug' field instead.")
     owner: Member! @tag(name: "public")
     me: Member!
-    members(first: Int @tag(name: "public"), after: String @tag(name: "public")): MemberConnection!
-      @tag(name: "public")
+    members(
+      first: Int @tag(name: "public")
+      after: String @tag(name: "public")
+      filters: MembersFilter
```
n1ru4l, line 895: filters
MembersFilter

Does singular make more sense here?

packages/services/api/src/modules/organization/providers/organization-manager.ts
jdolle: I don't have strong feelings here. I typically think if it's wrapped in an object then optional is fine b/c autocomplete tells you it's there.

packages/web/app/src/components/organization/members/list.tsx
jdolle: ok. I think the pattern would be to use a ref and scroll to that. But this sounds fine and we can adjust later if it's an issue for some reason

packages/services/api/src/modules/organization/module.graphql.ts
jdolle: It's named after `Query.members` like how we name inputs

packages/services/api/src/modules/organization/providers/organization-members.ts
jdolle: We have an index already that handles this:

```
"organization_member_pagination_idx" btree (organization_id DESC, user_id DESC, created_at DESC)
```

But this is important for the pagination -- so that the last element in the list has a cursor with the oldest timestamp

packages/services/api/src/modules/organization/providers/organization-members.ts
```diff
@@ -178,11 +188,12 @@ export class OrganizationMembers {
               `
             : sql``
         }
+        ${searching ? sql`AND to_tsvector("u"."display_name" || ' ' || "u"."email") @@ to_tsquery('simple', ${searchTerm + ':*'})` : sql``}
```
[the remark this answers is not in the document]
jdolle: prefix matching is a hard requirement

packages/services/api/src/modules/organization/module.graphql.ts
n1ru4l: So 
- `MembersFilter` and `Query.members(filter:)` 
- or `MembersFilters` and `Query.members(filters:)`?

jonathanawesome: I see `XFilter` and `XFilterInput` for `input` name and both `filter:` and `filters` in the codebase.

n1ru4l: For inputs we also kind of have the convention to append `Input` to the name. 

If in doubt on how to name things, I would always recommend to look for similar fields that are tagged with `@tag(name: "public")`, as everything that is part of the public API schema, in theory, is consistent and follows how we want the GraphQL schema to be exposed.

Having that said I am also guilty of my own feedback 😆 
image

image

n1ru4l: Since this is part of the private GraphQL API for now, I am okay with handling it later though. Just make sure to create a follow up task

packages/services/api/src/modules/organization/providers/organization-members.ts
n1ru4l: Now I get it, so in case of, we would need to update the index as well. 😢 Can be a follow-up PR, though!

```
 "om"."organization_id" DESC
        , "om"."created_at" DESC
        , "om"."user_id" DESC
```

packages/migrations/src/actions/2025.11.25T00-00-00.members-search.ts
```diff
@@ -0,0 +1,9 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.11.25T00-00-00.members-search.ts',
+  run: ({ sql }) => sql`
+    CREATE EXTENSION IF NOT EXISTS pg_trgm;
```
n1ru4l: Does this require additional configuration to our azure database? Please make sure to test this migration on the `dev` environment.

n1ru4l: 😢 
```
SQL Error [0A000]: ERROR: extension "pg_trgm" is not allow-listed for users in Azure Database for PostgreSQL
```

@jdolle we need to make sure we can introduce this extension without downtime

jdolle: Added to allow list per https://learn.microsoft.com/en-us/azure/postgresql/extensions/how-to-allow-extensions?tabs=allow-extensions-portal

There was no downtime when enabling this and there are no special considerations for the `pg_trgm` extension that would cause downtime, since it's not instantly changing behavior at install -- it's enabling more ways to index for searching.

For adding the index -- I'm not concerned due to the size of the table. The migration may take a few seconds in the background. And this will slightly increase create/update times for these records but again due to the size of the table, and also because of how infrequently these records are created/updated, this should not be an issue.

jonathanawesome:
One final tweak to the `useSearchParamsFilters` hook to ensure that it handles simple strings. This one's ready for a final review @n1ru4l .

packages/services/api/src/modules/organization/providers/organization-members.ts
jdolle: I've modified this index.

I was going back and forth on how to make this query as efficient as possible without introducing something like elastic search and I think this is now as good as it gets. The trigram index wasn't working out for a few reasons:

1. `organization_member` is already conditioned to where the organization_id matches. This is doing the bulk of the work for filtering users to just that org. No single organization has thousands of members yet.
2. `organization_member` and `users` is joining on user_id. We could keep the trigram but then we'd join on the display name / email here also to make use of it. But since a user only has one display name / email, then my understanding is that it won't benefit from this index.

We _could_ take the opposite route of first filtering a user by the display name / email, but it's much more efficient to filter the list how we have it for now... Maybe if a single organization had millions of users then it'd be different.

When we do hit performance limits with our data, or if the returned ordering becomes a more critical feature, then we should shift to cloning this data to something like elasticsearch.

packages/services/api/src/modules/organization/providers/organization-members.ts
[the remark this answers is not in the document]
jdolle: and this index wont get used on this specific query because the userId lookup will be used instead.

n1ru4l (approved):
Just played around with it, and seems solid 🎉

PR #7302, opened 2025-11-21
docs(router): Add authorization documentation and specification

kamilkisiela:
Reflects https://github.com/graphql-hive/federation-composition/pull/215 and https://github.com/graphql-hive/router/pull/538

### DIFF
--- a/packages/web/docs/src/content/router/configuration/authorization.mdx
+++ b/packages/web/docs/src/content/router/configuration/authorization.mdx
@@ -5,67 +5,18 @@ title: 'authorization'
 # authorization
 
 The `authorization` configuration lets you control fine-grained access to your GraphQL schema using
-directives. This allows you to restrict which fields authenticated users can access based on their
-authentication status or specific scopes.
+directives like `@authenticated` and `@requiresScopes`. This allows you to restrict which fields
+authenticated users can access based on their authentication status or specific scopes.
 
 For practical examples and a guide to authorization concepts, see
 **["Authorization"](../security/authorization)** in the documentation.
-
-## How Authorization Works
-
-The router supports two modes for handling unauthorized field access:
-
-- **`filter` (default mode):** This mode silently removes any fields from the incoming GraphQL
-  operation that the user is not authorized to access. For each field removed, a corresponding
-  authorization error is added to the `errors` list in the GraphQL response, while the rest of the
-  accessible data is returned as requested.
-- **`reject` mode:** In this mode, the router will reject any GraphQL operation that attempts to
-  access one or more unauthorized fields. The entire request is denied, and a descriptive error is
-  returned.
 
 ## Directives
 
 Access control is defined within the supergraph schema using the following directives:
 
-### `@authenticated`
-
-Restricts access to a field to only authenticated users. Any request without valid authentication
-(such as requests without a JWT token) will be prevented from accessing the field.
-
-**Usage:**
-
-```graphql
-type Query {
-  me: User @authenticated
-  publicData: String
-}
-```
-
-### `@requiresScopes(scopes: [[String]])`
-
-Provides more granular control by requiring the user to possess specific scopes. The directive
-supports:
-
-- **`AND` logic** for scopes within a nested list (e.g., `[["read:users", "write:users"]]`) - the
-  user must have all scopes in the list
-- **`OR` logic** for scopes across nested lists (e.g., `[["read:users"], ["admin:users"]]`) - the
-  user must have scopes from at least one of the nested lists
-
-**Usage:**
-
-```graphql
-type Query {
-  # User must have both scopes
-  users: [User] @requiresScopes(scopes: [["read:users", "write:users"]])
-
-  # User must have either scope
-  admin: AdminPanel @requiresScopes(scopes: [["admin:users"], ["admin:system"]])
-
-  # User must have read:users AND (admin:users OR admin:system)
-  reports: [Report]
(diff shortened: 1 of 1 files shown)


packages/web/docs/src/content/router/configuration/authorization.mdx
```diff
@@ -0,0 +1,176 @@
...
+---
+
+# authorization
+
+The `authorization` configuration lets you control fine-grained access to your GraphQL schema using
+directives. This allows you to restrict which fields authenticated users can access based on their
+authentication status or specific scopes.
+
+For practical examples and a guide to authorization concepts, see
+**["Authorization"](../security/authorization)** in the documentation.
```
jdolle, line 12: I like that authorization is covered both in the configuration and security section, but _this_ page is technically "authorization in the documentation" also.

I think it would be good to make a more clear delineation between the two documents. E.g. `security/authorization` could talk about more broadly about authorization and high-level motives behind our approach, and `configuration/authorization` can provide the details about configuration options in the router and cover the specific directive definitions.

This is true for most of the content in `Security` subpages. We could move important details to their related configuration page, and make `Security` a more high-level page that acts as an introduction to these features as concepts rather than provide too much detail up front.

packages/web/docs/src/content/router/configuration/authorization.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+```graphql
+type Query {
+  # User must have both scopes
+  users: [User] @requiresScopes(scopes: [["read:users", "write:users"]])
+
+  # User must have either scope
+  admin: AdminPanel @requiresScopes(scopes: [["admin:users"], ["admin:system"]])
+
+  # User must have read:users AND (admin:users OR admin:system)
```
jdolle: great example and explanations. This is what 99% of people will want to know.

packages/web/docs/src/content/router/configuration/authorization.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+- `"filter"`: Remove unauthorized fields and continue processing (returns errors for removed fields)
+- `"reject"`: Reject the entire request if any unauthorized fields are accessed
+
+## Examples
+
+### Filter Mode (Default)
+
+With `filter` mode, unauthorized fields are silently removed from the operation, but the query
+continues to execute and return the data the user can access:
```
jdolle, line 47: Why might one choose reject over filter? And should we mention how nullability impacts this? A lot of developers who are new to graphql don't realize how error bubbling works.

packages/web/docs/src/content/router/security/authorization.mdx
```diff
@@ -0,0 +1,430 @@
...
+  id: ID!
+  username: String @requiresScopes(scopes: [["user:read", "profile:read"]])
+  age: Int
+}
+```
+
+To access `User.username`, a client must have **both** `user:read` and `profile:read` scopes,
+ensuring they meet the access requirements from all subgraphs where the field is defined.
+
+### Merging `@requiresScopes`
```
jdolle, line 216: This is mostly redundant with the field-level composition. Could these be simplified to a single section about merging across subgraphs?

packages/web/docs/src/content/router/security/authorization.mdx
```diff
@@ -0,0 +1,430 @@
...
+  "errors": [
+    {
+      "message": "Unauthorized field or type",
+      "extensions": { "code": "UNAUTHORIZED_FIELD_OR_TYPE" }
+    }
+  ]
+}
+```
+
+Enable reject mode by setting `unauthorized.mode: reject` in your router configuration:
```
jdolle, line 419: as mentioned previously, it may be worth calling out issues with nullability and why you might choose one approach over the other

packages/web/docs/src/content/router/security/authorization.mdx
kamilkisiela: This section explains how scopes are merged. Nowhere in the field-level composition I cover that. It's part of higher "Combining Directives Across Federated Types" section.

packages/web/docs/src/content/router/security/authorization.mdx
kamilkisiela: Not sure it's worth explaining why pick one over another. It's obvious what they do and I feel like people will know when to use them, there are no rules when to use `filter` over `reject`, except one "If you want to reject request when at least one field is unauthorized you 'reject'", which is what I wrote:

> If a user tries to access any unauthorized field, the entire request is rejected and no data is
returned.

packages/web/docs/src/content/router/configuration/authorization.mdx
kamilkisiela: A separate section about nullability yes, good idea, but imo explaining when to use each mode is kind of obvious to people (they know their behaviour and there's nothing more to add really)

packages/web/docs/src/content/router/configuration/authorization.mdx
kamilkisiela: I want to avoid writing high-level content and link people to different pages. The idea here is to give people a comprehensive guide on authorization and only branch out to other pages when it involves a bigger/standalone topic (like JWT - as it's not only about authorization, but authentication too).

The only thing that is configuration-specific is filter/reject mode. I forgot to remove the directives from this page.

The configuration pages will be generated from code once Dotan is back and I'm fine with duplicating content a bit.

PR #7307, opened 2025-11-24
feat: hive-logger for JS SDK

n1ru4l:
### Background

Makes the Hive SDK ready for Hive Gateway providing a logger instance.

### Description

Use Hive Logger instead of our own home cooked logger instance for doing logging.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/libraries/yoga/src/index.ts
+++ b/packages/libraries/yoga/src/index.ts
@@ -205,7 +205,7 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
                       // Defer logs until yoga instance is initialized
                       // Ideally, onPluginInit would provide us access to the logger instance
                       // See https://github.com/graphql-hive/graphql-yoga/issues/4048#issuecomment-3576258603
-                      onYogaInitDefered.then(() => {
+                      void onYogaInitDefered.then(() => {
                         yoga?.logger[level](msg, attrs);
                       });
                     },


packages/libraries/core/src/client/utils.ts
```diff
@@ -249,3 +188,44 @@ export function isLegacyAccessToken(accessToken: string): boolean {
...
+
+  if (logger instanceof Logger) {
+    return logger;
+  }
+
+  return new Logger({
+    level: 'debug',
+    writers: [
+      {
+        write(level, _attrs, msg) {
```
n1ru4l: This writer implementation is "backwards-compatible" with how the legacy logger works.

packages/libraries/core/src/client/types.ts
```diff
@@ -219,7 +220,7 @@ export type HivePluginOptions = OptionalWhenFalse<
      *
      * Default: 'info'
      */
-    logger?: Logger | 'error' | 'info' | 'debug';
+    logger?: Logger | HiveLoggerLevel;
```
n1ru4l, line 223: Since the other PR introduces the logger, we can enforce that people pass a hive logger here (https://github.com/graphql-hive/console/pull/7290)

packages/libraries/yoga/src/index.ts
```diff
@@ -179,13 +179,10 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
         ? clientOrOptions
         : createHive({
             ...clientOrOptions,
+            // TODO: How to convert yoga.logger into HiveLogger here?
+            logger: clientOrOptions.logger ?? yoga.logger,
```
n1ru4l: @enisdenjo Any ideas here? 😓 

Basically, if the user provides the logger -> we use that one, otherwise we "convert" the yoga logger and pass it in here? Right now the yoga.logger types are incompatible..

packages/libraries/core/src/client/http-client.ts
```diff
@@ -58,6 +59,39 @@ export const http = {
...
+            logger.debug(msg);
+            return;
+          }
+          if (level === 'info' && msg) {
+            logger.info(msg);
+            return;
+          }
+          if (level === 'error' && msg) {
+            logger.error(msg);
+          }
```
enisdenjo, line 88: setting a level during logger instantiation will filter these out during write, you dont have to do all that. (unless typedefs)
```suggestion
          if (msg) {
            logger[level](msg);
          }
```

packages/libraries/core/src/client/http-client.ts
```diff
@@ -132,26 +166,26 @@ export async function makeFetchCall(
...
         if (isAggregateError(error)) {
           for (const err of error.errors) {
             if (isFinalAttempt) {
-              logger?.error(err);
+              logger.error({ error: err });
               continue;
             }
-            logger?.debug?.(String(err));
+            logger.debug(String(err));
           }
```
enisdenjo: Logger handles aggregate errors too, so you can just log everything instead of the last error.

packages/libraries/core/src/client/http-client.ts
```diff
@@ -132,26 +166,26 @@ export async function makeFetchCall(
...
 
           logErrorMessage();
           throw new Error(`Unexpected HTTP error. (x-request-id=${requestId})`, { cause: error });
         }
 
-        logger?.error(error);
+        logger.error({ error }, String(error));
```
enisdenjo:
```suggestion
        logger.error({ error });
```

packages/libraries/core/tests/reporting.spec.ts
```diff
@@ -272,10 +271,10 @@ test('should send data to app.graphql-hive.com/graphql by default', async () =>
...
 
   expect(logger.getLogs()).toMatchInlineSnapshot(`
-    [INF] [hive][reporting] Publish schema
-    [DBG] [hive][reporting] POST https://app.graphql-hive.com/graphql (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) Attempt (1/6)
-    [DBG] [hive][reporting] POST https://app.graphql-hive.com/graphql (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) succeeded with status 200 (666ms).
-    [INF] [hive][reporting] Published schema
+    [INF] [hive][reporting]Publish schema
+    [DBG] [hive][reporting]POST https://app.graphql-hive.com/graphql (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) Attempt (1/6)
+    [DBG] [hive][reporting]POST https://app.graphql-hive.com/graphql (x-request-id=xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx) succeeded with status 200 (666ms).
+    [INF] [hive][reporting]Published schema
```
enisdenjo, line 277: Hive Logger does not append a space, you have to `logger.child('[reporting] ')`.

packages/libraries/yoga/src/index.ts
enisdenjo: You convert a yoga logger to hive logger like this:
```ts
import { Logger } from '@graphql-hive/logger'
new Logger({
  writers: [
    {
      write(level, attrs, msg) {
        level = level === 'trace' ? 'debug' : level;
        yoga.logger[level](msg, attrs);
      },
    },
  ],
})

```

packages/libraries/core/src/client/http-client.ts
enisdenjo: Or do a `logger.error({ lastErr: errs[0] })` and `logger.debug({ errs })` in case you dont want everything on the error level.

packages/libraries/core/tests/reporting.spec.ts
n1ru4l: But then it would be `[hive] [reporting] `, because i would need to do the same for `[hive]`, or have multiple loggers etc... Ultimately, we want to remove these prefixes anyways in favor of attributes...

For now removing the space is an ok tradeoff here to not add more complexity and not change things too drastically.

packages/libraries/core/src/client/http-client.ts
n1ru4l: Would the legacy logger implementation in that case log anything here if there is no message? Or is the message auto-constructed from attributes if it includes an error object?

enisdenjo: Right, the legacy logger will log as if you did `console.error({ error })`.

packages/libraries/core/src/client/http-client.ts
n1ru4l: types :D

PR #7315, opened 2025-11-25
test(otel-collector): add integration tests for trace ingestion

adambenhassen:
### Description:
  Add integration tests to verify otel-collector starts correctly and inserts traces to ClickHouse.

### Tests cover:
- Accepting traces with valid authorization and inserting to ClickHouse
- Slug-based target references (`org/project/target` format)
- Rejecting unauthorized requests (401)
- Rejecting requests without `traces:report` permission
- Populating `otel_traces_normalized` materialized view for GraphQL span
- Querying through GraphQL API
- Querying through GraphQL API with filtering

CONSOLE-1562

### DIFF
--- a/integration-tests/tests/api/otel-collector.spec.ts
+++ b/integration-tests/tests/api/otel-collector.spec.ts
@@ -1,0 +1,649 @@
+import { clickHouseQuery } from '../../testkit/clickhouse';
+import { graphql } from '../../testkit/gql';
+import { GraphQlOperationType, ResourceAssignmentModeType } from '../../testkit/gql/graphql';
+import { execute } from '../../testkit/graphql';
+import { initSeed } from '../../testkit/seed';
+import { getServiceHost } from '../../testkit/utils';
+
+const TargetTracesQuery = graphql(`
+  query TargetTracesQuery($targetId: ID!, $traceIds: [ID!]) {
+    target(reference: { byId: $targetId }) {
+      id
+      viewerCanAccessTraces
+      traces(first: 10, filter: { traceIds: $traceIds }) {
+        edges {
+          node {
+            id
+            operationName
+            operationType
+            success
+            httpStatusCode
+            httpMethod
+            spans {
+              name
+            }
+          }
+        }
+      }
+    }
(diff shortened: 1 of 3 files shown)


integration-tests/tests/api/otel-collector.spec.ts
```diff
@@ -0,0 +1,431 @@
...
+  });
+
+  expect(response.status).toBe(200);
+
+  // Wait for trace to appear in otel_traces first
+  const found = await waitForTraceInClickHouse(traceId);
+  expect(found).toBe(true);
+
+  // Give a bit more time for the materialized view to process
+  await new Promise(resolve => setTimeout(resolve, 2000));
```
jdolle: I'd love for us to avoid waiting during tests in order to keep our suite fast, but I know when tests involve async queue processing, this is difficult.

We might consider in the future creating a synchronous version of our job processors specifically for testing... But I'm okay with this for now.

integration-tests/tests/api/otel-collector.spec.ts
```diff
@@ -0,0 +1,431 @@
...
+              name
+            }
+          }
+        }
+      }
+    }
+  }
+`);
+
+async function sendTrace(args: {
```
jdolle, line 90: Does this function belong in our testkit?

integration-tests/tests/api/otel-collector.spec.ts
n1ru4l: would also prefer polling here, timeouts were to flaky in past experience.

integration-tests/tests/api/otel-collector.spec.ts
adambenhassen: Could you please elaborate?

integration-tests/tests/api/otel-collector.spec.ts
adambenhassen: You're right. I already had a poll for the normalized table, which I added later on

n1ru4l (approved):
Great 😍

PR #7316, opened 2025-11-25
Console 1383 open non existing schema version results in infinite loading

### LINKED ISSUE #7052: open non-existing schema version results in infinite loading and backend exception
If I visit some url like `http://localhost:3000/the-guild/otel-demo/development/history/qjwghejkqwhenqwe` (note the invalid UUID), the API responds with the following error. Before we query for the UUID in postgres, we need to validate the input on the business logic layer and otherwise treat this scenario as "not found".

```
{
    "errors": [
        {
            "message": "Unexpected error.",
            "locations": [
                {
                    "line": 12,
                    "column": 7
                }
            ],
            "path": [
                "project",
                "target",
                "schemaVersion"
            ],
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR"
            }
        }
    ],
    "data": {
        "__typename": "Query",
        "project": {
            "__typename": "Project",
            "id": "7c4bcd42-32e7-4485-9f12-ce903acce4d9",
            "type": "FEDERATION",
            "target": {
                "__typename": "Target",
                "id": "219fedb6-8a9b-44d7-929f-b0b16750653c",
                "schemaVersion": null
            }
        }
    }
}
````

For a non existing schema version (but valid UUID) e.g. `http://localhost:3000/the-guild/otel-demo/development/history/5e52d300-b223-428f-9364-f10d31595f40`, we also get an infinite loading state and an exception thrown by our graphql server.

Image

### PR DESCRIPTION
**Pull request by jonathanawesome**
### Background

closes #7052 

### Description

This PR uses our NotFoundContent component to display notes to the user for 1) when an invalid UUID is detected and 2) when a valid UUID is detected but the API does not return a `schemaVersion`.

Additionally:

- Combines disparate `utils.ts` files into one and fixes up associated imports
- Pauses query when we detect an invalid UUID
- Updates `NotFoundContent` component with an optional `includeBackButton` prop, defaulted to true
- Fixes a `validateDOMNesting` error in `CopyText` component
- Removes unused field selections

### DIFF
--- a/packages/web/app/src/lib/utils.ts
+++ b/packages/web/app/src/lib/utils.ts
@@ -16,16 +16,6 @@ const darkChartStyles = {
 
 export function useChartStyles() {
   return darkChartStyles;
-  // TODO: fix it when Hive will have white theme
-  // useColorModeValue(
-  //   {
-  //     backgroundColor: '#fff',
-  //     textStyle: { color: '#52525b' },
-  //     legend: {
-  //       textStyle: { color: '#52525b' },
-  //     },
-  //   },
-  // );
 }
 
 // Strings

### CODE REVIEW

File: packages/web/app/src/lib/utils.ts [hunk 1]
```diff
@@ -1,10 +1,34 @@
...
+  backgroundColor: 'transparent',
+  textStyle: { color: '#fff' },
+  legend: {
+    textStyle: { color: '#fff' },
+  },
+};
+
+export function useChartStyles() {
+  return darkChartStyles;
+  // TODO: fix it when Hive will have white theme
```
jdolle: ^ prefer creating an issue in linear vs having comments

packages/web/app/src/pages/target-history-version.tsx
```diff
@@ -657,7 +654,28 @@ function ActiveSchemaVersion(props: {
   const schemaVersion = project?.target?.schemaVersion;
   const projectType = query.data?.project?.type;
 
-  if (isLoading || !schemaVersion || !projectType) {
+  // Order of these conditionals is important...relocate carefully!
+  if (!isValidVersionId) {
```
jdolle, line 658: for the future -- we should decide how much logic we want on the client vs server. I tend to prefer most validation being done on the server but for something that will change infrequently such as this, client side validation can provide a better user experience. But having a general guideline would help us with consistency and codebase maturity

PR #7317, opened 2025-11-25
feat(otel): enable otel tracing in development via environment variable

adambenhassen:
### Description
Enable the otelTracing feature flag via environment variable. This allows developers to use the `traces:report` permission with organization access tokens by just switching an environment variable.


docker/docker-compose.community.yml
```diff
@@ -244,6 +244,7 @@ services:
       SENTRY: '${SENTRY:-0}'
       SENTRY_DSN: '${SENTRY_DSN:-}'
       PROMETHEUS_METRICS: '${PROMETHEUS_METRICS:-}'
+      FEATURE_FLAGS_OTEL_TRACING_ENABLED: '1'
```
jdolle: should we also enable in `docker/docker-compose.dev.yml`?

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -20,11 +22,12 @@ export class Traces {
     private clickHouse: ClickHouse,
     private logger: Logger,
     private storage: Storage,
+    @Inject(OTEL_TRACING_ENABLED) private otelTracingEnabled: boolean,
   ) {}
 
   async viewerCanAccessTraces(organizationId: string) {
     const organization = await this.storage.getOrganization({ organizationId });
-    return organization.featureFlags.otelTracing;
+    return organization.featureFlags.otelTracing || this.otelTracingEnabled;
```
jdolle, line 30: this may be out of scope, but should this also check the user permissions or is that being done elsewhere already?

jdolle (approved):

docker/docker-compose.community.yml
adambenhassen: `docker-compose.dev.yml` contains the dependencies (databases, queues, etc.) that the graphql server needs to run, not the server itself. The server is run separately (typically via pnpm dev or similar)

packages/services/api/src/modules/operations/providers/traces.ts
adambenhassen: Yes, user permissions are already being checked elsewhere.

From what I've just analyzed, `viewerCanAccesTraces` is a field on the `Target` type. To access a `Target`, one must go through `target` query which calls `getTargetByReferenceInput`, which calls `getTarget`, which performs permission checking via `session.assertPerformAction` with the action `project:describe`. So basically by the time `viewerCanAccessTraces` is called, the user has already been permission checked to access target/project.

docker/docker-compose.community.yml
n1ru4l: We do not want to have this environment variable in the community yaml. The feature is not tested for us at scale, and before we have more customers using it I do not want to encourage self-hosters to use this feature and become dependent on it.

n1ru4l:
```suggestion
```

n1ru4l (approved):

PR #7320, opened 2025-11-25
chore(clickhouse): add size-based compression config for docker-compose.dev.yml

adambenhassen:
#### Description
Adds a ClickHouse compression configuration file for the local development environment. The configuration uses a size-based compression strategy where ZSTD compression is applied to large data parts (over 1GB) to optimize storage space, while LZ4 compression is used for smaller parts to prioritize read/write performance.

### DIFF
--- a/docker/configs/clickhouse/compression.xml
+++ b/docker/configs/clickhouse/compression.xml
@@ -1,0 +1,15 @@
+<clickhouse>
+  <compression>
+    <!-- Use ZSTD for large parts to save space. -->
+    <case>
+      <min_part_size>1073741824</min_part_size> <!-- 1GB -->
+      <min_part_size_ratio>0.01</min_part_size_ratio>
+      <method>zstd</method>
+      <level>1</level>
+    </case>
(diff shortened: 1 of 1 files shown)


docker/configs/clickhouse/compression.xml
```diff
@@ -0,0 +1,8 @@
+<clickhouse>
+  <compression>
+    <!-- Use LZ4 compression for better performance -->
```
jdolle, line 10: We have a lot of columns where we override the codec to zstd, and I believe lz4 is the default.
I don't think this change is necessary.

adambenhassen: That is true. Although, the main idea behind this change is to add a size based compression so as to use zstd for larger parts that I have created during my local load tests and trace seeding. Please don't rush into fresh PRs.

jdolle: Sounds interesting. Ok no worries. Sorry to have rushed the review

PR #7321, opened 2025-11-25
fix(cli): handle escaped single-quoted strings in schema change

adambenhassen:
### Description
Similar to https://github.com/graphql-hive/console/pull/7203, CLI also had the same issue.


jdolle (dismissed):
Ty. Please add a test so we don't accidentally break this in the future.

PR #7328, opened 2025-11-27
feat(api): expose Project.createdAt via public GraphQL API

adambenhassen:
### Description
Expose the existing database `created_at` column on the Project type through the GraphQL API.

CONSOLE-1571


adambenhassen:
> We coul also add a changeset that mentions that `Project.createdAt` is now part of the public API. But not thaaat important, I am fine with merging as is!

There should be one in the commit `.changeset/brave-falcons-glow.md`

PR #7340, opened 2025-11-29
add conditional for if an org exists

### LINKED ISSUE #6815: navigating to non-existing org results in infinite loading screen
Image

jonathanawesome:
### Background

closes #6815 

### Description

This PR adds a NotFoundContent component, with proper copy, when an organization is not found. 

Screenshot 2025-11-29 at 9 59 50 AM

### DIFF
--- a/packages/migrations/src/actions/2025.11.25T00-00-00.members-search.ts
+++ b/packages/migrations/src/actions/2025.11.25T00-00-00.members-search.ts
@@ -1,0 +1,17 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.11.25T00-00-00.members-search.ts',
+  run: ({ sql }) => sql`
+    -- The order was wrong. This was sorting by org_id, user_id, then created_at...
+    DROP INDEX IF EXISTS "organization_member_pagination_idx";
+
+    -- Replace "organization_member_pagination_idx" with a new index in the correct order
+    CREATE INDEX CONCURRENTLY IF NOT EXISTS "organization_member_pagination"
+    ON "organization_member" (
+      "organization_id" DESC
+      , "created_at" DESC
+      , "user_id" DESC
+    );
+  `,
+} satisfies MigrationExecutor;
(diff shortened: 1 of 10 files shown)


packages/web/app/src/components/layouts/organization.tsx
```diff
@@ -219,7 +220,16 @@ export function OrganizationLayout({
...
+
+        {currentOrganization ? (
+          <div className={className}>{children}</div>
+        ) : (
+          <NotFoundContent
+            heading="Organization not found"
+            subheading="Use the empty dropdown in the header to select an organization to which you have access."
+            includeBackButton={false}
+          />
+        )}
```
n1ru4l: Does this change work for users with `"users"."is_admin"` being set to `true` (hive "admin"), being able to see organizations they are not a member of?

n1ru4l: Yeah, seems like since `currentOriganization` is derived from `Query.me`, and admin members are not part of the organization, this would show them "not found", even though we sometimes use this to debug users organization with read-only permissions!

n1ru4l: We might need to have to query for the organization, if we introduce this in the org layout 🤔

jonathanawesome:
> We need a solution here that works with the ability of hive admins to see orgs they are not members of: [#7340 (comment)](https://github.com/graphql-hive/console/pull/7340#discussion_r2576653347)

Latest changes should be sufficient! Took me a while to find the `organizationBySlug` field and handle the situation where we see the null state when switching between orgs.

packages/web/app/src/components/layouts/organization.tsx
```diff
@@ -93,9 +97,14 @@ export function OrganizationLayout({
   const [isModalOpen, toggleModalOpen] = useToggle();
   const [query] = useQuery({
     query: OrganizationLayoutQuery,
-    requestPolicy: 'cache-first',
+    variables: {
+      organizationSlug: props.organizationSlug,
+    },
+    requestPolicy: 'cache-and-network',
```
n1ru4l: Doesn't this change cause that the query is now fired on every navigation?

jonathanawesome: great catch...this change was a leftover from my rendering troubleshooting. FIXED.

Issue #7153, opened 2025-10-22
Migration for 8.4.1 does not work


cosmincatalin:
I get this when applying the migrations for 8.4.1 (last successful migration is 8.4.0)

```
{
	"meta":
	[
	],
	"data":
	[
	],
	"rows": 0,
	"exception": "Code: 8. DB::Exception: SELECT query outputs column with name 'client_name', which is not found in the target table. Use 'AS' to assign alias that matches a column name. (THERE_IS_NO_COLUMN) (version 25.7.5.34 (official build))"
}
HTTPError: Response code 400 (Bad Request)
    at _Request.<anonymous> (file:///usr/src/app/index.js:89898:32)
    at Object.onceWrapper (node:events:639:26)
    at _Request.emit (node:events:536:35)
    at _Request._onResponseBase (file:///usr/src/app/index.js:89458:14)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async _Request._onResponse (file:///usr/src/app/index.js:89494:7) {
  input: undefined,
  code: 'ERR_NON_2XX_3XX_RESPONSE',
  timings: {
    start: 1761150529154,
    socket: 1761150529154,
    lookup: 1761150529154,
    connect: 1761150529154,
    secureConnect: undefined,
    upload: 1761150529154,
    response: 1761150529168,
    end: 1761150529168,
    error: undefined,
    abort: undefined,
    phases: {
      wait: 0,
      dns: 0,
      tcp: 0,
      tls: undefined,
      request: 0,
      firstByte: 14,
      download: 0,
      total: 14
    }
  },
  options: {
    request: undefined,
    agent: { http: undefined, https: undefined, http2: undefined },
    h2session: undefined,
    decompress: true,
    timeout: {
      connect: undefined,
      lookup: undefined,
      read: undefined,
      request: undefined,
      response: undefined,
      secureConnect: undefined,
      send: undefined,
      socket: undefined
    },
    prefixUrl: '',
    body: '\n' +
      '    CREATE MATERIALIZED VIEW IF NOT EXISTS "otel_subgraph_spans_mv" TO "otel_subgraph_spans" (\n' +
      '      "target_id" LowCardinality(String)\n' +
      '      , "subgraph_name" Array(String)\n' +
      '      , "trace_id" String\n' +
      '      , "span_id" String\n' +
      `      , "timestamp" DateTime('UTC')\n` +
      '      , "duration" UInt64\n' +
[...]

PR #7350, opened 2025-12-01
fix: clickhouse db migration typo

n1ru4l:
### Background

Seems to cause https://github.com/graphql-hive/console/issues/7153, even it does not for us. In any case the table is currently not consumed, but probably worth to fix it...

Also removes the table since it is unused and faulty in our deployment as well.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/.changeset/little-readers-return.md
+++ b/.changeset/little-readers-return.md
@@ -2,4 +2,4 @@
 'hive': patch
 ---
 
-Fix invalid materialized view definition.
+Fix invalid materialized view definition causing failing ClickHouse migrations


.changeset/little-readers-return.md
```diff
@@ -0,0 +1,5 @@
+---
+'hive': patch
+---
+
+Fix invalid materialized view definition.
```
jdolle: Minor: Consider adding the specific details of the error this solves.

n1ru4l:
```suggestion
Fix invalid materialized view definition causing failing ClickHouse migrations
```

PR #7352, opened 2025-12-01
fix app deployments pagination and feature flag

jdolle:
### Background

An app deployment version's operations is not able to load its next page due to a bug.
While setting up my local environment for testing, I also found that the feature flag wasn't being respected for app deployments for token permission creation also.

### Description

Correctly check the environment flag in `createFeatureFlagPermissionFilter` and adjust frontend to fix pagination.

### DIFF
--- a/packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
+++ b/packages/services/api/src/modules/organization/providers/organization-access-tokens.ts
@@ -898,7 +898,7 @@ export class OrganizationAccessTokens {
     const isOTELTracingEnabled = organization.featureFlags.otelTracing || this.otelTracingEnabled;
 
     return (id: Permission) =>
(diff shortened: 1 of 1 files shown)


.changeset/curly-birds-itch.md
```diff
@@ -0,0 +1,6 @@
+---
+'hive': patch
+---
+
+Correctly check the environment flag in createFeatureFlagPermissionFilter and adjust frontend to fix
+pagination.
```
n1ru4l:
```suggestion
Correctly check the global environment feature flag in for app deployments and fix app deployment pagination
```

n1ru4l: `createFeatureFlagPermissionFilter` is a technical implementation detail that provides no benefit to the reader imho.

n1ru4l (approved):
Approving with one minor comment: https://github.com/graphql-hive/console/pull/7352/files#r2584184126

PR #7364, opened 2025-12-03
fix(sdk-rs): send headers correctly for usage reporting

ardatan:
Ref CONSOLE-1600

- Sends headers correctly for usage reporting
- Sends operation name and client information correctly
- Adds unit tests for usage reporting to test the behavior overall including the ones above
- Runs Rust unit tests in CI
- Setup alpha version releases for Rust packages


n1ru4l (approved):
❤️

PR #7375, opened 2025-12-05
Explore route loaders and layout changes

jonathanawesome:
### Background

We're trying to get to one-query-per-page, but not quite there. 

### Description

This draft PR explores how we can use TanStack Router loaders and related functions to get to one-query-per-page. This is a proof-of-concept, not a proposal.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -19389,8 +19389,8 @@ snapshots:
     dependencies:
       '@aws-crypto/sha256-browser': 3.0.0
       '@aws-crypto/sha256-js': 3.0.0
-      '@aws-sdk/client-sso-oidc': 3.596.0(@aws-sdk/client-sts@3.596.0)
-      '@aws-sdk/client-sts': 3.596.0
+      '@aws-sdk/client-sso-oidc': 3.596.0
+      '@aws-sdk/client-sts': 3.596.0(@aws-sdk/client-sso-oidc@3.596.0)
       '@aws-sdk/core': 3.592.0
       '@aws-sdk/credential-provider-node': 3.596.0(@aws-sdk/client-sso-oidc@3.596.0)(@aws-sdk/client-sts@3.596.0)
       '@aws-sdk/middleware-host-header': 3.577.0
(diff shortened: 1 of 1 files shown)


packages/web/app/src/router.tsx
```diff
@@ -95,6 +102,20 @@ if (env.sentry) {
 
 const queryClient = new QueryClient();
 
+async function loadGraphQLData<TData, TVariables extends Record<string, any>>(
```
jonathanawesome: A helper function for using urql client in loaders.

packages/web/app/src/router.tsx
```diff
@@ -343,28 +364,46 @@ const organizationRoute = createRoute({
   errorComponent: ErrorComponent,
 });
 
-const organizationIndexRoute = createRoute({
+export const organizationLayoutRoute = createRoute({
```
jonathanawesome: This is a new pathless layout route. The component is the same component which we currently use to wrap org pages. Org pages call this layout route their parent.

packages/web/app/src/components/layouts/organization.tsx
```diff
@@ -50,181 +49,114 @@ export enum Page {
...
-                isOpen={isModalOpen}
-                toggleModalOpen={toggleModalOpen}
-                // reset the form every time it is closed
-                key={String(isModalOpen)}
-              />
-            </>
-          ) : null}
-        </div>
-      </div>
+      <PrimaryNavigation
```
jonathanawesome: We're no longer prop-drilling the Page enum member through, so this Tabs component was unaware of the active "Tab". We have an existing issue to clean up the inaccessibility of this Tabs usage, so this is a 2-for-1. Lots of interesting bits that were quick-to-build in the new `/navigation` files.

packages/web/app/src/components/v2/modals/transfer-organization-ownership.tsx
```diff
@@ -97,6 +97,8 @@ export const TransferOrganizationOwnershipModal = ({
         organizationSlug: organization.slug,
       },
     },
+    // don't fire this query until the modal opens
+    pause: !isOpen,
```
jdolle, line 101: clever

packages/web/app/src/router.tsx
```diff
@@ -95,6 +102,20 @@ if (env.sentry) {
...
+  query: TypedDocumentNode<TData, TVariables>,
+  variables: TVariables,
+): Promise<TData> {
+  const result = await urqlClient
+    .query(query, variables, {
+      requestPolicy: 'cache-first',
+    })
+    .toPromise();
+
+  if (result.error) throw result.error;
```
jdolle, line 115: I'd much rather handle an error result as data than as an exception. You lose all the typing info by throwing.

PR #7377, opened 2025-12-05
feat(api): retrieve app deployments based on last used data

adambenhassen:
### Description
Adds `activeAppDeployments` GraphQL query to find app deployments based on usage criteria, enabling teams to identify and clean up stale deployments

#### Filter options:
  - `lastUsedBefore` - find deployments used but not recently
  - `neverUsedAndCreatedBefore` - find old unused deployments
  - `name` - filter by app name (partial match)

Closes CONSOLE-1504

### DIFF
--- a/packages/services/api/src/modules/app-deployments/module.graphql.ts
+++ b/packages/services/api/src/modules/app-deployments/module.graphql.ts
@@ -11,7 +11,7 @@ export default gql`
       filter: AppDeploymentDocumentsFilterInput
     ): GraphQLDocumentConnection
     totalDocumentCount: Int!
-    status: AppDeploymentStatus! @tag(name: "public")
+    status: AppDeploymentStatus!
     """
     The timestamp when the app deployment was created.
     """
@@ -27,9 +27,9 @@ export default gql`
   }
 
   enum AppDeploymentStatus {
-    pending @tag(name: "public")
-    active @tag(name: "public")
-    retired @tag(name: "public")
+    pending
+    active
+    retired
   }
 
   type GraphQLDocumentConnection {


packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -806,6 +806,217 @@ export class AppDeployments {
...
+    );
+
+    // apply cursor-based pagination
+    let paginatedDeployments = filteredDeployments;
+    if (cursor) {
+      const cursorCreatedAt = new Date(cursor.createdAt).getTime();
+      paginatedDeployments = filteredDeployments.filter(deployment => {
+        const deploymentCreatedAt = new Date(deployment.createdAt).getTime();
+        return (
+          deploymentCreatedAt < cursorCreatedAt ||
```
jdolle, line 997: I think it'd be simpler to build this into the sql query.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -806,6 +806,217 @@ export class AppDeployments {
...
+          ${appDeploymentFields}
+        FROM
+          "app_deployments"
+        WHERE
+          "target_id" = ${args.targetId}
+          AND "activated_at" IS NOT NULL
+          AND "retired_at" IS NULL
+          ${args.filter.name ? sql`AND "name" ILIKE ${'%' + args.filter.name + '%'}` : sql``}
+        ORDER BY "created_at" DESC, "id"
+        LIMIT ${maxDeployments}
```
jdolle, line 886: To avoid setting this limit and to simplify this logic, I wonder if we should be flip the query order. So that you first collect, from Clickhouse, a set of the deployments that match the `lastUsed` filter, and then use those IDs to fetch the rest of the details from postgres. This should eliminate the overfetching and avoid any limit issues for incredibly large deployment sets

adambenhassen: Flipping the order would severely complicate the logic, including table joins across clickhouse, which is an expensive task without the data from postgres. I do not think this is worth it, nor do I think there would be 1000 deployments(?), especially considering that the idea is to tidy old deployments

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: The date filtering happens after the postgres query (using clickhouse data) so we can't paginate in sql without potentially skipping rows that would pass the filter

packages/web/docs/src/content/schema-registry/app-deployments.mdx
n1ru4l, line 1: This is best a follow up pr, as docs are deployed as time of merging the PR, before the code changes are deployed to production!

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -72,6 +102,18 @@ export default gql`
...
+    Date filter conditions (lastUsedBefore, neverUsedAndCreatedBefore) use OR semantics.
+    If no date filters are provided, all active deployments are returned.
+    The name filter uses AND semantics to narrow results.
+    Only active deployments are returned (not pending or retired).
+    """
+    activeAppDeployments(
+      first: Int
+      after: String
+      filter: ActiveAppDeploymentsFilter!
+    ): AppDeploymentConnection!
```
n1ru4l: The corresponding types need to be annotated with `@tag(name:"public")`, so they are part of the public GraphQL API.

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -2,30 +2,34 @@ import { gql } from 'graphql-modules';
...
   }
 
   enum AppDeploymentStatus {
-    pending
-    active
-    retired
+    pending @tag(name: "public")
+    active @tag(name: "public")
+    retired @tag(name: "public")
   }
```
n1ru4l, line 33: Since this enum type is added to the public API, we need to make sure they follow the naming conventions we decided for the public API, which is also a bit complicated since it is used within the `app:create` command.

 https://www.notion.so/theguildoss/Hive-Console-GraphQL-API-Design-Patterns-18ab6b71848a80678939c669699fe8e6?source=copy_link#18ab6b71848a80d5a1b1d16abd881d32

We could also omit the field for now since it is not required as an input type for the functionality introduced in this PR.

Your choice!

n1ru4l (approved):
There is one naming issue for the public GraphQL API: https://github.com/graphql-hive/console/pull/7377/changes#r2689652280

I recommend to either
- adjusting the schema to only expose types that follow our naming conventions
- remove the field from the public API for now

packages/services/api/src/modules/app-deployments/module.graphql.ts
adambenhassen: Went with omitting the field for now 👍

PR #7379, opened 2025-12-05
feat(sdk-rs): builder pattern, multiple endpoints & circuit breaker

ardatan:
Ref [CONSOLE-1605](https://linear.app/the-guild/issue/CONSOLE-1605)
Ref ROUTER-195

# Hive Rust SDK

## Breaking Changes to avoid future breaking changes, and better DX;

Switch to [Builder](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) pattern for `SupergraphFetcher`, `PersistedDocumentsManager` and `UsageAgent` structs. 

Benefits;

- No need to provide all parameters at once when creating an instance even for default values.

Example;
```rust
// Before
let fetcher = SupergraphFetcher::try_new_async(
        "SOME_ENDPOINT", // endpoint
         "SOME_KEY",
        "MyUserAgent/1.0".to_string(),
        Duration::from_secs(5), // connect_timeout
        Duration::from_secs(10), // request_timeout
        false, // accept_invalid_certs
        3, // retry_count
    )?;

// After
// No need to provide all parameters at once, can use default values
let fetcher = SupergraphFetcher::builder()
    .endpoint("SOME_ENDPOINT".to_string())
    .key("SOME_KEY".to_string())
    .build_async()?;
```

- Easier to add new configuration options in the future without breaking existing code, and increasing the number of parameters of `new` method;

Example;

```rust
let fetcher = SupergraphFetcher::try_new_async(
        "SOME_ENDPOINT", // endpoint
         "SOME_KEY",
        "MyUserAgent/1.0".to_string(),
        Duration::from_secs(5), // connect_timeout
        Duration::from_secs(10), // request_timeout
        false, // accept_invalid_certs
        3, // retry_count
        circuit_breaker_config, // Breaking Change -> new parameter added
    )?;

let fetcher = SupergraphFetcher::builder()
    .endpoint("SOME_ENDPOINT".to_string())
    .key("SOME_KEY".to_string())
    .build_async()?; // No breaking change, circuit_breaker_config can be added later if needed
```

## Circuit Breaker Implementation and Multiple Endpoints Support

Implementation of Circuit Breakers in Hive Console Rust SDK, you can learn more [here](https://the-guild.dev/graphql/hive/product-updates/2025-12-04-cdn-mirror-and-circuit-breaker)

Breaking Changes:

Now `endpoint` configuration accepts multiple endpoints as an array for `SupergraphFetcherBuilder` and `PersistedDocumentsManager`.

```diff
SupergraphFetcher::builder()
-    .endpoint(endpoint)
+    .add_endpoint(endpoint1)
+    .add_endpoint(endpoint2)
```

This change requires updating the configuration structure to accommodate multiple endpoints.

# Apollo Router Plugin
- Multiple endpoints support for `HiveRegistry` and `PersistedOperationsPlugin`

Breaking Changes:
- Now there is no `endpoint` field in the configuration, it has been replaced with `endpoints`, which is an array of strings. You are not affected if you use environment variables to set the endpoint.

```diff
HiveRegistry::new(
    Some(
        HiveRegistryConfig {
-            endpoint: String::from("CDN_ENDPOINT"),
+            endpoints: vec![String::from("CDN_ENDPOINT1"), String::from("CDN_ENDPOINT2")],
    )
)

### CODE REVIEW

File: packages/libraries/sdk-rs/src/persisted_documents.rs [hunk 1]
```diff
@@ -157,3 +130,145 @@ impl PersistedDocumentsManager {
...
+    _request_timeout: Duration,
+    /// Interval at which the Hive Console should be retried upon failure.
+    ///
+    /// By default, an exponential backoff retry policy is used, with 3 attempts.
+    _retry_policy: ExponentialBackoff,
+    /// Configuration for the size of the in-memory caching of persisted documents.
+    _cache_size: u64,
+    /// User-Agent header to be sent with each request
+    _user_agent: Option<String>,
+}
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist it is ok

packages/libraries/sdk-rs/src/supergraph_fetcher.rs
```diff
@@ -258,3 +198,182 @@ impl<AsyncOrSync> SupergraphFetcher<AsyncOrSync> {
...
+
+pub struct SupergraphFetcherBuilder {
+    _endpoint: Option<String>,
+    _key: Option<String>,
+    _user_agent: Option<String>,
+    _connect_timeout: Duration,
+    _request_timeout: Duration,
+    _accept_invalid_certs: bool,
+    _retry_policy: ExponentialBackoff,
+}
```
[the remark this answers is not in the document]
ardatan: @gemini-code-assist it is ok

.github/workflows/tests-integration.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
   integration:
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
```
dotansimha, line 26: why?

packages/libraries/router/src/main.rs
```diff
@@ -21,6 +21,7 @@ fn main() {
     register_plugins();
 
     // Initialize the Hive Registry and start the Apollo Router
+    // TODO: Look at builder pattern in Executable::builder().start()
```
dotansimha, line 24: ?

packages/libraries/sdk-rs/src/supergraph_fetcher/sync.rs
```diff
@@ -0,0 +1,191 @@
...
+impl SupergraphFetcher<SupergraphFetcherSyncState> {
+    pub fn fetch_supergraph(&self) -> Result<Option<String>, SupergraphFetcherError> {
+        let mut last_error: Option<SupergraphFetcherError> = None;
+        let mut last_resp = None;
+        for (endpoint, circuit_breaker) in &self.state.endpoints_with_circuit_breakers {
+            let resp = {
+                circuit_breaker
+                    .call(|| {
+                        let request_start_time = SystemTime::now();
+                        // Implementing retry logic for sync client
```
dotansimha, line 29: why implement it from scratch? https://docs.rs/reqwest-retry/latest/reqwest_retry/

dotansimha: if we have to, please move the whole retry flow into a fn, it's messy and code is super nested and hard to follow

dotansimha: ok i saw https://github.com/TrueLayer/reqwest-middleware/issues/32#issuecomment-1047677936
but maybe it can be an actual middleware?

packages/libraries/sdk-rs/src/supergraph_fetcher/mod.rs
```diff
@@ -0,0 +1,72 @@
...
+    ResponseParse(reqwest::Error),
+    ETagRead(TryLockError),
+    ETagWrite(TryLockError),
+    InvalidKey(InvalidHeaderValue),
+    MissingConfigurationOption(String),
+    RejectedByCircuitBreaker,
+    CircuitBreakerCreation(CircuitBreakerError),
+}
+
+impl Display for SupergraphFetcherError {
```
dotansimha: you can use `thiserror` to make it simpler and have the message next to each variant

packages/libraries/sdk-rs/src/supergraph_fetcher/builder.rs
```diff
@@ -0,0 +1,135 @@
...
+            connect_timeout: Duration::from_secs(5),
+            request_timeout: Duration::from_secs(60),
+            accept_invalid_certs: false,
+            retry_policy: ExponentialBackoff::builder().build_with_max_retries(3),
+            circuit_breaker: None,
+        }
+    }
+}
+
+impl SupergraphFetcherBuilder {
```
dotansimha, line 37: what about this? https://crates.io/crates/derive_builder
semms to simplify things

dotansimha: or https://bon-rs.com/guide/overview

dotansimha (changes requested):
please check comments

thanks

packages/libraries/sdk-rs/src/supergraph_fetcher/sync.rs
ardatan: I don't know what you mean by an actual middleware. Could you be more specific?

.github/workflows/tests-integration.yaml
ardatan: Because when an integration test matrix element(which is flakey) fails, it causes the others stop (which takes a lot of time). With this `fail-fast`, the others won't stop so we can retry only the failing one instead of all of them.

packages/libraries/sdk-rs/src/supergraph_fetcher/mod.rs
ardatan: It is moved from https://github.com/graphql-hive/console/pull/7379/changes#diff-51633905125dfd9a74bf6a25fee198e654c79d1b2b1d286895dae898bffd4d78L49 actually.

packages/libraries/sdk-rs/src/supergraph_fetcher/sync.rs
ardatan: Also the logic is moved from https://github.com/graphql-hive/console/pull/7379/changes#diff-51633905125dfd9a74bf6a25fee198e654c79d1b2b1d286895dae898bffd4d78L139.

packages/libraries/sdk-rs/src/supergraph_fetcher/mod.rs
ardatan: I changed it to thiserror anyways 👍

packages/libraries/router/src/main.rs
ardatan: Oh actually it was a leftover comment that means "We should check Apollo's internal supergraph loading logic to integrate with our fetcher in a better way" so that we don't have to implement sync and async fetchers etc.

packages/libraries/sdk-rs/src/supergraph_fetcher/builder.rs
ardatan: First I saw derive_builder but bon looks interesting. It supports even regular functions. I'll take a look later.

dotansimha: Great, thanks! 
Yeah `bon` seems to be popular and solves most of the use-cases in a reall nice way

packages/libraries/sdk-rs/src/supergraph_fetcher/sync.rs
dotansimha: I mean, just like `reqwest_retry` exposes the logic (for async) using the `reqwest_middleware`, we can create `ReqwestBlockingRetryMiddleware` and use it here, instead of having the retry code inside the code of the fetcher.

.github/workflows/tests-integration.yaml
dotansimha: Let's revert it here and raise it in the internal channel of Console, I think they have value in failing fast to avoid wasting CI times in some cases? (cc @n1ru4l )

packages/libraries/sdk-rs/src/supergraph_fetcher/sync.rs
ardatan: reqwest_middleware doesn't support blocking clients at all. Not sure how to do that?

dotansimha: Ah ok, I see now. I'm resolving this one for now. The effort of splitting is too high then

.github/workflows/tests-integration.yaml
n1ru4l: I agree with @ardatan here

packages/libraries/sdk-rs/src/supergraph_fetcher/builder.rs
ardatan: I just realized that our use cases are a bit unique. For example, SupergraphFetcher varies into Async and Sync, Circuit Breaker builder builds a circuit breaker instance which uses a struct from a 3rd party package. I think we can start with this then gradually improve to existing macros. I am sure there are ways to use that macros with tons of configuration but not sure if it is worth right now. What do you think?

dotansimha: Not worth right now

ardatan: So do you think this is ready to merge?

PR #7383, opened 2025-12-08
persistent job queue with postgraphile worker

n1ru4l:
### Background

If Redis dies we loose pending jobs for pending emails and webhooks

### Description

- Merges emails and webhook service + additional schema check purge "cron job" that was running in the server service into the workflow service powered by postgraphile worker
- we now use the pg database as the job backend, which ensure we are not loosing any jobs :)
- provides a simple interface so everyone can easily add new tasks and schedule them
- removes HTTP calls for enqueuing tasks and directly uses existing pg connection within services

#### Deployment Strategy

- In order to not loose any existing tasks in the redis queue, we will deploy this service along-side the existing emails and webhooks service. Then once the redis emails and webhooks queue is drained we continue with the next step
- A follow up PR will remove the emails and webhooks service and update the docker compose files
- We deploy the old service removal to production
- We release a new major self-hosted version with the new services

### Checklist

- [x] Input validation
- [x] Testing


packages/services/server/src/index.ts
```diff
@@ -209,50 +209,6 @@ export async function main() {
     }),
   }) as HivePubSub;
 
-  let dbPurgeTaskRunner: null | ReturnType<typeof createTaskRunner> = null;
```
n1ru4l, line 212: This is now a cron in the workflow service

packages/services/api/src/modules/alerts/providers/adapters/webhook.ts
```diff
@@ -1,37 +1,20 @@
...
 
   constructor(
     logger: Logger,
-    private http: HttpClient,
-    @Inject(WEBHOOKS_CONFIG) private config: WebhooksConfig,
-    @Inject(CONTEXT) context: GraphQLModules.ModuleContext,
+    private taskScheduler: TaskScheduler,
   ) {
     this.logger = logger.child({ service: 'WebhookCommunicationAdapter' });
-    this.webhooksService = createTRPCProxyClient<WebhooksApi>({
```
n1ru4l, line 24: HTTP RPC calls are no longer necessary, as we can just enqueue the job with the existing pg connection of the service.

packages/services/api/src/modules/shared/lib/task-runner.ts
n1ru4l, line 1: This weird construct is now just a cron job definition in the worker.

packages/services/workflows/src/lib/webhooks/send-webhook.ts
n1ru4l, line 1: this file is copied from webhooks service. Once the webhook service is removed, the duplication is removed.

packages/services/workflows/src/lib/expired-schema-checks.ts
n1ru4l, line 1: this was moved here from the server (graphql api) service

packages/services/workflows/src/lib/emails/providers.ts
```diff
@@ -0,0 +1,165 @@
...
+          'X-Postmark-Server-Token': config.token,
+        },
+        body: JSON.stringify({
+          From: emailFrom,
+          To: email.to,
+          Subject: email.subject,
+          HtmlBody: email.body,
+          MessageStream: config.messageStream,
+        }),
+      });
```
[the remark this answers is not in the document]
n1ru4l: This is just copy-pasting code. It worked like this before and we have built-int retries via the queue.

patches/slonik@30.4.4.patch
n1ru4l, line 1: This modification allows us to re-use the slonik pg pool with postgraphile-runner and avoid having multiple active connection pools.

packages/services/workflows/src/lib/emails/templates/audit-logs-report.ts
n1ru4l, line 1: All files in this folder are just copied over from the emails service. The duplication is eliminated after we remove the emails service.

packages/services/workflows/src/tasks/purge-expired-dedupe-keys.ts
n1ru4l, line 1: One thing that is not built into postgraphile-worker is deduplication, as completed jobs are automatically deleted. With a small layer we can achieve the same functionality and have postgraphile-worker clean up itself with a cron job.

We currently use this functionality for the rate limit emails for usage reporting.

integration-tests/testkit/emails.ts
```diff
@@ -17,5 +17,11 @@ export async function history(): Promise<Email[]> {
...
   });
 
-  return response.json();
+  const result: Email[] = await response.json();
+
+  if (!forEmail) {
+    return result;
+  }
+
+  return result.filter(result => result.to === forEmail);
```
jdolle, line 26: should we bother parsing/validating the payload here or assume the test will validate

packages/migrations/src/actions/2025.12.12T00-00-00.workflows-deduplication.ts
```diff
@@ -0,0 +1,17 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.12.12T00-00-00.workflows-deduplication.ts',
+  run: ({ sql }) => sql`
+      CREATE TABLE "graphile_worker_deduplication" (
+          "task_name"   text        NOT NULL,
+          "dedupe_key"  text        NOT NULL,
+          "expires_at"  timestamptz NOT NULL,
+          CONSTRAINT "dedupe_pk" PRIMARY KEY ("task_name", "dedupe_key")
```
jdolle, line 10: minor: if we use a uuid we shouldnt need to use the task name in the constraint.

packages/services/api/src/modules/alerts/providers/adapters/webhook.ts
```diff
@@ -41,20 +24,10 @@ export class WebhookCommunicationAdapter implements CommunicationAdapter {
...
-    } catch (error) {
-      const errorText =
-        error instanceof Error
-          ? error.toString()
-          : typeof error === 'string'
-            ? error
-            : JSON.stringify(error);
-      this.logger.error(`Failed to send Webhook notification`, errorText);
-    }
+    await this.taskScheduler.scheduleTask(SchemaChangeNotificationTask, {
```
jdolle, line 27: even though it's unlikely to fail, should we keep the try/catch logic?

packages/services/workflows/src/dev.ts
```diff
@@ -0,0 +1,15 @@
...
+config({
+  debug: true,
+  encoding: 'utf8',
+});
+
+await import('./index.js');
+
+// Not having this caused hell
+process.stdout.on('error', function (err) {
+  if (err.code == 'EPIPE') {
```
jdolle, line 12: this happens when writing to a closed stream. I suspect this means something isnt being shut down correctly.

packages/services/workflows/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+      if (stopHttpHeartbeat) {
+        logger.info('Stop HTTP heartbeat');
+        stopHttpHeartbeat();
+        logger.info('HTTP heartbeat stopped');
+      }
+      logger.info('Stopping HTTP server');
+      await server.close();
+      logger.info('HTTP server stopped');
+    } catch (error: unknown) {
+      logger.error({ error }, 'Unepected error occured');
```
jdolle:
```suggestion
      logger.error({ error }, 'Unexpected error occurred');
```

packages/services/workflows/src/index.ts
```diff
@@ -0,0 +1,154 @@
...
+
+registerShutdown({
+  logger: bridgeFastifyLogger(logger),
+  async onShutdown() {
+    try {
+      logger.info('Stopping task runner.');
+      await runner.stop();
+      logger.info('Task runner shutdown successful.');
+      logger.info('Shutdown postgres connection.');
+      await pg.end();
```
jdolle, line 134: This may be the cause of the EPIPE error mentioned before. I think you want to terminate the server and then the pg instance. Else another request could come in and pg will have been closed already and cannot be used to handle the request.

packages/services/workflows/src/kit.ts
```diff
@@ -0,0 +1,183 @@
...
+      {
+        'job.taskId': taskDefinition.name,
+        'job.id': job.id,
+      },
+      'task enqueued.',
+    );
+  }
+
+  async dispose() {
+    await (await this.tools).release();
```
jdolle, line 181: minor: should this return the promise to allow awaiting by the caller?

packages/services/workflows/src/lib/emails/providers.ts
```diff
@@ -0,0 +1,165 @@
...
+  provider: 'sendmail';
+};
+
+type EmailProviderConfig =
+  | SMTPEmailProviderConfig
+  | PostmarkEmailProviderConfig
+  | SendmailEmailProviderConfig
+  | MockEmailProviderConfig;
+
+export function createEmailProvider(config: EmailProviderConfig, emailFrom: string): EmailProvider {
```
jdolle, line 57: I might have missed it, but i dont see all of these documented in the settings. For self hosters, this would be good to have.

packages/migrations/src/actions/2025.12.12T00-00-00.workflows-deduplication.ts
n1ru4l: The task name is a string such as `sendEmail` and not an UUID. The constraint is here to avoid duplicated inserts for a task name and dedupe key combination. See the corresponding logic here: https://github.com/graphql-hive/console/pull/7383/changes#diff-8f629d86d9620c8d0ea5d15975a8bd80caee59069dbb8aeb7db86e35f0edd38eR138

packages/services/workflows/src/index.ts
n1ru4l: Only tasks use the database, this service only has an HTTP endpoint for prometheus and health checks.
The EPIPE is related to running the services in dev mode locally and seems to be an known issue in all services, but nothing that affects production.

integration-tests/testkit/emails.ts
```diff
@@ -6,8 +6,8 @@ export interface Email {
   body: string;
 }
 
-export async function history(): Promise<Email[]> {
-  const emailsAddress = await getServiceHost('emails', 3011);
+export async function history(forEmail?: string): Promise<Email[]> {
+  const emailsAddress = await getServiceHost('workflows', 3014);
```
dotansimha: nit: `const emailsAddress` could be `const workflowsEndpoint`

dotansimha (approved):
:chefkiss: 

looks great to me, API is clear and easier than what we had before. Also the fact that we persist tasks is a great win.  Also happy to see we are keep on reducing the services count, especially around those that doesn't make sense to have as standalone.

PR #7387, opened 2025-12-08
docs(router): plugin system

ardatan:
Documentation of https://github.com/graphql-hive/router/pull/482

### DIFF
--- a/packages/web/docs/src/content/router/extensibility/plugin_system.mdx
+++ b/packages/web/docs/src/content/router/extensibility/plugin_system.mdx
@@ -223,7 +223,7 @@ access the `context` field.
 For example, you can store some data in the context during the `on_graphql_params` hook and retrieve
 it later in the `on_subgraph_execute` hook.
 
-```rust filename="src/context_data.rs"
+```rust
 pub struct ContextData {
     incoming_data: &'static str,
 }
@@ -1450,11 +1450,10 @@ are also available to simplify creating early responses.
 To make it easier to create early responses, Hive Router provides some helper methods on the hook
 payloads.
 
-- `end_with_response(response: Response)` - Accepts a
-  `serde`-serializable body, automatically serializes it to JSON, and sets the appropriate headers.
-  To simplify the above example:
+- `end_with_response(response: Response)` - Accepts a `serde`-serializable body, automatically
+  serializes it to JSON, and sets the appropriate headers. To simplify the above example:
 
-```rust filename="src/forbid_anon_ops.rs" {33-52}
+```rust
                 let body = json!({
                     "errors": [
                         {
@@ -1465,7 +1464,7 @@ payloads.
                         }
                     ]
                 });
-                
+
                 // Here we short-circuit the request processing by returning an early response
                 return payload.end_with_response(
                     Response::with_body(
@@ -1478,7 +1477,7 @@ payloads.
 - `end_with_graphql_error(error: GraphQLError, status: StatusCode)` - Accepts a `GraphQLError`
   directly and constructs the appropriate GraphQL error response.
 
-```rust filename="src/forbid_anon_ops.rs" {33-52}
+```rust
                 let graphql_error = GraphQLError::from_message_and_code("Anonymous operations are not allowed", "ANONYMOUS_OPERATION");
                 // Here we short-circuit the request processing by returning an early response
                 return payload.end_with_graphql_error(


packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+## Custom Builds with Rust Plugins
+
+Hive Router is built using Rust, which allows for high performance and safety. One of the powerful
+features of Hive Router is the ability to create custom builds with your own Rust plugins. This
+enables you to add new capabilities or modify existing ones to better suit your requirements.
+
+### Create a new Rust project
+
+First, ensure you have the necessary development environment set up for
+[Rust 1.91.1 or later](https://rust-lang.org/tools/install/). Then, you need to create a new Rust
```
dotansimha: How did we decide 1.91.1, I mean, we don't have an official MSRV defined for the Crate (maybe we should?)

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+```bash
+cargo new --bin my_custom_router
+cd my_custom_router
+```
+
+Install `hive-router` as a dependency by adding it to your `Cargo.toml` file:
+
+```toml
+[dependencies]
+hive-router = "0.17"
```
dotansimha: `ntex` dependency is missing here, and it needs to be added because we don't expose it. The `main` example below assumes that you have `ntex` and then it fails.

dotansimha: Should be `ntex = "2"`

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+```
+
+You can use our example supergraph as a starting point;
+
+```bash
+curl -sSL https://federation-demo.theguild.workers.dev/supergraph.graphql > supergraph.graphql
```
dotansimha, line 85: What about subgraphs here? Are they already running somewhere? add a note please

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+```
+
+Then point to that supergraph in your `router.config.yaml`:
```
dotansimha: `Create a test/development config file to help you to easily run the router, call it  `router.config.yaml` and place it in the root of the project:`

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+
+```yaml filename="router.config.yaml"
+supergraph:
+  source: file
+  path: ./supergraph.graphql
+```
+
+Or you can use other ways to provide the supergraph, see
```
dotansimha: I think this is not really needed here, if someone already landed in the custom plugin system, they already know the docs and the fact they can use different sources.

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+### Create an entrypoint for your custom router
+
+Next, you need to create an entrypoint for your custom router. This is where you'll initialize the
+router and register your plugins. Create a new file `src/main.rs` and add the following code:
+
+```rust
+use hive_router::{PluginRegistry, router_entrypoint, BoxError};
+
+#[ntex::main]
+async fn main() -> Result<(), BoxError> {
```
dotansimha: `BoxError` -> `RouterInitError`

Also missing import for `RouterInitError`

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+
+### Create an entrypoint for your custom router
+
+Next, you need to create an entrypoint for your custom router. This is where you'll initialize the
+router and register your plugins. Create a new file `src/main.rs` and add the following code:
+
+```rust
+use hive_router::{PluginRegistry, router_entrypoint, BoxError};
+
+#[ntex::main]
```
dotansimha: In Hive Router, we are using a custom allocator, and users needs to know that and use it as well, otherwise it might cause unexpected runtime issues.

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+
+```bash
+cargo new --bin my_custom_router
+cd my_custom_router
+```
+
+Install `hive-router` as a dependency by adding it to your `Cargo.toml` file:
+
+```toml
+[dependencies]
```
dotansimha, line 41: I'm wondering about what needs to be exported from `hive-router` crate in order to make it easy for users. 

We already know that users needs 2 crucial things at the entry point, in order to be able to just create the first `main.rs`:

- ntex (v2)
- mimalloc

Should we export from `hive-router` some defaults in order to make it easier for users? I mean, it can be `#[hive_router::main]` that wraps `ntex::main`.

And it can be `hive_router::Allocator` to be used with `#[global_allocator]` for the sake of simplicty, or if we'll update this at some point in our codebase. 

I can imagine that users who are using a custom build won't deal with custom allocator/http server that's used by Hive-Router, and they will not want to deal with changing/upgrading those depedencies.

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+
+### Run your custom router
+
+Finally, you can build and run your custom router using Cargo:
+
+```bash
+cargo run
+```
+
+### Configure your plugins in `router.config.yaml`
```
dotansimha: Before this section, we need a dead simple example of creating `struct MyPlugin` and then implementing the trait for it. 

Then, an example snippet on how to register a plugin in the registry. 

Then, an example on how to add configuration to the plugin. 

And only then, an example on how to configure the plugin the config file.

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+`router.config.yaml` file like this:
+
+```yaml filename="router.config.yaml"
+plugins:
+  allow_client_id_from_file:
+    enabled: true
+    header: 'x-client-id'
+```
+
+## Registration of Plugins
```
dotansimha: Also, we need to clearly mention that registering a plugin doesn't mean it's activated. 

Consider the following case: a plugin is registered but doesn't have configuration. How will it be enabled in the config file?

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
dotansimha: Having a custom plugin involved 3 different parts and it should be very clear for the user that they need all 3:
- Create the plugin (actual implementation) 
- Register the plugin (to make it available for the router to load) 
- Enable the plugin and configure it

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+## Short Circuit Responses
+
+In many of the hooks mentioned above, you have the ability to short-circuit the request processing
+by providing a custom response.
+
+Let's say you want to implement a plugin that returns an early error response if it doesn't have a
+specific operation name. So basically this plugin rejects anonymous operations. See the highlighted
+code below;
+
+```rust filename="src/forbid_anon_ops.rs" {33}
```
dotansimha: While this example is great, it should be simpler. I would try to avoid copying a full plugin code here, and minimize the snippet so it will focus on the `end_with_response` call and how/when to use it.

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+
+Next, you need to create an entrypoint for your custom router. This is where you'll initialize the
+router and register your plugins. Create a new file `src/main.rs` and add the following code:
+
+```rust
+use hive_router::{PluginRegistry, router_entrypoint, BoxError};
+
+#[ntex::main]
+async fn main() -> Result<(), BoxError> {
+    /// This is where you can register your custom plugins
```
dotansimha: missing `init_rustls_crypto_provider` 
it fails with

```
Call CryptoProvider::install_default() before this point to select a provider manually, or make sure exactly one of the 'aws-lc-rs' and 'ring' features is enabled.

```

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+
+## Registration of Plugins
+
+Finally, to use your custom plugin, you need to register it with the `PluginRegistry` in your
+`main.rs` file.
+
+```rust filename="src/main.rs" {9-14}
+    let mut plugin_registry = PluginRegistry::new();
+    // Register your custom plugin
+    plugin_registry.register_plugin::<AllowClientIdFromFilePlugin>();
```
dotansimha: fn name is `register` and not `register_plugin`. 

Also, since it's returning `self`, is should really be: 

```rs
 let mut plugin_registry = PluginRegistry::new();
plugin_registry = plugin_registry.register_plugin::<AllowClientIdFromFilePlugin>();
```

But in any way, let's change it to be simpler and more readable: 
```rs
 let plugin_registry = PluginRegistry::new().register::<AllowClientIdFromFilePlugin>();
```

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
```diff
@@ -0,0 +1,972 @@
...
+`context.insert<T>`, `context.get_ref<T>`, `context.get_mut<T>` methods are used to store and
+retrieve data of type `T` in the request context.
+
+- `insert<T>(&self, data: T)` - Inserts data of type `T` into the context.
+- `get_ref<T>(&self) -> Option<&T>` - Retrieves a reference to the data of type `T` from the
+  context.
+- `get_mut<T>(&mut self) -> Option<&mut T>` - Retrieves a mutable reference to the data of type `T`
+  from the context.
+
+## Refresh State on Supergraph Reload
```
dotansimha: This section needs to be written differently in my opinion:
First, it needs to explain the lifecycle and the fact that a supergraph can change in background. 
Then, show the hook and how to respond to it if needed.

And then a seperate section called `Depending on Supergraph in your plugin state` that explains the usage of `ArcSwap` and how it can help to avoid these syncing issues

packages/web/docs/src/content/router/guides/extending-the-router/index.mdx
dotansimha: Also, after a very short while, users need `serde` and `Deserializable` for the config. It's not mentioned anywhere on how to install it.

ardatan: Actually Kamil and I discussed about it before, we will probably re-export things to avoid users to add extra dependencies

dotansimha: Yes, exactly

dotansimha: And it's not only to reduce the amount of dependencies, it's also to make it simpler for users to build plugins without maintaining the dependencies, or even knowing most of them.

ardatan: Avoiding extra dependencies is part of making it simpler

PR #7388, opened 2025-12-08
feat(api): conditional breaking changes based on app deployments

adambenhassen:
Closes CONSOLE-1276

### DIFF
--- a/integration-tests/tests/cli/schema.spec.ts
+++ b/integration-tests/tests/cli/schema.spec.ts
@@ -1098,9 +1098,7 @@ test.concurrent(
       expect.fail('Expected schema check to fail with breaking changes');
     } catch (error: any) {
       const output = stripAnsi(error.message || error.stderr || String(error));
-      expect(output).toContain('test-app@1.0.0');
-      expect(output).toContain('GetUserEmails');
-      expect(output).toContain('GetUserProfile');
+      expect(output).toContain('[1 app deployment affected]');
     }
   },
 );


packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -806,6 +829,359 @@ export class AppDeployments {
...
+          WHERE
+            "app_deployment_id" IN (${cSql.array(deploymentIds, 'String')})
+            AND hasAny("schema_coordinates", ${cSql.array(args.schemaCoordinates, 'String')})
+          ORDER BY "app_deployment_id", "document_hash"
+          LIMIT 1 BY "app_deployment_id", "document_hash"
+          LIMIT 10000
+        `,
+        queryId: 'get-affected-app-deployments-by-coordinates',
+        timeout: 30_000,
+      });
```
n1ru4l, line 1071: We can also get a list of active app deployments from only clickhouse using the `app_deplyments` table.

We can also "join" (or inner select) the clickhouse tables so we do not have to set magic number limits like `1000`, but can scan until we have at least x affected app deployments with at least x affected operating per affected app deployment

- We want to show all affected app deployments but maximum 10 app deployments (no not break UI and have a cazy pg schema check column payload)
- We want to show maximum 10 affected operations per affected app deployment

If users want to see more we could still send people to the app deployment view and filter there based on a passed schema coordinate 🤔

integration-tests/tests/api/app-deployments.spec.ts
```diff
@@ -1835,3 +1867,2014 @@ test('app deployment usage reporting', async () => {
   }).then(res => res.expectNoGraphQLErrors());
   expect(data.target?.appDeployment?.lastUsed).toEqual(expect.any(String));
 });
+
+test('activeAppDeployments returns empty list when no active deployments exist', async () => {
```
n1ru4l, line 1871: nit :D 
```suggestion
test('Target.activeAppDeployments returns empty list when no active deployments exist', async () => {
```

n1ru4l (changes requested):
The solution for getting the affected documents can be better and sorely on clickhouse data: https://github.com/graphql-hive/console/pull/7388/files#r2610298291

also see https://guild-oss.slack.com/archives/C03AQ477ECF/p1765456339391569?thread_ts=1765199461.488359&cid=C03AQ477ECF

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: Good point! I actually studied and implemented this to see how it would work.

The single clickhouse query approach over-complicates things because `app_deployments` is a `ReplacingMergeTree`, which requires `FINAL` to deduplicate rows (to make sure `is_active` is not `true` even after retiring, and not wait for background merge to happen at some point), then also the inner join for documents. That could be heavy on a large tables compared to just fetching 10 or 1000 deployment IDs from postgresql first.

The current approach (postgres for active deployment IDs, then clickhouse for documents) is both **cheaper** and simpler, we basically avoid deduplicating and inner joining a potentially huge table entirely.

I strongly think we should keep the current implementation, the single clickhouse query is not worth it.

n1ru4l: Couldn't we do a `GROUP BY` for the `app_deployments` table? We use it for a bunch of aggregations in other queries (though these queries are not using `ReplacingMergeTree`). I am fine with the query being slightly underperforming and later on improve once we hit scaling issues.

```
SELECT
  target_id,
  app_name,
  app_version,
  max(is_active) AS is_active
FROM
"app_deployments"
GROUP BY "target_id", "app_name", "app_version";
```

____

Aside from that the main issue I see in this implementation is that having more than 1000 app deployments can lead to a breaking change for one app deployment to be swallowed and not being reported, which would lead to wrong result and indicators for our customers, potentially leading to broken GraphQL clients.

I would therefore suggest one of the following:
- Using a single ClickHouse query (even though it is not the most performant one), as suggested above
- Use stable key-set based pagination for postgres and over-fetch by one (1001), and then execute against clickhouse until we have at least 10 affected clients (that have up to 10 operations referenced each) or there is no page left in postgres

adambenhassen: At first, I thought the query you wanted to improve was the postgres but then I realized it wasn't.

Ok, so, once again, the GROUP BY with max(is_active) addition you mentioned has an issue with ReplacingMergeTree. If a deployment was active then retired, max() would still return 1 (since there might be both is_active=1 and is_active=0 rows before merge), this needs the FINAL to get the deduplicated state. 

Then comes the issue with finding the documents, which results in an inner join, as well as nested array manipulations for the schema coordinates. 

If it's ok with you to bloat the query for potential memory issues instead of being stuck at 1000 for a non-existing case of concurrent active app deployments per a single target, then that's fine with me. Implementing.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -806,6 +829,314 @@ export class AppDeployments {
...
+        query: cSql`
+          SELECT
+            d."app_deployment_id" AS "appDeploymentId"
+            , d."app_name" AS "appName"
+            , d."app_version" AS "appVersion"
+            , groupArray((doc."document_hash", doc."operation_name")) AS "operations"
+            , arrayDistinct(arrayFlatten(groupArray(arrayIntersect(doc."schema_coordinates", ${cSql.array(args.schemaCoordinates, 'String')})))) AS "matchingCoordinates"
+          FROM (
+            SELECT * FROM "app_deployments" FINAL
+            WHERE "target_id" = ${args.targetId} AND "is_active" = true
```
n1ru4l: What is the difference of this query compared to the following:

```
SELECT
  target_id,
  app_name,
  app_version,
  max("is_active") AS "is_active"
FROM
  "app_deployments"
GROUP BY "target_id", "app_name", "app_version"
HAVING "is_active" = true
```

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -485,6 +485,21 @@ export default gql`
...
     usageStatistics: SchemaChangeUsageStatistics @tag(name: "public")
+    """
+    List of active app deployments that would be affected by this breaking change.
+    Only populated for breaking changes when app deployments are enabled.
+    """
+    affectedAppDeployments(
+      first: Int
+      after: String
+      firstOperations: Int
+    ): SchemaChangeAffectedAppDeploymentsConnection @tag(name: "public")
```
n1ru4l, line 496: Why the decision to make this a new top level field on the schema change instead of integrating it into the existing `usageStatistics` field, it seems like it would logically fit in there. 🤔 

As per the pagination arguments, I think we can safely omit these, as done with `SchemaChangeUsageStatistics.topAffectedOperations` and `SchemaChangeUsageStatistics.topAffectedClients`, as these represent a static "sample" snapshot at the time the schema check is being created, thus returning a list instead of a connection seems good enough. I don't really think we need to support offset pagination in here.

packages/services/storage/src/schema-change-model.ts
```diff
@@ -1299,6 +1299,24 @@ export const HiveSchemaChangeModel = z
         .nullable()
         .optional()
         .transform(value => value ?? null),
+      /** App deployments affected by this breaking change */
+      affectedAppDeployments: z
```
n1ru4l, line 1303: As mentioned in the GraphQL schema, doesn't this logically group within the `usageStatistics` property?

packages/services/api/src/modules/schema/resolvers/SchemaChange.ts
```diff
@@ -33,4 +34,81 @@ export const SchemaChange: SchemaChangeResolvers = {
...
   severityLevel: change => severityMap[change.criticality],
   severityReason: change => change.reason,
+  affectedAppDeployments: async (change, args, { injector }) => {
+    const coordinate = change.path;
+    if (!coordinate) {
+      return null;
+    }
+
+    // If targetId is available, query ClickHouse with limits
+    if (change.targetId) {
```
n1ru4l: A `SchemaCheck` is a point-in-time snapshot from when `hive schema:check` runs, so pulling the latest state from ClickHouse doesn’t align with that model - we can omit this part of the code.

packages/services/api/src/modules/schema/providers/models/composite.ts
```diff
@@ -487,13 +521,15 @@ export class CompositeModel {
       existingSdl: previousVersionSdl,
       incomingSdl: compositionCheck.result?.fullSchemaSdl ?? null,
       failDiffOnDangerousChange,
+      getAffectedAppDeployments: getAffectedAppDeploymentsForPublish,
```
n1ru4l, line 531: Is there a specific reason for passing through the function, instead of making AppDeployments a dependency in the `RegistryChecks` constructor and passing in the the ID as a string `affectedAppDeploymentsTargetId`, like with other parameters?

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -568,6 +593,102 @@ export class RegistryChecks {
...
+        } catch (error) {
+          this.logger.error(
+            'Failed to check affected app deployments (coordinateCount=%d): %s',
+            breakingCoordinates.size,
+            error instanceof Error ? error.stack : String(error),
+          );
+          Sentry.captureException(error, {
+            tags: { operation: 'app-deployment-check' },
+            extra: { coordinateCount: breakingCoordinates.size },
+          });
```
n1ru4l: We already have some kind of global sentry handler. I do not see a need to double report the exception here inline.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -568,6 +593,102 @@ export class RegistryChecks {
...
+          throw new Error(
+            `Unable to verify schema changes against app deployments. Please retry. If the issue persists, contact support.`,
+          );
```
n1ru4l: This error will never reach the end-user since we do [error masking](https://the-guild.dev/graphql/yoga-server/docs/features/error-masking).

We can just throw the original error here and treat it as an unexpected error to the outside world instead of instructing people to retry.

packages/services/api/src/modules/schema/providers/schema-check-manager.ts
```diff
@@ -40,23 +40,35 @@ export class SchemaCheckManager {
...
   getBreakingSchemaChanges(schemaCheck: SchemaCheck) {
     if (!schemaCheck.breakingSchemaChanges?.length) {
       return null;
     }
 
-    return schemaCheck.breakingSchemaChanges;
+    return schemaCheck.breakingSchemaChanges.map(change => ({
+      ...change,
+      targetId: schemaCheck.targetId,
+    }));
```
n1ru4l: Drilling in the `targetId` breaks conditional breaking change information, since we pass that through resolvers based on object identity.

While passing the information based on object identity is not optional:
Since we have no need to retrieve data from clickhouse after the creation of a schema check (and sorely rely on the static data stored in postgres "snapshot at the schema check run time"), we can omit passing the `targetId` here.

n1ru4l: The same applies for the other places where this is done.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -568,6 +593,102 @@ export class RegistryChecks {
       this.logger.debug('No conditional breaking change settings available');
     }
 
+    // Check against active app deployments if function is provided
+    if (args.getAffectedAppDeployments) {
```
n1ru4l, line 606: NIT: technically we are doing very similar logic to the loop above for conditional breaking changes. While I see no need to merge these loops, we could use the data-loader pattern, to make the code reading more "sync" (like above).

Not a big deal though.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -568,6 +593,102 @@ export class RegistryChecks {
...
+          (
+            change as {
+              affectedAppDeployments: Array<{
+                id: string;
+                name: string;
+                version: string;
+                affectedOperations: Array<{ hash: string; name: string | null }>;
+              }>;
+            }
+          ).affectedAppDeployments = [];
```
n1ru4l: Is the `as` cast here needed? The types should already be properly inferred from the change model definition?

packages/services/api/src/modules/schema/module.graphql.ts
n1ru4l: Okay, after digging deeper into the code, I realised that `usageStatistics` is deeply tied to the conditional breaking change configuration, which made me think whether an active app deployment existing is kind of a conditional breaking change, and whether we also use the "ignored client names" list for ignoring app deployments with the same name. 🤔

Thinking about our own set up (Hive Console Dashboard), we would still want to ignore breaking changes that would only affect our web app, as we deploy it together with our GraphQL API.

So the question we actually need to answer here is:
Do we want to treat conditional breaking changes about just being about actual schema coordinate usage data; or should it be a hybrid of the former and app deployments.

packages/services/api/src/modules/schema/providers/registry-checks.ts
adambenhassen: Agreed, a data-loader pattern would make this read more cleanly while keeping the batching. Will keep in mind for a future cleanup pass

packages/services/api/src/modules/schema/providers/models/composite.ts
adambenhassen: Tried this, but it creates a cyclic dependency: 
`SchemaVersionHelper -> RegistryChecks ->AppDeployments ->SchemaVersionHelper` 
With the callback approach, the function is created at call time rather than injection time, which avoids the cycle
Let me know if there's a better approach we can take

adambenhassen:
Documenting small user interface changes:
From:
Screenshot 2026-01-09 at 14 40 53

To:
Screenshot 2026-01-09 at 14 38 32

adambenhassen:
Documenting slight CLI changes: I think compacting it to just show the amount of app deployments affected is more useful than displaying each one.

Before:
```
ℹ Detected 2 changes

   Breaking changes:
   - Field users was removed from object type Query
      - test-app-02@1.0.2, payment-gateway@1.0.27, mobile-android@1.0.22, checkout-service@1.0.44, admin-panel@1.0.43 and 70 more
```

After:
Screenshot 2026-01-09 at 15 02 53

packages/services/api/src/modules/schema/module.graphql.ts
adambenhassen: I think adding a separate configuration field `excludedAppDeployments` rather than reusing `excludedClientNames` is the best way to do it, in my opinion. Mainly because:
- Client names (from usage reporting) and app deployment names are "technically" distinct, they may or may not match depending on how users set them up
- Explicit config gives users more cotrol and clearer understanding of what's being excluded, wether they want to exclude by app deployment or clientname
- Avoids confusion about why adding a "client name" would affect app deployment filtering

So basically implementing `excludedAppDeployments` would mean:
- New field in conditional breaking changes settings: `excludedAppDeployments: [String!]`
- New ui dropdown: `Allow breking change for these app deployments`
- App deployments with names matching entries in this list are filtered out from affected deployments

What do you think?

adambenhassen: Screenshot 2026-01-09 at 16 47 37

packages/services/storage/src/schema-change-model.ts
```diff
@@ -1460,6 +1485,7 @@ export const ConditionalBreakingChangeMetadataModel = z.object({
     requestCount: z.number().default(1),
     breakingChangeFormula: z.enum(['PERCENTAGE', 'REQUEST_COUNT']).default('PERCENTAGE'),
     excludedClientNames: z.array(z.string()).nullable(),
+    excludedAppDeploymentNames: z.array(z.string()).nullable(),
```
n1ru4l: For backwards-compat with existing configurations, we need the following, otherwise there will be a runtime exception.

```suggestion
    excludedAppDeploymentNames: z.array(z.string()).optional().nullable().default(null),
```

n1ru4l: Image

```
ZodError: [
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "array",
            "received": "undefined",
            "path": [
              "conditionalBreakingChangeMetadata",
              "settings",
              "excludedAppDeploymentNames"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "received": false,
            "code": "invalid_literal",
            "expected": true,
            "path": [
              "isSuccess"
            ],
            "message": "Invalid literal value, expected true"
          },
          {
            "code": "invalid_type",
            "expected": "null",
            "received": "array",
            "path": [
              "schemaCompositionErrors"
            ],
            "message": "Expected null, received array"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "null",
            "path": [
              "compositeSchemaSDL"
            ],
            "message": "Expected string, received null"
          },
          {
            "code": "invalid_type",
            "expected": "array",
            "received": "undefined",
            "path": [
              "conditionalBreakingChangeMetadata",
              "settings",
              "excludedAppDeploymentNames"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      }
    ],
    "path": [],
    "message": "Invalid input"
  }
]
    at get error (file:///Users/laurinquast/Projects/graphql-hive-3/node_modules/.pnpm/zod@3.25.76/node_modules/zod/v3/types.js:39:31)
    at ZodUnion.parse (file:///Users/laurinquast/Projects/graphql-hive-3/node_modules/.pnpm/zod@3.25.76/node_modules/zod/v3/types.js:114:22)
    at row (/Users/laurinquast/Projects/graphql-hive-3/packages/services/storage/src/index.ts:4279:39)
    at Array.map (<anonymous>)
    at Object.getPaginatedSchemaChecksForTarget (/Users/laurinquast/Projects/graphql-hive-3/packages/services/storage/src/index.ts:4278:26)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async SchemaManager.getPaginatedSchemaChecksForTarget (/Users/laurinquast/Projects/graphql-hive-3/packages/services/api/src/modules/schema/providers/schema-manager.ts:689:29)
    at async schemaChecks (/Users/laurinquast/Projects/graphql-hive-3/packages/services/api/src/modules/schema/resolvers/Target.ts:79:20)
    at async Promise.all (index 2)
    at async Promise.all (index 1)
    at async <anonymous> (/Users/laurinquast/Projects/graphql-hive-3/packages/services/server/src/graphql-handler.ts:284:26)
    at async Object.<anonymous> (/Users/laurinquast/Projects/graphql-hive-3/packages/services/server/src/graphql-handler.ts:279:5) {
  path: [ 'target', 'filteredSchemaChecks' ],
  locations: [ { line: 27, column: 5 } ],
  extensions: [Object: null prototype] {}
}
```

packages/services/api/src/modules/schema/module.graphql.ts
n1ru4l: I agree with these!

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -485,6 +485,21 @@ export default gql`
...
+  }
+
+  type SchemaChangeAffectedAppDeploymentsConnection {
+    nodes: [SchemaChangeAffectedAppDeployment!]! @tag(name: "public")
+    totalCount: Int! @tag(name: "public")
+    pageInfo: PageInfo! @tag(name: "public")
   }
```
n1ru4l, line 503: These do not follow the GraphQL Cursor Connection Specification. We follow it for all new paginated fields. https://relay.dev/graphql/connections.htm

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -548,6 +563,54 @@ export default gql`
...
+    """
+    Total count of operations within this app deployment that use the affected schema coordinate.
+    """
+    totalAffectedOperations: Int! @tag(name: "public")
+  }
+
+  type SchemaChangeAffectedAppDeploymentOperationsConnection {
+    nodes: [SchemaChangeAffectedAppDeploymentOperation!]! @tag(name: "public")
+    pageInfo: PageInfo! @tag(name: "public")
+  }
```
n1ru4l, line 603: Same as https://github.com/graphql-hive/console/pull/7388/changes#r2686779256

If it is a `XXXXConnection` type, it should follow the cursor connection specification.

n1ru4l (changes requested):
The UI for CLI and Dashboard now looks! 🔥 

I found one major issue, regarding the backwards-compatibility of the zod model for existing database records.

Also, since we decided to go with Connection types instead of lists, we should follow the pagination properly!

Last thing is to merge/rebase onto main.

Once these two issues are addressed, let's ship it! Great work 🥳

PR #7389, opened 2025-12-08
fix: http user input error

n1ru4l:
### Background

An increased amount of security researchers penetrate our API and invalid user input errors where reported as 500 errors on sentry. 

### Description

Closes https://linear.app/the-guild/issue/CONSOLE-1621/hive-service-fastify-sdk-error-handling

Changes the error handler to
- work the same whether sentry error reporting is enabled or not
- only report non 4XX errors and treat them as 500 errors

### DIFF
--- a/.changeset/yummy-ears-accept.md
+++ b/.changeset/yummy-ears-accept.md
@@ -2,6 +2,6 @@
 'hive': patch
 ---
 
-Do not swallow 4XX HTTP errors as 500 internal server errors when setry error reporting is enabled.
+Do not swallow 4XX HTTP errors as 500 internal server errors when sentry error reporting is enabled.
 
 Send the same predictable error responses with and without the sentry plugin enabled.


packages/services/service-common/src/http-error-handler.ts
```diff
@@ -3,10 +3,30 @@ import fp from 'fastify-plugin';
...
+const plugin: FastifyPluginAsync<{ isSentryEnabled: boolean }> = async (server, opts) => {
+  if (opts.isSentryEnabled) {
+    server.decorateReply('sentry', null);
+  }
 
   server.setErrorHandler((err, req, reply) => {
+    if (err.statusCode && err.statusCode < 500) {
+      req.log.warn(err.message);
+      void reply.status(err.statusCode).send(err.message);
+      return;
```
n1ru4l, line 19: Anything with status code below 500 is an user input error and we should not report it.

.changeset/yummy-ears-accept.md
```diff
@@ -0,0 +1,7 @@
+---
+'hive': patch
+---
+
+Do not swallow 4XX HTTP errors as 500 internal server errors when setry error reporting is enabled.
```
adambenhassen: sentry*

n1ru4l:
```suggestion
Do not swallow 4XX HTTP errors as 500 internal server errors when sentry error reporting is enabled.
```

Issue #470, opened 2022-10-11
Account Linking


kamilkisiela:
Allow end users to login via different methods but access the same account.

It depends on [SuperTokens's roadmap](https://supertokens.com/product-roadmap). Currently, it's marked as "in progress".

kamilkisiela:
#424

XiNiHa:
Status update:
Currently, there are Hive users who have the same email, while being separate users because of using different sign-in methods. This happened because SuperTokens do not support account linking in the OSS version. This task aims to solve this and unify those users into a single Hive user.

PR #7390, opened 2025-12-08
Introduce Account linking for new Users

XiNiHa:
### Background

Part of #470

We can easily introduce Account Linking for users who don't have any duplicate accounts yet.

### Description

- Reuse existing Hive user when there is a user with a matching email address, even with SuperTokens users with different `supertoken_user_id`
- Assert OIDC integration ID of sessions when performing access checks against OIDC-only organizations
- Remove existing guards around OIDC-only orgs and SuperTokens-Hive 1-1 relationships

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [x] Session management
- [x] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/server/src/supertokens.ts
```diff
@@ -189,17 +189,27 @@ export const backendConfig = (requirements: {
...
+                const internalUser = await internalApi.ensureUser({
+                  superTokensUserId: user.id,
+                  email: user.emails[0],
+                  oidcIntegrationId: input.userContext['oidcId'] ?? null,
+                  firstName: null,
+                  lastName: null,
+                });
+                payload.userId = internalUser.user.id;
+                payload.oidcIntegrationId = input.userContext['oidcId'] ?? null;
+              } catch {}
```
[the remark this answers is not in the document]
XiNiHa: It is intentional to create a partial session.

packages/services/storage/src/index.ts
```diff
@@ -609,14 +626,24 @@ export async function createStorage(
...
+          SELECT
+            ${userFields(sql`"users".`, sql`"stu".`)}
+          FROM
+            "users"
+          LEFT JOIN "supertokens_thirdparty_users" AS "stu"
+            ON ("stu"."user_id" = "users"."supertoken_user_id")
+          WHERE
+            "users"."email" = ${email};
+        `);
+        let internalUser = users.length > 0 ? UserModel.parse(users[0]) : null;
```
n1ru4l: How will this affect having more than one existing user already in the system? 

I think using the first occurring user is not stable enough. If we want to always use the first one, we need a stable order (e.g. sort ascending and use the oldest existing user).

But then, we could lock out existing users from organization memberships that went through the followin flow:
1. first accidentally logged in with the wrong method; immediately logged out (at `2025-12-19-00:00:00`)
2. logged in again with different method and correctly joined the org at (at `2025-12-19-00:01:00`)

With the logic as above, we would link new login methods to randomly (1.) or (2.) based on what the query planner feels like, which could lead to further confusion in an already confusing system. 😓

If we introduce a stable sorting and always pick the oldest existing user we would always pick (1.), even though the user has already the correct org membership permissions etc. with (2.).

All of these seem suboptimal.

For this first step, I think we should sorely focus on handling the scenario here:
- no user exists
- exactly one user exists

That way we eliminate confusion for all new users (unique email), and existing users that have exactly one login method right now.

Then after we deployed the changes for new users and users with exactly one account 
-> attempt to clean the database of "ghost" records (logged in but never joined an org; probably due to login method confusion)
-> see if we proceed with session merging or inform organization admins to perform the cleanup of duplicated accounts

n1ru4l: 🤦 I should have read further first till line 655 :D

n1ru4l: The default `internalUser` should probably be the user whose `supertokensUserId` matches, not the first one? 🤔

packages/services/storage/src/index.ts
```diff
@@ -625,6 +652,16 @@ export async function createStorage(
           action = 'created';
         }
 
+        if (users.length === 1 && internalUser.superTokensUserId != null) {
```
n1ru4l: Why must `internalUser.superTokensUserId` be `null` in this case, couldn't we do it for any one "record"?

packages/services/storage/src/index.ts
```diff
@@ -636,33 +673,18 @@ export async function createStorage(
           );
         }
 
-        return action;
+        return {
+          user: internalUser,
```
n1ru4l, line 733: Related to https://github.com/graphql-hive/console/pull/7390/changes#r2635220421, wouldn't this potentially return the wrong user in a scenario where we have more than one user in our database with that email?

And causing the wrong `user.id` being used here:

Image

packages/services/storage/src/index.ts
XiNiHa: It's more like an optimization, since we don't need to write nulls if it's already null 😆

packages/services/storage/src/index.ts
XiNiHa: Would https://github.com/graphql-hive/console/pull/7390/commits/4e0ab08212d64ce39587fdcb9aad2a050d41d0d7 be enough to resolve all of the issues you pointed out?

packages/web/app/src/components/layouts/organization-selectors.tsx
n1ru4l, line 1: Just an idea: We could introduce a new field on the viewer that just returns a stripped down organization type specific for this use-case.

```
type OrganizationForSelector {
  organizationId: ID!
  slug: String!
  requiresOIDC: Boolean
  # etc.
}
```

That would not require us having to introduce more granular permissions for specific `Organization` and `OIDCIntegration` fields.

This type could then be exposed as a field on the `User` type?

We did something similar for selecting projects/targets for access tokens (see `Organization.projectsForResourceSelector`/`Organization.projectForResourceSelector`).

packages/services/server/src/supertokens.ts
```diff
@@ -167,17 +167,27 @@ export const backendConfig = (requirements: {
...
```
n1ru4l: I was wondering why userId can be undefined when looking at `SuperTokensCookieBasedSession` and this action failing seems to be the only reason why it could happen? Am I missing something here?

Would we want to continue creating the session in case the ensure user call fails or instead retry? Is there any cases where it is expected to fail? Swallowing the error here seems a bit harsh in any case.

XiNiHa: I'm pretty sure that there was a reason for this but I can't remember and I don't see any reason 🤔 Let's get rid of this and see how things go 😇

packages/services/storage/src/index.ts
```diff
@@ -462,32 +462,60 @@ export async function createStorage(
...
+        identityId,
       }: {
-        superTokensUserId: string;
         email: string;
         fullName: string;
         displayName: string;
-        oidcIntegrationId: string | null;
+        identityId: string;
       },
       connection: Connection,
```
n1ru4l: nit: we could change this to type `DatabaseTransactionConnection`, to avoid re-use outside a transaction. I couldn't find any other usage though. 👍

packages/web/app/src/components/layouts/organization-selectors.tsx
XiNiHa: It was possible to solve it in a different way: https://github.com/graphql-hive/console/pull/7390/commits/c60647ec97e46a53c3697ef2d50a311449c6789b

packages/migrations/src/actions/2026.01.30T00-00-00.account-linking.ts
```diff
@@ -0,0 +1,27 @@
...
+          , "created_at" timestamptz NOT NULL DEFAULT now()
+          , UNIQUE ("user_id", "identity_id")
+        );
+      `,
+    },
+    {
+      name: 'rename `oidc_user_access_only` to `oidc_user_join_only` and re-add `oidc_user_access_only` column',
+      query: sql`
+        ALTER TABLE IF EXISTS "oidc_integrations"
+        RENAME COLUMN "oidc_user_access_only" TO "oidc_user_join_only";
```
n1ru4l, line 21: note: Renaming the column should be safe here as it only updates the metadata of the column and does not lock the table for a long time or result in rewriting all rows. Compared to other tables this one is quite small 👍

packages/services/api/src/modules/organization/resolvers/Query/myDefaultOrganization.ts
```diff
@@ -54,17 +38,41 @@ export const myDefaultOrganization: NonNullable<QueryResolvers['myDefaultOrganiz
...
+    const getPriority = (org: (typeof orgsWithOIDCConfig)[number]) => {
+      // prioritize user's own organization
+      if (org.ownerId === actor.user.id) {
+        return 1;
+      }
+      if (actor.oidcIntegrationId) {
+        // prioritize OIDC connected organization when user is authenticated with SSO
+        if (org.oidcIntegration?.id === actor.oidcIntegrationId) {
+          return 2;
+        }
```
n1ru4l, line 62: Shouldn't the default org be the oidc org if the user is logged in via that provider?

PR #7395, opened 2025-12-09
Generate a full list of changes on schema check, and add an argument to filter change lists

jdolle:
### Background

Schema proposals require a full list of changes to be able to patch the schemas. Currently, our schema checks are responsible for calculating this list of changes and saving it to the DB. However, checks don't include the nested changes. e.g. if a new type is added, checks don't care about showing that type's new fields.

### Description

This change modifies the check behavior so that it captures all changes and then the resolvers use the Rule from graphql-inspector to filter out the nested types.

### DIFF
--- a/packages/services/storage/src/schema-change-model.ts
+++ b/packages/services/storage/src/schema-change-model.ts
@@ -1012,16 +1012,16 @@ export const ObjectTypeInterfaceRemovedModel = implement<ObjectTypeInterfaceRemo
 export const SchemaQueryTypeChangedModel = implement<SchemaQueryTypeChangedChange>().with({
   type: SchemaQueryTypeChangedLiteral,
   meta: z.object({
-    oldQueryTypeName: z.string(),
-    newQueryTypeName: z.string(),
+    oldQueryTypeName: z.string().nullable(),
+    newQueryTypeName: z.string().nullable(),
   }),
 });
 
 export const SchemaMutationTypeChangedModel = implement<SchemaMutationTypeChangedChange>().with({
   type: SchemaMutationTypeChangedLiteral,
   meta: z.object({
-    oldMutationTypeName: z.string(),
-    newMutationTypeName: z.string(),
+    oldMutationTypeName: z.string().nullable(),
+    newMutationTypeName: z.string().nullable(),
   }),
 });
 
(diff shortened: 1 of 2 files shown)


packages/services/api/src/modules/schema/providers/inspector.ts
```diff
@@ -31,10 +31,10 @@ export class Inspector {
       'hive.diff.changes.count': result.length,
     }),
   })
-  async diff(existing: GraphQLSchema, incoming: GraphQLSchema) {
+  async diff(existing: GraphQLSchema, incoming: GraphQLSchema, rules?: Rule[]) {
     this.logger.debug('Comparing Schemas');
 
-    const changes = await diff(existing, incoming, [DiffRule.simplifyChanges]);
+    const changes = await diff(existing, incoming, rules);
```
jdolle, line 37: This change exposes the rules option up the callstack so it can be modifies.

packages/services/api/src/modules/schema/providers/schema-check-manager.ts
```diff
@@ -35,27 +36,70 @@ export class SchemaCheckManager {
...
     if (!schemaCheck.safeSchemaChanges?.length) {
       return null;
     }
 
+    if (simplifyChanges) {
+      /**
+       * @todo modify the simplifyChanges rule to loosen the argument requirements. This rule
+       * doesn't need more information than the change type and path fields.
+       */
+      return DiffRule.simplifyChanges({
```
jdolle: The rule's type definition is very strict and that our SchemaCheck model slightly modifies the change structure, but the only thing this specific rule is using is a change's `path` and a `type`, which are unmodified. So technically this is safe to use here even though the typing is a mess.

I'd be interested in hearing recommendations for how maybe we can avoid this typecasting.

packages/services/api/src/modules/schema/providers/schema-version-helper.ts
```diff
@@ -167,16 +167,26 @@ export class SchemaVersionHelper {
       'hive.safe-changes.count': changes?.safe?.length,
     }),
   })
-  @cache<SchemaVersion>(version => version.id)
-  private async _getSchemaChanges(schemaVersion: SchemaVersion) {
+  @cache<{ version: SchemaVersion; simplified: boolean }>(
```
jdolle: the cache typing requires everything be passed as a single argument.
I then cache these versions separately to avoid having to recalculate the simplified or full list.

packages/services/storage/src/db/types.ts
```diff
@@ -325,34 +326,35 @@ export interface schema_policy_config {
 }
 
 export interface schema_proposal_comments {
+  author: string;
```
jdolle, line 329: there are unrelated.. i was working on fixing types previously

PR #7403, opened 2025-12-10
add policy test for eslint-disable-next-line

jdolle:
### Background

We received a report that our linter is not using the "eslint-disable-next-line" feature.

### Description

This adds a test to verify the behavior.


jdolle:
We've determined that this is not the issue. graphqljs' parse method removes the comments.

PR #7405, opened 2025-12-11
enhance(sdk-rs): use usage report json schema as a source of truth

ardatan:
Instead of manually written types, using [usage-report-v2.schema.json](https://github.com/graphql-hive/console/compare/sasfer-structs?expand=1#diff-cb803bb5e61980e7e962d29bb5381d24a1b99bed20b2da22b42418c85f722096) as a source of truth, and generate structs from there is safer.

I also fixed the JSON schema to get unsigned integers instead of integers for relevant places

### DIFF
--- a/docker/docker.hcl
+++ b/docker/docker.hcl
@@ -363,7 +363,7 @@ target "apollo-router" {
   contexts = {
     router_pkg = "${PWD}/packages/libraries/router"
     sdk_rs_pkg = "${PWD}/packages/libraries/sdk-rs"
-    usage_service = "${PWD}/services/usage"
+    usage_service = "${PWD}/packages/services/usage"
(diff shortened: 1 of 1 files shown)


packages/libraries/sdk-rs/Cargo.toml
```diff
@@ -32,6 +32,8 @@ serde_json = "1"
 moka = { version = "0.12.10", features = ["future", "sync"] }
 sha2 = { version = "0.10.8", features = ["std"] }
 tokio-util = "0.7.16"
+typify = "0.5.0"
+regress = "0.10.5"
```
kamilkisiela, line 36: is it used?

ardatan: image
It uses it somehow 😅

PR #7410, opened 2025-12-11
feat(server): add hive tracing integration

EmrysMyrddin:
### Description

Adds Hive Tracing integration in addition to the existing OTEL setup.

I've also updated OTEL packages to the latest and used Hive Gateway setup helper instead of `NodeSDK`.

Ref CONSOLE-1646

### Breaking Changes

I've made breaking changes, but I can work around most of them if needed:
 - `TracingInstance` methods `build` and `start` have been replaced by `setup`
 - `TracingInstance` no longer have an `sdk` attribute
 - `TracingInstance` no longer offer a `tracerProvider` method. Use `trace.getTracerProvider` instead
 - Renamed `HIVE_USAGE_TARGET` to `HIVE_TARGET` (this is used by both Usage and Tracing)
 - Renamed `HIVE_USAGE_ACCESS_TOKEN` to `HIVE_ACCESS_TOKEN` (this is use by both Usage and Tracing)
 
### Usage

The integration is not enabled by default because it requires some configuration.
Provide the following env variables to enable it:

```env
HIVE_TRACING=1
HIVE_TARGET=<your-target>
HIVE_ACCESS_TOKEN=<your-access-token>

# Optionally, you can override the default collector endpoint
HIVE_TRACING_ENDPOINT=https://...
```

### Workarounds

Due to the fact our tsconfig uses the old module resolution mode, modules with multiple entrypoints can't be imported if they are defined using `exports` field in `package.json`.

To workaround this without making to much changes to the current setup, I've just added manually the plugin entrypoints in the tsconfig `path`, which is not ideal but the simplest solution in my opinion.

To make it work with unit tests, I also had to do a change in the way the plugin is bundled: https://github.com/graphql-hive/gateway/pull/1791

### DIFF
--- a/packages/services/server/package.json
+++ b/packages/services/server/package.json
@@ -54,6 +54,8 @@
     "zod": "3.25.76"
   },
   "buildOptions": {
-    "external": ["pg-native"]
+    "external": [
+      "pg-native"
+    ]
   }
 }


deployment/services/graphql.ts
```diff
@@ -134,8 +134,14 @@ export function deployGraphQL({
           GRAPHQL_PUBLIC_ORIGIN: `https://${environment.appDns}`,
           CDN_CF: '1',
           HIVE_USAGE: '1',
-          HIVE_USAGE_TARGET: hiveConfig.require('target'),
+          HIVE_TARGET: hiveConfig.require('target'),
           HIVE_USAGE_ENDPOINT: serviceLocalEndpoint(usage.service),
+          HIVE_TRACING: '1',
+          HIVE_TRACING_ENDPOINT: environment.isProduction
```
n1ru4l, line 140: note: using the public endpoints to avoid having to solve a circular pulumi dependency of graphql api and otel :D We can alter on fix it.

deployment/services/graphql.ts
```diff
@@ -134,8 +134,14 @@ export function deployGraphQL({
...
+            ? 'https://api.graphql-hive.dev/otel/v1/traces'
```
n1ru4l:
```suggestion
            ? 'https://api.graphql-hive.com/otel/v1/traces'
```

package.json
```diff
@@ -127,7 +127,8 @@
       "tar-fs": "2.1.4",
       "ip": "npm:neoip@2.1.0",
       "miniflare@3>undici": "5.29.0",
-      "tailwindcss": "3.4.17"
+      "tailwindcss": "3.4.17",
+      "@graphql-hive/plugin-opentelemetry": "1.3.0-alpha-9dd631a2329aa1d677cca642b666a55a497a691e"
```
EmrysMyrddin: Also let's not forget to update to release version before publishing !

package.json
```diff
@@ -127,7 +127,8 @@
       "tar-fs": "2.1.4",
       "ip": "npm:neoip@2.1.0",
       "miniflare@3>undici": "5.29.0",
-      "tailwindcss": "3.4.17"
+      "tailwindcss": "3.4.17",
+      "@graphql-hive/plugin-opentelemetry": "1.3.0-alpha-23838ba35fd408a4b5d1eea6368ca1d4874b955b"
```
n1ru4l: needs to be updated

PR #7414, opened 2025-12-12
enhance(js-sdk): support `graphql-client-name` and `graphql-client-version` as extra default in addition to `x-` ones

ardatan:
Today by default, `x-graphql-client-` ones are used by default but Apollo Router plugin and some other popular products are using the ones without `x-` by default. So in order to have an easier migration from other tools, this PR makes JS SDK to respect unprefixed version of those headers, too.


n1ru4l (approved):
Looks good!

PR #7421, opened 2025-12-13
Fix typo in pull request template

sondrateconsulting-ryan:


jdolle:
Thanks for finding that

PR #7432, opened 2025-12-16
feat: protected app deployment retiring

adambenhassen:
### Description 
Adds protection settings for app deployment retirement to prevent accidental removal of actively-used deployments.

### How It Works
1. When `retireAppDeployment` is called and protection is enabled:
  - App deployment must have been created at least `minDaysSinceCreation` days ago
  - If no usage data exists: retirement allowed (deployment was never used)
  - If usage exists, checks criteria based on `ruleLogic`:
    - **AND mode** (default): Both conditions must pass
      - Days inactive >= `minDaysInactive` 
      - **AND** Traffic percentage <= `maxTrafficPercentage`
     - **OR mode**: Either condition can pass
       - Days inactive >= `minDaysInactive` 
       - **OR** Traffic percentage <= `maxTrafficPercentage`
   - If check fails: retirement blocked with error
2. Using `force: true` (or `--force` in CLI) bypasses all checks

### Settings page
Screenshot 2026-01-27 at 11 51 07

Closes CONSOLE-1501

### DIFF
--- a/integration-tests/tests/api/app-deployments.spec.ts
+++ b/integration-tests/tests/api/app-deployments.spec.ts
@@ -172,6 +172,7 @@ const UpdateTargetAppDeploymentProtectionConfiguration = graphql(`
           appDeploymentProtectionConfiguration {
             isEnabled
             minDaysInactive
+            minDaysSinceCreation
             maxTrafficPercentage
             trafficPeriodDays
             ruleLogic
@@ -4821,6 +4822,7 @@ test('update app deployment protection configuration', async () => {
   expect(result.target?.appDeploymentProtectionConfiguration).toEqual({
     isEnabled: false,
     minDaysInactive: 7,
+    minDaysSinceCreation: 7,
     maxTrafficPercentage: 1.0,
     trafficPeriodDays: 30,
     ruleLogic: 'AND',
@@ -4855,7 +4857,7 @@ test('update app deployment protection configuration', async () => {
         appDeploymentProtectionConfiguration: {
           isEnabled: true,
           minDaysInactive: 7,
-          minDaysSinceCreation: 3,
+          minDaysSinceCreation: 7,
           maxTrafficPercentage: 5.0,
           trafficPeriodDays: 30,
           ruleLogic: 'AND',
@@ -4881,7 +4883,7 @@ test('update app deployment protection configuration', async () => {
   expect(result.target?.appDeploymentProtectionConfiguration).toEqual({
     isEnabled: true,
     minDaysInactive: 7,
-    minDaysSinceCreation: 3,
+    minDaysSinceCreation: 7,
     maxTrafficPercentage: 5.0,
     trafficPeriodDays: 30,
     ruleLogic: 'AND',


integration-tests/tests/cli/app.spec.ts
```diff
@@ -0,0 +1,228 @@
...
+    },
+    {},
+    'cli-force-app~1.0.0~hash',
+  );
+
+  await waitForOperationsCollected(1);
+
+  const targetSlug = `${organization.slug}/${project.slug}/${target.slug}`;
+
+  // CLI retire without --force should fail
```
jdolle, line 200: Nice job on these tests. I appreciate that you checked that it would fail without force before checking force passes.

packages/libraries/cli/src/commands/app/retire.ts
```diff
@@ -79,12 +83,31 @@ export default class AppRetire extends Command<typeof AppRetire> {
...
+        }
+        this.log(`  Required inactive days: ${details.requiredMinDaysInactive}`);
+        if (
+          details.currentTrafficPercentage !== null &&
+          details.currentTrafficPercentage !== undefined
+        ) {
+          this.log(`  Current traffic: ${details.currentTrafficPercentage.toFixed(2)}%`);
+        }
+        this.log(`  Max traffic threshold: ${details.maxTrafficPercentage}%`);
+        this.log('\nUse --force to bypass protection.\n');
```
jdolle, line 108: Again great job. You provide the important details and explain how they can bypass.

packages/migrations/src/actions/2026.01.21T00-00-00.app-deployment-protection.ts
```diff
@@ -0,0 +1,10 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.12.15T00-00-00.app-deployment-protection.ts',
+  run: ({ sql }) => sql`
+    ALTER TABLE targets ADD COLUMN app_deployment_protection_enabled BOOLEAN NOT NULL DEFAULT FALSE;
+    ALTER TABLE targets ADD COLUMN app_deployment_protection_min_days_inactive INT NOT NULL DEFAULT 30;
```
jdolle: This may get explained as I review the rest of the PR. But how do the min days work with our usage retention_in_days limit? E.g. if we retain only 7 days but have min_days_inactive as 30, then after 7 days of not being used will we think this is satisfied?

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -546,6 +561,91 @@ export class AppDeployments {
...
+          const diffMs = now.getTime() - lastUsed.getTime();
+          daysSinceLastUsed = Math.floor(diffMs / (1000 * 60 * 60 * 24));
+        }
+
+        // Get traffic percentage (use 30 days as period)
+        const trafficData = await this.getAppDeploymentTrafficPercentage({
+          targetId: args.targetId,
+          appName: args.appDeployment.name,
+          appVersion: args.appDeployment.version,
+          periodDays: 30,
```
jdolle: should this be configurable also?

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -806,6 +906,69 @@ export class AppDeployments {
...
+    const result = await this.clickhouse.query({
+      query: cSql`
+        SELECT
+          SUM(CASE WHEN "client_name" = ${args.appName} AND "client_version" = ${args.appVersion} THEN "total" ELSE 0 END) AS "appDeploymentOps"
+          , SUM("total") AS "totalOps"
+        FROM "operations_daily"
+        WHERE
+          "target" = ${args.targetId}
+          AND "timestamp" >= toDateTime(${formatClickHouseDate(periodFrom)}, 'UTC')
+          AND "timestamp" <= toDateTime(${formatClickHouseDate(periodTo)}, 'UTC')
```
jdolle: This utility method seems like a lot of work that the client or clickhouse should handle for us.

I dug a little into clickhouse docs and I think a better approach is to use `parseDateTimeBestEffort(${periodFrom.toISOString()})`. `parseDateTimeBestEffort` handles ISO format strings and toISOString always returns UTC, so it should be equivalent.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -806,6 +906,69 @@ export class AppDeployments {
...
+      `,
+      queryId: 'get-app-deployment-traffic-percentage',
+      timeout: 20_000,
```
jdolle, line 1522: Is 20s a default we use? My preference is to timeout quicker because 20s seems like a long time. If it's waiting that long then it's likely going to fail.

packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -325,6 +335,74 @@ export default gql`
...
+    minDaysInactive: Int! @tag(name: "public")
+    """
+    Maximum traffic percentage threshold. App deployments with traffic above this percentage cannot be retired.
+    """
+    maxTrafficPercentage: Float! @tag(name: "public")
+  }
+
+  """
+  Input for updating app deployment protection configuration.
+  Fields not provided (omitted) will retain the previous value.
```
jdolle, line 392: Do we error or return OK if no input fields are provided?

packages/services/api/src/modules/target/providers/target-manager.ts
```diff
@@ -601,8 +601,94 @@ export class TargetManager {
...
+  > {
+    this.logger.debug(
+      'Updating target app deployment protection settings (target=%o, configuration=%o)',
+      args.target,
+      args.configuration,
+    );
+
+    const selector = await this.idTranslator.resolveTargetReference({ reference: args.target });
+    if (!selector) {
+      this.session.raise('target:modifySettings');
```
jdolle, line 637: should the assertion be before the selection?
If you assert `target:modifySettings` before `resolveTargetReference`, then we can correctly return a more accurate error like "target doesnt exist"

packages/web/app/src/pages/target-settings.tsx
```diff
@@ -940,6 +975,228 @@ const BreakingChanges = (props: {
...
+  const { handleSubmit, isSubmitting, errors, touched, values, handleBlur, handleChange } =
+    useFormik({
+      enableReinitialize: true,
+      initialValues: {
+        minDaysInactive: configuration?.minDaysInactive ?? 30,
+        maxTrafficPercentage: configuration?.maxTrafficPercentage ?? 1.0,
+      },
+      validationSchema: Yup.object().shape({
+        minDaysInactive: Yup.number()
+          .min(0, 'Must be at least 0')
```
jdolle, line 1143: should a max be determined by `retentionInDays`?

integration-tests/tests/cli/app.spec.ts
adambenhassen: Thanks!

packages/libraries/cli/src/commands/app/retire.ts
adambenhassen: Thanks!

packages/migrations/src/actions/2026.01.21T00-00-00.app-deployment-protection.ts
adambenhassen: Currently if there's no usage data (e.g., it was purged due to retention limits), we allow retirement. This means setting `min_days_inactive` higher than `retention_in_days` would effectively just require waiting for the retention period.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: should should

packages/services/api/src/modules/target/module.graphql.ts
adambenhassen: returns OK - the storage uses `COALESCE` so omitted fields retain their existing values.

packages/services/api/src/modules/target/providers/target-manager.ts
adambenhassen: This pattern is used throughout the file, we return a permission error even when the target doesn't exist. I don't think you'd want to reveal if the target exists or not without permissions.

packages/web/app/src/pages/target-settings.tsx
adambenhassen: `minDaysInactive` uses the `app_deployment_usage` table, so it's not constrained by `retentionInDays`

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: good one!

adambenhassen:
A couple of changes:
  - Added `trafficPeriodDays` (default `30`) and `ruleLogic` (`AND`/`OR`) settings
  - Migration uses `IF NOT EXISTS`, `min_days_inactive` defaulted to `7`
  - Replaced `formatClickHouseDate` with `parseDateTimeBestEffort`
  - UI dropdown for `AND`/`OR` selection for flexibility
  - PR description updated, also the screenshot

packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -41,6 +41,12 @@ export default gql`
       input: UpdateTargetDangerousChangeClassificationInput! @tag(name: "public")
     ): UpdateTargetDangerousChangeClassificationResult! @tag(name: "public")
     """
+    Update the app deployment protection configuration of a target.
+    """
+    updateTargetAppDeploymentProtectionConfiguration(
+      input: UpdateTargetAppDeploymentProtectionConfigurationInput! @tag(name: "public")
```
n1ru4l: +1 on not making it public until someone asks for it explicitly

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -569,6 +584,99 @@ export class AppDeployments {
...
+          this.logger.debug(
+            'retire app deployment blocked by protection settings. (targetId=%s, appDeploymentId=%s, reason=%s)',
+            args.targetId,
+            appDeployment.id,
+            blockReason,
+          );
+
+          return {
+            type: 'error' as const,
+            message: `App deployment retirement blocked by protection settings. ${blockReason}. Use --force to override.`,
```
n1ru4l: Isn't `--force` a detail specific to the CLI client? The API should not return client specific error messages.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -569,6 +584,99 @@ export class AppDeployments {
...
+        const trafficData = await this.getAppDeploymentTrafficPercentage({
+          targetId: args.targetId,
+          appName: args.appDeployment.name,
+          appVersion: args.appDeployment.version,
+          periodDays: protectionConfig.trafficPeriodDays,
+        });
+        const trafficPercentage = trafficData?.trafficPercentage ?? null;
+
+        // Check protection rules
+        // If no usage data (lastUsed is null), we allow retirement per requirement
```
n1ru4l: AFAIK our customer also wanted to make sure that the app deployment was created at least X days ago, to not accidentally remove an app deployment that has not been used yet.

packages/migrations/src/actions/2026.01.21T00-00-00.app-deployment-protection.ts
```diff
@@ -0,0 +1,12 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2026.01.21T00-00-00.app-deployment-protection.ts',
+  run: ({ sql }) => sql`
+    ALTER TABLE targets ADD COLUMN IF NOT EXISTS app_deployment_protection_enabled BOOLEAN NOT NULL DEFAULT FALSE;
+    ALTER TABLE targets ADD COLUMN IF NOT EXISTS app_deployment_protection_min_days_inactive INT NOT NULL DEFAULT 7;
+    ALTER TABLE targets ADD COLUMN IF NOT EXISTS app_deployment_protection_max_traffic_percentage NUMERIC(5,2) NOT NULL DEFAULT 1.00;
+    ALTER TABLE targets ADD COLUMN IF NOT EXISTS app_deployment_protection_traffic_period_days INT NOT NULL DEFAULT 30;
+    ALTER TABLE targets ADD COLUMN IF NOT EXISTS app_deployment_protection_rule_logic TEXT NOT NULL DEFAULT 'AND';
```
n1ru4l: These should be grouped into a single statement, so there are less table locks

```sql
ALTER TABLE targets
  ADD COLUMN IF NOT EXISTS app_deployment_protection_enabled BOOLEAN NOT NULL DEFAULT FALSE
  , ADD COLUMN IF NOT EXISTS app_deployment_protection_min_days_inactive INT NOT NULL DEFAULT 7
  , ADD COLUMN IF NOT EXISTS app_deployment_protection_max_traffic_percentage NUMERIC(5,2) NOT NULL DEFAULT 1.00
  -- ...
;
```

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -1377,6 +1485,59 @@ export class AppDeployments {
...
+    appVersion: string;
+    periodDays: number;
+  }): Promise<{
+    trafficPercentage: number;
+    totalOperations: number;
+    appDeploymentOperations: number;
+  } | null> {
+    const periodFrom = new Date();
+    periodFrom.setDate(periodFrom.getDate() - args.periodDays);
+    periodFrom.setHours(0, 0, 0, 0);
```
n1ru4l: NIT: we have date-fns as a dependency already for these kind of things

```
import { subDays, startOfDay } from 'date-fns';

const periodFrom = startOfDay(subDays(new Date(), args.periodDays));
```

packages/services/api/src/modules/target/providers/target-manager.ts
n1ru4l: I agree with @adambenhassen here, we do not want to differentiate between does not exists and missing permissions to the outside world.

packages/web/app/src/pages/target-settings.tsx
```diff
@@ -1583,11 +1895,18 @@ function TargetSettingsContent(props: {
...
+              <BreakingChanges
+                organizationSlug={props.organizationSlug}
+                projectSlug={props.projectSlug}
+                targetSlug={props.targetSlug}
+              />
+              <AppDeploymentProtection
+                organizationSlug={props.organizationSlug}
+                projectSlug={props.projectSlug}
+                targetSlug={props.targetSlug}
+              />
```
n1ru4l: This section shows up when the app deployment feature flag is not enabled for the organization. We can conditionally render it based on the `Organization.isAppDeploymentsEnabled` schema coordinate.

packages/services/storage/src/index.ts
```diff
@@ -1988,11 +2014,83 @@ export async function createStorage(
...
+      projectId: string;
+      isEnabled?: boolean | null;
+      minDaysInactive?: number | null;
+      minDaysSinceCreation?: number | null;
+      maxTrafficPercentage?: number | null;
+      trafficPeriodDays?: number | null;
+      ruleLogic?: 'AND' | 'OR' | null;
+    }) {
+      return transformTargetSettings(
+        await tracedTransaction('updateTargetAppDeploymentProtectionSettings', pool, async trx => {
```
n1ru4l: Why is this single statement run in a transaction?

packages/services/storage/src/index.ts
```diff
@@ -1988,11 +2014,83 @@ export async function createStorage(
...
+                , array_agg(tv.destination_target_id) as targets
+              FROM targets AS it
+                LEFT JOIN target_validation AS tv ON (tv.target_id = it.id)
+              WHERE
+                it.id = ${target}
+                AND it.project_id = ${project}
+              GROUP BY
+                it.id
+              LIMIT 1
+            ) ret
```
n1ru4l: Why is this `FROM` statement needed? Seems to me like the query would work fine without it

Issue #7365, opened 2025-12-03
Inconsistent behavior between JS and Rust SDK


ardatan:
While porting the test from https://github.com/graphql-hive/console/blob/main/packages/libraries/core/tests/usage.spec.ts#L193 to Rust;
I realized that JS SDK doesn't send this one with `!`;
https://github.com/graphql-hive/console/pull/7364/files#diff-e95ef9e6aaeb5b11afa8ba0d1d6c31778caaaab15eef6f3259227a705dd55ac7R405

jdolle:
We can opt to make the JS implementation match the Rust implementation, or make Rust match JS.

---

This is the one case that isn't very well defined in the current logic.
We have a required input argument as a variable, but that variable isn't being passed when calling the operation. This request will fail anyway you look at it, but do we track this input as provided by the client, because it's required? Or do we track it as not provided, because it actually wasn't sent?...

The only time this matters is if you start with a required argument, send some data, push a breaking change to make it nullable, then want to go back to required. So, I consider this an extreme edge case, but it would be good to be consistent.

The function that determines whether or not this value is checked is https://github.com/graphql-hive/console/blob/main/packages/services/api/src/modules/schema/providers/registry-checks.ts#L905 

Due to this being such an edge case, I don't have any preference over which approach we choose.

PR #7439, opened 2025-12-16
Remove the usage flag from non-null, but unused variables to match js…

jdolle:
… sdk

### Background

https://github.com/graphql-hive/console/issues/7365

### Description

This is an extreme edge case where an operation variable is non-nullable but the variable isn't passed. In this case, the usage should not flag the value as used (!), since the operation will fail regardless and if we flag this as used, then the logic will be inconsistent with nested variable values. E.g. `input FooInput { f: String!, o: String! }` should not flag `FooInput.f` as used if it's not provided.

### DIFF
--- a/packages/libraries/sdk-rs/src/graphql.rs
+++ b/packages/libraries/sdk-rs/src/graphql.rs
@@ -497,6 +497,8 @@ impl<'a> OperationVisitor<'a, SchemaCoordinatesContext> for SchemaCoordinatesVis
                         ctx.schema_coordinates.insert(format!("{coordinate}!"));
                     }
 
+                    mark_as_used(ctx, &coordinate);
+
                     let field_type_name = field.value_type.inner_type();
 
                     match value {


packages/libraries/sdk-rs/src/graphql.rs
```diff
@@ -48,7 +47,6 @@ pub fn collect_schema_coordinates(
         used_input_fields: HashSet::new(),
         input_values_provided: HashMap::new(),
         used_variables: HashSet::new(),
-        non_null_variables: HashSet::new(),
```
jdolle, line 51: We don't need to track this because nullability doesn't impact whether or not a value was passed technically

packages/libraries/sdk-rs/src/graphql.rs
```diff
@@ -2306,16 +2269,10 @@ mod tests {
...
-            "User.name",
-            "Query.user",
-            "ID",
-            "Query.user.id!",
-            "Query.user.id",
-        ]
-        .into_iter()
-        .map(|s| s.to_string())
-        .collect::<HashSet<String>>();
+        let expected = vec!["User.name", "Query.user", "ID", "Query.user.id"]
```
jdolle, line 2272: Looking at the operation, you can see that `Query.user.id` is non-nullable and there is a variable defined for `$id`. However, no variable is actually passed in. This means we don't track it as used via the bang (!).

Same goes for pretty much all the other tests.

This matches the JS implementation

PR #7441, opened 2025-12-17
feat: allow customizing oidc scope per integration

### LINKED ISSUE #7355: Custom OIDC provider - can't define custom scopes
Hello,
we would like to use our own OIDC provider in Hive Console, but we also need to define our own authorization scopes.
Is it somehow possible now?
Are you planning to add such an option?
It would be enough for us if it were possible to enter the "authorization url" with defined `scope` query parameter something like `https://example.com/oauth2/v1/authorize?scope=xyz` and then it would be automatically merged with the default `openid` and `email` scope on your side and result would be `https://example.com/oauth2/v1/authorize?scope=xyz openid email`

Thank you in advance for your reply.
Tomáš Kroupa

XiNiHa:
### Background

Closes #7355

### Description

This PR includes a DB migration that adds a column for storing customized OIDC scope per each integration, and the ability to customize it from the console.

### Checklist

- [x] Input validation
- [x] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
+++ b/packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
@@ -447,16 +447,7 @@ function CreateOIDCIntegrationForm(props: {
       authorizationEndpoint: '',
       clientId: '',
       clientSecret: '',
-      additionalScopes: '[]',
-    },
-    validate(values) {
-      try {
-        JSON.parse(values.additionalScopes);
-      } catch {
-        return {
-          additionalScopes: 'Invalid JSON',
-        };
-      }
+      additionalScopes: '',
     },
     async onSubmit(values) {
       const result = await mutate({
@@ -467,7 +458,7 @@ function CreateOIDCIntegrationForm(props: {
           authorizationEndpoint: values.authorizationEndpoint,
           clientId: values.clientId,
           clientSecret: values.clientSecret,
-          additionalScopes: JSON.parse(values.additionalScopes),
+          additionalScopes: values.additionalScopes ? values.additionalScopes.split(' ') : [],
         },
       });
 
@@ -579,16 +570,14 @@ function CreateOIDCIntegrationForm(props: {
           <div>
             <Label htmlFor="additionalScopes">Additional Scopes</Label>
             <Input
-              placeholder="Additional Scopes"
+              placeholder="Separated by spaces"
               id="additionalScopes"
               name="additionalScopes"
               onChange={formik.handleChange}
-              onBlur={formik.handleBlur}
               value={formik.values.additionalScopes}
             />
             <FormError>
-              {(formik.touched.additionalScopes && formik.errors.additionalScopes) ||
-                mutation.data?.createOIDCIntegration.error?.details.additionalScopes}
+              {mutation.data?.createOIDCIntegration.error?.details.additionalScopes}
             </FormError>
           </div>
 
@@ -945,19 +934,7 @@ function UpdateOIDCIntegrationForm(props: {
       authorizationEndpoint: props.oidcIntegration.authorizationEndpoint,
       clientId: props.oidcIntegration.clientId,
       clientSecret: '',
-      additionalScopes: JSON.stringify(props.oidcIntegration.additionalScopes).replaceAll(
-        ',',
-        ', ',
-      ),
-    },
-    validate(values) {
-      try {
-        JSON.parse(values.additionalScopes);
-      } catch {
-        return {
-          additionalScopes: 'Invalid JSON',
-        };
-      }
+      additionalScopes: props.oidcIntegration.additionalScopes.join(' '),
     },
     async onSubmit(values) {
       const result = await oidcUpdateMutate({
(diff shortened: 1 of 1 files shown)


packages/migrations/src/actions/2025.12.17T00-00-00.oidc-scope.ts
```diff
@@ -0,0 +1,17 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2025.12.17T00-00-00.oidc-scope.ts',
+  noTransaction: true,
```
n1ru4l: This is safe to run in a transaction imho as it does not lock the table

packages/migrations/src/actions/2025.12.17T00-00-00.oidc-scope.ts
```diff
@@ -0,0 +1,17 @@
...
+  run: ({ sql }) => [
+    {
+      name: 'Create new "scope" column',
+      query: sql`
+        -- Alter the "oidc_integrations" table with the new column "scope"
+        ALTER TABLE "oidc_integrations"
+        ADD COLUMN "scope" TEXT[]
+        DEFAULT array['openid', 'email'];
```
n1ru4l: It might make sense to save only `additional_scopes` and automatically append `openid` and `email` on the application level instead of storing it in the database, as these are required for the OIDC flow.

packages/services/storage/src/index.ts
```diff
@@ -4823,6 +4833,7 @@ const OktaIntegrationBaseModel = zod.object({
   linked_organization_id: zod.string(),
   client_id: zod.string(),
   client_secret: zod.string(),
+  additional_scopes: zod.array(zod.string()),
```
n1ru4l: This also needs to be able to handle the `null` value of older database records. It caused this exception for me:
```
[15:59:51] ERROR: [
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "array",
            "received": "null",
            "path": [
              "additional_scopes"
            ],
            "message": "Expected array, received null"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "null",
            "path": [
              "oauth_api_url"
            ],
            "message": "Expected string, received null"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "array",
            "received": "null",
            "path": [
              "additional_scopes"
            ],
            "message": "Expected array, received null"
          }
        ],
        "name": "ZodError"
      }
    ],
    "path": [],
    "message": "Invalid input"
  }
]
```

packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
```diff
@@ -564,6 +576,22 @@ function CreateOIDCIntegrationForm(props: {
...
+              onChange={formik.handleChange}
+              onBlur={formik.handleBlur}
+              value={formik.values.additionalScopes}
+            />
+            <FormError>
+              {(formik.touched.additionalScopes && formik.errors.additionalScopes) ||
+                mutation.data?.createOIDCIntegration.error?.details.additionalScopes}
+            </FormError>
+          </div>
+
```
n1ru4l, line 583: NIT: It is functional but to me it feels a bit clunky having to write JSON in here.

Image

Image

What of we just allow passing a string instead, separate it by spaces (` `) and then truncate the single items?

Image

packages/services/storage/src/index.ts
n1ru4l: Something like the following works, to handle the default value on the application level:


```suggestion
  additional_scopes: zod
    .array(zod.string())
    .nullable()
    .transform(value => (value === null ? [] : value)),
```

packages/services/api/src/modules/oidc-integrations/providers/oidc-integrations.provider.ts
```diff
@@ -137,6 +138,7 @@ export class OIDCIntegrationsProvider {
         tokenEndpoint: tokenEndpointResult.data,
         userinfoEndpoint: userinfoEndpointResult.data,
         authorizationEndpoint: authorizationEndpointResult.data,
+        additionalScopes: args.additionalScopes,
```
n1ru4l: There should be proper input validation for these additional scopes. For example maximum amount of 10, and maximum length of 15 characters for each.

XiNiHa: https://github.com/graphql-hive/console/pull/7441/commits/2f10d5f9e49a975ead0f18b3b86426599226d40e

packages/web/app/src/components/organization/settings/oidc-integration-section.tsx
XiNiHa: https://github.com/graphql-hive/console/pull/7441/commits/eecf84c175e53ebf4ea486aa4610baeecca81d9b

PR #7451, opened 2025-12-18
Show diff on schema proposals editor

jdolle:
### Background

[Linear Issue](https://linear.app/the-guild/issue/CONSOLE-1675/schema-proposals-editor-should-highlight-which-parts-have-been-added)

Current UI:

Screenshot 2025-12-18 at 2 20 11 PM

### Description

Swaps the schema editor to the diff schema editor. This has built in "next" / "previous" buttons for quickly moving between changes, and adds highlighting so it's easier to see what's changed.

Proposed Change:
Screenshot 2025-12-18 at 7 36 14 PM

### DIFF
--- a/.changeset/polite-beds-kneel.md
+++ b/.changeset/polite-beds-kneel.md
@@ -1,0 +1,5 @@
+---
+'hive-console-sdk-rs': patch
+---
+
+export `minify_query` and `normalize_operation` functions (mainly for Hive Router)
--- a/packages/libraries/sdk-rs/src/graphql.rs
+++ b/packages/libraries/sdk-rs/src/graphql.rs
@@ -11,7 +11,7 @@ use std::collections::BTreeMap;
 use std::collections::HashMap;
 use std::collections::HashSet;
 
-use graphql_parser::minify_query;
+pub use graphql_parser::minify_query;
 use graphql_parser::parse_query;
 use graphql_parser::query::{
     Definition, Directive, Document, Field, FragmentDefinition, Number, OperationDefinition,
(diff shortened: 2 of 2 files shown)


packages/services/storage/src/index.ts
```diff
@@ -4325,7 +4325,7 @@ export async function createStorage(
           args.latest
             ? sql`
             INNER JOIN (
-              SELECT "service_name", "schema_proposal_id", max("created_at") as "maxdate"
+              SELECT COALESCE("service_name", '') as "service", "schema_proposal_id", max("created_at") as "maxdate"
```
jdolle, line 4328: This addresses an issue w/single schema where the service name can be null but I forced it to be an empty string for proposals to keep the column non-nullable.

packages/web/app/src/components/v2/diff-editor.tsx
```diff
@@ -13,15 +15,36 @@ export const DiffEditor = (props: {
...
+  /** Allow editing the after schema. Editable schemas don't allow toggling between diff and no-diff */
+  editable?: boolean;
+  lineNumbers?: boolean;
+  onMount?: (editor: editor.IStandaloneCodeEditor) => void;
+  onChange?: (source: string | undefined) => void;
 }): ReactElement => {
   const [showDiff, setShowDiff] = useState<boolean>(true);
   const sdlBefore = usePrettify(props.before);
-  const sdlAfter = usePrettify(props.after);
+  // runs once on mount then uses internal monaco state to manage
```
jdolle, line 26: This was important because as I was editing the SDL it was being formatted and the cursor was losing its position.

PR #7455, opened 2025-12-23
feat(core): add layer 2 cache support for persisted documents

adambenhassen:
### Description

Adds Layer 2 (L2) cache support for persisted documents, enabling Redis (or other external caches) as an intermediate layer between the in-memory LRU cache (L1) and the CDN.

Lookup flow: L1 (memory) -> L2 (Redis/alternative) -> CDN

### Example

``` ts
  import { createClient } from 'redis';
  import { createHive } from '@graphql-hive/core';

  const redis = createClient({ url: 'redis://localhost:6379' });
  await redis.connect();

  const hive = createHive({
    token: process.env.HIVE_TOKEN,
    experimental__persistedDocuments: {
      cdn: {
        endpoint: '',
        accessToken: '',
      },
      layer2Cache: {
        cache: {
          get: (key) => redis.get(`hive:pd:${key}`),
          set: (key, value, opts) =>
            redis.set(`hive:pd:${key}`, value, opts?.ttl ? { EX: opts.ttl } : {}),
        },
        ttlSeconds: 3600,        // 1 hour for found documents
        notFoundTtlSeconds: 60,  // 1 minute for notfound
      },
    },
  });
```

### DIFF
--- a/packages/libraries/core/src/client/persisted-documents.ts
+++ b/packages/libraries/core/src/client/persisted-documents.ts
@@ -6,8 +6,8 @@ import { defaultCircuitBreakerConfiguration } from './circuit-breaker.js';
 import { http, HttpCallConfig } from './http-client.js';
 import {
   PERSISTED_DOCUMENT_NOT_FOUND,
-  type PersistedDocumentsConfiguration,
   type PersistedDocumentsCache,
+  type PersistedDocumentsConfiguration,
 } from './types.js';
 
 type HeadersObject = {
--- a/packages/libraries/core/src/client/types.ts
+++ b/packages/libraries/core/src/client/types.ts
@@ -1,11 +1,11 @@
 import type { ExecutionArgs } from 'graphql';
 import type { PromiseOrValue } from 'graphql/jsutils/PromiseOrValue.js';
 import { LogLevel as HiveLoggerLevel, Logger } from '@graphql-hive/logger';
+import { MaybePromise } from '@graphql-tools/utils';
 import type { AgentOptions } from './agent.js';
 import { CircuitBreakerConfiguration } from './circuit-breaker.js';
 import type { autoDisposeSymbol, hiveClientSymbol } from './client.js';
 import type { SchemaReporter } from './reporting.js';
-import { MaybePromise } from '@graphql-tools/utils';
 
 type HeadersObject = {
   get(name: string): string | null;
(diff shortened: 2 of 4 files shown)


packages/libraries/core/src/client/persisted-documents.ts
```diff
@@ -154,33 +92,97 @@ export function createPersistedDocuments(
...
+      return;
+    }
+
+    const cacheValue = value === null ? PERSISTED_DOCUMENT_NOT_FOUND : value;
+    const ttl = value === null ? layer2NotFoundTtlSeconds : layer2TtlSeconds;
+
+    // Fire-and-forget. don't await, don't block
+    const setPromise = layer2Cache.set(documentId, cacheValue, ttl ? { ttl } : undefined);
+    if (setPromise && typeof setPromise.catch === 'function') {
+      setPromise.catch(error => {
```
ardatan: Floating promises are tricky in some environments(especially serverless ones). Those environments don't wait for event loop as Node does, and in this case we use `waitUntil`(from the context, available in Yoga so in Hive GW) to keep the worker open even if the response is returned to let the promise resolved, or explicit resource management to prevent a shutdown without resolving the promise/job.
https://the-guild.dev/graphql/yoga-server/docs/features/explicit-resource-management#background-jobs

packages/libraries/core/src/client/types.ts
```diff
@@ -296,6 +296,85 @@ export interface ServicesFetcherOptions {
...
+   * @returns The document body, PERSISTED_DOCUMENT_NOT_FOUND for negative cache hit, or null for cache miss
+   */
+  get(key: string): Promise<string | PersistedDocumentNotFound | null>;
+
+  /**
+   * Store a document in the cache.
+   * Optional - if not provided, the cache is read-only.
+   * @param key - The document ID
+   * @param value - The document body or PERSISTED_DOCUMENT_NOT_FOUND for negative caching
+   * @param options - Optional TTL configuration
```
ardatan: Better to specify if TTL is seconds or milliseconds etc?

packages/libraries/core/src/client/types.ts
```diff
@@ -296,6 +296,85 @@ export interface ServicesFetcherOptions {
...
+   */
+  set?(
+    key: string,
+    value: string | PersistedDocumentNotFound,
+    options?: { ttl?: number },
+  ): Promise<void>;
```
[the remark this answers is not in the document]
ardatan: We can relax the typings here, and allow user to return anything. These don't have to be `Promise` though, it can be `MaybePromise<unknown>` ?

PR #7459, opened 2026-01-05
Evaluate usageEstimation year input range at runtime

jdolle:
### Background

Org usage data is erroring due to being out of the valid range. https://linear.app/the-guild/issue/CONSOLE-1699/usage-metrics-missing-from-dashboard-since-2025-12-04

This happens because the range is defined at boot/deployment time.

### Description

This change modified the logic to validate the range inside the resolver. Although this adds some complexity, it prevents the date range from becoming stale.

I've also added a month to the current date to determine the valid year. This is to allow the date range to run over into the next year for December's usage report.

### DIFF
--- a/.changeset/five-mangos-give.md
+++ b/.changeset/five-mangos-give.md
@@ -1,0 +1,7 @@
+---
+'hive': patch
+---
+
+Set usageEstimation year validation range at runtime to avoid issues during the new year. This fixes
+an issue where the organization settings usage data was not loading for January until the service
+was deployed again.


packages/services/commerce/src/usage-estimator/index.ts
```diff
@@ -7,20 +7,25 @@ export const usageEstimatorRouter = router({
...
         .required(),
     )
     .query(async ({ ctx, input }) => {
+      // The range must be defined at runtime to avoid issues at the end of the year
+      const now = new Date();
+      const rangeValidation = z
+        .number()
+        .min(now.getFullYear() - 1)
+        .max(new Date(now.setMonth(now.getMonth() + 1)).getFullYear());
+      const year = rangeValidation.parse(input.year);
```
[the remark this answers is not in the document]
jdolle: These sideeffects are intentional.

packages/services/commerce/src/usage-estimator/index.ts
```diff
@@ -7,20 +7,25 @@ export const usageEstimatorRouter = router({
       z
         .object({
           month: z.number().min(1).max(12),
-          year: z
-            .number()
-            .min(new Date().getFullYear() - 1)
-            .max(new Date().getFullYear()),
+          year: z.any(),
```
dotansimha, line 10: maybe this could just be a more loose range, rather rhan a `any`? 
if we set the range between current year and next year, it should be fine and also fix this, right?

jdolle: The range validation happens inside the resolver, so the any is just a place holder.
But I do agree we _should_ be fine. I expect we'll deploy more than once a year. But my preference is still to check this at runtime.

PR #7460, opened 2026-01-06
fix(deployment): upgrade pulumi, solve race-condition between `Service` and `Deployment`, fix `Secrets` updating

dotansimha:
In this PR, I've made some refactoring to the Pulumi deployment, mostly around `ServiceDeployment`. With these changes, deployments are now faster, with zero downtime and no strong dependencies that might fail and cause a broken state. 

- [x] Prevent full `Deployment` replacement when `Secret` changes (by allowing Pulumi to set the `Secret`'s `metadata.name`)
- [x] Fix strong dependency between `Deployment` and `Service` (by avoiding using `kubernetex`'s `createService` function, and use a custom one - we have now only a dependency on `ports` and not full `spec`. Also no use of `parent` field of Pulumi). The relation now is loose and based on the `matchLabels` only. 
- [x] The change above also fixes a race condition we had while deploying the same commit ID - it was causing the `Service` to be removed because it was strictly bound to the `Deployment`. (to confirm)
- [x] Upgrade Pulumi CLI to latest 
- [x] Upgrade Pulumi K8s adapter to latest
- [x] Test on `dev`
- [x] Test on `staging` 

Closes https://github.com/graphql-hive/deployment/issues/1412

### DIFF
--- a/deployment/utils/service-deployment.ts
+++ b/deployment/utils/service-deployment.ts
@@ -369,7 +369,7 @@ export function createService(name: string, deployment: kx.Deployment) {
     containers.forEach(container => {
       if (container.ports) {
         container.ports.forEach(port => {
-          ports.push({ name: port.name, port: port.containerPort });
+          ports.push({ name: port.name || name, port: port.containerPort });
         });
       }
     });


deployment/utils/service-deployment.ts
```diff
@@ -360,3 +360,32 @@ export class ServiceDeployment {
     return { deployment, service };
   }
 }
+
+export function createService(name: string, deployment: kx.Deployment) {
```
dotansimha, line 364: The main difference from `@pulumi/kubernetesx` are:
- we evaluate `labels` and `ports` separately (instead of grouping them together and construct the `spec` field) - this leads to loose dependency and only change the `Service` if the port or labels has chaneged (instead of any change). 
- No `parent` set on the Pulumi resource, so no strict coupling between the `Service` and the `Deployment`.

deployment/utils/secrets.ts
```diff
@@ -11,9 +11,6 @@ export class ServiceSecret<T extends Record<string, string | Output<string>>> {
   ) {
     this.raw = data;
     this.record = new k8s.core.v1.Secret(this.name, {
-      metadata: {
```
dotansimha, line 14: If we don't set explicit name, then Pulumi will use the `this.name` (pass the the ctor) and will appends some suffix to it. Since `Secret`s are immutable, when we change the secret, it will re-create a new one with a new suffix, leading to a chain of reaction that causes a `RollingUpgrade` to the pod. 

If we explicitly set the name, then when the secret changes, we get a new one, and the `Deployment` needs to be re-created (instead of updated to point to the new Secret), and this leads to downtime as the pod is replaced without rolling upgrade.

PR #7473, opened 2026-01-08
Track proposal diff row numbers using state

jdolle:
### Background

We want to be able to collapse unchanged rows to reduce noise on schema proposal diff. The current way of adding row numbers is by counting the classes using css. This doesn't work when rows get hidden.

### Description

This adds logic to count the rows using context instead of relying on css. this allows us to show the row numbers even if rows are collapsed.

This may still yet change again depending on how rows are hidden.


jdolle:
This change is not necessary if we go the route of determining the lines/words first, then render.

PR #7475, opened 2026-01-08
Allow org owner to modify billing plan after downgrading

jdolle:
### Background

Currently, if a billing plan has previously been been configured, so there exists a "organizations_billing" record, and then the organization downgrades to a hobby plan, then it's impossible to select a new billing plan.

https://linear.app/the-guild/issue/CONSOLE-1698/cancelling-a-pro-subscription-ends-up-with-a-in-between-state-that

### Description

This could be solved a couple ways. We could delete this record when downgrading, which we've done manually in the past to fix this issue, or we could keep the record and adjust the "canUpdateSubscription" logic.

I've adjusted the "canUpdateSubscription" logic because it will keep the billing record which would be more convenient for organization owners who are looking to resubscribe, and it keeps invoices associated with this organization. (Stripe invoices are tied to "organizations_billing"."external_billing_reference".)


packages/services/api/src/modules/commerce/resolvers/Organization.ts
```diff
@@ -99,7 +100,7 @@ export const Organization: Pick<
 
     return {
       hasActiveSubscription: subscriptionInfo.subscription !== null,
-      canUpdateSubscription: subscriptionInfo.subscription !== null,
+      canUpdateSubscription: org.ownerId === user.id,
```
jdolle: I assume this is what we want to check. But I'm open to suggestions if we want to allow more than the org owner to update the subscription.

packages/web/app/src/components/organization/billing/BillingPaymentMethod.tsx
```diff
@@ -61,18 +63,23 @@ export const ManagePaymentMethod = (props: {
...
+                    window.location.href = result.data.generateStripePortalLink;
+                  }
+                })
+                .catch(() => {
+                  setLoadingDashboard(false);
+                });
             }}
           >
-            {mutation.fetching ? (
+            {mutation.fetching || loadingDashboard ? (
```
jdolle, line 88: Previously there was a period after the mutation resolves and before the page redirects where you can click this again and it'd mutate and try to redirect again.

packages/web/app/src/components/v2/radio-group.tsx
```diff
@@ -14,7 +14,8 @@ export const Radio = ({ children, className, ...props }: RadioGroupItemProps): R
   return (
     <Item
       className={clsx(
-        'relative overflow-hidden rounded-sm border text-left hover:border-orange-500/50 focus:ring data-[state=checked]:border-orange-500',
+        'relative overflow-hidden rounded-sm border text-left focus:ring data-[state=checked]:border-orange-500',
+        !props.disabled && 'hover:border-orange-500/50',
```
jdolle, line 18: previously highlighted the border on hover even if you couldn't update the subscription. This seemed confusing to me since it wasnt possible to interact with.

packages/web/app/src/pages/organization-subscription-manage.tsx
```diff
@@ -325,6 +325,11 @@ function Inner(props: {
     <div className="flex w-full flex-col gap-5">
       <Card className="w-full">
         <Heading className="mb-4">Choose Your Plan</Heading>
+        {organization.billingConfiguration.canUpdateSubscription ? (
+          <div className="mb-3 text-sm text-orange-500">
+            Only the organization owner can update the plan.
```
jdolle: Clarified the condition here in case there was confusion on why the subscription couldn't be updated.

packages/services/api/src/modules/commerce/resolvers/Organization.ts
n1ru4l: We have a `billing:update` permission which is also asserted within the `upgradeToPro` mutation resolver. That one should be used.

packages/web/app/src/pages/organization-subscription-manage.tsx
n1ru4l: This needs to be updated according to https://github.com/graphql-hive/console/pull/7475/changes#r2675422720

n1ru4l: Also this would show up for enterprise plans which is incorrect, as they cannot manually update their plan. see https://github.com/graphql-hive/console/pull/7475/changes#diff-c1c3c1eb960a82213570e95b97ef9e06f2d2a26d4360fa7687fc5140ff939f03R22

dotansimha:
>  We could delete this record when downgrading

btw, this is the least preferred way, because it means a new customer record will be created, and invoices history will be gone for the user (but still available in Stripe).

packages/web/app/src/components/organization/billing/BillingPaymentMethod.tsx
```diff
@@ -40,6 +40,8 @@ export const ManagePaymentMethod = (props: {
   const organization = useFragment(BillingPaymentMethod_OrganizationFragment, props.organization);
   const info = organization.billingConfiguration.paymentMethod;
 
+  const [loadingDashboard, setLoadingDashboard] = useState(false);
```
n1ru4l, line 48: Why is it necessary to track the loading state in addition to `mutation.fetching`?

jdolle: The fetch is to load the url, then the page tries to redirect. If the user has a slow connection, it looks like the page is loading and then stops. Tracking this as a state lets us show that the page is loading during the entire duration of the url generation + redirect.

PR #7477, opened 2026-01-09
fix: properly display insufficient github app permissions when creating a schema check

n1ru4l:
### Background

https://linear.app/the-guild/issue/CONSOLE-1284/missing-github-repository-permissions-causes-unexpected-error-for

### Description

This broke at some point, maybe through upgrading the GitHub SDK. 🤔 

### Checklist

- [x] Testing (internal video available)

### DIFF
--- a/.changeset/brave-buses-thank.md
+++ b/.changeset/brave-buses-thank.md
@@ -1,0 +1,6 @@
+---
+'hive': patch
+---
+
+Show correct error message for insufficient GitHub App installation permissions when attempting to
+create a check run as part of a schema check.


packages/services/api/src/modules/integrations/providers/github-integration-manager.ts
```diff
@@ -522,7 +522,7 @@ export class GitHubIntegrationManager {
 }
 
 function isOctokitRequestError(error: unknown): error is RequestError {
-  return !!error && typeof error === 'object' && 'code' in error && 'status' in error;
+  return error instanceof RequestError;
```
n1ru4l, line 525: This did not peoperly cover the error thrown from the github sdk, which resulted in the already existing error handlers to be skipped and the error being treated as unexpected 😓

Urigo:
is there a way to add a test for it?

n1ru4l:
@Urigo Would require mocking the GitHub API in an integration test. Can be considered as a follow up task - so far we shied away from testing third-party dependencies too much and relied on manual testing when we touch them. This case shows that it might not be optimal 😓

PR #7491, opened 2026-01-14
Rewrite proposal diff rendering to allow collapsing

jdolle:
### Background

https://linear.app/the-guild/issue/CONSOLE-1678/schema-proposals-diff-view-should-show-only-what-has-changed-and-the

For changes that touch very few lines compared to the total lines of SDL, it makes sense to collapse untouched portions of the schema. This reduces the amount of text people have to scan during a review and puts focus on exactly what has changed.

### Description

Using our previous method of parsing and rendering the SDL within components made it difficult to determine which lines are able to be hidden.

To accomplish our goal, parsing needs to happen first so that the lines can be categorized appropriately before render. I first created a builder instance that would be capable to capturing and then returning all the necessary data to group and render these lines.

The rest is rewriting every component to instead leverage this builder.

Screenshot 2026-01-14 at 6 01 40 PM

With this framework in place, we can fairly easily tweak the actual rendering logic of how we collapse and show lines.


packages/web/app/src/components/target/proposals/schema-diff/core.tsx
```diff
@@ -0,0 +1,1234 @@
...
+}) {
+  const changeType = determineChangeType(oldValue, newValue);
+  const name = oldValue?.name ?? newValue?.name ?? '';
+
+  diffDescription({
+    newNode: newValue!,
+    oldNode: oldValue!,
+    indent: 1,
+    builder,
+  });
```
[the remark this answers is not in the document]
jdolle: This will cause the typescript compiler to error because it isnt smart enough to know that if one is null then the other isnt.

packages/web/app/src/components/target/proposals/schema-diff/core.tsx
```diff
@@ -0,0 +1,1234 @@
...
+  if (
+    (isInputObjectType(oldType) || oldType === null) &&
+    (isInputObjectType(newType) || newType === null)
+  ) {
+    diffInputObject({
+      oldInput: oldType!,
+      newInput: newType!,
+      builder,
+    });
+    return;
```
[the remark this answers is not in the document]
jdolle: typescript isnt smart enough to know that if one is null then the other is defined. Removing `!` will cause a compiler error.

packages/web/app/src/components/target/proposals/schema-diff/lines.tsx
```diff
@@ -0,0 +1,81 @@
...
+          </td>
+        </tr>
+        {props.children?.slice(-3, props.children.length)}
+      </tbody>
+    );
+  }
+  return <tbody>{props.children}</tbody>;
+}
+
+// @todo annotations
```
n1ru4l: todo

packages/web/app/src/components/target/proposals/schema-diff/words.tsx
```diff
@@ -0,0 +1,99 @@
+// @TODO but how to know where the comment should land within that coordinate?
```
n1ru4l: todo

packages/web/app/src/components/target/proposals/index.tsx
```diff
@@ -509,7 +497,12 @@ export function Proposal(props: {
             </>
           )}
         />
-        {diff}
+        <SchemaDiff
+          className={props.className}
+          before={props.beforeSchema}
+          after={props.afterSchema}
+          // annotations={annotations}
```
n1ru4l: unused code?

PR #7492, opened 2026-01-14
feat: update Node.js to 24.13

n1ru4l:
### Background

https://nodejs.org/en/blog/release/v24.13.0

### Description

- Update to the latest LTS version of Node.js for our containers; github actions and CLIs
- Updates OCLIF
- no more windows x86 (x32) support; only 64-bit (x64)
- no more arm (x32) on linux; only arm64 (x64)
- drops the javascript xray scanning library for the preflight script, since it is not really needed and is incompatible with latest Node.js versions


integration-tests/fixtures/huge-schema.graphql
n1ru4l, line 1: had to delete some lines as it exhausted overall gothib action container resources. Now it is only big enough to exhaust the schema composition worker thread memory

Issue #7076, opened 2025-10-06
schema explorer: filter based on federation service name


n1ru4l:
In order to see which fields are unused within a service, filtering down the displayed types and fields based on a service name is a good addition.

We could put this filter next to the following elements:

<img width="458" height="214" alt="Image" src="https://github.com/user-attachments/assets/d3e85bef-bb06-4332-a074-0afbcdcaa879" />

PR #7497, opened 2026-01-15
inital commit of adding filtering by services on explore schema

egoodwinx:
### Background

https://github.com/graphql-hive/console/issues/7076

### Description

image

This commit introduces a new service name filter across all schema explorer pages (main explorer, deprecated, and unused views). The feature allows users to filter GraphQL types by the services that own them in a federated/supergraph setup.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/web/app/src/components/target/explorer/union-type.tsx
```diff
@@ -44,7 +80,7 @@ export function GraphQLUnionTypeComponent(props: {
       organizationSlug={props.organizationSlug}
     >
       <div className="flex flex-col">
-        {ttype.members.map((member, i) => (
+        {members.map((member: any, i: number) => (
```
jdolle: prefer type inference and avoiding `any`

packages/web/app/src/components/target/explorer/union-type.tsx
```diff
@@ -32,7 +38,37 @@ export function GraphQLUnionTypeComponent(props: {
...
+  const router = useRouter();
+  const searchObj = router.latestLocation.search;
+  const search =
+    'search' in searchObj && typeof searchObj.search === 'string'
+      ? searchObj.search.toLowerCase()
+      : undefined;
   const ttype = useFragment(GraphQLUnionTypeComponent_TypeFragment, props.type);
+  const { hasMetadataFilter, metadata: filterMeta } = useSchemaExplorerContext();
+
+  const members = ttype.members.filter((member: any) => {
```
jdolle: prefer removing the `any` cast.

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -199,6 +232,89 @@ export function DateRangeFilter() {
...
+    allTypes.forEach(type => {
+      type.supergraphMetadata?.ownedByServiceNames?.forEach((serviceName: string) => {
+        serviceNameSet.add(serviceName);
+      });
+    });
+
+    const extracted = Array.from(serviceNameSet).sort();
+
+    return extracted;
+  }, [query.data]);
```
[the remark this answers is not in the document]
jdolle: This suggestion completely ignores `type.fields` and while technically shorter, this is less performant since it's building an extra array in the process

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -199,6 +232,89 @@ export function DateRangeFilter() {
...
+      <DropdownMenuTrigger asChild>
+        <Button variant="secondary" className="data-[state=open]:bg-muted">
+          <FilterIcon className="size-4" />
+          &nbsp;Service
+          <span className="sr-only">Open menu to filter by service name.</span>
+        </Button>
+      </DropdownMenuTrigger>
+      <DropdownMenuContent
+        align="end"
+        className="max-h-[300px] min-w-[160px] max-w-[300px] flex-wrap overflow-y-auto"
```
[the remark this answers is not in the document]
jdolle: This might be valid-ish... How does this look when we overflow the x axis? Should we truncate instead?

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -33,6 +33,61 @@ import {
...
+          targetSlug: $targetSlug
+        }
+      }
+    ) {
+      __typename
+      id
+      latestValidSchemaVersion {
+        __typename
+        id
+        explorer(usage: { period: $period }) {
```
jdolle: It would be best if we could use fragments for this and `TypeFilter_AllTypes` since they're near identical requests.

I don't recall why `TypeFilter_AllTypes` wasn't already a fragment.

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -225,6 +279,102 @@ export function DateRangeFilter() {
...
+  const serviceNames = useMemo(() => {
+    const allTypes = query.data?.target?.latestValidSchemaVersion?.explorer?.types;
+    if (!allTypes) return [];
+
+    const serviceNameSet = new Set<string>();
+
+    for (const type of allTypes) {
+      // Add service names from type-level metadata
+      if (type.supergraphMetadata?.ownedByServiceNames) {
+        for (const serviceName of type.supergraphMetadata.ownedByServiceNames) {
```
jdolle: Minor: To avoid the extra if condition you could do:
```
for (const serviceName of type.supergraphMetadata?.ownedByServiceNames ?? []) { ... }
```

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -225,6 +279,102 @@ export function DateRangeFilter() {
...
+          serviceNameSet.add(serviceName);
+        }
+      }
+
+      // Add service names from field-level metadata
+      if ('fields' in type && Array.isArray(type.fields)) {
+        for (const field of type.fields) {
+          if (field.supergraphMetadata?.ownedByServiceNames) {
+            for (const serviceName of field.supergraphMetadata.ownedByServiceNames) {
```
jdolle: Minor: Here too it could be:
```
for (const serviceName of field.supergraphMetadata?.ownedByServiceNames ?? [])
```

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -225,6 +279,102 @@ export function DateRangeFilter() {
...
+
+    return extracted;
+  }, [query.data]);
+
+  return (
+    <DropdownMenu>
+      <DropdownMenuTrigger asChild>
+        <Button variant="secondary" className="data-[state=open]:bg-muted">
+          <FilterIcon className="size-4" />
+          &nbsp;Service
```
jdolle: Instead of `nbsp;`, if you want more control you could use the `mr-[number]` tailwind css class on the FilterIcon. E.g. `mr-2`

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -225,6 +279,102 @@ export function DateRangeFilter() {
...
+          <span className="sr-only">Open menu to filter by service name.</span>
```
jdolle, line 307: I like the use of `sr-only` 👍

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -225,6 +279,102 @@ export function DateRangeFilter() {
...
+                  unsetMetadataFilter('service', serviceName);
+                }
+              }}
+            >
+              {serviceName}
+            </DropdownMenuCheckboxItem>
+          ))
+        ) : (
+          <DropdownMenuCheckboxItem disabled>
+            No services found in metadata
```
jdolle, line 333: Should we hide the button entirely if there are no services?

packages/web/app/src/components/target/explorer/union-type.tsx
```diff
@@ -32,7 +38,37 @@ export function GraphQLUnionTypeComponent(props: {
...
+    let matchesFilter = true;
+    if (search) {
+      matchesFilter &&= member.name.toLowerCase().includes(search);
+    }
+    if (filterMeta.length && member.supergraphMetadata) {
+      // Check custom metadata attributes
+      const matchesMeta = member.supergraphMetadata.metadata?.some((m: any) =>
+        hasMetadataFilter(m.name, m.content),
```
jdolle, line 59: Am I understanding this correctly that this is fixing a bug w/unions, and unions weren't respecting the metadata filter?

packages/web/app/src/components/target/explorer/utils.ts
```diff
@@ -27,12 +27,21 @@ export function useExplorerFieldFiltering<
           doesMatchFilter &&= field.name.toLowerCase().includes(search);
         }
         if (filterMeta.length) {
+          const metadata =
+            field.supergraphMetadata as SupergraphMetadataList_SupergraphMetadataFragmentFragment;
```
jdolle, line 31: I dont think you should have to cast this.

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -199,6 +232,89 @@ export function DateRangeFilter() {
...
 
+export function ServiceNameFilter(props: {
+  organizationSlug: string;
+  projectSlug: string;
+  targetSlug: string;
+  period: {
+    to: string;
+    from: string;
+  };
+  metadataAttributes?: Array<{ name: string; values: string[] }> | null;
```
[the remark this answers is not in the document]
jdolle: I think AI is right here

packages/web/app/src/components/target/explorer/filter.tsx
egoodwinx: not sure - this will change the ui experience depending on data how do existing things work?

packages/web/app/src/components/target/explorer/utils.ts
egoodwinx:
```
Property 'metadata' does not exist on type '({ __typename: "SupergraphMetadata"; } & { ' $fragmentRefs'?: { SupergraphMetadataList_SupergraphMetadataFragmentFragment: SupergraphMetadataList_SupergraphMetadataFragmentFragment; } | undefined; }) | { ...; }'.
  Property 'metadata' does not exist on type '{ __typename: "SupergraphMetadata"; } & { ' $fragmentRefs'?: { SupergraphMetadataList_SupergraphMetadataFragmentFragment: SupergraphMetadataList_SupergraphMetadataFr
``` 
this error appears without the cast

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -225,6 +279,102 @@ export function DateRangeFilter() {
...
+  const [query] = useQuery({
+    query: ServiceNameFilter_ServiceNames,
+    variables: {
+      organizationSlug: props.organizationSlug,
+      projectSlug: props.projectSlug,
+      targetSlug: props.targetSlug,
+      period: props.period,
+    },
+    requestPolicy: 'cache-first',
+  });
```
n1ru4l: NIT: Is it necessary to fire of a query and hope it is covered by the cache here? Should we instead expose the list of service names via the GraphQL schema and pass it to this component as a fragment?

packages/web/app/src/components/target/explorer/union-type.tsx
egoodwinx: I guess it depends how you want it If you filtered by a union members with that metadata would still show alongside members without it but I guess this is the intended output right?

packages/web/app/src/components/target/explorer/filter.tsx
```diff
@@ -107,10 +144,12 @@ export function TypeFilter(props: {
...
   });
 
-  const allNamedTypes = query.data?.target?.latestValidSchemaVersion?.explorer?.types;
+  const explorer = query.data?.target?.latestValidSchemaVersion?.explorer;
+  const explorerFragment = useFragment(SchemaExplorerTypes_TypesFragment, explorer);
+  const allNamedTypes = explorerFragment?.types;
   const types = useMemo(
     () =>
-      allNamedTypes?.map(t => ({
+      allNamedTypes?.map((t: any) => ({
```
jdolle, line 152: Avoid using `any`

packages/web/app/src/components/target/explorer/filter.tsx
jdolle: We've solved this on a case-by-case basis so far.

When I consider whether or not to disable or hide an element such as this--I evaluate whether or not it's likely for the condition to change.

If a project is likely to go from no services to services, then it'd disable it rather than hide it, so that it sets the expectation that this is an option when services are available. If the project is unlikely to ever have services, such as for a monolith, then I'd rather it be hidden.

I think that's the behavior I would like: Keep disabled for distributed project types, and hide for single schema projects.

packages/web/app/src/components/target/explorer/union-type.tsx
jdolle: I think filtering is correct. Please remove the `any` cast though.

packages/web/app/src/components/target/explorer/utils.ts
jdolle: That has me thinking that the fragment is not being extracted properly. You may be missing a `useFragment` call.

PR #7504, opened 2026-01-15
Replace email verification with custom implementation

### LINKED ISSUE #420: ensure enforce email verification is enforced when making a direct request to the API
The email verification check of super tokens is only done via the `supertokens-node` SDK and on the front end.
In the `server` service (graphql API), we call the `supertokens-core` directly. We need to make sure the email verification is checked there as well.

XiNiHa:
Closes #420

### Background

The existing email verification relied on SuperTokens' implementation, which was not flexible enough to be used in wider use cases like email verification for SSO accounts during account linking.

### Description

- Creates the `email_verifications` table and fills it based on the existing `supertokens_emailverification_verified_emails` table
- Implements the email verification system through the GraphQL API
- Replaces the existing email verification system from SuperTokens with the new implementation
- Updates the frontend to use the new system 

### Checklist

- [x] Input validation
- [x] Output encoding
- [x] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/auth/resolvers/Mutation/sendVerificationEmail.ts
+++ b/packages/services/api/src/modules/auth/resolvers/Mutation/sendVerificationEmail.ts
@@ -1,15 +1,17 @@
+import { FORWARDED_IP_HEADER_NAME } from '../../../../modules/shared/providers/tokens';
 import { EmailVerification } from '../../providers/email-verification';
 import type { MutationResolvers } from './../../../../__generated__/types';
 
 export const sendVerificationEmail: NonNullable<
   MutationResolvers['sendVerificationEmail']
 > = async (_, { input }, { injector, req }) => {
+  const forwardedIPHeaderName = injector.get(FORWARDED_IP_HEADER_NAME);
   const result = await injector.get(EmailVerification).sendVerificationEmail(
     {
       userIdentityId: input.userIdentityId,
       resend: input.resend ?? undefined,
     },
-    req.ip,
+    req.headers[forwardedIPHeaderName]?.toString() ?? req.ip,
   );
 
   if (!result.ok) {
--- a/packages/services/api/src/create.ts
+++ b/packages/services/api/src/create.ts
@@ -63,7 +63,7 @@ import { HivePubSub, PUB_SUB_CONFIG } from './modules/shared/providers/pub-sub';
 import { REDIS_INSTANCE } from './modules/shared/providers/redis';
 import { S3_CONFIG, type S3Config } from './modules/shared/providers/s3-config';
 import { Storage } from './modules/shared/providers/storage';
-import { WEB_APP_URL } from './modules/shared/providers/tokens';
+import { FORWARDED_IP_HEADER_NAME, WEB_APP_URL } from './modules/shared/providers/tokens';
 import { supportModule } from './modules/support';
 import { provideSupportConfig, SupportConfig } from './modules/support/providers/config';
 import { targetModule } from './modules/target';
@@ -154,6 +154,7 @@ export function createRegistry({
   encryptionSecret: string;
   app: {
     baseUrl: string;
+    forwardedIPHeaderName: string;
   } | null;
   schemaConfig: SchemaModuleConfig;
   supportConfig: SupportConfig | null;
@@ -296,6 +297,11 @@ export function createRegistry({
     {
       provide: WEB_APP_URL,
       useValue: app?.baseUrl.replace(/\/$/, '') ?? 'http://localhost:3000',
+      scope: Scope.Singleton,
+    },
+    {
+      provide: FORWARDED_IP_HEADER_NAME,
+      useValue: app?.forwardedIPHeaderName,
       scope: Scope.Singleton,
     },
     {
--- a/packages/services/api/src/modules/shared/providers/tokens.ts
+++ b/packages/services/api/src/modules/shared/providers/tokens.ts
@@ -1,3 +1,4 @@
 import { InjectionToken } from 'graphql-modules';
 
 export const WEB_APP_URL = new InjectionToken<string>('WEB_APP_URL');
+export const FORWARDED_IP_HEADER_NAME = new InjectionToken<string>('FORWARDED_IP_HEADER_NAME');
--- a/packages/services/server/src/index.ts
+++ b/packages/services/server/src/index.ts
@@ -295,6 +295,7 @@ export async function main() {
       app: env.hiveServices.webApp
         ? {
             baseUrl: env.hiveServices.webApp.url,
+            forwardedIPHeaderName: env.supertokens.rateLimitIPHeaderName,
           }
         : null,
       tokens: {


packages/services/storage/src/index.ts
```diff
@@ -681,6 +682,154 @@ export async function createStorage(
...
+            , "expires_at" = NULL
+          WHERE "id" = ${emailVerification.id}
+        `);
+
+        return {
+          ok: true,
+          verified: true,
+        };
+      });
+    },
```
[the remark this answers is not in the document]
XiNiHa: Oops, had no idea about this and actually was feeling weird dealing with the Storage module 🫠

packages/services/storage/src/index.ts
```diff
@@ -681,6 +682,154 @@ export async function createStorage(
...
+          `);
+        }
+
+        return {
+          ok: true,
+          userId: user.id,
+          token: emailVerification.token,
+          expiresAt: new Date(emailVerification.expiresAt!),
+        };
+      });
```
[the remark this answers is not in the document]
XiNiHa: @n1ru4l what do you think about this?

packages/services/api/src/modules/auth/resolvers/Mutation/sendVerificationEmail.ts
```diff
@@ -0,0 +1,35 @@
+import zod from 'zod';
+import { EmailVerification } from '../../providers/email-verification';
+import type { MutationResolvers } from './../../../../__generated__/types';
+
+export const sendVerificationEmail: NonNullable<
+  MutationResolvers['sendVerificationEmail']
+> = async (_, { input }, { injector }) => {
+  const parseResult = zod.string().email().safeParse(input.email);
```
n1ru4l: This should happen within `EmailVerification.sendVerificationEmail`.

packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -0,0 +1,190 @@
...
+    const user = await this.storage.getUserBySuperTokenId({
+      superTokensUserId: input.superTokensUserId,
+    });
+    if (!user) {
+      throw new Error('User not found.');
+    }
+
+    const emailVerification = await this.pool.maybeOne<{
+      verifiedAt: number | null;
+    }>(sql`
```
n1ru4l: Please use `zod` models for de-coding DB I/O instead of using type annotations.

packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    if (emailVerification?.verifiedAt) {
+      return {
+        ok: false,
+        message: 'Your email address has already been verified.',
+      };
+    }
+
+    if (emailVerification && emailVerification.token == null) {
+      throw new HiveError('Database is in invalid state');
```
n1ru4l: Why would we want to let the end-user know this? This should be exposed as an external server error to the outside world. Whether the record is in a valid/invalid state should be dealt with via the `zod` model (e.g. an union type).  The `zod` model will raise an exception of the columns in the database are in an incompatible state.

packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    await this.taskScheduler.scheduleTask(EmailVerificationTask, {
+      user: { id: user.id, email: input.email },
+      emailVerifyLink: `${this.appBaseUrl}/auth/verify-email?superTokensUserId=${input.superTokensUserId}&token=${emailVerification.token}`,
+    });
+
+    return {
+      ok: true,
+      expiresAt: new Date(emailVerification.expiresAt!),
```
n1ru4l:
```suggestion
      expiresAt: new Date(emailVerification.expiresAt),
```

Using `!` should be avoided outside of tests and is usually a sign that the types can be better.

packages/services/storage/src/index.ts
```diff
@@ -625,6 +625,18 @@ export async function createStorage(
...
+        if (internalUser.provider === 'GOOGLE' || internalUser.provider === 'GITHUB') {
+          await t.query(sql`/* ensureUserExists */
+					  INSERT INTO "email_verifications" ("user_id", "provider", "email", "verified_at")
+					  VALUES (
+							${internalUser.id}
+							, ${internalUser.provider}
+							, ${internalUser.email}
+							, now()
+						);
+          `);
```
n1ru4l: Could the business logic simply not require a `email_verification` to be preent in the database if the provider is either google or github? That way we do not need to insert these in the first place.

packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    return {
+      verified: emailVerification?.verifiedAt != null,
+    };
+  }
+
+  async sendVerificationEmail(input: {
+    superTokensUserId: string;
+    email: string;
+  }): Promise<{ ok: true; expiresAt: Date } | { ok: false; message: string }> {
```
n1ru4l: We should rate-limit this function to avoid email spamming.

packages/services/storage/src/index.ts
```diff
@@ -622,6 +622,19 @@ export async function createStorage(
...
+            await t.query(sql`/* ensureUserExists */
+              INSERT INTO "email_verifications" ("user_id", "provider", "email", "verified_at")
+              VALUES (
+                ${internalUser.id}
+                , ${internalUser.provider}
+                , ${internalUser.email}
+                , now()
+              );
+            `);
+          }
```
n1ru4l: Instead of inserting these records for the social providers that guarantee the email was already verified, we could simply skip the check/lookup on whether the email verification record exists. That way we only need to insert the OIDC and email/password records into the database.

packages/web/app/src/pages/auth-verify-email.tsx
n1ru4l, line 1: It seems like we are automatically attempting to send the verification email when visiting this route (`http://localhost:3000/auth/verify-email`). We should only send the verification email once, and then require an additional user action to resend in case it did not arrive.

n1ru4l: Also when I visit this route (`http://localhost:3000/auth/verify-email`) without being signed in/any parameters passed I get the following error:

Image

packages/migrations/src/actions/2026.01.09T00-00-00.email-verifications.ts
```diff
@@ -0,0 +1,48 @@
...
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4()
+        , "user_id" uuid NOT NULL REFERENCES "users" ("id") ON DELETE CASCADE
+        , "provider" text NOT NULL
+        , "email" text NOT NULL
+        , "token" text
+        , "created_at" timestamptz NOT NULL DEFAULT now()
+        , "expires_at" timestamptz
+        , "verified_at" timestamptz
+        , UNIQUE ("user_id", "provider", "email")
+      );
```
n1ru4l, line 18: Given to context that we do this as a pre-requisite for introducing account linking I have the following concern:

Why are the rows linked to the `"users"."id"` column here?

Isn't the goal here to verify the used sign in method "owns" the email (so we can later on link it to the same `"users"` account)?

A sign-in method is the following:
- `password_email`
- `oidc` (for each `"oidc_integrations"."id"` + oidc `sub`/`email` payload combination)
- google/github (these can probably be omitted from needing to be inserted into this table since the email is already verified through the user being signed up through these providers)

In my understanding we want to do the email verification BEFORE we even create our record within the `users` table, so I am a bit confused on why the choice of linking it to `users.id`.

n1ru4l: The flow we want to achieve here is to:
- User signs in -> supertoken records are created
- We prompt the user to verify the email
- Email is verified -> create our user account (in `"users"` table; later on: we would instead first attempt to link it to existing user)

packages/services/api/src/modules/auth/providers/email-verification.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/77fb300a59a0a4ae23acf0b4dd6fadef65723ffa

packages/services/api/src/modules/auth/providers/email-verification.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/77fb300a59a0a4ae23acf0b4dd6fadef65723ffa

packages/services/api/src/modules/auth/providers/email-verification.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/77fb300a59a0a4ae23acf0b4dd6fadef65723ffa

packages/migrations/src/actions/2026.01.09T00-00-00.email-verifications.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/77fb300a59a0a4ae23acf0b4dd6fadef65723ffa

packages/services/api/src/modules/auth/resolvers/Mutation/sendVerificationEmail.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/2992fd7b2d5db8722276df5504d9eb24bd58ead2

packages/services/api/src/modules/auth/providers/email-verification.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/1043111a23ae174223ca530bb0086b0037e648c0

packages/services/storage/src/index.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/59124b00d460c6006553a011a1c5e44face2ea1f

packages/services/storage/src/index.ts
XiNiHa: Looks like a duplicate :sweat_smile:

packages/web/app/src/pages/auth-verify-email.tsx
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/c2a37fac16efd39facd5551b0adcebef8d3c69a9, https://github.com/graphql-hive/console/pull/7504/commits/6dd70713a55bf1b0eec693166d312892e68f879e

packages/services/api/src/modules/auth/resolvers/Mutation/sendVerificationEmail.ts
XiNiHa: Actually it was a security hole 😅 https://github.com/graphql-hive/console/pull/7504/commits/c9a63b7cb9be0fe954bce1b92b2dcbd4d6131c63

packages/migrations/src/actions/2026.01.09T00-00-00.email-verifications.ts
```diff
@@ -0,0 +1,30 @@
...
+ * This migration establishes the tables for email verifications.
+ */
+export default {
+  name: '2026.01.09T00-00-00.email-verifications.ts',
+  run: ({ sql }) => sql`
+    DO $$
+    BEGIN
+      CREATE TABLE IF NOT EXISTS "email_verifications" (
+        "id" uuid PRIMARY KEY DEFAULT uuid_generate_v4()
+        , "supertokens_user_id" text NOT NULL UNIQUE
```
n1ru4l: Given that the goal is to ultimately remove supertokens, would it might make sense to instead have a different column name that is not `supertokens` specific e.g. `user_identity_id` or `user_login_method_id`?

packages/services/workflows/src/tasks/email-verification.ts
```diff
@@ -5,21 +5,18 @@ import { renderEmailVerificationEmail } from '../lib/emails/templates/email-veri
...
   schema: z.object({
-    user: z.object({
-      email: z.string(),
-      id: z.string(),
-    }),
-    emailVerifyLink: z.string(),
+    email: z.string(),
+    verificationLink: z.string(),
   }),
 });
```
n1ru4l, line 13: Changing this task schema introduces a breaking change that could existing tasks that are enqueued mid deployment to fail, as the new version of the service now expects a different task schema.

In order to mitigate such a risk and us having to manually tinker with failed tasks in our database and cleaning them up, I would rather recommend to remove the fields, that are no longer being used, and re-use the existing ones.

```diff
  export const EmailVerificationTask = defineTask({
    name: 'emailVerification',
    schema: z.object({
      user: z.object({
        email: z.string(),
-       id: z.string().optional(),
      }),
      emailVerifyLink: z.string(),
    }),
  });
```

packages/web/app/src/pages/auth-verify-email.tsx
```diff
@@ -1,69 +1,169 @@
...
+        verified
+      }
+      error {
+        message
+      }
+    }
+  }
+`);
+
 function AuthVerifyEmail() {
```
n1ru4l, line 39: When visiting the route `/auth/verify-email`, while the email was already verified I see the following screen, with the error that the email was already verified.

Image

While it is a very specific an edge case, it would be good if we just redirect the user to `/` in that case, so he is not stuck on that page.

packages/migrations/src/actions/2026.01.09T00-00-00.email-verifications.ts
```diff
@@ -0,0 +1,30 @@
...
+        , "token" text
```
n1ru4l: We are storing the token as clear text. 🤔

I wonder whether we instead should only store the hashed version of it and compare the input to the hash. 

An email verification token can:
- Prove ownership of an email
- (in the future) finish account linking

If our database leaks and these tokens are stored in plain text an attacker could confirm the email on behalf of a user, but it would not really grant him access to the user account. 🤔 

In combination with a forged OIDC provider this could work like this:

- Malicious user gains read access to database, identifies user that is part of an organization
- Creates OIDC provider that pretends to own the email
- Malicious user signs in with the forged oidc provider, which results in a confirmation email being sent
- Malcious user grabs the newly issues token from the database
- Malicious user confirms the email and completes the account linking flow; thus gaining access to performing actions on the users behalf.


A malicious user gaining access to the database is a major breach and would most likely result in him being able to modify the database itself, but I still think it is worth to reduce attack surface here and only store a token hash in the database (similar to how we do it with access tokens; [example](https://github.com/graphql-hive/console/blob/b9065c95d07e2ad000650f269708bde9278cde6c/packages/services/api/src/modules/organization/lib/organization-access-key.ts#L95-L118)).

We can still use the user id to lookup the token, we just then need to compare the input token against the database record hash within our business logic.

packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    if (!existingVerification || input.resend) {
+      if (actorId) {
+        await this.rateLimiter.check(
+          'sendVerificationEmail',
+          actorId,
+          60_000,
+          2,
+          `Exceeded rate limit for sending verification emails.`,
+        );
```
n1ru4l: The rate limiting should happen before any logic runs in this method and be based on the requestors IP, since this action is never done as part of a session (as the client sends the `ignore-session` header?

n1ru4l: We already have something similar [here](https://github.com/graphql-hive/console/blob/b9065c95d07e2ad000650f269708bde9278cde6c/packages/services/server/src/supertokens.ts#L262-L271)

packages/services/storage/src/index.ts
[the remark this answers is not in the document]
XiNiHa: Resolved in the other way

packages/migrations/src/actions/2026.01.09T00-00-00.email-verifications.ts
XiNiHa: I was also thinking about that but wanted to be more specific. Agreed though! https://github.com/graphql-hive/console/pull/7504/commits/9f7d3a36c44da9d21233dfcb2fd42a5ebe0b1b86

packages/services/workflows/src/tasks/email-verification.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/4b73b1e6b4d03ac4b3452b78c1bbf4c0f5255e0a

packages/web/app/src/pages/auth-verify-email.tsx
XiNiHa: I went the simpler way, but let me know if you want a clearer one with enums/unions etc https://github.com/graphql-hive/console/pull/7504/commits/5828af4b5da90fef579ba37f204a232cf20535d6

packages/migrations/src/actions/2026.01.09T00-00-00.email-verifications.ts
XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/bf62fb06ac009a35483ca37185a01d49464a0564

packages/services/api/src/modules/auth/providers/email-verification.ts
XiNiHa: The `actorId` was actually an IP address already :joy: renamed the argument!
https://github.com/graphql-hive/console/pull/7504/commits/1f1412cade66b0f4fd0319611045e5d30eacb092

packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -0,0 +1,249 @@
...
+          email: superTokensUser.email,
+        },
+        emailVerifyLink: `${this.appBaseUrl}/auth/verify-email?userIdentityId=${emailVerification.userIdentityId}&token=${token}`,
+      });
+    }
+
+    return {
+      ok: true,
+      expiresAt: new Date(emailVerification.expiresAt),
+    };
```
n1ru4l, line 195: This block feels quite messy and unreadable to me, can we just early return if `existingVerification` exists and  `input.resend` is falsy - and then instead run all the logic for creating hash, inserting, enqueing the mail afterwards?

XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/5b40bfeebdc922442d14f4c2d78d698d33534d11

XiNiHa:
Looks like this also clearly closes #420

packages/services/api/src/modules/auth/resolvers/Mutation/sendVerificationEmail.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const sendVerificationEmail: NonNullable<
+  MutationResolvers['sendVerificationEmail']
+> = async (_, { input }, { injector, req }) => {
+  const result = await injector.get(EmailVerification).sendVerificationEmail(
+    {
+      userIdentityId: input.userIdentityId,
+      resend: input.resend ?? undefined,
+    },
+    req.ip,
```
n1ru4l: Since we run behind a load-balancer in production we actually have to use the forwarded IP header here. We already have this in the environment configuration, that we can re-use here (see https://github.com/graphql-hive/console/pull/7281).

XiNiHa: https://github.com/graphql-hive/console/pull/7504/commits/8a16f9fec779536e04d6cd28114e5a202066aacb

n1ru4l (approved):

PR #7531, opened 2026-01-21
Store schema proposal changes on a separate column in the schema chec…

jdolle:
…k; update graphql inspector

### Background

https://linear.app/the-guild/issue/CONSOLE-1547/schema-proposals-resolve-difference-between-simplifiedchanges-in

The existing changeset is calculated from the API Schema. This excludes directives and runs on more than just what is in the one service change. Also, it is intended to be a minimal set of changes -- meaning no nested changes (e.g. new type added should not list new fields in the changeset).

Patching schemas requires the full changeset and must include directives.

### Description

This requires https://github.com/graphql-hive/graphql-inspector/pull/2930 so that the order of the directive changes is correct. (Must add a directive before you use it.)

This change also adds a new column to the schema check object to hold this full set of changes. Some of these are redundant with the existing changes stored, but the order is critical to be able to patch correctly and efficiently.

### Checklist

- [x] Testing

### DIFF
--- a/integration-tests/tests/api/schema/publish.spec.ts
+++ b/integration-tests/tests/api/schema/publish.spec.ts
@@ -1807,7 +1807,7 @@ describe('schema publishing changes are persisted', () => {
     `,
     schemaAfter: /* GraphQL */ `
       type Query {
-        a(a: String): String!
+        a(a: String): String
       }
     `,
     equalsObject: {


packages/libraries/cli/package.json
```diff
@@ -49,18 +49,18 @@
   },
   "dependencies": {
     "@graphql-hive/core": "workspace:*",
-    "@graphql-inspector/core": "7.1.0",
+    "@graphql-inspector/core": "7.1.1-alpha-20260121180707-eb09941cd292b5c7c0a7526b9ceee77d69591952",
```
n1ru4l: This should not point to the alpha, but I am pretty sure the intension is to update the dependency when the inspector pr is merged and release. I am just flagging it so it is not forgotten accidentally.

packages/services/api/src/modules/schema/providers/inspector.ts
```diff
@@ -36,16 +36,13 @@ export class Inspector {
...
-    return changes
-      .filter(dropTrimmedDescriptionChangedChange)
-      .map(change =>
-        HiveSchemaChangeModel.parse({
-          type: change.type,
-          meta: change.meta,
-          isSafeBasedOnUsage: change.criticality.isSafeBasedOnUsage,
-        }),
-      )
-      .sort((a, b) => a.criticality.localeCompare(b.criticality));
```
n1ru4l, line 48: Why did we drop the criticality sorting? Is it because the directive definition changes need to be first?

packages/services/api/package.json
```diff
@@ -17,7 +17,7 @@
     "@date-fns/utc": "2.1.1",
     "@graphql-hive/core": "workspace:*",
     "@graphql-hive/signal": "1.0.0",
-    "@graphql-inspector/core": "7.1.0",
+    "@graphql-inspector/core": "7.1.1-alpha-20260121180707-eb09941cd292b5c7c0a7526b9ceee77d69591952",
```
n1ru4l: Same as https://github.com/graphql-hive/console/pull/7531/changes#r2715892626

n1ru4l (approved):
Added a few comments, overall looking good.

I am still not convinced that we should re-use schema checks for doing schema proposal checks.

In general, I also like the idea of storing the diff of the subgraph schema and could even consider that we in the future should start doing that for schema versions, so we can show a diff history for a single subgraph over time 🤔

packages/libraries/cli/package.json
jdolle: Good callout.
That's the intention. 👍

packages/services/api/src/modules/schema/providers/inspector.ts
jdolle: yeah. The order of the changes matters when applying them (using `patch`). Sorting them breaks the order and can cause changes to be flagged as conflicting when they aren't.

PR #7535, opened 2026-01-22
refactor schema publishing logic

n1ru4l:
### Background

The main reason for these changes is to clean up the schema check, publish and delete logic a bit before introducing graph variants.

### Description

- Cleaning up a bit of the logic
- fail early on validations where possible (e.g. schema name provided or not)
- Remove `schema_log_to_version` db table type dependency from `CompositeModel`. I want to call `CompositeModel` methods with input from a different origin (for graph variants), I narrowed down the passed attribute types that are not used within any of the `CompositeModel` methods, so it is more generic.
- Get rid of some ugly query in `Storage.getLatestSchemas` that joined too many tables for no good reason

I am aware that I now introduced new composite and single schema types along side the old types we have for records that are loaded from the database. Unifying these should probably be done at later stage when we isolate the schema version database logic from the legacy storage class. Also there are various other places throughout the code base where the old database model is still being used (schema proposals and other graphql resolvers). I did not want to expand this refactor to these parts.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-publisher.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-publisher.ts
@@ -1667,7 +1667,7 @@ export class SchemaPublisher {
       }),
     ]);
 
-    const [latestComposable, latestVersion] = await Promise.all([
+    const [latestVersion, latestComposable] = await Promise.all([
       this.schemaManager.getLatestSchemaVersionWithSchemaLogs({
         target,
       }),


packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -1136,26 +1148,38 @@ export class SchemaPublisher {
...
 
-    // If trying to push with a service name and there are existing services
-    if (input.service) {
+    if (project.type !== Types.ProjectType.SINGLE) {
+      if (!input.service) {
+        return {
+          __typename: 'SchemaPublishMissingServiceError' as const,
+          message: 'Missing service name',
+        } as const;
+      }
```
n1ru4l, line 1220: This check was previously done within `SchemaPublisher.internalPublish` -> `CompositeModel.publish` -> `RegistryChecks.serviceName`.

We can let this fail early before even attempting to acquire a lock.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -1551,12 +1577,6 @@ export class SchemaPublisher {
           DeleteFailureReasonCode.CompositionFailure,
         )?.compositionErrors;
 
-        if (getReasonByCode(deleteResult.reasons, DeleteFailureReasonCode.MissingServiceName)) {
-          errors.push({
-            message: 'Service name is required',
-          });
-        }
```
n1ru4l, line 1606: This could have never happened in the first place since, we kind of validate the service name before, by checking whether an existing service with that name exists before attempting to delete it.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: This also allows simplifying the input parameters for `CompositeModel.publish` and `SingleModel.publish` and overall making these a bit easier to read and reason about.

packages/services/api/src/modules/schema/providers/models/shared.ts
```diff
@@ -180,7 +180,6 @@ export const PublishIgnoreReasonCode = {
 
 export const PublishFailureReasonCode = {
   MissingServiceUrl: 'MISSING_SERVICE_URL',
-  MissingServiceName: 'MISSING_SERVICE_NAME',
```
n1ru4l, line 183: We now assert this on `SchemaPublisher` level.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -677,8 +677,8 @@ export class RegistryChecks {
     if (args.includeUrlChanges) {
       inspectorChanges.push(
         ...detectUrlChanges(
-          args.includeUrlChanges.schemasBefore.filter(isCompositeSchema),
-          args.includeUrlChanges.schemasAfter.filter(isCompositeSchema),
+          args.includeUrlChanges.schemasBefore,
+          args.includeUrlChanges.schemasAfter,
```
n1ru4l, line 730: This was unnecessary

PR #7542, opened 2026-01-23
Fix minor spacing or highlight details on proposal diff

jdolle:
### Background

Recently rewrote schema proposal diff rendering logic. After merging and testing features together -- a few minor details were missed.

### Description

Fixes some minor spacing and highlighting issues discovered during testing.

Screenshot 2026-01-22 at 4 39 50 PM


packages/web/app/src/components/target/proposals/schema-diff/core.tsx
```diff
@@ -1095,7 +1096,7 @@ function diffDirectiveUsage(
 
   props.builder.write(
     SPACE,
-    type(name, lineToWordChange(changeType)),
+    typeName(name, lineToWordChange(changeType)),
```
jdolle, line 1099: subtle, but `type` is intended to be for the a type definition rather than a reference to the type. The main differences is that `type`'s second argument is a coordinate, which is planned to be used for rendering annotations/comments.

PR #7374, opened 2025-12-04
feat: new hive lab

mskorokhodov:
### Background

Replacement for GraphiQL

### Description

New hive laboratory as replacement for new GraphiQL

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Authentication management
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Error handling and logging
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

PR #7558, opened 2026-01-26
chore: upgrade fastify and fastify vite

n1ru4l:
### Background

Main motivation for this is to resolve the following alert. https://github.com/graphql-hive/console/security/dependabot/444

### Description

- Upgrades fastify and fastify vite to the latest major version.
- Not really a lot of breaking API changes for us 🥳 
- A noteable improvement is is that the pino logging API is now type-safe, which helped finding a bunch of incorrect log interpolations.
- Removes the need to patch the vite fastify plugin, as the underlying issue that produced invalid regex expressions being inlined in the `index.html` in dev mode was resolved.
- did a quick run through our dev environment; to make sure app works as expected after the vite fastify upgrade

### DIFF
--- a/packages/web/app/src/server/index.ts
+++ b/packages/web/app/src/server/index.ts
@@ -73,7 +73,7 @@ async function main() {
       // ├── /client
       // │   └── index.html
       // └── index.js
-      root: resolve(__dirname, 'client', 'client'),
+      root: resolve(__dirname, 'client'),
       // Prevent fastify/static from creating '*' route.
       // We want to define our own '*' route to handle all requests.
       wildcard: false,
--- a/packages/web/app/package.json
+++ b/packages/web/app/package.json
@@ -3,7 +3,7 @@
   "type": "module",
   "private": true,
   "scripts": {
-    "build": "tsx ../../../scripts/runify.ts src/server/index.ts && vite build --outDir dist/client",
+    "build": "tsx ../../../scripts/runify.ts src/server/index.ts && vite build",
     "build-storybook": "storybook build",
     "dev": "tsx watch --clear-screen=false --exclude \"./**/*.mjs\" src/server/dev.ts",
     "generate-changelog": "node ../../../scripts/generate-changelog.js",
--- a/packages/web/app/vite.config.ts
+++ b/packages/web/app/vite.config.ts
@@ -43,6 +43,7 @@ export default {
     }),
   ],
   build: {
+    outDir: 'dist',
     rollupOptions: {
       input: {
         index: resolve(__dirname, 'index.html'),


packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
n1ru4l, line 108: A side effect of upgrading pino:

It now has strict type checking based on the provided message string and asserts provided values. That uncovered a bunch of cases where the passed arguments had an incorrect type or a wrong interpolation character was being used.

packages/services/service-common/src/fastify-tracing.ts
```diff
@@ -54,12 +54,7 @@ declare module 'fastify' {
 
 function defaultFormatSpanName(request: FastifyRequest) {
   const { method } = request;
-  let path;
-  if (request.routeOptions) {
-    path = request.routeOptions.url;
-  } else {
-    path = request.routerPath;
-  }
```
n1ru4l, line 62: `routerPath` does no longer exist.

packages/web/app/src/server/index.ts
```diff
@@ -73,7 +73,7 @@ async function main() {
       // ├── /client
       // │   └── index.html
       // └── index.js
-      root: resolve(__dirname, 'client'),
+      root: resolve(__dirname, 'client', 'client'),
```
n1ru4l: The directory structure changed slightly.

Image

All other folders within `dist/client` are empty though... Only `dist/client/client` actually contains stuff we want.

patches/@fastify__vite.patch
n1ru4l, line 1: @mskorokhodov You introduced this file as part of https://github.com/graphql-hive/console/pull/7374. Since the patch was conflicting with the current version, I tried to run it without and noticed that I could not find any issues.

What exactly was the reason for adding this and what kind of issue should this patch solve?  With that in mind, is it safe to remove the patch?

packages/web/app/src/server/index.ts
n1ru4l: Okay I fixed the issue in https://github.com/graphql-hive/console/pull/7558/changes/22dd643d6c6a503ea3f0ace8521f39cfc7ba995c

patches/@fastify__vite.patch
mskorokhodov: Issue i've got is that in dev mode `fastify-vite` had broken regex caused by escaping in index.html, that led me to broken `monaco-editor` with `vite dev`

n1ru4l: Note: We confirmed that the regex is correctly embedded in the index.html file

packages/services/server/src/index.ts
```diff
@@ -156,7 +156,10 @@ export async function main() {
       return res.status(403).send(err.message);
     }
 
-    return supertokensErrorHandler()(err, req, res);
+    // We can not upgrade Supertokens Node as it removed some APIs we rely on for
```
dotansimha, line 159: supertokens 😭

packages/services/service-common/src/fastify-tracing.ts
dotansimha: @n1ru4l did you had the chance to try one of the services with tracing running? Just to make sure this thing still works, as it's usually not used in develpoment and we use it only when deployed.

PR #7565, opened 2026-01-27
feat(app-deployments): improve app:create performance with v2 format

adambenhassen:
### Background
  App deployment uploads can be slow when deploying many documents, especially when most documents haven't changed between versions. Currently, every document is re-uploaded on each deployment, even if it already exists from a previous version.

  ### Description
  Implements a new v2 storage format for app deployments that enables delta uploads: only uploading documents that don't already exist. This dramatically improves deployment speed for incremental updates.

  ### Notable Changes

  **CLI:**
  - Auto-detects v2 format when all hashes are SHA256 (no `--format` flag needed). Logs a message pointing to docs if hashes aren't SHA256.
  - `--format` flag is still available as an explicit override.
  - Delta optimization is built into `createAppDeployment`: existing hashes are returned in the mutation response, no separate query needed. Single round-trip.
  - `--target` is no longer required for delta optimization (server resolves from access token).
  - Added `--showTiming` flag to display per-batch timing breakdown (requires server-side `FEATURE_FLAGS_APP_DEPLOYMENT_TIMINGS_ENABLED=1`).

  **Server:**
  - `createAppDeployment` accepts optional `hashes` input and returns `existingHashes`: enables delta uploads in a single round-trip.
  - `appDeploymentDocumentHashes` query on `Target` type accepts `hashes` input and returns only the matching subset (server-side filtering via ClickHouse).
  - Uses `cSql.longArray` for safe handling of large hash arrays.
  - Delta query includes documents from pending (not yet activated) deployments, fixing dedup on re-runs of `app:create`.
  - Prevents mixing v1 and v2 formats on the same deployment version.
  - Upload timings use generic `label`/`duration` pairs, gated behind `FEATURE_FLAGS_APP_DEPLOYMENT_TIMINGS_ENABLED` env var + client `showTimings` opt-in.
  - Renamed `S3_UPLOAD_CONCURRENCY` to `APP_DEPLOYMENT_S3_UPLOAD_CONCURRENCY`.
  - Format and hash manifest written during `createAppDeployment` (handles 100% and partial dedup)

  **CDN:**
  - Format is encoded in the `apps-enabled` S3 key body (`v1-inactive`/`v2-inactive` during upload, `v1`/`v2` after activation). CDN reads the format via GET (previously HEAD) and resolves documents directly: no more v2-then-v1 fallback.
  - V2 hash manifest written during activation (`app-v2-manifest/{targetId}/{appName}/{appVersion}`) for version isolation. CDN fetches document and manifest in parallel, rejects hashes not belonging to the requested version.
  - Manifest is written before the `apps-enabled` active flag to prevent a race condition.

  ### Backwards Compatibility
  - Existing deployments with `apps-enabled` body `'1'` are treated as active v1.
  - Legacy v2 deployments without a hash manifest skip the version isolation check (breadcrumb logged for monitoring).
  - V1 format continues to work unchanged.

  ### Security
  - V2 documents are now version-isolated: a hash from a retired version cannot be accessed via an active version's path. The CDN verifies each hash against a per-version manifest.

  ### Limitations
  - documents are only deduped for published apps; not created ones, due to clickhouse's `app_deployments` table only containing activated deployments, so basically we can’t know which `app_deployment_id` values belong to certain target/appName for cross-version pending deployments. requires postgres data joining.

  ### Benchmarks
  All changes were benchmarked with the script provided in `load-tests/app-deployments`.

  #### 1000 Documents

  | Scenario | Docs Uploaded | Parse | Validate | Coords | ClickHouse | S3 | **Total** |
  |----------|---------------|-------|----------|--------|------------|-----|-----------|
  | V1 | 1000 | 17ms | 401ms | 45ms | 1790ms | 8274ms | **9.1s** |
  | V2 Initial | 1000 | 10ms | 276ms | 48ms | 936ms | 7040ms | **7.6s** |
  | V2 Delta | 50 (950 skipped) | 1ms | 25ms | 5ms | 51ms | 323ms | **376ms** |

  ## Format Comparison

  | Feature | V1 | V2 (SHA256) |
  |---------|-------------|-------------|
  | Hash format | Any alphanumeric | SHA256 only |
  | Hash validation | Format only | Format + content match |
  | S3 key | `app/{target}/{name}/{version}/{hash}` | `app-v2/{target}/{name}/{hash}` |
  | Cross-version dedup | No | Yes |
  | Delta uploads | No | Yes |
  | Version isolation | Built-in (version in key) | Hash manifest per version |

## Integration tests added

- v2 format rejects non-sha256 hashes
- v2 format rejects sha256 hash that does not match document content
- v2 format accepts sha256 hash with sha256: prefix
- v1 format (legacy) accepts any hash format
- v2 format accepts sha256 hashes and CDN access works
- appDeploymentDocumentHashes returns existing hashes for delta upload
- v2 format prevents hash collision by rejecting non-sha256 hashes
- v2 format enables cross-version document sharing
- v1 format documents are accessible via CDN using format from apps-enabled key
- appDeploymentDocumentHashes returns empty array for app with no previous documents
- rejects mixing v1 and v2 document formats on the same deployment
- re-running app:create deduplicates documents from the pending deployment
- v2 version isolation: CDN rejects hash not belonging to the requested version
- CDN uses format from apps-enabled key to resolve documents without fallback
- v2 deployment with 100% dedup still resolves correctly via CDN
- v2 deployment with partial dedup resolves both shared and new hashes

## Manual e2e tests ran

**CLI:**

- v2 auto-detection: detected sha256 hashes, used v2 format
- Initial upload: 2 new documents, 0 skipped
- Re-run dedup (same version): all documents already exist, nothing uploaded
- Publish: published successfully
- Cross-version dedup: 1 new, 2 skipped (shared with previous version)
- Explicit v1 format: uploaded without auto-detect message
- v1 auto-detection (non-sha256 hashes): detected as v1, logged docs link
- Format mixing prevention: rejected v2 upload on v1 deployment
- 100% dedup (v3.0.0): all docs exist, nothing uploaded, publish succeeds
- 100% dedup (v4.0.0): all docs exist, nothing uploaded, publish succeeds
- Timing output only when flag is enabled
- Hash content mismatch: rejected with clear error shoing expected vs provided

**CDN:**

- v2 document accessible via correct version: 200
- Hash from another version rejected (version isolation): 404
- Shared hash accessible via both versions: 200
- Hash not belonging to version rejected: 404
- v1 inactive deployment not accessible: 404
- v2 inactive deployment not accessible: 404
- Backward compat (pre-format-encoding deployment): 200
- v1 document accessible via CDN after publish: 200

**To summarize what was tested:**
- Auto-detection (v1 and v2)
- Delta uploads (same version, cross-version, 100% dedup)
- Format mixing prevention
- Hash content validation
- CDN access (v1, v2, shared hashes)
- Version isolation (reject hashes from other versions)
- Inactive deployments rejected
- Backward compatibility

Closes CONSOLE-1582


docker/docker-compose.dev.yml
```diff
@@ -42,7 +42,8 @@ services:
 
   s3:
     image: quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z
-    mem_limit: 200m
+    mem_limit: 400m
+    cpus: 1
```
adambenhassen, line 46: this thing loves to hog all cpu after being overloaded, most likely gc thrash

docker/docker-compose.dev.yml
```diff
@@ -75,6 +76,7 @@ services:
   clickhouse:
     image: clickhouse/clickhouse-server:24.8-alpine
     mem_limit: 4096m
+    cpus: 1
```
adambenhassen, line 79: also loves to do random stuff in the background

packages/libraries/cli/src/commands/app/create.ts
```diff
@@ -119,10 +175,57 @@ export default class AppCreate extends Command<typeof AppCreate> {
...
+          'The --target flag is required when using --format=v2 for delta optimization.',
+        );
+      }
+      const hashesResult = await this.registryApi(endpoint, accessToken).request({
+        operation: GetExistingDocumentHashesQuery,
+        variables: {
+          target,
+          appName: flags['name'],
+        },
+      });
```
n1ru4l: Why are we fetch a list of all items instead of using the hashes we have locally as input to confirm whether they exist or not?

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -118,6 +134,27 @@ export default gql`
       after: String @tag(name: "public")
       filter: ActiveAppDeploymentsFilter! @tag(name: "public")
     ): AppDeploymentConnection! @tag(name: "public")
+    """
+    Get list of document hashes that already exist for this app (across all versions).
+    Used for delta uploads - CLI can skip uploading documents that already exist.
+    """
+    appDeploymentDocumentHashes(appName: String!): AppDeploymentDocumentHashesResult!
```
n1ru4l: Does it make sensee take a list of input hashes from the client and input them here? Then, we could for example only send back a list of the hashes of the operations that need to be uploaded.

Might be overkill though, but for high frequent operation changes on one app deployment, the list of hashes that would be returned currently could become quite long.

WDYT?

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -118,6 +134,27 @@ export default gql`
...
+  type AppDeploymentDocumentHashesResult {
+    ok: AppDeploymentDocumentHashesOk
+    error: AppDeploymentDocumentHashesError
+  }
+
+  type AppDeploymentDocumentHashesOk {
+    """
+    List of document hashes that already exist for this app.
+    """
+    hashes: [String!]!
```
n1ru4l: The error state is never used, so we could just return a list and not use the mutation return type pattern.

packages/services/cdn-worker/src/artifact-storage-reader.ts
```diff
@@ -360,25 +372,26 @@ export class ArtifactStorageReader {
...
-
     const headers: Record<string, string> = {};
     if (etagValue) {
       headers['if-none-match'] = etagValue;
     }
 
-    const response = await this.request({
-      key,
+    // Try v2 key first
+    const keyV2 = buildOperationS3BucketKeyV2(targetId, appName, hash);
```
n1ru4l: Instead of trying v2 first and then fallback back to v1, maybe we encode the format within the is app deployment active key to avoid the extra roundtrip?

packages/libraries/cli/src/commands/app/create.ts
n1ru4l, line 1: For a better user experience, we could auto-detect whether the hashes are sha256 and then automatically use the new format for uploading. If the format is not identified as sha256 we could print some information about this (e.g. mention the speed improvements and pointing to our documentation that shows how to configure relay or codegen).

We can still keep the format parameter, so the user could enforce the old format if they need to, just in case.

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -271,8 +313,43 @@ export default gql`
...
+    clickhouseMs: Int!
+    """
+    Time spent uploading documents to S3.
+    """
+    s3Ms: Int!
+    """
+    Number of documents successfully processed.
+    """
+    documentsProcessed: Int!
+  }
```
n1ru4l, line 328: These fields are internals, I kind of feel uncomfortable with exposing these metrics on our GraphQL schema to end users. Maybe we could instead create a generic timings type:

```graphql
type DocumentUploadTiming {
  label: String!
  duration Int!
}

 type AddDocumentsToAppDeploymentOk {
  timings: [DocumentUploadTiming!]
}
```

That way, the type is kind of proof if we change the underlying implementation. Furthermore, we could return these timings only for benchmarks (via a feature flag) and not make them accessible by customers

packages/services/server/src/environment.ts
```diff
@@ -174,6 +174,7 @@ const S3Model = zod.object({
   S3_SECRET_ACCESS_KEY: zod.string(),
   S3_SESSION_TOKEN: emptyString(zod.string().optional()),
   S3_BUCKET_NAME: zod.string(),
+  S3_UPLOAD_CONCURRENCY: emptyString(NumberFromString.optional()),
```
n1ru4l: NIT: this environment variable is not documented and also a bit misleading as this concurrency is applied per app deployment creation and not kind of a global one for all S3 writes. So a better name would be `APP_DEPLOYMENT_S3_UPLOAD_CONCURRENCY`?

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -1543,6 +1552,55 @@ export class AppDeployments {
...
+          SELECT document_hash AS hash
+          FROM app_deployment_documents
+          PREWHERE app_deployment_id IN (
+            SELECT app_deployment_id
+            FROM app_deployments
+            PREWHERE target_id = ${args.targetId}
+              AND app_name = ${args.appName}
+            GROUP BY app_deployment_id
+          )
+          GROUP BY document_hash
```
n1ru4l: Just out of curiosity, how would this query compare to using `DISTINCT` - is it identical?

```suggestion
          SELECT DISTINCT document_hash AS hash
          FROM app_deployment_documents
          PREWHERE app_deployment_id IN (
            SELECT DISTINCT app_deployment_id
            FROM app_deployments
            PREWHERE target_id = ${args.targetId}
              AND app_name = ${args.appName}
          )
```

packages/libraries/cli/src/commands/app/create.ts
```diff
@@ -37,6 +38,16 @@ export default class AppCreate extends Command<typeof AppCreate> {
...
+    showTiming: Flags.boolean({
+      description: 'Show timing breakdown for each batch',
+      default: false,
+    }),
+    format: Flags.string({
+      description:
+        'Storage format version. "v1" (default) uses per-version storage and allows any hash format. "v2" enables cross-version deduplication and requires sha256 hashes.',
+      default: 'v1',
+      options: ['v1', 'v2'],
+    }),
```
n1ru4l, line 49: One thing I noticed is possible is to first publish v1 documents and then v2 documents

```bash
➜  cli git:(chore-use-sha256) ✗ pnpm start app:create \
  --target "the-guild/graphql-hive/development" \
  --name "web-app" \
  --version "0.0.2-example-version" --format v1 \
  /Users/laurinquast/Projects/graphql-hive-3/packages/web/app/src/gql/persisted-documents.json
```

then do 

```bash
➜  cli git:(chore-use-sha256) ✗ pnpm start app:create \
  --target "the-guild/graphql-hive/development" \
  --name "web-app" \
  --version "0.0.2-example-version" --format v2 \
  /Users/laurinquast/Projects/graphql-hive-3/packages/web/app/src/gql/persisted-documents.json
```

Could this potentially cause problems? 🤔

packages/libraries/cli/src/commands/app/create.ts
n1ru4l, line 1: When I run the create command twice in a row, all documents are uploaded twice. 

```
pnpm start app:create \
  --target "the-guild/graphql-hive/development" \
  --name "web-app" \
  --version "0.0.2-example-version" --format v2 \
  /Users/laurinquast/Projects/graphql-hive-3/packages/web/app/src/gql/persisted-documents.json

> @graphql-hive/cli@0.57.4 start /Users/laurinquast/Projects/graphql-hive-3/packages/libraries/cli
> ./bin/dev app:create --target the-guild/graphql-hive/development --name web-app --version 0.0.2-example-version --format v2 /Users/laurinquast/Projects/graphql-hive-3/packages/web/app/src/gql/persisted-documents.json

App deployment "web-app@0.0.2-example-version" is created pending document upload. Uploading 177 new documents (0 already exist)...
100 / 177 (56%) documents uploaded...

App deployment "web-app@0.0.2-example-version" (177 new, 0 skipped) created.
Activate it with the "hive app:publish" command.

 pnpm start app:create \
  --target "the-guild/graphql-hive/development" \
  --name "web-app" \
  --version "0.0.2-example-version" --format v2 \
  /Users/laurinquast/Projects/graphql-hive-3/packages/web/app/src/gql/persisted-documents.json

> @graphql-hive/cli@0.57.4 start /Users/laurinquast/Projects/graphql-hive-3/packages/libraries/cli
> ./bin/dev app:create --target the-guild/graphql-hive/development --name web-app --version 0.0.2-example-version --format v2 /Users/laurinquast/Projects/graphql-hive-3/packages/web/app/src/gql/persisted-documents.json

App deployment "web-app@0.0.2-example-version" is created pending document upload. Uploading 177 new documents (0 already exist)...
100 / 177 (56%) documents uploaded...

App deployment "web-app@0.0.2-example-version" (177 new, 0 skipped) created.
Activate it with the "hive app:publish" command.
```

The reason is that we only create the `app_deployment` clickhouse record when the app deployment is published. I guess it is okay since this only happens if this is the first ever app deployment or a re-upload of the existing app deployment in case it failed. 🤔

packages/services/cdn-worker/src/artifact-storage-reader.ts
n1ru4l: My proposal would be that we store a list of all hashes that belong for a given app deployment version and check against that before resolving the actual document.

n1ru4l: With the current change, it seems possible to resolve and execute a persisted document hash that does not actually belong to the requested app deployment version.

Example path:

```
/appName/version/hash
```

Because the document is resolved solely by the hash, the version segment no longer enforces that the hash actually belongs to that specific deployment version. 🤔

This means a client could potentially execute a document that only existed in a retired deployment version, as long as the corresponding object still exists in the S3 bucket.

The only requirement would be knowing the version of any non-retired deployment for the same app.

Example scenario:

- some-version → retired
- another-version → active

If the hash H only existed in some-version, it may still be possible to execute it using:

```
/my-app/another-version/H
```

as long as the underlying S3 object for `H` still exists.

This effectively bypasses the intended version isolation and could allow execution of documents that should no longer be accessible. This could therefore be considered a security vulnerability.

______

To ensure version integrity, we could store a list of allowed document hashes for each app deployment version and verify that the requested hash belongs to that version before resolving the document.

To minimize latency, both lookups could be performed in parallel:

```
parallel(
  GET app-v/app-name/hash          -> returns the document
  GET app-v2/app-name/version      -> returns the list of valid hashes
)
```

Even in a worst-case scenario where an app deployment contains ~10,000 documents:

```
10,000 SHA-256 hashes ≈ ~300 KB
```

This is small enough to load quickly and cache efficiently.

This would enforce that only documents explicitly associated with a deployment version can be executed, preventing access to hashes from retired versions.

packages/libraries/cli/src/commands/app/create.ts
adambenhassen: Agreed. fetching the full list doesn't scale well as the hash set grows for a target/app. Sending our local hashes as input and having the server return only the ones that need uploading (as you suggested) is the better approach. Will rework this.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: They are effectively identical for this use case (no aggregation, just deduplication). I tested locally and the difference is negligible for our usecase. Per [Altinity's benchmarks](https://kb.altinity.com/altinity-kb-queries-and-syntax/distinct-vs-group-by-vs-limit-by/) at 100M rows, GROUP BY is ~2.3x faster and with nearly identical memory. Either works, switched to DISTINCT for readability.

packages/services/api/src/modules/app-deployments/module.graphql.ts
adambenhassen: Agreed, same comment as [above](#discussion_r2797399675)

packages/services/api/src/modules/app-deployments/module.graphql.ts
adambenhassen: Makes sense

packages/services/cdn-worker/src/artifact-storage-reader.ts
adambenhassen: Yeah, that sounds good. let's have that. 

**Update:** The only problem is that `apps-enabled` is only written during activation, and at that point we don't know the document format. The format is only known during `addDocumentToAppDeployment`.

I propose the following approach:

1. **Upload**: write `v1-inactive` or `v2-inactive` to `apps-enabled`
2. **Activation**: overwrite with `v1` or `v2` (read existing value, strip `-inactive`, write back)
3. **CDN**: `HEAD` -> `GET`, read body. If it contains `-inactive` or is `0` -> not active. If `v1`/`v2`/`1` -> active, use format accordingly.

**Backward compat:** existing deployments have `1` -> treated as active `v1`.

packages/libraries/cli/src/commands/app/create.ts
adambenhassen: I agree, this automates it. Also requires some changes to automatically resolve `target` as well, but it's worth it. Implementing.

packages/services/api/src/modules/app-deployments/module.graphql.ts
adambenhassen: Agreed

packages/services/server/src/environment.ts
adambenhassen: Sure, let's do that

packages/libraries/cli/src/commands/app/create.ts
adambenhassen: Ok, I like this find. I added a server-side check, so if documents already exist for a deployment, the format of the incoming batch must match. If you upload v1 first and then try v2 (or vice versa), the server rejects it with a clear error message. The check samples one existing hash and determines format by whether it's SHA-256.

packages/libraries/cli/src/commands/app/create.ts
adambenhassen: Ok, I fixed it regardless. The delta query should now also check documents from the current pending deployment, so re-running `app:create` on the same version corectly skips already-uploaded documents even before activation.

packages/services/cdn-worker/src/artifact-storage-reader.ts
adambenhassen: Okay. Solid observation. To summarize for implementation:

**Write side (Activation):**
- Before setting `apps-enabled` for v2 deployment, queries all document hashes from clichouse and writes them as a newline-separated manifest to `app-v2-manifest/{targetId}/{appName}/{appVersion}`

**Read side (CDN):**
- Fetches the document and manifest in parallel (Promise.all)
- If manifest exists, verifies the requested hash is in the allowed set
- If hash isn't in the manifest: returns notFound
- If no manifest (legacy deployments): skips the check for backward compat

~~**One thing to note:** Manifest for v1 is also fetched with the parallel execution. I think an extra 404 is an acceptable tradeoff for your improvement suggestion.~~

Edit: Following the implementation of proposed changes in the [prior discussion](https://github.com/graphql-hive/console/pull/7565#discussion_r3052609703), the tradeoff is eliminated.

adambenhassen: Agreed. check discussion below.

adambenhassen: Check my proposed changes above with the latest changes: https://github.com/graphql-hive/console/pull/7565/commits/5cc6655e01bf911c9d218eb87c93e596633023ad

adambenhassen:
Pending conflict resolution once all conversations are resolved.

packages/services/api/src/modules/app-deployments/providers/app-deployments-manager.ts
```diff
@@ -67,6 +67,7 @@ export class AppDeploymentsManager {
       name: string;
       version: string;
     };
+    hashes?: readonly string[] | null;
```
n1ru4l, line 70: Should there be an assertion that makes the `hashes` property mandatory for the v2 format?

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -431,28 +453,74 @@ export class AppDeployments {
...
+        appDeployment.targetId,
+        appDeployment.name,
+        appDeployment.version,
+      );
+      // Check if manifest already exists (written during createAppDeployment)
+      const existingManifest = await this.s3[0].client.fetch(
+        [this.s3[0].endpoint, this.s3[0].bucket, manifestKey].join('/'),
+        { method: 'HEAD', aws: { signQuery: true } },
+      );
+      if (existingManifest.statusCode !== 200) {
```
n1ru4l, line 470: Why wouldn't there be an existing manifest at that stage? It should have been written there during app create according to the logic?

If we then create it only based on clickhouse - it would not contain all hashes so produce an invalid manifest?

So my question is, why do we have this logic here for falling back in the first place?

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -194,47 +167,53 @@ export default gql`
...
     target: TargetReferenceInput
     appName: String!
     appVersion: String!
+    """
+    Optional list of document hashes the client intends to upload.
+    If provided, the response will include the subset that already exist on the server,
+    enabling delta uploads (skip uploading documents that already exist).
+    """
+    hashes: [String!]
   }
```
n1ru4l, line 217: The way I understand the implementation providing hashes indicates that this will be v2 (sha256) app deployment version. In that case, should this be documented more explicitly here?

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -151,6 +134,11 @@ export default gql`
       after: String @tag(name: "public")
       filter: ActiveAppDeploymentsFilter! @tag(name: "public")
     ): AppDeploymentConnection! @tag(name: "public")
+    """
+    Given a list of document hashes, returns the subset that already exist for this app (across all versions).
+    Used for delta uploads - CLI can skip uploading documents that already exist.
+    """
+    appDeploymentDocumentHashes(appName: String!, hashes: [String!]!): [String!]!
```
n1ru4l: Is this still needed if it there is now `CreateAppDeploymentOk.existingHashes`?

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -41,20 +33,19 @@ export default gql`
...
+    V1 format: version-scoped storage. Each version stores documents separately.
+    Allows any hash format. No cross-version deduplication.
+    """
+    V1
+    """
+    V2 format: content-addressed storage with SHA256 hashes.
+    Enables cross-version deduplication and delta uploads.
+    """
+    V2
   }
```
n1ru4l, line 49: NIT: Should we use more descriptive naming like `CUSTOM`/`ARBITRARY`, `SHA256` instead of `V1` and `V2`?

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -1886,9 +1617,141 @@ export class AppDeployments {
...
+    );
+
+    return hashes;
+  }
+
+  /**
+   * Check the format of existing documents for a deployment by sampling one hash.
+   * Returns 'v1', 'v2', or null if no documents exist.
+   */
+  async getDeploymentDocumentFormat(appDeploymentId: string): Promise<'v1' | 'v2' | null> {
```
n1ru4l: Why not store the app deployment version format (whether it is using arbitrary hashes or sha256 hashes) on the pg record (`app_deployments` table) when initially created via `Mutation.createAppDeployment`?

That way on both `createAppDeployment.addDocumentToAppDeployment` and `createAppDeployment.activeAppDeployments` we avoid the roundtrip to s3 or clickhouse in order to compute the format.

packages/services/cdn-worker/src/artifact-handler.ts
```diff
@@ -406,10 +406,13 @@ export const createArtifactRequestHandler = (deps: ArtifactRequestHandler) => {
...
-        false ===
-        (await deps.isAppDeploymentActive(params.targetId, params.appName, params.appVersion))
-      ) {
+      const deploymentStatus = await deps.getAppDeploymentStatus(
+        params.targetId,
+        params.appName,
+        params.appVersion,
+      );
+
+      if (!deploymentStatus.enabled) {
```
n1ru4l, line 415: The `apps-enabled` key can be repurposed  to store more information at the same time:
- whether the app is enabled
- what format it contains (arbitary hashes or sha256 hashes)
- which documents belong to this app version

That means for legacy versions created before these adjustments it would always contain `"0"` or `"1"` (or be non-existent). For new ones it could be a JSON object that contains the above information e.g.

```json5
{
  "isActive": true,
  // arbitrary/custom or sha256
  "hashFormat": "sha256",
  "documents": [
    "hash1",
    "hash2",
    "hash3",
    // ....
  ]
}
```

_______

This information could then be sufficient for the CDN handler to figure out on how to resolve the documents and it would reduce the need for one more S3 read.

Furthermore, it can act as a foundation for [other parallel ongoing efforts](https://guild-oss.slack.com/docs/TAYJ1FSUA/F0ASURR7CQ1).

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -431,28 +453,74 @@ export class AppDeployments {
...
+      const manifestKey = buildV2HashManifestKey(
+        appDeployment.targetId,
+        appDeployment.name,
+        appDeployment.version,
+      );
+      // Check if manifest already exists (written during createAppDeployment)
+      const existingManifest = await this.s3[0].client.fetch(
+        [this.s3[0].endpoint, this.s3[0].bucket, manifestKey].join('/'),
+        { method: 'HEAD', aws: { signQuery: true } },
+      );
```
n1ru4l, line 469: how do we guarantee that all documents that were specified in the manifest were uploaded/are available?

packages/services/api/src/modules/app-deployments/providers/app-deployments-manager.ts
adambenhassen: No separate format field exists on `CreateAppDeploymentInput`; hashes presence is the v2 signal. Without hashes, the code skips v2 writes and returns empty `existingHashes`, safely degrading to v1. The CLI enforces this: v2 always sends hashes, v1 sends undefined. Should I add an explicit format field to make the contract stricter?

EDIT: Added.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: Ok, with the `format` field now, from comment above, required on `CreateAppDeploymentInput`, v2 deployments always provide hashes at create time, so the manifest is always written during `createAppDeployment`. I'll remove the fallback and throw if the manifest is missing for a v2 activation, that surfaces the bug instead of silently producing an invalid manifest.

packages/services/api/src/modules/app-deployments/module.graphql.ts
adambenhassen: Addressed: added an explicit `format` field to `CreateAppDeploymentInput`

packages/services/api/src/modules/app-deployments/module.graphql.ts
adambenhassen: It's redundant now indeed. removing.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: Ok. I tended to avoid changes involving postgres. Implementing.

packages/services/cdn-worker/src/artifact-handler.ts
adambenhassen: Agreed, but out of scope for this PR. Merging status + format + hashes into one S3 key is a separate refactor.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: We don't currently verify that. The manifest controls which hashes are allowed for version isolation, but if a document wasn't uploaded, the CDN returns 404 naturally. Adding an activation-time check (compare manifest hashes against uploaded documents) would catch incomplete deployments, but that's a separate concern from this PR.

packages/services/api/src/modules/app-deployments/module.graphql.ts
adambenhassen: Agreed, it makes more sense indeed

adambenhassen:
> Aside from that I also noticed that skipped uploads documents are not visible via the Hive Console UI since we skip the Clickhouse insert, which would also kind of lead of versions with re-used documents not showing up in the conditional breaking change detection? 😅 Do I understand that correct?

Correct. in custom format, every document goes through `addDocumentsToAppDeployment` which inserts into clickhouse, so the document list and breaking change detection are always complete. With SHA256 dedup, the CLI skips uploading existing documents, so they never get clickhouse records for the new version. Which is the whole point of dedup.

**Proposal:** after `getExistingDocumentHashes` identifies deduped hashes, copy their clickhouse records (body, operation name, coordinates) to the new deployment ID via `INSERT...SELECT`. One clickhouse query during `createAppDeployment`, no client involvement. This gives SHA256 the same clickhouse behavior as custom format.

> Maybe we should step away on the code iteration on the PR for a moment have a bit of a more high-level RFC document, that makes it faster to iterate on while thinking about all involved components? Digging through the code to understand the implementation details and discover flaws is quite time consuming for me.

I'd rather if you continue reviewing this PR to the completion of its scope. the remaining items are additive improvements, not flaws. The core functionality is covered by extensive tests and the review hasn't surfaced any critical issues. The clickhouse copy for deduped documents and the merged S3 key (status + format + hashes) can be follow-ups without blocking what's already working.

PR #7584, opened 2026-01-27
remove compare to latest schema version feature flag

n1ru4l:
### Background

[Few years](https://github.com/graphql-hive/console/pull/1884) back we introduced `compareToPreviousComposableVersion` to introduce two things:
- allow publishing schemas that would result in a failed schema version (to work around annoying timings that one might need to do sometimes when publishing a service schema)
- provide a proper diff between the actual functional schema versions

We gradually enabled this behaviour for most of our customers back then.

Then in [April 2024](https://github.com/graphql-hive/console/pull/4394), we enabled this behaviour for all new federation and stitching projects

### Description

This removes the organization scoped feature flag and makes it the default behaviour for all stitching and federation projects. Therefore, it is now possible for stitching and projects project types to publish service schemas to the schema registry that break composition.

For monolithic projects, we still still reject an invalid schema.

### DIFF
--- a/.changeset/slimy-trams-search.md
+++ b/.changeset/slimy-trams-search.md
@@ -4,6 +4,6 @@
 
 Enable automatic retrieval of schema changes by comparing with the latest composable version. This has already been the default for new projects created after April 2024. 
 
-The previous behavior was preserved only for legacy compatibility.
+Federation and schema stitching projects can now publish service schemas to the registry even if those schemas would break composition. This has also been the default behavior for new projects created after April 2024.
 
-To ensure every version publishd to the schema registry is composable, we recommend to first check the schema against the registry before publishing.
+To ensure every version publishd to the schema registry is composable, we recommend to first check the schema against the registry **before** publishing.


.changeset/slimy-trams-search.md
```diff
@@ -0,0 +1,9 @@
+---
+'hive': minor
+---
+
+Enable automatic retrieval of schema changes by comparing with the latest composable version. This has already been the default for new projects created after April 2024. 
+
+The previous behavior was preserved only for legacy compatibility.
+
+To ensure every version publishd to the schema registry is composable, we recommend to first check the schema against the registry before publishing.
```
jdolle:
```suggestion
To ensure every version published to the schema registry is composable, we recommend to first check the schema against the registry before publishing.
```

packages/services/api/src/modules/schema/providers/models/single.ts
```diff
@@ -276,8 +270,23 @@ export class SingleModel {
...
+      return {
+        conclusion: SchemaPublishConclusion.Reject,
+        reasons: [
+          {
+            code: PublishFailureReasonCode.CompositionFailure,
+            compositionErrors: compositionCheck.reason.errorsBySource.graphql,
+          },
+        ],
+      };
+    }
```
n1ru4l, line 286: For single schemas we keep the previous behaviour, since it does not really make sense to adopt the one from federation/stitching where there are multiple services.

Issue #7557, opened 2026-01-26
hive-console-sdk 0.3.4 is not compatible with hive-apollo-router-plugin 3.0.1


yanns:
Hi,

you've released hive-console-sdk 0.3.4 that is using graphql-tools 0.5.0.
hive-apollo-router-plugin is using hive-console-sdk but is using graphql-tools 0.4.2.

Both versions of graphql-tools are sadly not compatible:
```
error[E0308]: mismatched types
   --> /Users/yannsimon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hive-apollo-router-plugin-3.0.1/src/usage.rs:317:53
    |
317 | ...                   schema,
    |                       ^^^^^^ expected `Document<'static, String>`, found a different `Document<'static, String>`
    |
note: there are multiple different versions of crate `graphql_tools` in the dependency graph
   --> /Users/yannsimon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/graphql-tools-0.5.0/src/parser/schema/ast.rs:9:1
    |
  9 | pub struct Document<'a, T: Text<'a>> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this is the expected type
    |
   ::: /Users/yannsimon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/graphql-tools-0.4.2/src/parser/schema/ast.rs:9:1
    |
  9 | pub struct Document<'a, T: Text<'a>> {
    | ------------------------------------ this is the found type
    = help: you can use `cargo tree` to explore your dependency tree
```

kamilkisiela:
```
hive-apollo-router-plugin = "3.0.1"
  ^ graphql-tools = "0.4.2"
  ^ hive-console-sdk = "0.3.3"

hive-console-sdk = "0.3.3"
  ^ graphql-tools = "0.4.2"
```

The `hive-apollo-router-plugin` on its latest version (see above) depends on graphql-tools v0.4.2 and hive-console-sdk v0.3.3.
The `hive-console-sdk@0.3.3` depends on graphql-tools v0.4.2

The latest version of `hive-apollo-router-plugin` is not supporting latest sdk, but why is that a problem? Can't you use the version that's supported?

yanns:
Your link does not seem to work.
`hive-apollo-router-plugin` is depending on `hive-console-sdk = "0.3.3"`. That's true.
[...]

PR #7585, opened 2026-01-27
fix(hive-apollo-router-plugin): Upgrade to latest `hive-console-sdk` and drop direct dependency on `graphql-tools`

dotansimha:
Closes https://github.com/graphql-hive/console/issues/7557


yanns:
nice!

PR #7592, opened 2026-01-27
Implement new color themes

jonathanawesome:
This PR introduces a new color palette for our previously-default dark mode and adds a light mode palette.

My original ambition here was not only to replace the gray palette with swappable neutral palettes for light and dark, but to also update the non-gray/neutral colors. While I have the neutral palette in Figma variable modes, I did _not_ have most of the other colors figured out in the same fashion. In the end, I've kept the non-neutral colors as they're mostly sufficient and enable this light-mode focused PR to get merged.

### DIFF
--- a/.changeset/thin-animals-smell.md
+++ b/.changeset/thin-animals-smell.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+added light mode to console


n1ru4l (approved):
Looking nice 💯

mskorokhodov (approved):

PR #7601, opened 2026-01-28
fix: inline app deployment lookup

n1ru4l:
### Background

Passing 10_000+ input ids of all affected deployments to each of the queries rejects them from ClickHouse.

### Description

Changes the implementation to use an inlined query for retrieving the affected app deployments. This will slightly affect query performance, but allow the query to be executed in the first place.

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
+++ b/packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
@@ -987,7 +987,13 @@ export class AppDeployments {
             any(app_name) AS "appName",
             any(app_version) AS "appVersion"
           FROM app_deployments
-          PREWHERE target_id = ${args.targetId}
+          PREWHERE
+            target_id = ${args.targetId}
+            ${
+              args.excludedAppDeploymentNames?.length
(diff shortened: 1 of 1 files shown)


packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -1011,27 +1017,11 @@ export class AppDeployments {
...
-      activeDeployments = activeDeployments.filter(d => !excludedNamesSet.has(d.appName));
-      if (activeDeployments.length !== originalCount) {
-        this.logger.debug(
-          'Filtered out %d app deployments by excluded names (targetId=%s, excludedNames=%o)',
-          originalCount - activeDeployments.length,
-          args.targetId,
-          args.excludedAppDeploymentNames,
-        );
-      }
-    }
```
n1ru4l, line 1027: We can do this filtering as part of the initial query

jdolle (approved):

PR #7605, opened 2026-01-29
fix(yoga): handle malformed persisted document IDs with proper error instead of 500

adambenhassen:
### Background

Malformed `document_id` values cause 500 errors instead of helpful validation errors. The core library throws `PersistedDocumentValidationError` but the yoga plugin wasn't catching it.

### Description
  Added error handling in `@graphql-hive/yoga` to catch validation errors and return a proper GraphQL error with `INVALID_DOCUMENT_ID` code.

### DIFF
--- a/packages/libraries/yoga/src/index.ts
+++ b/packages/libraries/yoga/src/index.ts
@@ -239,9 +239,26 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
               return null;
             },
             async getPersistedOperation(key, _request, context) {
-              const document = await experimentalPersistedDocs.resolve(key, {
-                waitUntil: context.waitUntil,
-              });
+              let document: string | null;
+              try {
+                document = await experimentalPersistedDocs.resolve(key, {
+                  waitUntil: context.waitUntil,
+                });
+              } catch (error) {
+                if (
+                  error &&
+                  typeof error === 'object' &&
+                  'code' in error &&
+                  error.code === 'INVALID_DOCUMENT_ID' &&
+                  'message' in error &&
+                  typeof error.message === 'string'
+                ) {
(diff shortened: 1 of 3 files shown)


packages/libraries/yoga/src/index.ts
```diff
@@ -239,9 +239,26 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
               return null;
             },
             async getPersistedOperation(key, _request, context) {
-              const document = await experimentalPersistedDocs.resolve(key, {
-                waitUntil: context.waitUntil,
-              });
```
n1ru4l, line 244: @santino You introduced `PersistedDocumentValidationError`, any specific reason for that one not being a `GraphQLError`? - if there isn't I would propose to instead just make that error be a `GraphQLError`, so the catch statement here is not needed.

santino: @n1ru4l My reason was to keep the validation error intentionally framework-agnostic. Ideally core shouldn't depend on `graphql-js`. The Yoga plugin seems like a good place to translate core errors into GraphQL errors, just like we already do with customErrors.notFound. The catch handler to convert `PersistedDocumentValidationError` to a `GraphQLError` with HTTP 400 seems like a good approach to me.
This keeps core clean while properly handling the error in the framework-specific layer.

packages/libraries/yoga/src/index.ts
santino, line 1: This approach to read the error object works. Alternatively we could export `PersistedDocumentValidationError` from core and import it in the yoga library to use it like this `if (err instanceof PersistedDocumentValidationError)`. Just a comment if you think it's worth

n1ru4l: I do not have a strong opinion here. Either, solution works with me!

PR #7620, opened 2026-02-04
Enhancement: New lab preflight prompt metadata

mskorokhodov:
### Description

Ability to call prompt with: title, defaultValue, placeholder and markdown description

Example:

```
await lab.prompt('Email', '', {
  placeholder: 'example@example.com',
  description: 'Read more in [Docs](https://google.com)'
})
```
Screenshot 2026-02-04 at 10 18 31

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/web/app/src/laboratory/lib/preflight.ts
+++ b/packages/web/app/src/laboratory/lib/preflight.ts
@@ -3,6 +3,13 @@ import cryptoJsSource from 'crypto-js/crypto-js.js?raw';
 import type { LaboratoryEnv, LaboratoryEnvActions, LaboratoryEnvState } from '@/laboratory/lib/env';
 import { LaboratoryPlugin } from '@/laboratory/lib/plugins';
 
+export interface LaboratoryPreflightPromptField {
+  title?: string;
+  defaultValue?: string;
+  description?: string;
+  placeholder?: string;
+}
+
 export interface LaboratoryPreflightLog {
   level: 'log' | 'warn' | 'error' | 'info' | 'system';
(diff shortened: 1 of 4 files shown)


packages/web/app/src/laboratory/lib/preflight.ts
```diff
@@ -3,6 +3,13 @@ import cryptoJsSource from 'crypto-js/crypto-js.js?raw';
 import type { LaboratoryEnv, LaboratoryEnvActions, LaboratoryEnvState } from '@/laboratory/lib/env';
 import { LaboratoryPlugin } from '@/laboratory/lib/plugins';
 
+export interface LaboratoryPreflightPromptField {
+  title?: string;
+  defaultValue?: string;
+  description?: string;
+  placeholder?: string;
+}
```
adambenhassen, line 11: (Minor observation) I think this is currently unused, but it seems to have potential to be used in `context.tsx:75` and other locations to dry up the repeated inline type definitions. Maybe that was the initial intention?

mskorokhodov: Thanks, looks like i've had good idea it use it and then forgot:) I actually moved it to interface in order to support multiple fields rendering in same modal, so will implement it in next PR.

PR #7624, opened 2026-02-04
fix(app-deployments): resolve hard limit cutoff in stale deployments query

adambenhassen:
### Background
The `activeAppDeployments` query with date filters returned empty results for targets with >1000 deployments. The postgres `LIMIT 1000` was cutting off older stale deployments before filtering.

### Description
Implement clickhouse-first query strategy when date filters are present:
  - Query clickhouse first to identify stale deployment IDs
  - Fetch only those deployments from postgres (bypassing LIMIT 1000)

Also fixed Joda date format: YYYY -> yyyy, SS -> ss.

### DIFF
--- a/integration-tests/tests/api/app-deployments.spec.ts
+++ b/integration-tests/tests/api/app-deployments.spec.ts
@@ -2030,6 +2030,72 @@ test('activeAppDeployments filters by neverUsedAndCreatedBefore', async () => {
   expect(result.target?.activeAppDeployments.edges[0].node.createdAt).toBeTruthy();
 });
 
+test('activeAppDeployments works for > 1000 records with a date filter (neverUsedAndCreatedBefore) set', async () => {
+  const { createOrg, ownerToken } = await initSeed().createOwner();
+  const { createProject, setFeatureFlag, organization } = await createOrg();
+  await setFeatureFlag('appDeployments', true);
+  const { createTargetAccessToken, project, target } = await createProject();
+  const token = await createTargetAccessToken({});
+
+  // seed 1,200 app deployments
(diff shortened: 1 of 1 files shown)


packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -1391,12 +1495,16 @@ export class AppDeployments {
         FROM
           "app_deployments"
         WHERE
-          "target_id" = ${args.targetId}
+          ${
+            staleDeploymentIds
+              ? sql`"id" = ANY(${sql.array(staleDeploymentIds, 'uuid')})`
```
jdolle: we recently had an issue where the list of IDs passed to a query was too long and was causing the request to fail.
It looks like we're not capping the number of stale deployments that could be returned and therefore this could also have issues.

This was resolved previously by inlining the select. Should we also do that here?

adambenhassen: That's a very neat observation! Inlining the select would have been a great solution, but the problem is that we're doing separate database queries here between postgres and clickhouse, so can't really do a true subquery here. 

One approach I just thought of, is to process in chunks of 1000 (or more) whenever they exceed that threshold, to avoid the potential query too long failure. Or, we could just add `retired_at` and `created_at` columns (and maybe others) and get rid of the postgres calls at once.

I think we can do both, but without migrating old data, so we basically populate the `retired_at` and `created_at` columns as we move forward, and at some point we just remove the postgres calls.

jdolle (commented):
Please add an integration test that covers this case if possible

n1ru4l (approved):

jdolle (approved):

PR #7635, opened 2026-02-05
add `Mutation.updateSchemaComposition` to the public GraphQL API schema

n1ru4l:
### Background

CONSOLE-1790

### Description

Reshapes the `Mutation.updateSchemaComposition` types a bit to leverage `ProjectReferenceInput` and follow other design practices and adds it to the public API schema.

Note: The Hive check against live traffic is failing. The reason is that some folks started using it from our private GraphQL schema. However, the usage is quite small, thus I decided it is okay to break the API before adding it to the true public schema!

### Checklist

- [x] Input validation
- [x] Output encoding
- [x] Session management

### DIFF
--- a/packages/services/api/src/modules/schema/module.graphql.ts
+++ b/packages/services/api/src/modules/schema/module.graphql.ts
@@ -64,6 +64,7 @@ export default gql`
     external: SchemaCompositionExternalMethodInput
     """
     Use the legacy federation v1 schema composition.
+    """
     legacy: Boolean
   }
 


packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -43,55 +48,69 @@ export default gql`
...
   }
 
-  input UpdateSchemaCompositionNativeInput {
-    organizationSlug: String!
-    projectSlug: String!
+  input SchemaCompositionMethodInput @oneOf @tag(name: "public") {
+    """
+    Use the native federation schema composition. This is the default method for new projects.
+    """
+    native: Boolean
```
n1ru4l, line 60: Some thoughts on the schema design here:

Drawback of using a `Boolean` here is that we could not add additional properties specifically for the native composition via this existing config option in a backwards compatible way. A more future-proof solution would be to do something like the following:

```graphql
input SchemaCompositionNativeMethodInput {
  _: Boolean # noop field since a type without fields can not exist :(
}

input SchemaCompositionMethodInput @oneOf @tag(name: "public") {
  native: SchemaCompositionConfigurationNativeMethodInput
}
```

Though having a `_` field is also a bit ugly, but GraphQL types must have a at least one field (aside from `__typename`)...

In case the (right now very unlikely) scenario where we would have sub configuration for either `native` or `legacy` composition, we could deprecate existing fields and introduce new ones with the sub configuration.

```graphql
input SchemaCompositionNativeCompositionMethodInput {
  someConfigOptionICannotImagineRightNow: Boolean
}

input SchemaCompositionMethodInput @oneOf @tag(name: "public") {
  native: Boolean @deprecated(reason: "use 'SchemaCompositionMethodInput.nativeComposition' instead.")
  nativeComposition: SchemaCompositionConfigurationNativeCompositionMethodInput
}
```

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -43,55 +48,69 @@ export default gql`
...
-    organizationSlug: String!
-    projectSlug: String!
-    endpoint: String!
-    secret: String!
+  input UpdateSchemaCompositionInput @tag(name: "public") {
+    """
+    Reference to the project on which the schema composition should be applied.
+    """
+    project: ProjectReferenceInput!
+    """
```
n1ru4l, line 76: I moved the input that was shared in all input types to be it's standalone input.

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -43,55 +48,69 @@ export default gql`
...
     error: UpdateSchemaCompositionError
   }
 
-  interface UpdateSchemaCompositionError implements Error {
+  interface UpdateSchemaCompositionError @tag(name: "public") {
     message: String!
   }
 
-  type UpdateSchemaCompositionNativeError implements UpdateSchemaCompositionError & Error {
+  type UpdateSchemaCompositionNativeError implements UpdateSchemaCompositionError
```
n1ru4l, line 98: Federation stuff, all things that implement `Error` must be public, otherwise we cannot compose the public schema contract.

PR #7660, opened 2026-02-09
fix: sentry graphql api

n1ru4l:
### Background

Our GraphQL API stopped reporting errors to sentry 😓 

### Description

This PR solves these things:
- Resume reporting unexpected errors to sentry
- Strip out HTTP auth credentials from reported URLs

It does not solve logging http auth credentials to the logs. This I will be addressed in a follow up task.

### DIFF
--- a/packages/services/api/src/modules/token/resolvers/Mutation/createToken.ts
+++ b/packages/services/api/src/modules/token/resolvers/Mutation/createToken.ts
@@ -10,7 +10,6 @@ export const createToken: NonNullable<MutationResolvers['createToken']> = async 
   { input },
   { injector },
 ) => {
-  throw new Error('BUBUBUBBU');
   const CreateTokenInputModel = z.object({
     name: TokenNameModel,
   });


packages/services/service-common/src/sentry.ts
n1ru4l, line 1: There was some config that we applied on the graphql api sentry instance but did not apply to other services. I moved the config here to be shared everywhere as the additional context seems helpful for other services as well.

packages/web/app/src/env/backend.ts
```diff
@@ -243,10 +242,4 @@ Sentry.init({
   dsn: env.sentry?.dsn,
   release: env.release,
   environment: env.environment,
-  integrations: [
-    // HTTP integration is only available on the server
-    new Integrations.Http({
-      tracing: false,
-    }),
-  ],
```
n1ru4l, line 251: We do not use sentry for tracing? 😕

packages/services/server/src/graphql-handler.ts
```diff
@@ -82,25 +73,73 @@ function hasFastifyRequest(ctx: unknown): ctx is {
...
+  return useErrorHandler(({ errors, context: unsafeContest }): void => {
     // Not sure what changed, but the `context` is now an object with a contextValue property.
     // We previously relied on the `context` being the `contextValue` itself.
-    const ctx = ('contextValue' in context ? context.contextValue : context) as Context;
+    const context: {
+      operationName: string | null;
+      variableValues: Record<string, unknown> | null | undefined;
+      document: DocumentNode;
+      contextValue: Context;
+    } = unsafeContest as any;
```
n1ru4l: I tried to use the generic of `useErrorHandler`, but seems like there is a typing flaw in envelop/yoga itself. I used this reassign hack for declaring the types. I am not proud of it but it works. If anyone has a better solution I am happy to hear that.

packages/services/server/src/graphql-handler.ts
```diff
@@ -120,49 +159,6 @@ function useNoIntrospection(params: {
   };
 }
 
-export function useHiveSentry() {
```
n1ru4l, line 123: I spend quite some time typing to figure out within the envelop sentry plugin why the error reporting is not working. In the end I decided to just inline the logic that we use into our own error handler as @dotansimha proposed.

packages/services/service-common/src/sentry.ts
```diff
@@ -0,0 +1,64 @@
...
+      }
+
+      return event;
+    },
+    beforeBreadcrumb(breadcrumb) {
+      if (breadcrumb.message) {
+        breadcrumb.message = scrubBasicAuth(breadcrumb.message);
+      }
+      return breadcrumb;
+    },
```
n1ru4l, line 59: 👍 

Image

packages/services/service-common/src/sentry.ts
n1ru4l: This also now holds the additional masking logic.

packages/services/api/src/modules/shared/providers/http-client.ts
```diff
@@ -88,17 +102,27 @@ export class HttpClient {
           );
           span.setAttribute('error.message', details || '');
 
-          logger.error(error);
+          logger.error('HTTP request failed (message=%s, code=%s)', error.message, error.code);
```
n1ru4l, line 105: This was logging the full error object, which is super unhelpful, noisy and takes years to scroll through in case any http error occurs.

 Image

packages/services/api/src/modules/shared/providers/http-client.ts
```diff
@@ -88,17 +102,27 @@ export class HttpClient {
...
+
+class HiveHttpClientError extends Error {
+  constructor(
+    message: string,
+    public code: string,
+  ) {
+    super(message);
+    this.name = 'HiveHttpClientError';
+  }
+}
```
n1ru4l, line 128: Stripped down, less verbose version of the error raised by `got`. Reason for this mainly is that within `useErrorHandler` we always print the error - leading to a second version of https://github.com/graphql-hive/console/pull/7660#discussion_r2783182488

packages/services/api/src/modules/shared/providers/http-client.ts
```diff
@@ -14,6 +14,20 @@ interface HttpClientOptions extends OptionsOfJSONResponseBody {
 
 type HttpOptions = Omit<HttpClientOptions, 'method' | 'throwHttpErrors' | 'resolveBodyOnly'>;
 
+const gotInstance = got.extend({
+  headers: {
+    'user-agent': '',
```
n1ru4l, line 19: We do not need to tell everyone what http client we use.

packages/services/api/src/modules/shared/providers/http-client.ts
dotansimha: Yes, thank you for that.

packages/services/service-common/src/sentry.ts
```diff
@@ -0,0 +1,61 @@
...
+  dedupeIntegration,
+  extraErrorDataIntegration,
+} from '@sentry/integrations';
+import { httpIntegration, init, linkedErrorsIntegration } from '@sentry/node';
+import { scrubBasicAuth } from './scrub';
+
+/**
+ * Initialize Sentry SDK with our commong configuration options.
+ */
+export function sentryInit(args: {
```
dotansimha, line 13: nice!

packages/services/service-common/src/sentry.ts
dotansimha: 😍

PR #7661, opened 2026-02-09
show all linked auth methods in members list

XiNiHa:
### Background

In #7390, the organization member list page has been changed to display no information about users' authentication methods due to having no way to figure that out after the introduction of account linking. However the feature got requested to be brought back, therefore this PR reimplements it.

### Description

The `User.providers` field is added in the GraphQL API, and the existing `User.provider` field has changed to return the first element of the `providers` array.

2026-02-09T22:11:15,611219440+09:00

### Checklist

- [ ] Input validation
- [x] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing

### DIFF
--- a/packages/web/app/src/components/organization/members/list.tsx
+++ b/packages/web/app/src/components/organization/members/list.tsx
@@ -37,24 +37,24 @@ import { MemberRolePicker } from './member-role-picker';
 export const authProviderToIconAndTextMap: Record<
   GraphQLSchema.AuthProviderType,
   {
-    icon: IconType;
+    Icon: IconType;
     text: string;
   }
 > = {
   [GraphQLSchema.AuthProviderType.Google]: {
-    icon: FaGoogle,
+    Icon: FaGoogle,
     text: 'Google OAuth 2.0',
   },
   [GraphQLSchema.AuthProviderType.Github]: {
-    icon: FaGithub,
+    Icon: FaGithub,
     text: 'GitHub OAuth 2.0',
   },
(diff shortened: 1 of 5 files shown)


packages/web/app/src/components/organization/members/list.tsx
```diff
@@ -135,9 +164,21 @@ const OrganizationMemberRow = memo(function OrganizationMemberRow(props: {
...
+                    const Icon = authProviderToIconAndTextMap[provider].icon;
+                    return <Icon key={provider} className="mx-auto size-5" />;
+                  })}
+                </div>
+              </TooltipTrigger>
+              <TooltipContent>
+                User's authentication methods: {providerTexts.join(', ')}
+              </TooltipContent>
+            </Tooltip>
+          </TooltipProvider>
```
n1ru4l: While this is functional, it will not look optimal if there are 2+ providers. I propose that we move the list of login methods to a column behind the name. Furthermore, based on whether OIDC-only login is enabled we could grey out the methods and provide info in the tooltip on why this method is greyed out (with the exception being the admin as he can login via any method). What do you think about this?

n1ru4l: Here some quick AI mockups of how it could look like

Image

Image

n1ru4l (approved):

PR #7662, opened 2026-02-09
Feature/insights filters [CONSOLE-1607]

jonathanawesome:
This PR implements a saved filters feature for Insights.

The feature has the following requirements as per internal and external stakeholders:
- Permissions
   - All users should be able to save filters that are scoped for their own use
   - Users with appropriate permissions should also be able to save filters to a _project_, making them available across _targets_
- Users must be able to _manage_ saved filters
   - rename saved filters
   - delete saved filters
   - update saved filters in a similar way that a user would generate a new filter (all filter options available)
   - updating saved filters must be context aware (UI should notify users when modified filters will produce unintended results)
- The existing UI does not allow for filtering client _versions_
   - Users should be able to filter for versions within a given client
   - This addition should work seamlessly within the current search param pattern

As a reminder, we decided early on in this discussion that we should invest in this opportunity to improve our filtering UI. This resulted in significant UI changes with the following implications:
- Introduce new/replacement components that demonstrate alternative patterns with the goal of making feature implementation faster and more consistent
   - new components in `components/base` 
   - introduce `@base-ui/react` as an alternative to headless/radix/etc
   - we restrict the component APIs from allowing `className` in favor of providing strict style variants
- Replace the existing Insights operations/clients filter `Sheets` with a Linear-like approach to filtering
   - Include DateRange filtering as a core/default filter

### DIFF
--- a/deployment/services/db-migrations.ts
+++ b/deployment/services/db-migrations.ts
@@ -47,6 +47,7 @@ export function deployDbMigrations({
         // Since K8s job are immutable, we can't edit or ask K8s to re-run a Job, so we are doing a
         // pseudo change to an env var, which causes Pulumi to re-create the Job.
         IGNORE_RERUN_NONCE: force ? Date.now().toString() : '0',
+        SUPERTOKENS_AT_HOME: '1',
       },
     },
     [clickhouse.deployment, clickhouse.service, ...(dependencies || [])],
--- a/deployment/services/graphql.ts
+++ b/deployment/services/graphql.ts
@@ -74,6 +74,7 @@ export function deployGraphQL({
   sentry: Sentry;
 }) {
   const apiConfig = new pulumi.Config('api');
+  const supertokensConfig = new pulumi.Config('supertokens');
   const apiEnv = apiConfig.requireObject<Record<string, string>>('env');
 
   const hiveConfig = new pulumi.Config('hive');
@@ -93,6 +94,11 @@ export function deployGraphQL({
   });
   const hiveUsageSecret = new ServiceSecret('hive-usage', {
     usageAccessToken: hiveConfig.requireSecret('usageAccessToken'),
+  });
+  const supertokensSecrets = new ServiceSecret('supertokens-at-home', {
+    refreshTokenKey: supertokensConfig.requireSecret('refreshTokenKey'),
+    accessTokenKey: supertokensConfig.requireSecret('accessTokenKey'),
+    bypassRateLimitKey: supertokensConfig.requireSecret('bypassRateLimitKey'),
   });
 
   return (
@@ -149,6 +155,7 @@ export function deployGraphQL({
               ? observability.tracingEndpoint
               : '',
           S3_MIRROR: '1',
+          SUPERTOKENS_AT_HOME: '1',
         },
         exposesMetrics: true,
         port: 4000,
@@ -215,6 +222,10 @@ export function deployGraphQL({
         persistedDocumentsSecret,
         'cdnAccessKeyId',
       )
+      // Supertokens
+      .withSecret('SUPERTOKENS_REFRESH_TOKEN_KEY', supertokensSecrets, 'refreshTokenKey')
+      .withSecret('SUPERTOKENS_ACCESS_TOKEN_KEY', supertokensSecrets, 'accessTokenKey')
+      .withSecret('SUPERTOKENS_RATE_LIMIT_BYPASS_KEY', supertokensSecrets, 'bypassRateLimitKey')
       // Zendesk
       .withConditionalSecret(zendesk.enabled, 'ZENDESK_SUBDOMAIN', zendesk.secret, 'subdomain')
       .withConditionalSecret(zendesk.enabled, 'ZENDESK_USERNAME', zendesk.secret, 'username')
--- a/packages/services/api/src/create.ts
+++ b/packages/services/api/src/create.ts
@@ -159,6 +159,7 @@ export function createRegistry({
     baseUrl: string;
     rateLimit: null | {
       ipHeaderName: string;
+      bypassKey: string | null;
     };
   } | null;
   schemaConfig: SchemaModuleConfig;
--- a/packages/services/api/src/modules/shared/providers/redis-rate-limiter.ts
+++ b/packages/services/api/src/modules/shared/providers/redis-rate-limiter.ts
@@ -1,6 +1,8 @@
+import { timingSafeEqual } from 'node:crypto';
 import { Inject, Injectable, Scope } from 'graphql-modules';
 import { type Redis } from 'ioredis';
 import { type FastifyRequest } from '@hive/service-common';
+import { sha256 } from '../../auth/lib/supertokens-at-home/crypto';
 import { Logger } from './logger';
 import { REDIS_INSTANCE } from './redis';
 import { RateLimitConfig } from './tokens';
@@ -9,7 +11,8 @@ import { RateLimitConfig } from './tokens';
   scope: Scope.Singleton,
 })
 export class RedisRateLimiter {
-  logger: Logger;
+  private logger: Logger;
+  private bypassKey: Buffer | null;
 
   constructor(
     @Inject(REDIS_INSTANCE) private redis: Redis,
@@ -17,6 +20,7 @@ export class RedisRateLimiter {
     logger: Logger,
   ) {
     this.logger = logger.child({ module: 'RateLimiter' });
+    this.bypassKey = config.config?.bypassKey ? Buffer.from(config.config.bypassKey) : null;
   }
 
   /**
@@ -27,11 +31,25 @@ export class RedisRateLimiter {
     /** duration of the time window */
     timeWindowSeconds = 5 * 60,
     /** maximum amount of requests allowed in the time window */
-    maxActionsPerTimeWindow = 10,
+    maxActionsPerTimeWindow = 30,
   ) {
     if (!this.config.config) {
       this.logger.debug('rate limiting is disabled');
       return false;
+    }
+
+    const cookies: undefined | Record<string, undefined | string> = (req as any).cookies;
+
+    if (this.bypassKey !== null && cookies?.['sBypassRateLimitKey']) {
+      const incomingBypassKey = Buffer.from(cookies['sBypassRateLimitKey']);
+
+      if (
+        this.bypassKey.length === incomingBypassKey.length &&
+        timingSafeEqual(incomingBypassKey, this.bypassKey)
+      ) {
+        this.logger.debug('rate limit bypassed via provided key');
+        return false;
+      }
     }
 
     req.routeOptions.url;
@@ -46,18 +64,20 @@ export class RedisRateLimiter {
       ip = req.headers[this.config.config.ipHeaderName] as string;
     }
 
-    const key = `server-rate-limiter:${req.routeOptions.url}:${ip}`;
+    const ipHash = sha256(ip);
+
+    const key = `server-rate-limiter:${req.routeOptions.url}:${ipHash}`;
 
     const current = await this.redis.incr(key);
     if (current === 1) {
       await this.redis.expire(key, timeWindowSeconds);
     }
     if (current > maxActionsPerTimeWindow) {
-      this.logger.debug('request is rate limited');
+      this.logger.debug('request is rate limited (ip_hash=%s)', ipHash);
       return true;
     }
 
-    this.logger.debug('request is not rate limited');
+    this.logger.debug('request is not rate limited (ip_hash=%s)', ipHash);
     return false;
   }
 }
--- a/packages/services/api/src/modules/shared/providers/tokens.ts
+++ b/packages/services/api/src/modules/shared/providers/tokens.ts
@@ -7,6 +7,7 @@ export class RateLimitConfig {
   constructor(
     public readonly config: null | {
       ipHeaderName: string;
+      bypassKey: string | null;
     },
   ) {}
 }
--- a/packages/services/server/src/environment.ts
+++ b/packages/services/server/src/environment.ts
@@ -106,6 +106,7 @@ const SuperTokensModel = zod.union([
     SUPERTOKENS_API_KEY: zod.string(),
     SUPERTOKENS_RATE_LIMIT: emptyString(zod.union([zod.literal('1'), zod.literal('0')]).optional()),
     SUPERTOKENS_RATE_LIMIT_IP_HEADER_NAME: emptyString(zod.string().optional()),
+    SUPERTOKENS_RATE_LIMIT_BYPASS_KEY: emptyString(zod.string().optional()),
   }),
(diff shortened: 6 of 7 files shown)


packages/migrations/src/actions/2026.02.07T00-00-00.saved-filters.ts
```diff
@@ -0,0 +1,44 @@
...
+export default {
+  name: '2026.02.07T00-00-00.saved-filters.ts',
+  run: ({ sql }) => sql`
+CREATE TABLE "saved_filters" (
+  "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+  "project_id" uuid NOT NULL REFERENCES "projects"("id") ON DELETE CASCADE,
+  "type" text NOT NULL,
+  "created_by_user_id" uuid NOT NULL REFERENCES "users"("id") ON DELETE CASCADE,
+  "updated_by_user_id" uuid REFERENCES "users"("id") ON DELETE SET NULL,
+  "name" varchar(100) NOT NULL,
```
n1ru4l: nit: I would keep this text, from past experience changing the length later on can be annoying as it requires rewriting the whole table. Pg does not store text or varchar differently, so it makes not really a difference.

packages/migrations/src/actions/2026.02.07T00-00-00.saved-filters.ts
```diff
@@ -0,0 +1,44 @@
...
+  "description" text,
+  "filters" jsonb NOT NULL,
+  "visibility" text NOT NULL,
+  "views_count" integer NOT NULL DEFAULT 0,
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "updated_at" timestamptz NOT NULL DEFAULT now(),
+  PRIMARY KEY ("id"),
+  CONSTRAINT "saved_filters_name_length" CHECK (char_length("name") >= 1 AND char_length("name") <= 100),
+  CONSTRAINT "saved_filters_type_check" CHECK ("type" IN ('INSIGHTS')),
+  CONSTRAINT "saved_filters_visibility_check" CHECK ("visibility" IN ('private', 'shared'))
```
n1ru4l: enums exist in pg: https://www.postgresql.org/docs/current/datatype-enum.html

packages/migrations/src/actions/2026.02.07T00-00-00.saved-filters.ts
```diff
@@ -0,0 +1,44 @@
...
+  "name" varchar(100) NOT NULL,
+  "description" text,
+  "filters" jsonb NOT NULL,
+  "visibility" text NOT NULL,
+  "views_count" integer NOT NULL DEFAULT 0,
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "updated_at" timestamptz NOT NULL DEFAULT now(),
+  PRIMARY KEY ("id"),
+  CONSTRAINT "saved_filters_name_length" CHECK (char_length("name") >= 1 AND char_length("name") <= 100),
+  CONSTRAINT "saved_filters_type_check" CHECK ("type" IN ('INSIGHTS')),
```
n1ru4l: we never really did constraints for input lengths as the business logic layer should already cover this. Covering it in the business logic layer only also makes it easier as loosening the constraints will not require locking the whole database in the future and strictening them will not conflict with existing records.

packages/services/api/src/modules/audit-logs/providers/audit-logs-types.ts
```diff
@@ -354,6 +354,33 @@ export const AuditLogModel = z.union([
       organizationAccessTokenId: z.string().uuid(),
     }),
   }),
+  z.object({
```
n1ru4l, line 357: great you also thought about audit logs!🥇

packages/services/api/src/modules/saved-filters/module.graphql.ts
```diff
@@ -0,0 +1,151 @@
+import { gql } from 'graphql-modules';
+
+export const typeDefs = gql`
+  enum SavedFilterType {
+    INSIGHTS
+  }
+
+  enum SavedFilterVisibility {
```
n1ru4l:
```suggestion
  enum SavedFilterVisibilityType {
```

to follow our "style guide" https://www.notion.so/theguildoss/Hive-Console-GraphQL-API-Design-Patterns-18ab6b71848a80678939c669699fe8e6?source=copy_link

packages/services/api/src/modules/saved-filters/module.graphql.ts
```diff
@@ -0,0 +1,151 @@
...
+      search: String = null
+    ): SavedFilterConnection!
+    viewerCanCreateSavedFilter: Boolean!
+  }
+
+  extend type Mutation {
+    createSavedFilter(
+      selector: TargetSelectorInput!
+      input: CreateSavedFilterInput!
+    ): CreateSavedFilterResult!
```
n1ru4l: For mutation types we usually only take on `input`, argument and for referencing the target we have the `TargetReferenceInput` input type.

packages/services/api/src/modules/saved-filters/module.graphql.ts
```diff
@@ -0,0 +1,151 @@
...
+    createdAt: DateTime!
+    updatedAt: DateTime!
+    createdBy: User
+    updatedBy: User
+    viewerCanUpdate: Boolean!
+    viewerCanDelete: Boolean!
+  }
+
+  type InsightsFilterConfiguration {
+    operationIds: [ID!]!
```
n1ru4l: nit: is it ids or names?

packages/services/api/src/modules/operations/module.graphql.ts
```diff
@@ -18,6 +32,11 @@ export default gql`
     Filter by only showing operations performed by specific clients.
     """
     clientNames: [String!] @tag(name: "public")
```
n1ru4l: Should we deprecated these?

packages/services/api/src/modules/saved-filters/resolvers/Target.ts
```diff
@@ -0,0 +1,38 @@
...
+        target,
+        args.type,
+        args.first,
+        args.after,
+        mapVisibility(args.visibility ?? null),
+        args.search ?? null,
+      ),
+  viewerCanCreateSavedFilter: (target, _args, { session }) => {
+    return session.canPerformAction({
+      action: 'project:modifySettings',
```
n1ru4l: Should we introduce a extra permission for this one?

packages/services/api/src/modules/saved-filters/providers/saved-filters.provider.ts
n1ru4l, line 1: a bit of logging can help observability in prod when pinning down issues!

packages/services/api/src/modules/saved-filters/module.graphql.ts
jonathanawesome: ahh...it's actually hashes.

packages/services/api/src/modules/saved-filters/resolvers/Target.ts
jonathanawesome: I'm not yet familiar enough with the overall permission setup to make this call. @jdolle do you have an opinion here?

packages/services/api/src/modules/operations/resolvers/Target.ts
```diff
@@ -54,6 +58,23 @@ export const Target: Pick<
     };
   },
   operationsStats: async (target, args, _ctx) => {
+    // Validate clientVersionFilters size limits to prevent DoS via large SQL IN clauses
+    const clientVersionFilters = args.filter?.clientVersionFilters;
+    if (clientVersionFilters) {
+      if (clientVersionFilters.length > MAX_CLIENT_VERSION_FILTERS) {
+        throw new GraphQLError(
+          `clientVersionFilters must contain at most ${MAX_CLIENT_VERSION_FILTERS} elements`,
```
jdolle: Ideally, this should be a typed error. But `operationsStats` is public and the current schema doesnt support it.

Should we deprecate and create a new field?

packages/services/api/src/modules/saved-filters/providers/saved-filters-storage.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    items = items.slice(0, limit);
+
+    return {
+      edges: items,
+      pageInfo: {
+        hasNextPage,
+        hasPreviousPage: cursor !== null,
+        get endCursor() {
+          return items[items.length - 1]?.cursor ?? '';
```
jdolle, line 177: I believe you want to get this before you slice the `items` array, else youll fetch the last item again.

packages/services/api/src/modules/saved-filters/resolvers/Target.ts
jdolle: We should.
It's a valid case to want a user to be able to create a saved filter, but not modify the project settings. I expect this to be the normal case actually.

packages/services/api/src/modules/saved-filters/module.graphql.ts
```diff
@@ -0,0 +1,157 @@
+import { gql } from 'graphql-modules';
+
+export const typeDefs = gql`
+  enum SavedFilterType {
+    INSIGHTS
```
jdolle: if we don't already have another use case in mind, consider removing this enum and field.

packages/services/api/src/modules/saved-filters/module.graphql.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  extend type Target {
+    savedFilter(id: ID!): SavedFilter
+    savedFilters(
+      type: SavedFilterType!
+      first: Int = 50
+      after: String = null
+      visibility: SavedFilterVisibilityType = null
+      search: String = null
```
jdolle: minor: `= null` isn't doing anything

packages/services/api/src/modules/saved-filters/providers/saved-filters-storage.ts
jonathanawesome: The cursor query uses strict <, so computing endCursor before slicing would skip an item. This matches the pattern across the codebase.

packages/services/api/src/modules/saved-filters/module.graphql.ts
jonathanawesome: Trying to plan ahead here. Explorer has filters that could adapt this UI, possibly Traces.

packages/services/api/src/modules/operations/resolvers/Target.ts
jonathanawesome: I wasn't comfortable making changes to the public API in an already complicated PR. Are you good with doing this in a follow up?

packages/migrations/src/actions/2026.02.19T00-00-00.saved-filter-permission.ts
```diff
@@ -0,0 +1,12 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2026.02.19T00-00-00.saved-filter-permission.ts',
+  run: ({ sql }) => sql`
+    UPDATE "organization_member_roles"
+    SET "permissions" = array_append("permissions", 'sharedSavedFilter:modify')
+    WHERE "permissions" @> ARRAY['project:modifySettings']
+      AND NOT ("permissions" @> ARRAY['sharedSavedFilter:modify'])
+      AND "locked" = false;
```
n1ru4l, line 10: This will cause a full table scan rewrite many database records rows, we need to verify that is not blocking the table for too long.

Generally I would say we prefer to handle such backwards-compatibility on the business logic layer. However, I am okay with this, if it will not impact the live service too long.

packages/services/api/src/modules/operations/resolvers/Target.ts
n1ru4l: NIT: we have GraphQL primitives to describe these fields, schema coordinates.

`'OperationStatsFilterInput.clientVersionFilters'` seems better than 'clientVersionFilters'

packages/services/api/src/modules/operations/resolvers/Target.ts
```diff
@@ -54,6 +58,23 @@ export const Target: Pick<
...
+        );
+      }
+      for (const filter of clientVersionFilters) {
+        if (filter.versions && filter.versions.length > MAX_VERSIONS_PER_FILTER) {
+          throw new GraphQLError(
+            `Each clientVersionFilter.versions must contain at most ${MAX_VERSIONS_PER_FILTER} elements`,
```
n1ru4l: NIT: we have GraphQL primitives to describe these fields, schema coordinates. `'ClientVersionFilterInput.versions'`

packages/services/api/src/modules/saved-filters/module.graphql.ts
n1ru4l: We cannot know whether all places where we use filters will have the same filter criterias (looking at you `InsightsFilterConfiguration`). So we could make this more abstract using interfaces and what not. Or as @jdolle suggested keep it simple and adapt later if needed - this part is in the private schema for now anyways.

packages/services/api/src/modules/saved-filters/resolvers/SavedFilter.ts
```diff
@@ -0,0 +1,105 @@
...
+  clientFilters: Array<{ name: string; versions: string[] | null }>;
+  dateRange: { from: string; to: string } | null;
+};
+
+function extractFilters(filterData: unknown): ParsedFilters {
+  const filters = filterData as {
+    operationHashes?: string[];
+    clientFilters?: Array<{ name: string; versions?: string[] | null }>;
+    dateRange?: { from: string; to: string } | null;
+  };
```
n1ru4l: Why the need to cast the thing? Why are the filters not typed properly?

packages/migrations/src/actions/2026.02.19T00-00-00.saved-filter-permission.ts
n1ru4l: we have around 12.000 roles records in our production database, so we could start with seeding some more locally and see runtime of the migration there. Then we could seed the same amount on staging and test there. Let me know if you need help.

packages/services/api/src/modules/saved-filters/module.graphql.ts
jonathanawesome: removed

n1ru4l:
Some additional feedback, nothing that should be part of this PR though.

I am spoiled from the linear "type to search" feature in the context menu and would love to see the same in ours. 😁

I could see that happen in several ways.

E.g. when I am in a context submenu menu with a search field, startign typing would automatically focus and populate the `Search...` input 🤔 
image

Or more like linear "AI Filters", where on the top level context menu, we already had a search input that can find operations, clients, filters etc. 
image

Issue #3133, opened 2023-10-23
Issue with newlines in schema checks using hive cli


platzhersh:
I'm currently trying to set up schema checks for one of our subgraphs using the hive CLI and our self-hosted graphql hive instance, where we are tracking our supergraph (federated).

Technically, it is working. Meaning, I can see the schema checks listed in the right target.

However, there seem to be some issues with newlines in the graphQL comments.

We are using pothos to define our schema and `printSubgraphSchema` to generate the schema we are using to check against GraphQL Hive.

If I generate the schema locally, it does look fine.

We have a lot of multiline block comments containing Markdown, similiar to this:

```graphql
"""
**some heading**

some description

**some subheading:** some text

|some table header|some table header|some table header|
|---|---|---|
|	-   **list object**|text|
|	-   **list object**|text|
|	-   **list object**|text|

"""
```

The description in the reported schema in graphql hive looks the same.

BUT: in the reported schema check, it looks like the multiline block comment is getting sent / received as a one-line block comment:

![Bildschirm­foto 2023-10-23 um 22 41 41](https://github.com/kamilkisiela/graphql-hive/assets/1445827/f0e5213d-cbf2-44a0-a11c-2bdbc2b3c874)

Is this a CLI issue? Or maybe Hive itself?

Using 

```
"@apollo/subgraph": "2.5.6",
"@graphql-hive/cli": "0.30.2",
"graphql": "16.8.1",
```

platzhersh:
Is there any progress, or anyone else having this problem?

przemyslawkowalczyk:
Do you have any plans to implement the functionality that would allow to add multi-line comments?

kroupacz:
Hello. We just migrated our federated GraphQL schema from Apollo Studio to Hive Console registry and we encountered the same issue where removing "new lines" from multiple line GraphQL comments/descriptions is breaking a large part of our documentation and we would like the `hive schema:publish` command to not remove "new lines" from multiple line GraphQL comments/descriptions.

kroupacz:
@jdolle Just to clarify. Our goal is to ensure that the formatting of multiple line GraphQL descriptions is preserved also in the supergraph and our clients can also see properly formatted descriptions when they introspect our GraphQL schema.

jdolle:
Thank you. I've addressed the issue with the service SDL being on a single line in the UI as well as multiline comments specifically losing their spacing when published using the CLI.

[...]

PR #7664, opened 2026-02-09
Prettify service sdl

jdolle:
### Background

https://linear.app/the-guild/issue/CONSOLE-414/issue-with-newlines-in-schema-checks-using-hive-cli

Related to https://github.com/graphql-hive/console/issues/3133

### Description

This change pretty prints the SDL before passing it to the schema check and publish functions. This causes us to store the pretty printed SDL instead of the raw format, which could be minified.

### DIFF
--- a/integration-tests/testkit/flow.ts
+++ b/integration-tests/testkit/flow.ts
@@ -1207,6 +1207,8 @@ export function fetchLatestSchema(token: string) {
               deletedService
             }
           }
+          sdl
+          supergraph
           schemas {
             nodes {
               ... on SingleSchema {
--- a/integration-tests/tests/cli/schema.spec.ts
+++ b/integration-tests/tests/cli/schema.spec.ts
@@ -1107,7 +1107,7 @@ test.concurrent('schema:publish ignores SDL formatting', async ({ expect }) => {
   const { createOrg } = await initSeed().createOwner();
   const { inviteAndJoinMember, createProject, organization } = await createOrg();
   await inviteAndJoinMember();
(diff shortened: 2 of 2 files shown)


n1ru4l (changes requested):
this change seems to break quite a few tests. When was this bug introduced? Or was this the behaviour all along? This PR is also missing a test that shows how it resolves the situation of makrdown tables being destroyed when the schema is published with the Hive CLI.

kamilkisiela:
If that doesn’t fix it, it could be tied to how graphql-tools/loaders handle schema loading - specifically how our CLI loads schemas and then forwards them to Hive’s GraphQL API.

jdolle:
> When was this bug introduced?

This appears to have always been the case. The CLI minifies the schema, which removes the newlines, and even though the API schema and supergraph are pretty-printed, the original SDL was never formatted in the resolver.

We could technically remove the minification from the CLI but I believe a more appropriate method is to format the schema and store that version in our system.

n1ru4l:
@jdolle The issue you linked mentions markdown tables specifically, it seems like this specifically requires a solution on the CLI?

jdolle:
Ty. I had missed the requirement for multiline descriptions and focused on the preview in the UI.

I've adjusted the minify logic to keep multiline descriptions as is

n1ru4l:
Is the comment formatting also retained in the supergraph/public schema SDL? Should we add a test for it specifically?

jdolle:
We dont run minify anywhere in the backend -- we just use the graphql schema print function to format the schemas. So the supergraph isnt impacted, but I can cover that in the test for regression.

PR #7669, opened 2026-02-10
feat(app-deployments): add createdAt, activatedAt, retiredAt fields and UI/UX improvements

adambenhassen:
### Background
Users lacked temporal context when reviewing app deployments, no visibility into when deployments were created, activated, or retired. This made it difficult to assess deployment age during schema check reviews with breaking changes.

### Description
Added `createdAt`, `activatedAt` and `retiredAt` fields to App Deployment pages. Additionally, some UI elements on deployment pages were broken after the recent UI color palette overhaul.

Closes CONSOLE-1858

Screenshot 2026-02-10 at 14 27 45
Screenshot 2026-02-10 at 14 58 07
Screenshot 2026-02-10 at 14 27 13
Screenshot 2026-02-10 at 14 33 00
Screenshot 2026-02-10 at 14 33 14

### Note
Server-side sorting is planned in a next PR after this merges.

### DIFF
--- a/packages/web/app/src/components/target/history/errors-and-changes.tsx
+++ b/packages/web/app/src/components/target/history/errors-and-changes.tsx
@@ -11,7 +11,6 @@ import {
   AccordionItem,
   AccordionTrigger,
 } from '@/components/ui/accordion';
-import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
 import { Heading } from '@/components/ui/heading';
 import { PulseIcon } from '@/components/ui/icon';


adambenhassen:
UI issues/inconsistencies encountered and fixed:
Screenshot 2026-02-10 at 14 49 11
Screenshot 2026-02-10 at 14 49 20
Screenshot 2026-02-10 at 14 49 39
Screenshot 2026-02-10 at 14 46 39
Screenshot 2026-02-10 at 11 17 57
Screenshot 2026-02-10 at 14 51 53
Screenshot 2026-02-10 at 14 51 58

packages/web/app/src/pages/target-checks-affected-deployments.tsx
```diff
@@ -277,6 +300,10 @@ function TargetChecksAffectedDeploymentsContent(props: {
                   <TableRow>
                     <TableHead className="w-[200px]">App Name</TableHead>
                     <TableHead>Version</TableHead>
+                    <TableHead>Status</TableHead>
```
n1ru4l: Nit: Isn't the status always `active` if the app deployment shows up in this list? Does this column provide any additional value?

packages/web/app/src/pages/target-checks-affected-deployments.tsx
```diff
@@ -277,6 +300,10 @@ function TargetChecksAffectedDeploymentsContent(props: {
...
+                    <TableHead>Created</TableHead>
+                    <TableHead>Activated</TableHead>
```
n1ru4l: NIT: Does it provide any benefit to the user to differentiate between created and activated? Could the activation date be enough information?

packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -26,10 +34,10 @@ export default gql`
...
   }
 
-  enum AppDeploymentStatus {
-    pending
-    active
-    retired
+  enum AppDeploymentStatus @tag(name: "public") {
+    pending @tag(name: "public")
+    active @tag(name: "public")
+    retired @tag(name: "public")
```
n1ru4l: NIT: as per https://www.notion.so/theguildoss/Hive-Console-GraphQL-API-Design-Patterns-18ab6b71848a80678939c669699fe8e6?source=copy_link#18ab6b71848a80d5a1b1d16abd881d32 enums should follow a specific pattern before we add them to the public API.

So my question here is: 
- Can we avoid adding this type to the public API just yet or did someone specifically ask for it
- If we rename it and would it break any existing clients (mainly CLI)

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -1097,6 +1097,46 @@ export class AppDeployments {
...
+        createdAt: string;
+        activatedAt: string | null;
+        retiredAt: string | null;
+      }>(
+        sql`
+          SELECT
+            "id",
+            to_json("created_at") AS "createdAt",
+            to_json("activated_at") AS "activatedAt",
+            to_json("retired_at") AS "retiredAt"
```
jdolle, line 1114: This is very similar to the recent issue where we needed to populate data in clickhouse (or run in batches) to avoid additional lookups that could be problematic.

Alternatively, we could add a limit to the first query to avoid ever having to worry about exceeding the payload threshold. I think this would be a reasonable solution if we ordered the conflicting active deployments by either their creation time or last used time. I dont foresee teams needing to know that 100+ app deployments failed the check.

packages/services/api/src/modules/schema/providers/registry-checks.ts
```diff
@@ -46,6 +46,9 @@ export type AffectedAppDeployment = {
     id: string;
     name: string;
     version: string;
+    createdAt: string | null;
```
jdolle, line 49: created at should never be null

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
adambenhassen: No action needed here.

Postgres handles large `ANY(array)` differently, as a single parametrized query, not HTTP form fields. There's no payload size limit like Clickhouse Cloud has.

Additionally, this was tested with 6000+ deployments on dev cluster and works fine.

packages/services/api/src/modules/schema/providers/registry-checks.ts
adambenhassen: `createdAt` inherits `createdAt: string | null | undefined` in `SchemaChange.ts:77`, this is out of scope of this PR. CreatedAt is never null regardless. will not change.

PR #7672, opened 2026-02-10
fix: broken "view org as admin" functionality

n1ru4l:
### Background

Tried to view an organization to debug some issue and noticed this was no longer possible.

### Description

The ability to do that was broken in https://github.com/graphql-hive/console/pull/7639. This PR restores the functionality by making the `Organization.me` field nullable and returning `null` in case the viewer is not part of the organization.

I tested this change locally, but it might be worth adding some e2e test, to avoid an regression.

### Checklist

- [x] Testing

### DIFF
--- a/packages/services/api/src/modules/organization/resolvers/Organization.ts
+++ b/packages/services/api/src/modules/organization/resolvers/Organization.ts
@@ -54,10 +54,6 @@ export const Organization: Pick<
       userId: me.id,
     });
 
-    if (!member) {
-      return null;
-    }
-
     return member;
   },
   members: (organization, args, { injector }) => {


packages/web/app/src/components/organization/settings/personal-access-tokens/create-personal-access-token-sheet-content.tsx
```diff
@@ -250,7 +250,8 @@ export function CreatePersonalAccessTokenSheetContent(
                               <Form.FormControl>
                                 <PermissionSelector
                                   permissionGroups={
-                                    organization.me.availablePersonalAccessTokenPermissionGroups
+                                    organization.me?.availablePersonalAccessTokenPermissionGroups ??
+                                    []
```
n1ru4l, line 254: For types sake, user can actually not get here ever in the first place if they are not member of the org.

PR #7673, opened 2026-02-10
fix(server): handle OIDC token exchange errors gracefully & fix debug log modal

adambenhassen:
### Background

OIDC token exchange errors (i.e. expired Azure AD client secret) cause unhandled 500 responses instead of actionable user-facing messages.

### Description

Catches OIDC sign-in errors and returns a `GENERAL_ERROR` with a user-safe message classified by OAuth 2.0 error code (RFC 6749 5.2), network failure type, or userinfo endpoint error. Full errors are still logged server-side.

This PR also fixes the broken debug modal not displaying the log area.

Closes CONSOLE-1794

### DIFF
--- a/packages/services/server/src/supertokens.ts
+++ b/packages/services/server/src/supertokens.ts
@@ -369,7 +369,7 @@ const getEnsureUserOverrides = (
         }
         return null;
       }
-      
+
       try {
         const response = await originalImplementation.thirdPartySignInUpPOST(input);
 


packages/services/server/src/supertokens/oidc-provider.ts
```diff
@@ -303,6 +303,66 @@ async function getOIDCConfigFromInput(
...
 
+/**
+ * Classify an OIDC sign-in error into a user-safe description.
+ * Avoids leaking sensitive details (app IDs, trace IDs, internal URLs)
+ * while still pointing administrators toward the likely cause.
+ */
+export function describeOIDCSignInError(error: unknown): string {
+  const message = error instanceof Error ? error.message : String(error);
+
+  if (message.includes('invalid_client')) {
```
jdolle, line 314: generally I try to avoid matching on message text. Are there any other fields we can match on that have more guarantees than the message?

packages/services/server/src/supertokens/oidc-provider.ts
```diff
@@ -303,6 +303,66 @@ async function getOIDCConfigFromInput(
...
+  if (message.includes('unsupported_grant_type')) {
+    return 'Your OIDC provider does not support the authorization code grant type. Please verify the provider supports the OAuth 2.0 authorization code flow.';
+  }
+
+  if (message.includes('invalid_scope')) {
+    return 'Your OIDC provider rejected the requested scopes. Please review the additional scopes configured in your OIDC integration settings.';
+  }
+
+  if (message.includes('Could not find OIDC integration')) {
+    return 'The OIDC integration could not be found. It may have been removed or misconfigured. Please contact your organization administrator.';
```
jdolle, line 339: why contact org admin instead of `review your OIDC integration settings` here?

n1ru4l: On a note we should also consider that this is end user facing, people seeing this might not even have access to OIDC settings. So I think it might be okay to say, something like "please contact your organization admin about the OIDC configuration"?

packages/services/server/src/supertokens/oidc-provider.ts
adambenhassen: Valid concern in general, but I've investigated the supertokens source and `exchangeAuthCodeForOAuthTokens` throws plain Error objects with this format `{"error":"invalid_client","error_description":"..."}`

There are no structured fields (error.code, typed subclasses, etc..), everything is embedded in the message string. the OAuth2 error codes (invalid_client, invalid_grant, etc) only exist inside the stringified response body within the message.

adambenhassen:
Also fix the debug modal

Screenshot 2026-02-12 at 16 14 21

Screenshot 2026-02-12 at 16 14 28

PR #7676, opened 2026-02-10
Correct the public url for http callback in documentation

enisdenjo:
HTTP callback config for Hive Gateway public_url should not also have callback path.


enisdenjo:
Closed because the GW is in the wrong, documentation is ok.

See https://github.com/graphql-hive/gateway/pull/1944.

PR #7677, opened 2026-02-11
Increase service keepAliveTimeout from 72s to 905s

jdolle:
### Background

One of our customers was encountering upstream disconnects. This should help us reduce the likelihood that these 503s happen. Note that these errors do get retried by the hive-client, but it's much more efficient to maintain connections slightly longer. This also matches the idle timeout for a number of AWS services such as NAT Gateway.

### Description

Increase keepAliveTimeout for services and contour (envoy) proxy. And I added an inline comment because this configuration is often misunderstood or overlooked.


n1ru4l:
We are using Cloudflare, which [recommends a keep alive of 900 seconds](https://developers.cloudflare.com/load-balancing/load-balancers/#http-keep-alive-persistent-http-connection). Is it sufficient to only configure this on the Node.js services - or ~~does this also require adjusting the envoy proxy configuration~~ (did have a look before coffee)?

Should we increase the keep alive of envoy for all services behind cloudflare?

jdolle:
Great find. I didn't realize all of our requests were going through Cloudflare.

Knowing that now, I believe we should increase the keep alive of all our services to 905_000 (905 seconds).

packages/services/service-common/src/fastify.ts
```diff
@@ -71,6 +71,16 @@ export async function createServer(options: {
     },
     requestIdHeader: 'x-request-id',
     trustProxy: true,
+    // If a connection is idle for 905 seconds or more, the connection times out.
+    // The default for fastify is 72_000, but this is meant for more dynamic clients.
+    // Requests to Hive's services are proxied through Cloudflare, which has a 900s
```
dotansimha, line 77: While CloudFlare has 900s configured, it's still only 1 part in the chain: `cf -> cluster (envoy) -> usage-service`. 

Envoy has a limit of 1h by default, so adjusting this to 905s, should be ok, i think? (900s -> 1h -> 905s) 

Also, this change assumes all users go through CF, which might not be correct for self-hosters so maybe it should be adjustable?

jdolle: 1hr seems incredibly long.
In reviewing envoy's docs, it looks like we can adjust this but they combined the idle and keepalive timeouts into a single [`idle_timeout` option](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-field-config-core-v3-httpprotocoloptions-idle-timeout).

How do we feel about lowering the envoy idle_timeout to 905s and setting our services to 910s? Also I will make this adjustable.

jdolle: I'm still uncovering some details. It looks like we actually use Contour for our proxy and it has a 60s idle timeout
https://projectcontour.io/docs/main/configuration/#timeout-configuration

dotansimha: Contour is just the K8s integration layer for Envoy, but it's good to know that they configure some things differently than enovy's defaults.

dotansimha (approved):
LGTM, this should align the configs across the pipeline. 
Can we please test it on `dev` / `staging` cluster before merging? :)

jdolle:
Tested on dev/staging for around 12hrs and saw no 503s
However, I didnt see any 503s prior to the deploy either. These environments receive little traffic so we're unlikely to encounter this race condition on those environments.

PR #7689, opened 2026-02-12
feat: use Hive Router for public GraphQL API, along side Hive Gateway

dotansimha:
Hive Router now supports everything we do in Hive Gateway today. Including:

- [x] Use public API using contract from Hive CDN 
- [x] Export Hive Traces 
- [x] Export internal OTEL trace
- [ ] Security plugins / operation complexity? 

In addition, this PR extends the deployment code with a new option for weighted deployments that can split the traffic across different pods.

- [x] Add deployment for Hive Router 0.0.39
- [x] Add Hive Router config file mount
- [x] Extend deployment code with the ability to run multiple services and split traffic
- [x] Configure public api to use 50/50 traffic between Hive Router and Hive Gateway

Testing:

- [x] test locally
- [ ] test on dev/staging

### DIFF
--- a/deployment/utils/reverse-proxy.ts
+++ b/deployment/utils/reverse-proxy.ts
@@ -158,8 +158,8 @@ export class Proxy {
             ],
             services: isWeightBasedUpstream(route.service)
               ? route.service.map(serviceDef => ({
-                  name: serviceDef.service.metadata.name,
-                  port: serviceDef.service.spec.ports[0].port,
+                  name: serviceDef.upstream.metadata.name,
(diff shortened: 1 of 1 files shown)


dotansimha:
@n1ru4l thoughts on what security plugins should we enable? https://the-guild.dev/graphql/hive/docs/router/security/operation-complexity

n1ru4l:
@dotansimha for now I would just mirror what we have for gateway https://github.com/graphql-hive/console/blob/f6daaf12c5dca1e7c5e6b222a55325f646f9bab2/deployment/config/public-graphql-api/gateway.config.ts#L139-L144

PR #7690, opened 2026-02-12
feat: adding target option to schema reporting

nopor:
### Background
When using `useHive` for publishing the schema, it implies the usage of the old tokens.

### Description
This change adds the option to add the target (slug or id) to the publish/reporting schema functionality.
It affects the core client, e.g. the `useHive` function that is exposed through `@graphql-hive/yoga`.

### Checklist

- [X] Input validation


nopor:
As discussed separately: due to the function being deprecated, this won't be added/needed.

PR #7691, opened 2026-02-12
Speed up app deployment 1000+ test

jdolle:
### Background

I wrote this test to provide a regression on a recent change to the pagination of app deployments, but the test regularly times out in CI.

### Description

I've adjusted the app deployment creation to perform a lower level DB call to insert all records at once to avoid the slow insert loops.

Alternatively, we can delete it entirely.

### DIFF
--- a/.changeset/cool-ties-tell.md
+++ b/.changeset/cool-ties-tell.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/cli': patch
+---
+
+Do not minify multiline descriptions in SDL on check/push
--- a/.changeset/curly-dodos-retire.md
+++ b/.changeset/curly-dodos-retire.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+Fix service SDL being printed on a single line in check and publish views
--- a/.prettierignore
+++ b/.prettierignore
@@ -18,6 +18,7 @@ __snapshots__/
 
 # test fixtures
 integration-tests/fixtures/init-invalid-schema.graphql
+integration-tests/fixtures/whitespace-oddity.graphql
 /target
 tmp
 
--- a/integration-tests/fixtures/whitespace-oddity.graphql
+++ b/integration-tests/fixtures/whitespace-oddity.graphql
@@ -1,0 +1,13 @@
+"""
+Multi line comment:
+1. Foo
+2. Bar
+3. Should stay in a list format
+"""
+type Query { status: Status }
+
+# Another comment here
+enum Status {
+  ACTIVE INACTIVE
+  PENDING
+}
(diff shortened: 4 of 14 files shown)


integration-tests/testkit/clickhouse.ts
```diff
@@ -30,3 +30,23 @@ export async function clickHouseQuery<T>(query: string): Promise<{
 
   return response.json();
 }
+
+export async function clickHouseInsert<T>(query: string): Promise<void> {
```
jdolle, line 34: It took me way longer than I'd like to admit to figure out that the reason why my insert was failing was because clickhouse inserts dont return json...

The only diff between this and the clickHouseQuery function is that this doesnt call .json on the response

integration-tests/tests/api/app-deployments.spec.ts
```diff
@@ -2031,45 +2033,56 @@ test('activeAppDeployments filters by neverUsedAndCreatedBefore', async () => {
 });
 
 test('activeAppDeployments works for > 1000 records with a date filter (neverUsedAndCreatedBefore) set', async () => {
+  const seed = await initSeed();
+  const conn = await seed.createDbConnection();
```
adambenhassen: Missing `using` to properly dispose of db connection when the test finishes
```suggestion
  await using conn = await seed.createDbConnection();
```

integration-tests/tests/api/app-deployments.spec.ts
```diff
@@ -2031,45 +2033,56 @@ test('activeAppDeployments filters by neverUsedAndCreatedBefore', async () => {
...
   const { createOrg, ownerToken } = await initSeed().createOwner();
```
adambenhassen: Why not reuse existing seed instance?

```suggestion
  const { createOrg, ownerToken } = await seed.createOwner();
```

integration-tests/tests/api/app-deployments.spec.ts
adambenhassen, line 2112: This loop is not paginating, it's running the same query, fetching the same first page 60 times (there's no `after` cursor).

Either add the cursor or run once.

I suggest running this once instead, since this is the bottleneck of the test, it reduced the run time from 20 seconds to 3 seconds:

```suggestion
  const deploymentsResult = await execute({
    document: GetActiveAppDeployments,
    variables: {
      targetSelector: {
        organizationSlug: organization.slug,
        projectSlug: project.slug,
        targetSlug: target.slug,
      },
      first: 20,
      filter: {
        neverUsedAndCreatedBefore: tomorrow.toISOString(),
      },
    },
    authToken: ownerToken,
  }).then(res => res.expectNoGraphQLErrors());
  expect(deploymentsResult.target?.activeAppDeployments.edges).toHaveLength(20);
```

jdolle: Good catch. Added the cursor to the request.

Issue #7110, opened 2025-10-14
Replace supertokens with our own authentication


n1ru4l:
We want to remove Supertokens from Hive Console.

The main reasons:
- hard/impossible to upgrade
  - https://github.com/graphql-hive/console/pull/5896
  - https://github.com/graphql-hive/console/pull/5894
- no account linking in OSS version
- we maintain some code for supporting multi-tenant OIDC login, which feels like quite a mess

In Feburary 2025 @kamilkisiela evaluated and started working on replacing supertokens with better-auth, but due to other priorities he abandoned the project.

Basically the replacement needs to support everything we have today:
- Sign in/up via Email (with email verification and forgot password etc.)
- Sign up via GitHub
- Sign up via organization OIDC provider
- The transition should be seamless so users do not have to sign-in again

It makes sense to re-use the existing supertoken pg tables and at the same cleanup the tables (there are a lot of tables created that are actually unused). In addition to that we also need to create a db migration that seeds the tables, for users that start using hive console self-hosted after we remove the super tokens service.

In the next step we could then introduce account linking.

______

Resources/ Prior work:
- https://github.com/better-auth/better-auth
- https://github.com/graphql-hive/console/tree/kamil-better-auth

PR #7699, opened 2026-02-16
supertokens at home

n1ru4l:
Closes https://github.com/graphql-hive/console/issues/7110

### Background

- supertokens-core is one more service we need to maintain
- adding additional functionality around the supertokens-node SDKs feels hacky and results in hard to maintain code
- upgrading supertokens and it's SDKs is always breaks things

### Description

This adds a drop in solution that makes the supertokens-core service obsolete, while providing the same external facing functionality.

- Implements the relevant [FDI](https://app.swaggerhub.com/apis/supertokens/FDI/1.18.0) (Frontend ---> Server) endpoints on our `server` service
  - instead of using the [CDI](https://app.swaggerhub.com/apis/supertokens/CDI/4.0.0) (Server ---> Supertokens Core Service) endpoints to communicate with supertokens-core, the relevant logic for managing access tokens, refresh tokens, sessions and user login methods was ported to the `server` service
- Re-using existing database tables
- Supports rollout without re-login for users
- Supports easy rollback via environment variables (until we feel confident to fully remove supertokens-core)
- DOES NOT remove supertokens-react and co; this can be done in a follow up task

[A video runthrough on our Slack](https://guild-oss.slack.com/archives/C01E8ATBQGN/p1771591785582129?thread_ts=1771346501.869319&cid=C01E8ATBQGN)

##### Access Token Key Rotation

The supertoken access tokens are meant to be stateless to allow authentication without having the supertokens core. Because of that supertokens core comes with built-in RS256 JWT key rotation to mitigate risk when one JWT key leaks.

Since we are currenty only using the supertokens access keys to verify users on the web app and double verify the active session via our database (aka not stateless), I omitted this part as it would introduce some more complexity where not needed - Instead we have a single static RSA256 key for access token signing.

If we introduce stateless auth, we can re-consider adding this functionality.

##### Access Token Signing Key Incompatibility

Since supertokens core does not immediately write RS256 keys to the database, we cannot do a "session refresh" free rollout. That means we need to generate a new RSA256 key pair and pass it as a environment variable to the server instance.

For existing users that have already a `sRefreshToken` and `sAccessToken` cookie set this has the following implications:
- The next request with the `sAccessToken` cookie will yield an error with `NEEDS_REFRESH`
- The session refresh flow is started, calling `/auth-api/session/refresh`, generating a new token and then pending requests are retried

No re-login needed for the user.

#### How to try this at home

Clone this branch, setup the local running services.

Add the following to your `packages/services/server/.env`

```
SUPERTOKENS_AT_HOME=1
SUPERTOKENS_REFRESH_TOKEN_KEY=1000:15e5968d52a9a48921c1c63d88145441a8099b4a44248809a5e1e733411b3eeb80d87a6e10d3390468c222f6a91fef3427f8afc8b91ea1820ab10c7dfd54a268:39f72164821e08edd6ace99f3bd4e387f45fa4221fe3cd80ecfee614850bc5d647ac2fddc14462a00647fff78c22e8d01bc306a91294f5b889a90ba891bf0aa0
SUPERTOKENS_ACCESS_TOKEN_KEY=s-aaa5da0d-8678-46ef-b56d-9cd19e1cdb5b|MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAjdpx/LNIAPKDCNgszZIzz2tepThK9/3mCWTrOjAJY8KI8YgMo5Gf+IwYvJ2VcpKYa36UJ70bD283P2O/yE/IjtXI6S9cJyd5LrYs+QDENc8au63iDy2iAiFpR1kO7cWQPDKK3OrD+hc5ZEHA3LN82Kb4ZnEA6tAulPfULVDU+RJfSWZOCE+LnkSZ8obvJjiMeknhNSSJko6V3WVuL5ToYfRIiOnueoTywB+3O3Mtp6lBj1j2rpQfO/qvLdRYLpDmLaoaScAyymWfeBp0hpwxd5Jm4vexyHgit2sK0S+tFl0pmh37iVGsRqPJEPISpEwQBcHOhKRj0uW+t/feK6U0WQIDAQAB|MIIEvgIBADANBgkqhkiG9w0BAQEFAASCBKgwggSkAgEAAoIBAQCN2nH8s0gA8oMI2CzNkjPPa16lOEr3/eYJZOs6MAljwojxiAyjkZ/4jBi8nZVykphrfpQnvRsPbzc/Y7/IT8iO1cjpL1wnJ3kutiz5AMQ1zxq7reIPLaICIWlHWQ7txZA8Morc6sP6FzlkQcDcs3zYpvhmcQDq0C6U99QtUNT5El9JZk4IT4ueRJnyhu8mOIx6SeE1JImSjpXdZW4vlOhh9EiI6e56hPLAH7c7cy2nqUGPWPaulB87+q8t1FgukOYtqhpJwDLKZZ94GnSGnDF3kmbi97HIeCK3awrRL60WXSmaHfuJUaxGo8kQ8hKkTBAFwc6EpGPS5b63994rpTRZAgMBAAECggEBAIl96/IFS4svg/Z0oah3RySKa2g1EeUhAXClkqIJoXBCRD3nomiAY8+i6u8Wxp4QnQ/D1pJV5v6ky6XzZxYezsQzTtNGBkolJn4yMZEAPy3wmXbD6VLQ5jCudb6kAaZRUaYnTxUlr+Kd1BDq8qZ4ik/sNuQEL+Fo+12EgPGTYXov7lWwWjNbzuzMQMm7b1BDU7D/8s/lGg4wimJffVSd4C++buN4Jxm1n1hWWREl7jkJC0sp2J50cpt9IhIIhi8DOnGAcJ4aTtABEJdZyXlO0QllN/D5FEbZBC0Jkbl3lmaIo1WVEYdDpcbSLZxGeYD0CkH4CF/BzUpeHq7FU0HkqOkCgYEA5tgLRFC4CS/FtR3JQ1YifHN4J2bI3BEyje6eiI/n9jmm5zUN/WjCQ6ui2fPzbKAC3yD60lcCgrP7YAn4KFmOv9RS7ApamUH+AX5UBfHlhvwzi4U9eenu7UHH8XrxEHlAwUC9mQbaqzoR/A7jEg8qqincMDUCkk1kjP4nNgQSBFcCgYEAnU/KSQf79m57nXMv+ge9QWkAggnHO7baRox5qlMM6l1gTZB5yELaLNXeik9D2mhVwcpezcQo2Uf+B4MviVeqpoTzYwgdxYg+ebYPUzhd3Ya8ANRDwKCB7SSoRULEDpWebV6ngOc+ruv9ii3ZbVEi7ribtHo6w7rVVJc2bMEKns8CgYB9ssp/yoxLxE2dz7hWCEMDDUUx/1AENQEYNATzS5j9hGsTntodULvnaUBl+eZlEcQ+h5DMlEBzt1l79DHClvGaFx2IFiM7LKoJWiaajhtzo0TWBhlxlyZY3ubm4RD+7WeLU5tqBkdv0VEVtW2D2epbeivBvDvIOog0Ffh3+0NsRQKBgGPA8w84hugPy0dega/VNIfD49SSCsqs+uD9tzDwlSIQsD6/PNpmuh7wR7wA45Ad1TOb9l4Y46ZU5psw7vXyp34MlKHZxbc63BMmBbXJ6ovNIm6MK6J8pacRNbslyVlOOzYzbZhqCu+1KgNza4rMhpBGdEYPtC/ly91mPdbc2rU1AoGBALl6eZqBMahE0S19X5SO/xykGe4ALf74UWCsfKrVEO4Zd3IcELnir0uEPABWvd5C/EAaGoi/a2xgdwuKG32GMpinjoXJywzsquQC6N8CcFIzDiQXaL4j4lFztjgowqNs/YwpOGbm1Dyr3Av072jDPajQGP/xX4fFxBZFnyk1vnXT
```

That is it. You can kill the `supertokens` docker container to verify that everything is indeed working without the service.

There is also some more instructions available over here for setting up supertokens at home without requiring users to re-login: https://github.com/graphql-hive/console/blob/4896531c415854aee0a11680cf891d0efa4a6534/.changeset/light-lions-wear.md

### Resources

- https://app.swaggerhub.com/apis/supertokens/FDI/1.18.0
- https://app.swaggerhub.com/apis/supertokens/CDI/4.0.0
- https://github.com/supertokens/supertokens-core/blob/4224caad84c74f8a208da6c288a50859f4277ed1/src/main/java/io/supertokens/session/refreshToken/RefreshToken.java
- https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/session/Session.java
- https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/session/accessToken/AccessToken.java
- https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/utils/Utils.java
- https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/webserver/api/session/RefreshSessionAPI.java
- https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps

### Checklist

- [x] Input validation
- [x] Output encoding
- [x] Session management
- [x] Access control
- [x] Cryptographic practices
- [x] System configuration
- [ ] Database security
- [x] Testing

### DIFF
--- a/packages/services/api/src/modules/auth/lib/supertokens-at-home/crypto.ts
+++ b/packages/services/api/src/modules/auth/lib/supertokens-at-home/crypto.ts
@@ -6,6 +6,7 @@
  */
 
 import * as c from 'node:crypto';
+import { promisify } from 'node:util';
 import bcrypt from 'bcryptjs';
 import jwt from 'jsonwebtoken';
 import z from 'zod';
@@ -26,7 +27,9 @@ export async function comparePassword(password: string, hash: string) {
   return await bcrypt.compare(password, hash);
 }
 
-function decryptRefreshToken(encodedData: string, masterKey: string) {
+const pbkdf2Async = promisify(c.pbkdf2);
+
+async function decryptRefreshToken(encodedData: string, masterKey: string) {
   // 1. Decode the incoming string (URL -> Base64 -> Buffer).
   const urlDecodedData = decodeURIComponent(encodedData);
   const buffer = Buffer.from(urlDecodedData, 'base64');
@@ -45,7 +48,7 @@ function decryptRefreshToken(encodedData: string, masterKey: string) {
   const keylen = 32; // 32 bytes = 256 bits
   const digest = 'sha512'; // NOTE: This is a guess. See explanation below.
 
-  const secretKey = c.pbkdf2Sync(masterKey, iv, iterations, keylen, digest);
+  const secretKey = await pbkdf2Async(masterKey, iv, iterations, keylen, digest);
 
   // 4. Separate the encrypted data from the authentication tag.
   const authTagLength = 16; // 128 bits
@@ -126,7 +129,7 @@ const RefreshTokenPayloadModel = z.object({
 
 type RefreshTokenPayloadType = z.TypeOf<typeof RefreshTokenPayloadModel>;
 
-export function parseRefreshToken(refreshToken: string, masterKey: string) {
+export async function parseRefreshToken(refreshToken: string, masterKey: string) {
   const [payload, nonce, version] = refreshToken.split('.');
 
   if (version !== 'V2') {
@@ -139,7 +142,7 @@ export function parseRefreshToken(refreshToken: string, masterKey: string) {
   let refreshTokenPayload: RefreshTokenPayloadType;
   try {
     refreshTokenPayload = RefreshTokenPayloadModel.parse(
-      JSON.parse(decryptRefreshToken(payload, masterKey)),
+      JSON.parse(await decryptRefreshToken(payload, masterKey)),
     );
   } catch (err) {
     return {
--- a/packages/services/server/src/supertokens-at-home.ts
+++ b/packages/services/server/src/supertokens-at-home.ts
@@ -466,7 +466,7 @@ export async function registerSupertokensAtHome(
         return rep.status(404).send();
       }
 
-      const parseResult = parseRefreshToken(refreshToken, secrets.refreshTokenKey);
+      const parseResult = await parseRefreshToken(refreshToken, secrets.refreshTokenKey);
 
       if (parseResult.type === 'error') {
         req.log.debug('Wrong refresh token version provided. err=%s', parseResult.code);


integration-tests/testkit/auth.ts
```diff
@@ -155,31 +213,51 @@ const createSession = async (
...
-): Promise<{ access_token: string; refresh_token: string }>;
-export function authenticate(
-  email: string,
-  oidcIntegrationId?: string,
-): Promise<{ access_token: string; refresh_token: string }>;
-export function authenticate(
-  email: string | string,
   oidcIntegrationId?: string,
 ): Promise<{ access_token: string; refresh_token: string }> {
+  if (process.env.SUPERTOKENS_AT_HOME === '1') {
```
n1ru4l, line 234: I wanted to retain the ability to use the supertokens-core and our implementation for seeding users.

integration-tests/.env
```diff
@@ -21,3 +21,6 @@ CLICKHOUSE_ASYNC_INSERT_MAX_DATA_SIZE=1000
...
...RySKa2g1EeUhAXClkqIJoXBCRD3nomiAY8+i6u8Wxp4QnQ/D1pJV5v6ky6XzZxYezsQzTtNGBkolJn4yMZEAPy3wmXbD6VLQ5jCudb6kAaZRUaYnTxUlr+Kd1BDq8qZ4ik/sNuQEL+Fo+12EgPGTYXov7lWwWjNbzuzMQMm7b1BDU7D/8s/lGg4wimJffVSd4C++buN4Jxm1n1hWWREl7jkJC0sp2J50cpt9IhIIhi8DOnGAcJ4aTtABEJdZyXlO0QllN/D5FEbZBC0Jkbl3lmaIo1WVEYdDpcbSLZxGeYD0CkH4CF/BzUpeHq7FU0HkqOkCgYEA5tgLRFC4CS/FtR3JQ1YifHN4J2bI3BEyje6eiI/n9jmm5zUN/WjCQ6ui2fPzbKAC3yD60lcCgrP7YAn4KFmOv9RS7ApamUH+AX5UBfHlhvwzi4U9eenu7UHH8XrxEHlAwUC9mQbaqzoR/A7jEg8qqincMDUCkk1kjP4nNgQSBFcCgYEAnU/KSQf79m57nXMv+ge9QWkAggnHO7baRox5qlMM6l1gTZB5yELaLNXeik9D2mhVwcpezcQo2Uf+B4MviVeqpoTzYwgdxYg+ebYPUzhd3Ya8ANRDwKCB7SSoRULEDpWebV6ngOc+ruv9ii3ZbVEi7ribtHo6w7rVVJc2bMEKns8CgYB9ssp/yoxLxE2dz7hWCEMDDUUx/1AENQEYNATzS5j9hGsTntodULvnaUBl+eZlEcQ+h5DMlEBzt1l79DHClvGaFx2IFiM7LKoJWiaajhtzo0TWBhlxlyZY3ubm4RD+7WeLU5tqBkdv0VEVtW2D2epbeivBvDvIOog0Ffh3+0NsRQKBgGPA8w84hugPy0dega/VNIfD49SSCsqs+uD9tzDwlSIQsD6/PNpmuh7wR7wA45Ad1TOb9l4Y46ZU5psw7vXyp34MlKHZxbc63BMmBbXJ6ovNIm6MK6J8pacRNbslyVlOOzYzbZhqCu+1KgNza4rMhpBGdEYPtC/ly91mPdbc2rU1AoGBALl6eZqBMahE0S19X5SO/xykGe4ALf74UWCsfKrVEO4Zd3IcELnir0uEPABWvd5C/EAaGoi/a2xgdwuKG32GMpinjoXJywzsquQC6N8CcFIzDiQXaL4j4lFztjgowqNs/YwpOGbm1Dyr3Av072jDPajQGP/xX4fFxBZFnyk1vnXT
```
n1ru4l, line 26: These are my localhost secrets, don't worry.

packages/migrations/src/actions/2026.02.18T00-00-00.ensure-supertokens-tables.ts
n1ru4l, line 1: So this attempts to create all the supertokens tables that our implementation needs.

packages/services/api/src/modules/auth/lib/supertokens-at-home/crypto.ts
n1ru4l, line 1: this contains all the encryption stuff needed, harvested from various sources
- https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/utils/Utils.java
- https://github.com/supertokens/supertokens-core/blob/master/src/main/java/io/supertokens/session/accessToken/AccessToken.java
- https://github.com/supertokens/supertokens-core/blob/4224caad84c74f8a208da6c288a50859f4277ed1/src/main/java/io/supertokens/session/refreshToken/RefreshToken.java

packages/services/api/src/modules/auth/lib/supertokens-strategy.ts
```diff
@@ -240,13 +338,37 @@ export class SuperTokensUserAuthNStrategy extends AuthNStrategy<SuperTokensCooki
...
+    this.logger.debug('Attempt verifying SuperTokens session');
+
+    if (args.req.headers['ignore-session']) {
+      this.logger.debug('Ignoring session due to header');
+      return null;
+    }
+
+    const sessionData = this.accessTokenKey
+      ? await this._verifySuperTokensAtHomeSession(args, this.accessTokenKey)
+      : await this._verifySuperTokensCoreSession(args);
```
n1ru4l, line 357: Like everything I to keep it in a way so we could easily rollback to supertokens-core if needed

packages/services/api/src/modules/auth/lib/supertokens-strategy.ts
```diff
@@ -219,16 +226,107 @@ export class SuperTokensUserAuthNStrategy extends AuthNStrategy<SuperTokensCooki
...
+
+    args.req.log.debug('attempt parsing access token from cookie');
+
+    const cookie = parseCookie(args.req.headers.cookie ?? '');
+    let rawAccessToken: string | undefined = cookie['sAccessToken'];
+
+    if (!rawAccessToken) {
+      args.req.log.debug('attempt parsing access token authorization header');
+      rawAccessToken = args.req.headers.authorization?.replace('Bearer ', '')?.trim();
+    }
```
n1ru4l, line 265: I actually noticed that we use the authorization header instead of cookies in our integration tests.

packages/services/api/src/modules/auth/providers/oauth-cache.ts
n1ru4l, line 1: purpose of this is to store on our end the state of an authorization attempt while the user is redirected to the authorization endpoint and then comes back to us. Redis with some TTL seemed like a good solution here.

packages/services/api/src/modules/auth/providers/supertokens-store.ts
n1ru4l, line 1: So this bad boy contains all the CRUD logic needed for interacting with the supertokens tables.

packages/services/api/src/modules/shared/providers/redis-rate-limiter.ts
n1ru4l, line 1: I took the opportunity to reimplement the rate limiting we already had for supertokens in a bit of a nicer way for fastify 😇

packages/services/server/src/supertokens-at-home/shared.ts
n1ru4l, line 1: Just some small utilities that are used in a few places

packages/services/server/src/graphql-handler.ts
```diff
@@ -36,10 +36,6 @@ export interface GraphQLHandlerOptions {
   registry: Registry;
   signature: string;
   tracing?: TracingInstance;
-  supertokens: {
-    connectionUri: string;
-    apiKey: string;
-  };
```
n1ru4l, line 42: this wasnt used before and is some legacy leftover

packages/services/server/src/index.ts
n1ru4l, line 1: here we conditionally glue in euther our implementation or the supertokens-node implementation - based on the environment variables

packages/services/server/src/supertokens-at-home.ts
n1ru4l, line 1: This contains all the route handlers and is our implementation of the FDI endpoints. I avoided abstracting things so it is easier to understand the steps involved. I found it quite hard jumping around the supertokens code-base understanding the different steps.

packages/web/app/src/lib/supertokens/thirdparty.ts
```diff
@@ -3,9 +3,7 @@ import { env } from '@/env/frontend';
 type Provider = 'google' | 'okta' | 'github' | 'oidc';
 const providers: Provider[] = ['google', 'okta', 'github', 'oidc'];
 
-export const enabledProviders: Provider[] = providers.filter(
-  provider => env.auth[provider] === true,
-);
+export const enabledProviders: Provider[] = providers.filter(provider => !!env.auth[provider]);
```
n1ru4l, line 6: the okta integration was broken 😓

packages/web/app/src/lib/urql.ts
```diff
@@ -292,3 +292,6 @@ type GraphQLPayload = {
       documentId: string;
     }
 );
+
+// @ts-expect-error for testing purposes ok
+window.__YOU_ARE_FIRED_attemptSessionRefresh = () => Session.attemptRefreshingSession();
```
n1ru4l, line 297: just some nice helper to check the session refresh flow

packages/migrations/src/run-pg-migrations.ts
```diff
@@ -183,5 +184,8 @@ export const runPGMigrations = async (args: { slonik: DatabasePool; runTo?: stri
       await import('./actions/2026.01.30T00-00-00.account-linking'),
       await import('./actions/2026.02.06T00-00-00.zendesk-unique'),
       await import('./actions/2026.01.30T10-00-00.oidc-require-invitation'),
+      ...(env.useSupertokensAtHome
+        ? [await import('./actions/2026.02.18T00-00-00.ensure-supertokens-tables')]
+        : []),
```
n1ru4l, line 189: The reason for this migration to be opt-in is that there can always be a race condition between supertokens-core creating the tables and our migrations service creating the tables. If we win, supertokens-core will be in an inconsistent state. Unfortunately, there is no way to prevent supertokens-core from creating tables.

For new self hosters the flow is
1. start without supertokens-core (run as is)
2. start without supertokens-core (remove supertokens service from docker compose and set the `SUPERTOKENS_AT_HOME` environment variable

For the latter you could not easily migrate back to supertokens-core without creating the missing tables, as this one does only create the tables explicitly used by us.

Maybe instead we should create all the tables for now, and then later on delete the unused ones 🤔

jdolle: > Maybe instead we should create all the tables for now, and then later on delete the unused ones 🤔

Funny -- I just commented about this.
If we can avoid this race condition then we absolutely should.

Reiterating what I mentioned before, the table creation should be dated in the past so it runs before any migrations that reference it.

I assume the only concern around us creating the tables is that if we upgrade supertokens then it wont migrate for us. But we already know we won't be doing that, so it isn't a real concern.

packages/web/app/src/lib/urql.ts
dotansimha: Remove before merge? 😆

packages/web/app/src/pages/auth-sign-in.tsx
```diff
@@ -73,7 +73,7 @@ const SignInFormSchema = z.object({
       required_error: 'Email is required',
     })
     .email('Invalid email address'),
-  password: z.string().min(8, 'Password must be at least 8 characters'),
+  password: z.string(),
```
dotansimha, line 76: No need because you moved it to backend?

dotansimha: Ah, because it's sign-in

docker/docker-compose.end2end.yml
```diff
@@ -37,5 +37,9 @@ services:
     ports:
       - '3567:3567'
 
+  db:
```
dotansimha, line 40: still needed? or this was used for testing?

.changeset/light-lions-wear.md
```diff
@@ -0,0 +1,76 @@
+---
+'hive': minor
```
dotansimha, line 2: that's great 😍

docker/docker-compose.end2end.yml
n1ru4l: Yep it is needed for seeding user accounts in integration tests

packages/web/app/src/lib/urql.ts
n1ru4l: We can keep it it is a useful helper.

packages/web/app/src/pages/auth-sign-in.tsx
n1ru4l: yeah no validation needed for sign in!

PR #7700, opened 2026-02-16
feat(app-deployments): add server-side sorting by created, activated, and last used

adambenhassen:
### Background
The app deployments table has no sorting capability, deployments were always listed by creation date descending. Users need to sort by different columns (created, activated, last used) to find relevant deployments quickly, especially in targets with many deployments.

### Description
Adds server-side sorting to the app deployments table.

### DIFF
--- a/packages/services/api/src/modules/app-deployments/providers/app-deployments-manager.ts
+++ b/packages/services/api/src/modules/app-deployments/providers/app-deployments-manager.ts
@@ -226,8 +226,8 @@ export class AppDeploymentsManager {
       cursor: string | null;
       first: number | null;
       sort: {
-        field: 'CREATED_AT' | 'ACTIVATED_AT' | 'LAST_USED';
-        direction: 'ASC' | 'DESC';
+        field: GraphQLSchema.AppDeploymentsSortField;
+        direction: GraphQLSchema.SortDirectionType;
       } | null;
     },
   ) {
--- a/packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
+++ b/packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
@@ -797,13 +797,12 @@ export class AppDeployments {
       try {
         cursor = decodeAppDeploymentSortCursor(args.cursor);
       } catch (error) {
-        this.logger.error(
+        this.logger.debug(
           'Failed to decode cursor for getPaginatedAppDeployments (targetId=%s, cursor=%s): %s',
           args.targetId,
           args.cursor,
           error instanceof Error ? error.message : String(error),
         );
-        throw new Error('Invalid cursor format for getPaginatedAppDeployments.');
       }
     }
     if (cursor && cursor.sortField !== sortField) {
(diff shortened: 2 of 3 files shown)


packages/services/api/src/modules/app-deployments/module.graphql.ts
```diff
@@ -108,7 +129,11 @@ export default gql`
     """
     The app deployments for this target.
     """
-    appDeployments(first: Int, after: String): AppDeploymentConnection
+    appDeployments(
+      first: Int
```
jdolle, line 133:
```suggestion
      first: Int! = 20
```

Minor: I like to set the default in the SDL so that it's clearly visible to API consumer and so that although the argument doesnt need passed, youll always get a value in the resolver.

packages/services/api/src/modules/app-deployments/providers/app-deployments-manager.ts
```diff
@@ -222,13 +222,38 @@ export class AppDeploymentsManager {
...
   ) {
-    return await this.appDeployments.getPaginatedAppDeployments({
-      targetId: target.id,
-      cursor: args.cursor,
-      first: args.first,
-    });
+    const { sort, cursor, first } = args;
+
+    const pagePromise =
+      sort?.field === 'LAST_USED'
```
jdolle, line 237: Do we have an exported enum from codegen we could use?

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -768,13 +770,77 @@ export class AppDeployments {
...
+    const sortField = args.sort?.field ?? 'CREATED_AT';
+    const sortDirection = args.sort?.direction ?? 'DESC';
+
+    let cursor = null;
+    if (args.cursor) {
+      try {
+        cursor = decodeAppDeploymentSortCursor(args.cursor);
+      } catch (error) {
+        this.logger.error(
+          'Failed to decode cursor for getPaginatedAppDeployments (targetId=%s, cursor=%s): %s',
```
jdolle, line 801: I think the only time we should receive an invalid cursor is if the user manually changes it. Because of this, I'd prefer making the log a debug log and instead of throwing, default to returning the first page.

packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -824,6 +883,221 @@ export class AppDeployments {
...
+      args.direction,
+    );
+    const limit = args.first ? (args.first > 0 ? Math.min(args.first, 20) : 20) : 20;
+    const isDesc = args.direction === 'DESC';
+    let cursor = null;
+    if (args.cursor) {
+      try {
+        cursor = decodeAppDeploymentSortCursor(args.cursor);
+      } catch (error) {
+        this.logger.error(
```
jdolle: ^ here too

packages/services/api/src/modules/app-deployments/module.graphql.ts
adambenhassen: I agree but, there's already a code check for this that clamps and sets it to 20 when null. If there's any needed change, it has to also change all fields that utilize the (nullable, no default) `first` parameter pattern and not just `appDeployments`. For consistency, this should be a codebase-wide decision, not a one-off on this field.

This is based on the project patterns document: https://www.notion.so/theguildoss/Hive-Console-GraphQL-API-Design-Patterns-18ab6b71848a80678939c669699fe8e6#18ab6b71848a80d5a1b1d16abd881d32

packages/services/api/src/modules/app-deployments/providers/app-deployments-manager.ts
adambenhassen: Nope. codegen produces string union types, not typescript enums, so there's no enum member to reference

packages/services/api/src/modules/app-deployments/module.graphql.ts
n1ru4l: https://github.com/graphql-hive/console/issues/7750

PR #7705, opened 2026-02-18
remove supertokens-core and supertokens-node

n1ru4l:
this PR removes the supertokens core depndency, making supertokens-at-home the default implementation being used. The migration steps are documented as part of the changeset.

### DIFF
--- a/packages/migrations/src/run-pg-migrations.ts
+++ b/packages/migrations/src/run-pg-migrations.ts
@@ -179,9 +179,9 @@ export const runPGMigrations = async (args: { slonik: DatabasePool; runTo?: stri
       await import('./actions/2026.01.30T00-00-00.account-linking'),
       await import('./actions/2026.02.06T00-00-00.zendesk-unique'),
       await import('./actions/2026.01.30T10-00-00.oidc-require-invitation'),
+      await import('./actions/2026.02.18T00-00-00.ensure-supertokens-tables'),
       await import('./actions/2026.02.19T00-00-00.saved-filter-permission'),
       await import('./actions/2026.02.24T00-00-00.proposal-composition'),
       await import('./actions/2026.02.25T00-00-00.oidc-integration-domains'),
-      await import('./actions/2026.02.18T00-00-00.ensure-supertokens-tables'),
     ],
   });


docs/DEVELOPMENT.md
```diff
@@ -46,64 +46,6 @@ Add "user" field to ./docker/docker-compose.dev.yml
 - Open the UI (`http://localhost:3000` by default) and Sign in with any of the identity provider
 - Once this is done, you should be able to log in and use the project
 
-#### Possible Issues During Setup
```
jdolle, line 49: Removing this makes me happy

packages/migrations/src/run-pg-migrations.ts
```diff
@@ -186,10 +180,8 @@ export const runPGMigrations = async (args: { slonik: DatabasePool; runTo?: stri
       await import('./actions/2026.02.06T00-00-00.zendesk-unique'),
       await import('./actions/2026.01.30T10-00-00.oidc-require-invitation'),
       await import('./actions/2026.02.19T00-00-00.saved-filter-permission'),
-      ...(env.useSupertokensAtHome
-        ? [await import('./actions/2026.02.18T00-00-00.ensure-supertokens-tables')]
-        : []),
       await import('./actions/2026.02.24T00-00-00.proposal-composition'),
       await import('./actions/2026.02.25T00-00-00.oidc-integration-domains'),
+      await import('./actions/2026.02.18T00-00-00.ensure-supertokens-tables'),
```
jdolle: minor: prefer the dates being in order and not reordering migrations

packages/services/server/src/supertokens-at-home.ts
```diff
@@ -70,6 +71,57 @@ export async function registerSupertokensAtHome(
       });
   }
 
+  const OIDCIdLookupSchema = z.object({
+    slug: z.string({
+      required_error: 'Slug is required',
+    }),
+  });
+
+  server.post('/auth-api/oidc-id-lookup', async (req, res) => {
```
jdolle, line 80: Registering the endpoint here is much neater than registering in server/src/index 🎉

packages/services/server/src/supertokens-at-home.ts
```diff
@@ -735,7 +794,7 @@ export async function registerSupertokensAtHome(
 
         const oidClientConfig = new oidClient.Configuration(
           {
-            issuer: oidcIntegration.id,
+            issuer: 'noop',
```
jdolle, line 797: Is this intentional? And if so, what does it accomplish?

n1ru4l: This library requires providing it, but it is not really used for the methods we use. To avoid follow up person to confuse to think it needs to be the integration id I changed it to some noop value to make it more obvious.

packages/migrations/src/run-pg-migrations.ts
n1ru4l: done!

PR #7706, opened 2026-02-19
Add schema proposal composition background job

jdolle:
### Background

Schema proposals need to run composition when they're updated. By creating a workflow, composition can be ran in the background on proposal update.

### Description

Screenshot 2026-02-25 at 3 02 58 PM

This adds a background job to run composition when proposals are updated, and adds more columns to the schema proposal record to store the latest composition state.

A subscription was added to allow async updating of the composition state but the frontend logic for the subscription is not yet added because this PR is getting large.

Also, the redis pubsub instance was factored out into a separate package to avoid a circular dependency between the workflows and api packages.

### Checklist

- [x] Testing

### DIFF
--- a/integration-tests/tests/api/proposals/subscribe.spec.ts
+++ b/integration-tests/tests/api/proposals/subscribe.spec.ts
@@ -109,7 +109,7 @@ describe('Schema Proposals', () => {
         .then(r => r.expectNoGraphQLErrors());
       expect(checkResultErrors.schemaCheck.__typename).toBe(`SchemaCheckSuccess`);
       const { value } = await query.next();
-      expect(value.data.schemaProposalComposition.status).toBe(`success`);
+      expect(value.data.schemaProposalComposition.status).toBe(`SUCCESS`);
       await expect(query.return?.()).resolves.toMatchObject({ done: true });
     },
   );


packages/migrations/src/actions/2026.02.24T00-00-00.proposal-composition.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export default {
+  name: '2026.02.24T00-00-00.proposal-composition.ts',
+  run: ({ sql }) => [
+    {
+      name: 'add schema proposal composition state',
+      query: sql`
+        ALTER TABLE IF EXISTS "schema_proposals"
+            ADD COLUMN IF NOT EXISTS "composition_status" TEXT
+          , ADD COLUMN IF NOT EXISTS "composition_timestamp" TIMESTAMPTZ
```
jdolle, line 11: Looking for feedback on this.
Should the composition state for the proposal be per-save or is it acceptable to be per-proposal (latest state only)?

jonathanawesome: I think the current/latest state is the only relevant composition state in a proposal.

deployment/index.ts
```diff
@@ -201,6 +190,18 @@ const schemaPolicy = deploySchemaPolicy({
...
+deployWorkflows({
+  image: docker.factory.getImageId('workflows', imagesTag),
+  docker,
+  environment,
+  postgres,
+  postmarkSecret,
+  observability,
+  sentry,
+  heartbeat: heartbeatsConfig.get('webhooks'),
+  schema,
```
jdolle, line 202: moved after the schema instance because i need to reference it.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -728,6 +728,18 @@ export class SchemaPublisher {
     const retention = await this.rateLimit.getRetention({ targetId: target.id });
     const expiresAt = retention ? new Date(Date.now() + retention * millisecondsPerDay) : null;
 
+    if (input.schemaProposalId) {
+      // @todo use saved composition settings
```
jdolle: This needs addressed before merging.

packages/services/workflows/package.json
```diff
@@ -10,15 +10,23 @@
   },
   "devDependencies": {
     "@graphql-hive/logger": "1.0.9",
+    "@graphql-hive/pubsub": "workspace:*",
+    "@graphql-inspector/core": "7.1.2",
+    "@graphql-inspector/patch": "0.1.3-alpha-20260225183305-b1e68e3f363401fe16845d7f731ff8ed3467f5a7",
```
jdolle: merge + release https://github.com/graphql-hive/graphql-inspector/pull/2937 before this PR.

packages/services/workflows/src/environment.ts
n1ru4l, line 1: Adding these additional environment variables need to be documented in a changeset for self-hosters.

packages/services/api/src/modules/proposals/providers/schema-proposal-storage.ts
```diff
@@ -43,6 +45,23 @@ export class SchemaProposalStorage {
...
+      endpoint?: string | null;
+      encryptedSecret?: string | null;
+    };
+    native: boolean;
+  }) {
+    const tools: WorkerUtils = await makeWorkerUtils({
+      pgPool: this.pool.pool,
+      logger: bridgeGraphileLogger(this.logger),
+    });
+    await tools.addJob('schemaProposalComposition', { input });
```
n1ru4l: We already have a typesafe way for scheduling a job:  `TaskSchedule`

https://github.com/graphql-hive/console/blob/d777e3252ae54f2b2ee10bb431a73e3a9489e590/packages/services/api/src/modules/alerts/providers/adapters/webhook.ts#L27-L30

packages/services/api/src/modules/proposals/providers/schema-proposal-storage.ts
```diff
@@ -388,6 +433,17 @@ const SchemaProposalModel = z.object({
   stage: StageModel,
   targetId: z.string(),
   author: z.string(),
+  compositionStatus: z.string().nullable(),
```
n1ru4l: Should this be an enum?

PR #7747, opened 2026-02-26
add Insights filters product update

jonathanawesome:
This PR adds a product update for the new insights filters UI.


packages/web/docs/src/authors/index.ts
```diff
@@ -280,6 +280,11 @@ export const authors = {
     link: 'https://github.com/XiNiHa',
     github: 'XiNiHa',
   },
+  jonathanawesome: {
```
jdolle, line 283: 🥇

PR #7756, opened 2026-02-28
Console 1584 denied access to screen even though i can see the nav link

jonathanawesome:
This PR solves an issue where users could see "Settings" navigation links in the organization and project layouts even when they lacked the permissions to access the Settings page. Clicking the link would silently redirect them back, creating a confusing "denied access" experience. The target layout already had the correct permission gating — this PR brings the organization and project layouts in line by querying the relevant viewerCan* fields and conditionally showing the Settings tab.

Additionally, this PR takes the opportunity to DRY up our layout files (organization.tsx, project.tsx, target.tsx), which previously duplicated ~150 lines of nearly identical navigation boilerplate (tabs, loading skeletons, permission-gated links). This is replaced with:

- A new SecondaryNavLink component (secondary-nav-link.tsx) built with TanStack Router's createLink for full route type safety
- An updated SecondaryNavigation component that accepts a typed links array prop, encapsulating tabs, skeleton loading state, and actions
- Each layout now declares its navigation as a simple array of link objects with visible flags for permission gating, reducing duplication and making it easier to add or modify nav links going forward.

### DIFF
--- a/packages/web/app/src/components/layouts/organization.tsx
+++ b/packages/web/app/src/components/layouts/organization.tsx
@@ -130,7 +130,7 @@ export function OrganizationLayout({
       </Header>
       <SecondaryNavigation
         page={page}
-        loading={query.fetching}
+        loading={!currentOrganization}
         className="min-w-[600px]"
         links={
           currentOrganization
--- a/packages/web/app/src/components/layouts/target.tsx
+++ b/packages/web/app/src/components/layouts/target.tsx
@@ -189,7 +189,7 @@ export const TargetLayout = ({
         <>
           <SecondaryNavigation
             page={page}
-            loading={query.fetching}
+            loading={!currentOrganization || !currentProject || !currentTarget}
             className="flex h-full grow flex-col"
             links={
               currentOrganization && currentProject && currentTarget
--- a/packages/web/app/src/components/layouts/project.tsx
+++ b/packages/web/app/src/components/layouts/project.tsx
@@ -115,7 +115,7 @@ export function ProjectLayout({
         <>
           <SecondaryNavigation
             page={page}
-            loading={query.fetching}
+            loading={!currentOrganization || !currentProject}
             links={
               currentOrganization && currentProject
                 ? [


packages/web/app/src/components/layouts/organization.tsx
```diff
@@ -125,84 +128,72 @@ export function OrganizationLayout({
...
+                  params: { organizationSlug: currentOrganization.slug },
+                  search: { page: 'list' },
+                },
+                {
+                  value: Page.Settings,
+                  label: 'Settings',
+                  visible:
+                    currentOrganization.viewerCanAccessSettings ||
+                    currentOrganization.viewerCanManageAccessTokens ||
+                    currentOrganization.viewerCanManagePersonalAccessTokens,
```
jdolle, line 158: I'm okay with this as is but I had previously interpreted `viewerCanAccessSettings` as an inclusive case meaning the user can access any one of the things on the settings page.

Whatever the case is, if it's clear within the descriptions of the fields then I'm okay with it.

packages/web/app/src/components/layouts/target.tsx
```diff
@@ -184,153 +187,87 @@ export const TargetLayout = ({
...
-                <div className="flex flex-row gap-x-8 border-b-2 border-b-transparent px-4 py-3">
-                  <div className="bg-neutral-5 h-5 w-12 animate-pulse rounded-full" />
-                  <div className="bg-neutral-5 h-5 w-12 animate-pulse rounded-full" />
-                  <div className="bg-neutral-5 h-5 w-12 animate-pulse rounded-full" />
-                </div>
-              )}
-              {currentTarget && isCDNEnabled && (
+          <SecondaryNavigation
+            page={page}
+            loading={!currentOrganization || !currentProject || !currentTarget}
```
jdolle, line 192: loading should use the loading state of the query.
If one of these fields errors then this logic would show the spinner forever rather than indicating an error occurred during fetch

jdolle (commented):
Good stuff. I just had some minor comments.

jonathanawesome:
> looks good with me, please review the comments from @jdolle and address if needed!

Thanks! It's in my queue, will get to it this evening.

packages/web/app/src/components/layouts/organization.tsx
jonathanawesome: `viewerCanManagePersonalAccessTokens` isn't in the resolver on the backend...this is an easier front-end fix. Should we prefer the resolver update?

packages/web/app/src/components/layouts/target.tsx
jonathanawesome: Great suggestion. I did the thing I always complain about and just copied the pattern...same pattern exists across the other layouts. I'll fix all 3.

packages/web/app/src/components/layouts/organization.tsx
jdolle: If this gets the behavior we need then I'm okay with this. We don't have clearly defined requirements anywhere to push logic to the backend for things like this.

jdolle (approved):

n1ru4l:
I noticed one thing here after testing this on dev, not using the `fetching` property was there for a reason, to prevent sub nav flickering 😓 

https://github.com/user-attachments/assets/3adc2266-1dc5-4e3d-9326-eed928b8f1e5

PR #7761, opened 2026-03-03
fix: no assigned resources on permission group

n1ru4l:
### Background

See https://linear.app/the-guild/issue/CONSOLE-1910/access-token-detail-view-shows-wrong-resolved-resources

### Description

Show the correct UI state in case no resources are assigned. Previously we showed "granted on all resources" which is incorrect.

image


packages/services/api/src/modules/organization/module.graphql.ts
```diff
@@ -1468,6 +1468,7 @@ export default gql`
     - **App Deployment** "the-guild/graphql-hive/production/appDeployment/production"
 
     These ids can also contain wildcards, e.g. "the-guild/graphql-hive/*", to reference all targets in a project.
+    If the field resolves to 'null', the permission group is granted on no resources.
```
n1ru4l:
```suggestion
    If the field resolves to 'null', the permission group is not granted on any resource.
```

PR #7768, opened 2026-03-04
Console 1903 is it possible to have a not rule in filters

jonathanawesome:
This PR adds exclude ("is not") filter support to insights, allowing operations and clients to be negated in both ad-hoc and saved filters. Filter chips are redesigned as segmented multi-button components (Linear-style) with distinct label, operator, count, and remove segments. Also fixes a browser lockup in DateRangePicker caused by a bidirectional useEffect re-render loop — onUpdate is now called directly in click handlers instead.

https://github.com/user-attachments/assets/70fc639f-053f-4c42-8745-f62f504c3b8f


jdolle (approved):
🚀

PR #7773, opened 2026-03-04
Add schemaCheck:approve to the organization access token ui

jdolle:
### Background

Personal access tokens grant permission to `schemaCheck:approve`, but this option is not available to organization access tokens.

This permission is used on schema checks with the `--forceSafe` argument.

### Description

Adds the permission.

### DIFF
--- a/.changeset/fast-cities-serve.md
+++ b/.changeset/fast-cities-serve.md
@@ -2,4 +2,4 @@
 'hive': patch
 ---
 
-Add schemaCheck:approve permission to organization access token list
+Add `schemaCheck:approve` permission to organization access token list


packages/services/api/src/modules/organization/lib/organization-access-token-permissions.ts
```diff
@@ -146,6 +146,12 @@ export const permissionGroups: Array<PermissionGroup> = [
         title: 'Delete service',
         description: 'Deletes a service from the schema registry.',
       },
+      {
+        id: 'schemaCheck:approve',
+        title: 'Approve schema check',
+        description: 'Grant access to approve failed schema checks.',
+        dependsOn: 'project:describe',
```
jdolle: I'd like feedback on this--
I added the dependsOn to match the organization member permission UI, but admittedly I don't know why we have this dependsOn.

I assume because the resolvers make calls that require `project:describe`, but I'd be open to removing this restriction and adjusting the resolvers to avoid having permissions depend on other permissions.

n1ru4l:
Can access tokens approve schema changes today? Also `Mutation.approveFailedSchemaCheck` is not really part of the public GraphQL schema. 

Okay..., nevermind, just realized we support approving failed schema checks via the CLI since https://github.com/graphql-hive/console/pull/7193. 🤦 

Okay, in that case this makes sense!

packages/services/api/src/modules/organization/lib/organization-access-token-permissions.ts
n1ru4l: The depends on is mostly for user permissions, so you do not assign `manage project` without `project describe`. We could drop the dependency here.

n1ru4l:
```suggestion
```

PR #7788, opened 2026-03-05
fix the organization members list not updating after removing a user

jonathanawesome:
This PR fixes the organization members list not updating after removing a user. It passes refetchMembers down to each member row and calls it after a successful deletion, matching the existing pattern used by the invitations component.

### DIFF
--- a/.changeset/lemon-planes-eat.md
+++ b/.changeset/lemon-planes-eat.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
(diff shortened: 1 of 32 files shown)


n1ru4l (approved):
💯

n1ru4l:
@jonathanawesome A side note here, we did the refetching when we did not have many users in each organization. Since that has now changed and we have quite a few organizations with few hundred to thousand users - Should we reconsider going for cache update handlers instead of refetching for a better UX? IF so - let's create a follow up issue.

PR #7813, opened 2026-03-09
Fix colliding monaco editor bundles

jonathanawesome:
This PR fixes the Monaco diff editor rendering bug (text overlapping between panes, TextModel got disposed errors) caused by a version mismatch between the editor core loaded from CDN and the workers bundled locally by vite-plugin-monaco-editor. The fix switches @monaco-editor/react to use the locally bundled Monaco instance — matching the approach already used by the laboratory library.

### DIFF
--- a/packages/web/app/src/lib/preflight/graphiql-plugin.tsx
+++ b/packages/web/app/src/lib/preflight/graphiql-plugin.tsx
@@ -145,6 +145,21 @@ export const enum PreflightWorkerState {
   ready,
 }
 
+function PromiseWithResolvers<T>() {
+  let resolve: (value: T) => void;
+  let reject: (error: any) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+
+  return {
+    promise,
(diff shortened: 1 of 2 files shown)


jdolle (commented):
Tested locally and this looks great now.

The cypress e2e test isn't finding the correct instances of monaco though

packages/web/app/src/lib/preflight/graphiql-plugin.tsx
```diff
@@ -233,7 +248,7 @@ export function usePreflight(args: {
       );
 
       let isFinished = false;
-      const isFinishedD = Promise.withResolvers<void>();
+      const isFinishedD = PromiseWithResolvers<void>();
```
n1ru4l, line 251: This is super confusing to me why `Promise.withResolvers` was raising an error for me - I checked whether `https://cdn.jsdelivr.net/npm/monaco-editor` might polyfill it, but couldn't find any hint... I do not have Chrome installed though on my machine and was using Electron v118 for running cypress tests - which probably does not have this API yet 😅

PR #7816, opened 2026-03-10
docs: give some examples for contract composition

n1ru4l:
As requested by a client - I had some time to fill while waiting for CI, so I used some LLM magic and altered the output a bit to be more correct. 😄


jonathanawesome (approved):

jonathanawesome:
Nice upgrade, the Gemini suggestions are good!

PR #7819, opened 2026-03-10
Console 1943 figure out why were getting so many monaco related errors in

jonathanawesome:
This PR fixes a Sentry error "TextModel got disposed before DiffEditorWidget model got reset" caused by a race condition when Monaco DiffEditor components unmount during tab switches. The fix uses useLayoutEffect to call setModel(null) before the library's cleanup runs, detaching models from the widget and preventing the disposal order conflict.


n1ru4l (approved):
😍 nice!

PR #7826, opened 2026-03-11
fix: improve clickhouse queries for trace span lookups

n1ru4l:
### Background

https://guild-oss.slack.com/archives/C01E8ATBQGN/p1773238875524199?thread_ts=1773214117.824749&cid=C01E8ATBQGN

### DIFF
--- a/packages/services/api/src/modules/operations/providers/traces.ts
+++ b/packages/services/api/src/modules/operations/providers/traces.ts
@@ -41,12 +41,21 @@ export class Traces {
     this.logger.debug('looking up traces by id (traceIds=%o)', traceIds);
     const result = await this.clickHouse.query<unknown>({
       query: sql`
+        WITH trace_window AS (
+          SELECT
+            min("Start") AS start_ts,
+            max("End") AS end_ts
+          FROM "otel_traces_trace_id_ts"
+          WHERE "TraceId" IN (${sql.array(traceIds, 'String')})
+        )
         SELECT
           ${traceFields}
         FROM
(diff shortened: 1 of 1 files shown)


kamilkisiela:
Add:
- https://guild-oss.slack.com/archives/C01E8ATBQGN/p1773243085639649?thread_ts=1773214117.824749&cid=C01E8ATBQGN
- https://guild-oss.slack.com/archives/C01E8ATBQGN/p1773243591679119?thread_ts=1773214117.824749&cid=C01E8ATBQGN

please and we're hella fast

packages/services/api/src/modules/operations/providers/traces.ts
```diff
@@ -105,15 +105,15 @@ export class Traces {
           "otel_traces"
         PREWHERE
           "TraceId" = ${traceId}
-        WHERE
-          "SpanAttributes"['hive.target_id'] = ${targetId}
```
jdolle, line 109: Could you elaborate on why this has such an impact in the query performance? I'd expect filtering in the DB to be as performant as post filtering

n1ru4l: Data is stored in columns, clickhouse needs to read the whole "SpanAttributes" column, deserialize and look for the `hive.target_id` key.

n1ru4l: By the way this is only to mitigate the unlikely scenario where someone knows the TraceId of another customer and attempts to read the spans of these

PR #7828, opened 2026-03-11
Run clickhouse app deployment checks in parallel to speed up schema c…

jdolle:
…heck resolution

### Background

https://linear.app/the-guild/issue/CONSOLE-1949/network-error-during-graphql-hive-cli-schema-check

Some checks are timing out for orgs with a large number of app deployments and a large number of breaking changes.

### Description

This runs three of the clickhouse queries for app deployments in parallel. Although this means overall we'll be making more queries to clickhouse, it will speed up the resolution of schema.

### DIFF
--- a/.changeset/eight-carpets-smoke.md
+++ b/.changeset/eight-carpets-smoke.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+Adjust app deployment schema check to perform database queries in parallel


packages/services/api/src/modules/app-deployments/providers/app-deployments.ts
```diff
@@ -1492,11 +1424,109 @@ export class AppDeployments {
...
+        retiredAt: string | null;
+      }>(
+        sql`
+          SELECT
+            "id",
+            to_json("created_at") AS "createdAt",
+            to_json("activated_at") AS "activatedAt",
+            to_json("retired_at") AS "retiredAt"
+          FROM "app_deployments"
+          WHERE "id" = ANY(${sql.array(deploymentIds, 'uuid')})
```
jdolle, line 1528: This depends on one of the prior queries and so cannot be parallelized

PR #7836, opened 2026-03-12
suppress monaco related errors in sentry + replace deprecated method

jonathanawesome:
This PR adds ignoreErrors to the Sentry configuration to filter known non-critical Monaco editor internal errors (setStart/setEnd on Range, TextModel got disposed) from being reported to Sentry. It also replaces the deprecated configureScope API with getCurrentScope().


jdolle (approved):
please add a changeset

PR #7935, opened 2026-03-30
console-1968-improve-most-valuable-alert-types

jonathanawesome:
This PR adds a metric-based alerting system to Console. Users can define rules that fire when traffic, reliability, or latency on a target breaches a threshold (fixed or % change) and get notified via the existing Slack/webhook/MS Teams channels. Ships behind a two-tier feature flag (cluster kill-switch + per-org enable, both default off) so we can selectively enroll our own org for validation without exposing the feature to other customers, then flip a single Pulumi config value to GA the feature for everyone.

#### Backend 

- **5 PG tables:** `metric_alert_rules`, `_rule_channels`,` _incidents`, `_state_log`, `_notifications_sent`. State log has plan-gated expires_at (7d HOBBY/PRO, 30d ENTERPRISE); daily 4am purge.
- **API:** `MetricAlertRule` with cursor-paginated incidents/state-log; add/update/delete mutations; Target-side queries; cross-scope validation (channels + saved filters must match rule's project).
- **Workflows:** `evaluateMetricAlertRules` cron runs every minute, executes `NORMAL → PENDING → FIRING → RECOVERING` state machine with a "Hold minutes" debounce. Per-channel notification fan-out via independent retryable `sendMetricAlertChannelNotification` jobs with three idempotency layers (pre-send dedup, `ON CONFLICT` post-send, `Idempotency-Key` header). `purgeExpiredAlertStateLog` daily cron.
- **Observability:** new ClickHouse-query histogram, Metric-Alerts Grafana dashboard (4 panels), and two Pulumi-provisioned Grafana alert rules (`ClickHouseSlow` warning, `ClickHouseErrors` critical) protecting against silent staleness of the alerter.

#### Frontend 

- **Alerts area under Target with subroutes:** Activity, Rules, Create, Detail
- **Detail page:** rule conditions panel, state-transitions timeline bar, metric over-time chart, events table
- **Rules table:** activity table with severity-bucketed activity chart
- **Notification preview:** users see what Slack/webhook payloads look like before saving
- **Refactor of filter components** (filter-dropdown to floating/filter-menu)...new reusable base components (button, card, accordion, data-table, form, input, page-lead, description-list, select, etc.) with stories

#### Two-tier feature flag, mirroring the existing schemaProposals pattern

- Cluster-wide kill switch (`FEATURE_FLAGS_METRIC_ALERT_RULES_ENABLED`): defaults off; flipping on enables the feature cluster-wide.
- Per-org enable (`organizations.feature_flags.metricAlertRules`): defaults `false`; can be set via direct PG UPDATE to enable for specific orgs (no admin mutation in this PR; same operational pattern as how schemaProposals is enabled today).
- OR semantics: enabled = clusterFlag || orgFlag. Matches every existing flag in the codebase. The cluster flag short-circuits...when off, resolvers fall back to checking the org flag.
- Wired through Pulumi config so a single `featureFlags:metricAlertRulesEnabled` stack value flips both API and workflows.
- Workflows cron filters rules whose org isn't enrolled when the cluster flag is off; runs everything when it's on. `purgeExpiredAlertStateLog` runs unconditionally so opted-in orgs' state-log tables stay bounded.

#### Seed script
`scripts/seed-insights-and-alerts/` replaces the old `seed-insights.mts` with metric-first alert history (per-rule incident windows + matching ClickHouse ops + PG state-log rows)

### Notable decisions

-Evaluation anchored to `helpers.job.run_at`, not `Date.now()`. Wall-clock would shift the window forward when the worker is backed up and miss the spike that should have fired.
-Rule grouping by (`target, window, savedFilter`) — one ClickHouse query per group; query count scales with groups, not rule count.
-OR-semantics feature flag, resolver-layer gate (mirrors schemaProposals).
-Atomic firing transition. State row + log row + incident row + per-channel job enqueues all in one PG transaction
-Plan-gated `expires_at` snapshotted at insert time, so plan changes only affect new rows.

### Worth a closer look in review

A few sub-features that touch shared infrastructure or have cost / blast-radius implications and deserve focused attention:

- **Feature flag plumbing.** `alerts/resolvers/Target.ts`, the three mutation files, and `workflows/src/lib/metric-alert-evaluator.ts` all have to agree on the OR-gate semantics. If a future contributor copies the resolver pattern but forgets to mirror the SQL predicate (or vice versa), you'd get a state where mutations are gated but the cron evaluates rules anyway, or vice versa. Worth confirming the four call sites are consistent.
- **ClickHouse query cost.** The evaluator runs `evaluateMetricAlertRules` every minute against `operations_minutely` / `operations_hourly`, batched by (`targetId`, `timeWindowMinutes`, `savedFilterId`). Worst case: every enabled rule with a unique grouping key is its own ClickHouse round-trip. The query is light (covered by the (target, timestamp) index, returns 2 rows), but it's worth keeping an eye on aggregate ClickHouse load when we widen rollout. Consider checking once we've enabled for a handful of orgs whether `system.query_log` shows acceptable patterns.
- **Workflows service now reaches ClickHouse.** `packages/services/workflows/src/lib/clickhouse-client.ts` is a new dependency edge. The workflows service previously only talked to Postgres. New env vars (`CLICKHOUSE_HOST`, etc.) are wired through the workflows deployment in `deployment/services/workflows.ts`. Confirm the Pulumi stack actually has these set in non-dev environments before flipping the env var.
- **State-log retention.** The `metric_alert_state_log` table is the highest-volume new table. Each rule transition writes a row, and the table has plan-gated TTL. The `purgeExpiredAlertStateLog` cron is the only thing keeping it bounded.
- **Notification fan-out is silent on partial failure.** `metric-alert-notifier.ts` sends to each channel in sequence. If Slack succeeds and the webhook fails, today the failure is logged but doesn't surface to the user...they just get a partial notification. Future observability work will add metrics here, but in the meantime any review feedback on whether we should retry / surface failures more prominently is welcome.


jdolle:
> ClickHouse query cost. The evaluator runs evaluateMetricAlertRules every minute against operations_minutely / operations_hourly, batched by (targetId, timeWindowMinutes, savedFilterId). Worst case: every enabled rule with a unique grouping key is its own ClickHouse round-trip. The query is light (covered by the (target, timestamp) index, returns 2 rows), but it's worth keeping an eye on aggregate ClickHouse load when we widen rollout. Consider checking once we've enabled for a handful of orgs whether system.query_log shows acceptable patterns.

Let's be sure to create an alert on this and add it to our dashboard(s).

jdolle:
> Plan-gated state-log retention. Hobby/Pro 7d, Enterprise 30d. expires_at is snapshotted at insert time so plan changes only affect new rows.

This is an interesting decision. If they have 30d retention and change to 7d, then there could be a gap in logs if it's 7-30 days after the change in subscription.

Long term, I feel like we need an "on subscription change" event that we can do a variety of things with

jdolle:
> Notification fan-out is silent on partial failure. metric-alert-notifier.ts sends to each channel in sequence. If Slack succeeds and the webhook fails, today the failure is logged but doesn't surface to the user...they just get a partial notification. Future observability work will add metrics here, but in the meantime any review feedback on whether we should retry / surface failures more prominently is welcome.

This is the first followup feature I'd address. Firing exactly once is always a challenge. Many of these notification platforms have idempotency built in so we can fire at least once. But we should definitely not partially fulfil

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,85 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2026.04.15T00-00-01.metric-alert-rules.ts',
+  run: ({ psql }) => psql`
+CREATE TYPE "metric_alert_type" AS ENUM ('LATENCY', 'ERROR_RATE', 'TRAFFIC');
+CREATE TYPE "metric_alert_metric" AS ENUM ('avg', 'p75', 'p90', 'p95', 'p99');
```
jdolle: minor: all other types here are upper case

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,85 @@
...
+
+CREATE TABLE "metric_alert_rules" (
+  "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+  "organization_id" uuid NOT NULL REFERENCES "organizations"("id") ON DELETE CASCADE,
+  "project_id" uuid NOT NULL REFERENCES "projects"("id") ON DELETE CASCADE,
+  "target_id" uuid NOT NULL REFERENCES "targets"("id") ON DELETE CASCADE,
+  "created_by_user_id" uuid REFERENCES "users"("id") ON DELETE SET NULL,
+  "updated_by_user_id" uuid REFERENCES "users"("id") ON DELETE SET NULL,
+  "type" "metric_alert_type" NOT NULL,
+  "time_window_minutes" integer NOT NULL DEFAULT 30,
```
jdolle: minor: Id prefer not setting a default at the DB level. We just don't know what the best option is for every case

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,85 @@
...
+CREATE TYPE "metric_alert_severity" AS ENUM ('INFO', 'WARNING', 'CRITICAL');
+CREATE TYPE "metric_alert_state" AS ENUM ('NORMAL', 'PENDING', 'FIRING', 'RECOVERING');
+
+CREATE TABLE "metric_alert_rules" (
+  "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+  "organization_id" uuid NOT NULL REFERENCES "organizations"("id") ON DELETE CASCADE,
+  "project_id" uuid NOT NULL REFERENCES "projects"("id") ON DELETE CASCADE,
+  "target_id" uuid NOT NULL REFERENCES "targets"("id") ON DELETE CASCADE,
+  "created_by_user_id" uuid REFERENCES "users"("id") ON DELETE SET NULL,
+  "updated_by_user_id" uuid REFERENCES "users"("id") ON DELETE SET NULL,
```
jdolle, line 19: Everywhere there's an ON DELETE -- if we need our our delete operations to be somewhat performant, then you need to index by that column.

That way, in this case, when a user is deleted, the `updated_by_user_id` column is indexed and can easily find the correct user_id value. Otherwise, this will cause a complete table scan.

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,85 @@
...
+  "type" "metric_alert_type" NOT NULL,
+  "time_window_minutes" integer NOT NULL DEFAULT 30,
+  "metric" "metric_alert_metric",
+  "threshold_type" "metric_alert_threshold_type" NOT NULL,
+  "threshold_value" double precision NOT NULL,
+  "direction" "metric_alert_direction" NOT NULL DEFAULT 'ABOVE',
```
jdolle: same here im not convinced the default should be at the DB level. This seems like it belongs in the application logic.

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,85 @@
...
+  "state_changed_at" timestamptz,
+  "confirmation_minutes" integer NOT NULL DEFAULT 0,
+  "saved_filter_id" uuid REFERENCES "saved_filters"("id") ON DELETE SET NULL,
+  PRIMARY KEY ("id"),
+  CONSTRAINT "metric_alert_rules_metric_required" CHECK (
+    ("type" = 'LATENCY' AND "metric" IS NOT NULL) OR ("type" != 'LATENCY' AND "metric" IS NULL)
+  )
+);
+
+CREATE INDEX "idx_metric_alert_rules_enabled" ON "metric_alert_rules" ("enabled") WHERE "enabled" = true;
```
jdolle: I'm yet to get to this part of your code, but if we need to scale this up and fan out the processing, then it may be natural to fan out by org, target, or rule ID.

Maybe we can discuss this in more depth, but consider indexing a "paginatable" column where enabled is true.

packages/services/api/src/modules/alerts/providers/metric-alert-rules-storage.ts
```diff
@@ -0,0 +1,621 @@
...
+  .object({
+    id: zod.string(),
+    organizationId: zod.string(),
+    projectId: zod.string(),
+    targetId: zod.string(),
+    createdByUserId: zod.string().nullable(),
+    updatedByUserId: zod.string().nullable(),
+    type: zod.enum(['LATENCY', 'ERROR_RATE', 'TRAFFIC']),
+    timeWindowMinutes: zod.number(),
+    metric: zod.enum(['avg', 'p75', 'p90', 'p95', 'p99']).nullable(),
```
jdolle: minor: same enum casing comment

packages/services/api/src/modules/alerts/providers/metric-alert-rules-storage.ts
```diff
@@ -0,0 +1,621 @@
...
+    lastEvaluatedAt?: Date;
+    lastTriggeredAt?: Date;
+  }): Promise<void> {
+    await this.pool.query(psql`/* updateRuleState */
+      UPDATE "metric_alert_rules"
+      SET
+        "state" = ${args.state}
+        , "state_changed_at" = COALESCE(${args.stateChangedAt?.toISOString() ?? null}, "state_changed_at")
+        , "last_evaluated_at" = COALESCE(${args.lastEvaluatedAt?.toISOString() ?? null}, "last_evaluated_at")
+        , "last_triggered_at" = COALESCE(${args.lastTriggeredAt?.toISOString() ?? null}, "last_triggered_at")
```
jdolle: It would potentially be useful to not update the triggered at if the rule is already triggered. But I'm yet to get to the app logic around this.

packages/services/api/src/modules/alerts/resolvers/Mutation/addMetricAlertRule.ts
```diff
@@ -0,0 +1,110 @@
...
+    thresholdType: input.thresholdType,
+    thresholdValue: input.thresholdValue,
+    direction: input.direction,
+    severity: input.severity,
+    name: input.name,
+    confirmationMinutes: input.confirmationMinutes ?? 0,
+    savedFilterId: input.savedFilterId ?? null,
+  });
+
+  await storage.setRuleChannels({
```
jdolle: what if `addMetricAlertRule` succeeds and `storage.setRuleChannels` errors.
Can this request be retried idempotently?

Consider making this a transaction

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,85 @@
...
+
+CREATE TABLE "metric_alert_rule_channels" (
+  "metric_alert_rule_id" uuid NOT NULL REFERENCES "metric_alert_rules"("id") ON DELETE CASCADE,
+  "alert_channel_id" uuid NOT NULL REFERENCES "alert_channels"("id") ON DELETE CASCADE,
```
jdolle, line 65: So if a channel is deleted, then it's also removed from the alerting list of rule channels.

This could result in an empty list of channels for an alert -- which I believe you validate against. Might be worth allowing zero channels for an alert. This also lets users test out the feature to see how it really evaluates before any alerts go out.

packages/services/api/src/modules/alerts/resolvers/Mutation/updateMetricAlertRule.ts
```diff
@@ -0,0 +1,120 @@
...
+      return {
+        error: {
+          message: 'All notification channels must belong to the same project as the target.',
+        },
+      };
+    }
+  }
+
+  if (input.savedFilterId) {
+    const filterProjectId = await storage.getSavedFilterProjectId(input.savedFilterId);
```
jdolle: I somehow completely missed the tie-in to saved filters before this. This is very interesting... I'd like to understand it a bit more but will keep my eyes peeled through the rest of the PR also

packages/services/api/src/modules/alerts/resolvers/MetricAlertRuleIncident.ts
```diff
@@ -0,0 +1,14 @@
...
+/*
+ * Note: This object type is generated because "MetricAlertRuleIncidentMapper" is declared. This is to ensure runtime safety.
+ *
+ * When a mapper is used, it is possible to hit runtime errors in some scenarios:
+ * - given a field name, the schema type's field type does not match mapper's field type
+ * - or a schema type's field does not exist in the mapper's fields
+ *
+ * If you want to skip this file generation, remove the mapper or update the pattern in the `resolverGeneration.object` config.
+ */
+export const MetricAlertRuleIncident: MetricAlertRuleIncidentResolvers = {
```
jdolle, line 4: should this be implemented?

packages/services/api/src/modules/alerts/resolvers/MetricAlertRuleStateChange.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { MetricAlertRuleStateChangeResolvers } from './../../../__generated__/types';
+
+export const MetricAlertRuleStateChange: MetricAlertRuleStateChangeResolvers = {
+  rule: async (entry, _, { injector }) => {
+    const rule = await injector
+      .get(MetricAlertRulesStorage)
+      .getMetricAlertRule({ id: entry.metricAlertRuleId });
+
+    if (!rule) {
+      throw new Error(`Metric alert rule ${entry.metricAlertRuleId} not found`);
```
jdolle: Should this return null instead of throwing?

packages/services/api/src/modules/alerts/module.graphql.ts
```diff
@@ -154,4 +154,239 @@ export default gql`
...
+    ERROR_RATE
+    TRAFFIC
+  }
+
+  enum MetricAlertRuleMetric {
+    avg
+    p75
+    p90
+    p95
+    p99
```
jdolle: I assume this is why the casing for this enum was different in the DB.
I think everywhere should be consistent but the enums in the schema dont need to match 1:1 what's in the DB. There's frequently a translation layer built

packages/services/api/src/modules/alerts/module.graphql.ts
```diff
@@ -154,4 +154,239 @@ export default gql`
...
+    PENDING
+    FIRING
+    RECOVERING
+  }
+
+  type MetricAlertRule {
+    id: ID!
+    name: String!
+    type: MetricAlertRuleType!
+    target: Target!
```
jdolle: maybe controversial but I'd prefer non-null here since it's another DB call. I always favor partial results

packages/services/api/src/modules/alerts/module.graphql.ts
```diff
@@ -154,4 +154,239 @@ export default gql`
...
+    """
+    currentIncident: MetricAlertRuleIncident
+    """
+    Past incidents for this alert rule, paginated newest-first.
+    """
+    incidents(first: Int, after: String): MetricAlertRuleIncidentConnection!
+    """
+    State change history for this rule (powers the state timeline).
+    """
+    stateLog(from: DateTime!, to: DateTime!): [MetricAlertRuleStateChange!]!
```
jdolle, line 242: should this be paginated?

packages/services/api/src/modules/alerts/module.graphql.ts
```diff
@@ -154,4 +154,239 @@ export default gql`
...
+    thresholdType: MetricAlertRuleThresholdType!
+    thresholdValue: Float!
+    direction: MetricAlertRuleDirection!
+    severity: MetricAlertRuleSeverity!
+    confirmationMinutes: Int
+    channelIds: [ID!]!
+    savedFilterId: ID
+  }
+
+  input UpdateMetricAlertRuleInput {
```
jdolle, line 353: very minor: It may be a nice convenience to have a specialized `enableMetricAlertRule` also. I'm envisioning a toggle in the UI

packages/services/api/src/modules/commerce/constants.ts
```diff
@@ -15,3 +15,15 @@ export const USAGE_DEFAULT_LIMITATIONS: Record<
...
   },
 };
+
+/**
+ * How long alert state-log rows are kept before being purged by the
+ * `purgeExpiredAlertStateLog` cron task. The evaluation engine snapshots this
+ * value into `metric_alert_state_log.expires_at` at insert time based on the
+ * organization's plan, so changing the constant only affects new rows.
+ */
+export const ALERT_STATE_LOG_RETENTION_DAYS: Record<'HOBBY' | 'PRO' | 'ENTERPRISE', number> = {
```
jdolle, line 25: Retention can be set in the DB by org. But I'm fine with us specifying that this is the retention for this specific feature for these plans.
Let's be sure that, once released, this is reflected on our pricing page

package.json
```diff
@@ -44,7 +44,7 @@
     "release": "pnpm build:libraries && changeset publish",
     "release:version": "changeset version && pnpm build:libraries && pnpm --filter @graphql-hive/cli oclif:readme",
     "seed:app-deployments": "tsx scripts/seed-app-deployments.mts",
-    "seed:insights": "tsx scripts/seed-insights.mts",
+    "seed:insights-and-alerts": "tsx scripts/seed-insights-and-alerts/seed-insights-and-alerts.mts",
```
jdolle, line 51: minor: prefer keeping the script name the same. I see this as an enhancement of the existing script

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -0,0 +1,476 @@
...
+  window: 'current' | 'previous';
+  total: string;
+  total_ok: string;
+  average: number;
+  percentiles: [number, number, number, number];
+};
+
+type GroupKey = string;
+
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
```
jdolle, line 52: Is this overkill?

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -0,0 +1,476 @@
...
+
+  return rule.direction === 'ABOVE'
+    ? compareValue > rule.thresholdValue
+    : compareValue < rule.thresholdValue;
+}
+
+/**
+ * Alert state-log retention per plan, in days. Workflows doesn't depend on
+ * `@hive/api` so this is mirrored from the canonical source at
+ * `packages/services/api/src/modules/commerce/constants.ts`
```
jdolle, line 119: I would love for us to have a shared package that we can use with both workflows and the api without having to duplicate. I ran into this issue also before

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -0,0 +1,476 @@
...
+}
+
+function hasElapsed(stateChangedAt: string | null, minutes: number): boolean {
+  if (!stateChangedAt) return true;
+  const changedAt = new Date(stateChangedAt).getTime();
+  return Date.now() - changedAt >= minutes * 60_000;
+}
+
+function formatClickHouseDate(date: Date): string {
+  return date.toISOString().replace('T', ' ').replace('Z', '').slice(0, 19);
```
jdolle: I think clickhouse has some utility functions built in so you dont have to do this

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -0,0 +1,476 @@
...
+  const breached = isThresholdBreached(currentValue, previousValue, rule);
+
+  // State-log retention is derived from the rule's organization plan, which
+  // is included on the row by `fetchEnabledRules` (single JOIN) — no extra
+  // database round-trip per rule.
+  const retentionDays = getAlertStateLogRetentionDays(rule.organizationPlanName);
+
+  // Notification side-effects fire AFTER the DB transaction commits. Sending
+  // a Slack/webhook for a transition that didn't actually persist would be
+  // worse than missing one, so we collect the intent here and fan out below.
```
jdolle: Add idempotency and let the workflow retry.

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -0,0 +1,476 @@
...
+export async function evaluateRule(args: {
+  rule: MetricAlertRuleRow;
+  current: ClickHouseWindowRow;
+  previous: ClickHouseWindowRow;
+  pg: PostgresDatabasePool;
+  logger: Logger;
+  onTransition?: OnStateTransition;
+}): Promise<void> {
+  const { rule, current, previous, pg, logger, onTransition } = args;
+  const now = new Date();
```
jdolle: I think we should pass in the start and end times rather than rely on relative times. If we use Date.now and the workflow gets backed up, then we potentially could miss windows where an alert would fire just because our processing was too slow. This would cast doubt on the validity of the entire feature

packages/services/workflows/src/lib/metric-alert-notifier.ts
```diff
@@ -0,0 +1,274 @@
...
+        blocks: [
+          {
+            type: 'section',
+            text: {
+              type: 'mrkdwn',
+              text: [
+                `*${event.rule.name}* — ${action}`,
+                `Type: ${event.rule.type} | Severity: ${event.rule.severity}`,
+                changeText,
+                `Target: \`${event.targetSlug}\` in \`${event.projectSlug}\``,
```
jdolle, line 88: Once released, let's be sure to show a screenshot of an example of each alert

packages/services/workflows/src/tasks/evaluate-metric-alert-rules.ts
```diff
@@ -0,0 +1,130 @@
...
+              requestBroker: context.requestBroker,
+              logger,
+            });
+          }
+        },
+      });
+    }
+  }
+
+  logger.info('Metric alert evaluation complete');
```
jdolle: debug or we can provide more context for it to be useful

packages/services/workflows/src/tasks/purge-expired-alert-state-log.ts
```diff
@@ -0,0 +1,25 @@
...
+  const result = await args.context.pg.oneFirst(psql`
+    WITH "deleted" AS (
+      DELETE FROM "metric_alert_state_log"
+      WHERE "expires_at" < NOW()
+      RETURNING 1
+    )
+    SELECT COUNT(*)::int FROM "deleted";
+  `);
+  const amount = z.number().parse(result);
+  args.logger.debug({ purgedCount: amount }, 'finished purging expired alert state log entries');
```
jdolle, line 24: Minor: It might be really nice to log all the purged states so that we have a history if customers ask about which alerts fired during certain windows. I could see this being an important historical artifact when doing incident reviews.

packages/services/workflows/src/index.ts
```diff
@@ -43,13 +44,19 @@ const modules = await Promise.all([
...
 
 const crontab = `
   # Purge expired schema checks every Sunday at 10:00AM
   0 10 * * 0 purgeExpiredSchemaChecks
   # Every day at 3:00 AM
   0 3 * * * purgeExpiredDedupeKeys
+  # Evaluate metric alert rules every minute
+  * * * * * evaluateMetricAlertRules
+  # Purge expired alert state log entries daily at 4:00 AM
+  0 4 * * * purgeExpiredAlertStateLog
```
jdolle: minor: maybe review when we perform backups and purge slightly before that to keep the size slightly more optimized

jdolle (commented):
I got through about half. Will finish asap

packages/web/app/src/components/base/floating/filter-menu/url-filter.ts
```diff
@@ -0,0 +1,104 @@
...
+    onChange,
+    onRemove,
+    valuesLabel,
+    excludeMode,
+    onExcludeModeChange,
+  };
+}
+
+/** Encode `FilterSelection[]` as `string[]` of ids — the common flat case. */
+export function encodeIdSelections(selections: FilterSelection[]): string[] | undefined {
```
jdolle, line 98: why not return an empty array if length is 0?

packages/web/app/src/components/target/alerts/alert-activity-chart.tsx
```diff
@@ -0,0 +1,205 @@
...
+  // unambiguous when hovering, even if the x-axis only shows the date.
+  const tooltipTimeFormatter = new Intl.DateTimeFormat('en-US', {
+    month: 'short',
+    day: 'numeric',
+    hour: 'numeric',
+    minute: '2-digit',
+  });
+
+  // Aim for ~7 visible x-axis labels regardless of bucket count.
+  const targetLabelCount = 7;
```
jdolle: i'm curious why 7? Is this based on our page width and standard resolution?
Should it differ for mobile resolutions?

packages/web/app/src/components/target/alerts/alert-activity-table.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+const TIMESTAMP_FORMAT = new Intl.DateTimeFormat('en-US', {
+  month: 'long',
+  day: 'numeric',
+  year: 'numeric',
+  hour: '2-digit',
+  minute: '2-digit',
+  second: '2-digit',
+  hour12: true,
+  timeZone: 'UTC',
```
jdolle: minor/future improvement:
existing insights charts use local timezones. It'd be nice to be able to adjust the timezone

packages/web/app/src/components/target/alerts/alert-conditions-panel.tsx
```diff
@@ -0,0 +1,319 @@
...
+  );
+}
+
+export function AlertConditionsPanel({
+  rule,
+  organizationSlug,
+  projectSlug,
+  targetSlug,
+}: AlertConditionsPanelProps) {
+  console.log('AlertConditionsPanel', { rule });
```
jdolle:
```suggestion
```

packages/web/app/src/components/target/alerts/alert-event-detail.tsx
```diff
@@ -0,0 +1,289 @@
...
+  metric?: string | null;
+  direction: MetricAlertRuleDirection;
+  thresholdType: MetricAlertRuleThresholdType;
+  thresholdValue: number;
+  timeWindowMinutes: number;
+  savedFilter?: SavedFilter | null;
+};
+
+const METRIC_LABEL_BY_TYPE: Record<MetricAlertRuleType, string> = {
+  [MetricAlertRuleType.ErrorRate]: 'error rate',
```
jdolle, line 74: it might be nicer to return the label from the backend to guarantee if we add a rule type in the future that it's shown correctly.

packages/web/app/src/components/target/alerts/alert-event-detail.tsx
```diff
@@ -0,0 +1,289 @@
...
+  previous: number | null | undefined,
+): string {
+  if (current === null || current === undefined) return '—';
+  const currentText = formatRawValue(rule.type, current);
+  if (previous === null || previous === undefined) return currentText;
+  const previousText = formatRawValue(rule.type, previous);
+  if (rule.type === MetricAlertRuleType.ErrorRate) {
+    const delta = current - previous;
+    const sign = delta >= 0 ? '+' : '';
+    return `${currentText} (was ${previousText}, ${sign}${delta.toFixed(1)}pp)`;
```
jdolle: I prefer `%` over `pp` because it's more commonly used

packages/web/app/src/components/target/alerts/alert-events-table.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const TIMESTAMP_FORMAT = new Intl.DateTimeFormat('en-US', {
+  month: 'long',
+  day: 'numeric',
+  year: 'numeric',
+  hour: '2-digit',
+  minute: '2-digit',
+  second: '2-digit',
+  hour12: true,
+  timeZone: 'UTC',
```
jdolle: ^ here again i'd prefer local or allowing switching at least between utc and local

packages/web/app/src/components/target/alerts/alert-form.tsx
```diff
@@ -0,0 +1,837 @@
...
+  const { toast } = useToast();
+
+  const [channelsQuery] = useQuery({
+    query: AlertForm_ChannelsQuery,
+    variables: { organizationSlug, projectSlug },
+  });
+
+  const [savedFiltersQuery] = useQuery({
+    query: AlertForm_SavedFiltersQuery,
+    variables: { organizationSlug, projectSlug, targetSlug },
```
jdolle, line 370: `savedFilters` is paginated to the first 50. Should this be passing in a page number variable?

packages/web/app/src/components/target/alerts/alert-form.tsx
```diff
@@ -0,0 +1,837 @@
...
+              <div className="flex items-start gap-6">
+                <FormField
+                  control={form.control}
+                  name="metricSelection"
+                  render={({ field }) => (
+                    <FormItem>
+                      <FormLabel label="Metric" />
+                      <FormControl>
+                        <Select
+                          options={[...METRIC_OPTIONS]}
```
jdolle: why spread?

```suggestion
                          options={METRIC_OPTIONS}
```

packages/web/app/src/components/target/alerts/alert-form.tsx
```diff
@@ -0,0 +1,837 @@
...
+                />
+                <FormField
+                  control={form.control}
+                  name="timeWindowMinutes"
+                  render={({ field }) => (
+                    <FormItem>
+                      <FormLabel label="Range" />
+                      <FormControl>
+                        <Select
+                          options={[...RANGE_OPTIONS]}
```
jdolle:
```suggestion
                          options={RANGE_OPTIONS}
```

packages/web/app/src/components/target/alerts/alert-form.tsx
```diff
@@ -0,0 +1,837 @@
...
+                <div className="flex items-start gap-6">
+                  <FormField
+                    control={form.control}
+                    name="direction"
+                    render={({ field }) => (
+                      <FormItem>
+                        <FormLabel label="Condition" />
+                        <FormControl>
+                          <Select
+                            options={[...CONDITION_OPTIONS]}
```
jdolle:
```suggestion
                            options={CONDITION_OPTIONS}
```

packages/web/app/src/components/target/alerts/alert-form.tsx
```diff
@@ -0,0 +1,837 @@
...
+                  />
+                  <FormField
+                    control={form.control}
+                    name="thresholdType"
+                    render={({ field }) => (
+                      <FormItem>
+                        <FormLabel label="Threshold type" />
+                        <FormControl>
+                          <Select
+                            options={[...THRESHOLD_TYPE_OPTIONS]}
```
jdolle:
```suggestion
                            options={THRESHOLD_TYPE_OPTIONS}
```

packages/web/app/src/components/target/alerts/alert-form.tsx
```diff
@@ -0,0 +1,837 @@
...
+                    )}
+                  />
+                </div>
+
+                <AlertMetricChart
+                  organizationSlug={organizationSlug}
+                  projectSlug={projectSlug}
+                  targetSlug={targetSlug}
+                  metricSelection={watchedValues.metricSelection}
+                  timeWindowMinutes={parseInt(watchedValues.timeWindowMinutes, 10) || 10_080}
```
jdolle: This is 7 days right?
If there's a blip that causes an alert, is this a long enough period to obscure the blip's metric to make it unclear why it alerted?
We may want to determine the time window based on the alert's value calculation period. I.e. if alerting when error rate is > 2% in the last 15min, then maybe we display the last 30min?

packages/web/app/src/components/target/alerts/alert-metric-chart.tsx
```diff
@@ -0,0 +1,260 @@
...
+    query: AlertMetricChart_Query,
+    variables: {
+      targetSelector: { organizationSlug, projectSlug, targetSlug },
+      period,
+      resolution,
+    },
+  });
+
+  const stats = result.data?.target?.operationsStats;
+  const isLatency = metricSelection.startsWith('LATENCY:');
```
jdolle: not a fan of string matching. Should another parameter be passed in to indicate the metric type?

packages/web/app/src/components/target/alerts/alert-notification-preview.tsx
```diff
@@ -0,0 +1,261 @@
...
+  const threshold = formatThreshold(props.direction, props.thresholdValue, props.thresholdType);
+
+  return (
+    <div className="space-y-1">
+      <div className="text-neutral-10 mb-2 text-xs font-medium">Slack preview</div>
+      <div className="bg-neutral-2 dark:bg-neutral-3 border-neutral-5 rounded-md border p-3">
+        {/* Bot header */}
+        <div className="mb-2 flex items-center gap-2">
+          <div className="bg-accent text-accent flex size-5 items-center justify-center rounded-sm text-[10px] font-bold">
+            H
```
jdolle, line 107: im not sure why there's an H here.

packages/web/app/src/pages/target-alerts-activity.tsx
```diff
@@ -0,0 +1,206 @@
...
+        title="Alert activity"
+        description="Monitor alert firings, recoveries, and state changes."
+      />
+
+      <div className="mt-6">
+        <Filters
+          dimensions={dimensions}
+          pinnedControls={
+            <Select
+              options={[...VIEW_RANGE_OPTIONS]}
```
jdolle:
```suggestion
              options={VIEW_RANGE_OPTIONS}
```

packages/web/app/src/pages/target-alerts-detail.tsx
```diff
@@ -0,0 +1,264 @@
...
+            },
+            to: '/$organizationSlug/$projectSlug/$targetSlug/alerts/rules',
+          }}
+        />
+
+        <PageLead description={buildDescription(rule)} title={rule.name} />
+
+        <div className="flex">
+          <Select
+            options={[...VIEW_RANGE_OPTIONS]}
```
jdolle:
```suggestion
            options={VIEW_RANGE_OPTIONS}
```

jonathanawesome:
> > Plan-gated state-log retention. Hobby/Pro 7d, Enterprise 30d. expires_at is snapshotted at insert time so plan changes only affect new rows.
> 
> This is an interesting decision. If they have 30d retention and change to 7d, then there could be a gap in logs if it's 7-30 days after the change in subscription.
> 
> Long term, I feel like we need an "on subscription change" event that we can do a variety of things with

Good idea. Follow up issue added to Linear triage.

jonathanawesome:
> > Notification fan-out is silent on partial failure. metric-alert-notifier.ts sends to each channel in sequence. If Slack succeeds and the webhook fails, today the failure is logged but doesn't surface to the user...they just get a partial notification. Future observability work will add metrics here, but in the meantime any review feedback on whether we should retry / surface failures more prominently is welcome.
> 
> This is the first followup feature I'd address. Firing exactly once is always a challenge. Many of these notification platforms have idempotency built in so we can fire at least once. But we should definitely not partially fulfil

Follow up issue created!

packages/services/api/src/modules/alerts/providers/metric-alert-rules-storage.ts
jonathanawesome: This was dead code, forgot to remove it from a previous change. Removed and added a comment for future users.

packages/services/api/src/modules/alerts/resolvers/MetricAlertRuleIncident.ts
jonathanawesome: This was a poorly-worded comment. This body is intentionally empty to satisfy codegen.

packages/services/api/src/modules/alerts/resolvers/MetricAlertRuleStateChange.ts
jonathanawesome: `MetricAlertRule` is non-null in the schema. This is a guard against a race condition where a rule might be deleted but there's an in-flight request for the log rows. My thinking here is that the error is surfaced in user land and can be ignored or addressed.

packages/services/api/src/modules/alerts/module.graphql.ts
jonathanawesome: I don't believe so. We have a max of 30 days retention on the state logs as well as a UI requirement to, potentially, show all 30 days of the data in the state transitions bar and the accompanying chart. 

If we decide, in the future, to add an unbounded state log view across all/multiple rules, then we would paginate.

packages/services/api/src/modules/alerts/module.graphql.ts
jonathanawesome: Yeah, this is a great idea. We'll have to add a toggle to the table and another on the detail screen. I'll create a followup Linear issue to track.

packages/services/api/src/modules/commerce/constants.ts
jonathanawesome: Follow up Linear issue created in the overall project.

packages/services/workflows/src/lib/metric-alert-notifier.ts
jonathanawesome: added to content follow up PR

packages/services/workflows/src/tasks/purge-expired-alert-state-log.ts
jonathanawesome: We keep a separate record for incidents that isn't purged. Does that cover your ask?

packages/services/workflows/src/lib/metric-alert-evaluator.ts
jonathanawesome: Follow up Linear issue [created](https://linear.app/the-guild/issue/CONSOLE-2047/share-clickhouse-client-across-services).

packages/services/workflows/src/lib/metric-alert-evaluator.ts
jonathanawesome: For now, no. Follow up Linear issue [created](https://linear.app/the-guild/issue/CONSOLE-2047/share-clickhouse-client-across-services).

packages/web/app/src/components/base/floating/filter-menu/url-filter.ts
jonathanawesome: `undefined` strips the key from the URL entirely.

packages/web/app/src/components/target/alerts/alert-activity-chart.tsx
jonathanawesome: 7 just because it's a good fit for our generally supported resolution. We don't officially support mobile across all screens, so that wasn't a factor.

Point taken though and I've adjusted the tick count based on the series size.

packages/web/app/src/components/target/alerts/alert-event-detail.tsx
jonathanawesome: Returning directly from the backend is a larger piece of work, but fair point regarding the UI keeping up with changes in the API. I changed the objects to typed Records using the type...will fail typecheck.

packages/web/app/src/components/target/alerts/alert-form.tsx
jonathanawesome: GOOD CATCH! This surfaces the deeper issue that also affects the manage filters route. For this PR, the Select that holds these max50 `savedFilters` also has a search box...so for now I think this is OK.

This PR contains new table UI will help on the manage filters route. Follow up PR created to address this.

packages/web/app/src/components/target/alerts/alert-form.tsx
jonathanawesome: It's currently set to the alert range selection (default 7 days)..when the range is changed the chart updates.

Good call re: context. I updated the chart to display 2x the range.

packages/web/app/src/components/target/alerts/alert-notification-preview.tsx
jonathanawesome: This is a placeholder in the Slack alert preview:
Screenshot 2026-05-06 at 7 30 25 PM

package.json
jonathanawesome: It's a significant upgrade to the script that specifically includes alerts seed data which depends on the usage/insights seed data.

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,120 @@
...
+  "last_evaluated_at" timestamptz,
+  "last_triggered_at" timestamptz,
+  "state" "metric_alert_state" NOT NULL,
+  "state_changed_at" timestamptz,
+  "confirmation_minutes" integer NOT NULL,
+  "saved_filter_id" uuid REFERENCES "saved_filters"("id") ON DELETE SET NULL,
+  PRIMARY KEY ("id"),
+  CONSTRAINT "metric_alert_rules_metric_required" CHECK (
+    ("type" = 'LATENCY' AND "metric" IS NOT NULL) OR ("type" != 'LATENCY' AND "metric" IS NULL)
+  )
```
n1ru4l: Should we introduce this constraint on the DB level? Lets say we add a new alert type. We would need to update the constraint which is a blocking operation.

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,120 @@
...
+CREATE TABLE "metric_alert_incidents" (
+  "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+  "metric_alert_rule_id" uuid NOT NULL REFERENCES "metric_alert_rules"("id") ON DELETE CASCADE,
+  "started_at" timestamptz NOT NULL DEFAULT now(),
+  "resolved_at" timestamptz,
+  "current_value" double precision NOT NULL,
+  "previous_value" double precision,
+  "threshold_value" double precision NOT NULL,
+  PRIMARY KEY ("id")
+);
```
n1ru4l, line 80: We could write these to a clickhouse table with TTL instead. That way we do not need a postgres cronjob for cleaning up the database tables

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
```diff
@@ -0,0 +1,120 @@
...
+  "incident_id" uuid REFERENCES "metric_alert_incidents"("id") ON DELETE CASCADE,
+  "from_state" "metric_alert_state" NOT NULL,
+  "to_state" "metric_alert_state" NOT NULL,
+  "value" double precision,
+  "previous_value" double precision,
+  "threshold_value" double precision,
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "expires_at" timestamptz NOT NULL,
+  PRIMARY KEY ("id")
+);
```
n1ru4l, line 102: Same as https://github.com/graphql-hive/console/pull/7935/changes#r3333150762 - though we kinda would loose the consitency benefits.

packages/services/api/src/modules/alerts/resolvers/Mutation/addMetricAlertRule.ts
```diff
@@ -0,0 +1,147 @@
...
+      metric: input.metric ?? null,
+      thresholdType: input.thresholdType,
+      thresholdValue: input.thresholdValue,
+      direction: input.direction,
+      severity: input.severity,
+      name: input.name,
+      confirmationMinutes: input.confirmationMinutes ?? 0,
+      savedFilterId: input.savedFilterId ?? null,
+      channelIds: input.channelIds,
+    });
```
n1ru4l, line 40: All this logic should be within a provider class not the resolver. We keep resolvers as lean as possible.

packages/services/api/src/modules/alerts/resolvers/Mutation/deleteMetricAlertRules.ts
```diff
@@ -0,0 +1,46 @@
...
+  await injector.get(Session).assertPerformAction({
+    action: 'alert:modify',
+    organizationId,
+    params: { organizationId, projectId },
+  });
+
+  const deleted = await injector.get(MetricAlertRulesStorage).deleteMetricAlertRules({
+    projectId,
+    ruleIds: [...input.ruleIds],
+  });
```
n1ru4l: Same as https://github.com/graphql-hive/console/pull/7935/changes#r3333172407

packages/services/api/src/modules/alerts/resolvers/Mutation/updateMetricAlertRule.ts
```diff
@@ -0,0 +1,118 @@
...
+    thresholdType: input.thresholdType ?? undefined,
+    thresholdValue: input.thresholdValue ?? undefined,
+    direction: input.direction ?? undefined,
+    severity: input.severity ?? undefined,
+    name: input.name ?? undefined,
+    confirmationMinutes: input.confirmationMinutes ?? undefined,
+    savedFilterId: input.savedFilterId ?? undefined,
+    enabled: input.enabled ?? undefined,
+    channelIds: input.channelIds ?? undefined,
+  });
```
n1ru4l: Same as https://github.com/graphql-hive/console/pull/7935/changes#r3333172407

packages/services/api/src/modules/alerts/resolvers/MetricAlertRule.ts
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      return await injector.get(TargetManager).getTarget({
+        targetId: rule.targetId,
+        projectId: rule.projectId,
+        organizationId: rule.organizationId,
+      });
+    } catch {
+      return null;
+    }
+  },
```
n1ru4l: Is this resolver ever used?

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -0,0 +1,523 @@
...
+  organizationPlanName: string | null;
+};
+
+type ClickHouseWindowRow = {
+  window: 'current' | 'previous';
+  total: string;
+  total_ok: string;
+  average: number;
+  percentiles: [number, number, number, number];
+};
```
n1ru4l: We use zod for decoding IO. Unless there is a proof that this would become a bottle-neck; please use zod.

packages/services/workflows/src/lib/metric-alert-evaluator.ts
n1ru4l: We have an isUUID function already in the codebase that can be used

packages/services/workflows/src/lib/metric-alert-notifier.ts
```diff
@@ -0,0 +1,218 @@
...
+
+export type NotificationEvent = {
+  state: 'firing' | 'resolved';
+  rule: MetricAlertRuleRow;
+  currentValue: number;
+  previousValue: number;
+  organizationSlug: string;
+  projectSlug: string;
+  targetSlug: string;
+};
```
n1ru4l, line 34: Same as https://github.com/graphql-hive/console/pull/7935/changes#r3333210090

packages/services/workflows/src/tasks/evaluate-metric-alert-rules.ts
```diff
@@ -0,0 +1,226 @@
...
+          'rules.in_group': groupRules.length,
+          time_window_minutes: representative.timeWindowMinutes,
+        },
+      },
+      async span => {
+        try {
+          let windows;
+          try {
+            windows = await queryClickHouseWindows(
+              context.clickhouse!,
```
n1ru4l: Please do not use the `!` oeprator outside test code.

packages/services/workflows/src/tasks/send-metric-alert-channel-notification.ts
```diff
@@ -0,0 +1,258 @@
...
+  channelId: string;
+  channelType: AlertChannelRow['type'];
+  channelName: string;
+  slackChannel: string | null;
+  webhookEndpoint: string | null;
+  // slugs
+  organizationSlug: string;
+  projectSlug: string;
+  targetSlug: string;
+};
```
n1ru4l: https://github.com/graphql-hive/console/pull/7935/changes#r3333210090

packages/services/workflows/src/tasks/send-metric-alert-channel-notification.ts
```diff
@@ -0,0 +1,258 @@
...
+        const event: NotificationEvent = {
+          state: row.toState === 'FIRING' ? 'firing' : 'resolved',
+          rule: {
+            id: row.ruleId,
+            organizationId: row.organizationId,
+            // Only fields the notifier reads. Other MetricAlertRuleRow fields
+            // aren't needed for message formatting and are omitted to keep the
+            // payload narrow.
+            projectId: '',
+            targetId: '',
```
n1ru4l: These attributes should not be passed if they are not needed.

packages/services/workflows/src/environment.ts
```diff
@@ -92,6 +92,20 @@ const RedisModel = zod.object({
...
+  }),
+  zod.object({
+    CLICKHOUSE: zod.literal('1'),
+    CLICKHOUSE_HOST: zod.string(),
+    CLICKHOUSE_PORT: NumberFromString,
+    CLICKHOUSE_USERNAME: zod.string(),
+    CLICKHOUSE_PASSWORD: emptyString(zod.string().optional()),
+    CLICKHOUSE_PROTOCOL: emptyString(zod.string().optional()),
+  }),
+]);
```
n1ru4l, line 107: I assume the reason why we introduce it optional is so that self-hosters do not need to provide it? In that case in the code where we use `!`, lets instead just do an assertion on whether clickhouse exists and/or start service startup if `FEATURE_FLAGS_METRIC_ALERT_RULES_ENABLED` is enabled but the clickhouse config does not exist?

packages/services/workflows/src/index.ts
```diff
@@ -43,17 +56,25 @@ const modules = await Promise.all([
...
+  import('./tasks/purge-expired-alert-state-log.js'),
 ]);
 
 const crontab = `
   # Purge expired schema checks every Sunday at 10:00AM
   0 10 * * 0 purgeExpiredSchemaChecks
   # Every day at 3:00 AM
   0 3 * * * purgeExpiredDedupeKeys
+  # Evaluate metric alert rules every minute
+  * * * * * evaluateMetricAlertRules
```
n1ru4l: This should be added conditionally based on the config?

n1ru4l (commented):
I reviewed the code and added some comments

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
jonathanawesome: Fair point. Validation already happens in both the `add` and `update` resolvers.

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
jonathanawesome: The cron job sweeps for expired `metric_alert_state_log`s, not for `metric_alert_incidents`. My thinking here is that incidents would remain indefinitely for historical recording (`started_at` and `resolved_at` could give us a total incident duration in the a future UI update), but the state logs themselves and their detail are purged per the plan type.

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
jonathanawesome: One issue with moving this out of pg is that we'd lose the benefits of the atomic transaction (update the rule status, write to the incident, notification queue).

Another is that this table is potentially heavy on both the read (15s polling in the UI) and write (one row per state transition, on every cron tick, for every active rule) sides. Would there be a concern about Clickhouse cost there?

packages/services/api/src/modules/alerts/resolvers/MetricAlertRule.ts
jonathanawesome: Nope. Early code...good catch!

packages/migrations/src/actions/2026.04.15T00-00-01.metric-alert-rules.ts
n1ru4l: reading from clickhouse is almost never an issue unless the query is ass. The main reason for vouching for clickhouse here is to avoid us having a cleanup job that needs to run. But I guess it is okay and I agree with the transactional stuff - lets keep things as is and improve later on if it becomes an issue.

packages/services/api/src/modules/alerts/module.graphql.ts
```diff
@@ -154,4 +154,270 @@ export default gql`
...
+    organizationSlug: String!
+    projectSlug: String!
```
n1ru4l: We have the input type `ProjectReferenceInput` that we use in other parts that we could also use here instead. Nothing major - we can still change this after we shipped it as it is private API for now 👍

n1ru4l (approved):

integration-tests/tests/api/project/metric-alerts.spec.ts
n1ru4l, line 1: As a followup, I would like to see some tests that ensure you cannot create alerts for projects of other orgs you are not a member of etc. to avoid regressions

PR #7745, opened 2026-02-26
feat: oidc domain registration and verification 

n1ru4l:
### Background

Closes https://github.com/graphql-hive/console/issues/7735

### Description

This PR addresses the following things:
- Make the OIDC integration management section look better than before
- Allow admins to register and verify domains associated with their organization via a DNS challenge (by querying google and cloudflare for a TXT record)
- If the admins verified a domain, the users attempting to log in via OIDC do not need to verify their email
- Alter the end2end tests to be more like cloud Hive Console by requiring email verification
- Add end2end test to verify the register domain and user flow by using a hard-coded DNS record on `buzzcheck.dev`

A full run through of the PR can be found on on this internal Slack video: https://guild-oss.slack.com/archives/CAY2119MX/p1772221874651489?thread_ts=1772218438.245559&cid=CAY2119MX

Originally I wanted to have https://github.com/graphql-hive/console/issues/7763 as part of the scope of this PR. But it got a bit bigger, so I will do that in a follow up PR.

Here are some screenshots of the functionality, but I recommend to watch the video instead.

image

image

image

### Rollout Strategy

This change is fully backwards-compatible. I will probably manually create records for our known and trusted enterprise users to spare them some time running through the DNS challenge. In case we would need to rollback the pre-existing functionality continues to work. The only regression would be that some people need to verify their email again.

### Checklist

- [x] Input validation
- [x] Output encoding
- [x] Authentication management
- [x] Session management
- [x] Access control
- [x] Error handling and logging
- [x] Testing

PR #7966, opened 2026-04-10
feat: add Azure Federated Workload Identity support for OIDC SSO login

cosmincatalin:
### Background

When deployed in Kubernetes with Azure Workload Identity, the OIDC SSO integration should be able to authenticate to the identity provider (e.g., Microsoft Entra ID) using a federated token (`client_assertion`) instead of storing a `client_secret`. This eliminates the need for secrets management for the OAuth2 token exchange.

### Description

Adds a `useFederatedCredential` option to OIDC integrations. When enabled, the token exchange uses the `AZURE_FEDERATED_TOKEN_FILE` (injected by the Azure Workload Identity webhook) as a `client_assertion` JWT bearer instead of `client_secret`.

- **DB migration**: `use_federated_credential` boolean column on `oidc_integrations`, `client_secret` made nullable
- **Storage/Entity**: `OIDCIntegration.encryptedClientSecret` now `string | null`, added `useFederatedCredential: boolean`
- **GraphQL schema**: `clientSecret` optional in `CreateOIDCIntegrationInput`, `useFederatedCredential` field on type + inputs
- **OIDC provider**: Validates that either `clientSecret` or `useFederatedCredential` is provided on create
- **Auth server token exchange** (`supertokens-at-home.ts`): Reads federated token file and sends `client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer` + `client_assertion=<token>` instead of `client_secret`
- **UI**: Shows federated credential status badge, handles nullable `clientSecretPreview`

Token exchange with federated credentials:
```
POST /token
  grant_type=authorization_code
  client_id=<app-registration-client-id>
  client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer
  client_assertion=<contents-of-AZURE_FEDERATED_TOKEN_FILE>
  code=<auth-code>
  code_verifier=<pkce-verifier>
```


packages/services/server/src/supertokens-at-home.ts
```diff
@@ -1304,20 +1306,49 @@ export async function registerSupertokensAtHome(
...
+            return rep.status(200).send({
+              status: 'SIGN_IN_UP_NOT_ALLOWED',
+              reason: 'Sign in failed. Please contact your organization administrator.',
+            });
+          }
+
+          const { readFileSync } = await import('node:fs');
+          const federatedToken = readFileSync(tokenFilePath, 'utf-8');
+          grantParams['client_assertion_type'] =
+            'urn:ietf:params:oauth:client-assertion-type:jwt-bearer';
```
n1ru4l: Can you please move the token file reading to the `environment.ts` file and model the configuration in a way that would allow also configuring other providers in the future (for self-hosters)?

E.g. the environment variable could look like the following:

```
OIDC_WORKLOAD_FEDERATION_IDENTITY_PROVIDER="azure"
```

Which then implies that we have to setup some job that polls the path at that env[`AZURE_FEDERATED_TOKEN_FILE`] in intervals?

Also I think it would be safe to fail service startup if `OIDC_WORKLOAD_FEDERATION_IDENTITY_PROVIDER` is provided but `AZURE_FEDERATED_TOKEN_FILE` is not set to a file path or cannot be read?

Attempting to read the `tokenFilePath` every time a request is authenticated seems a lot, so maybe we can have a container class/service that just polls or watches the file path?

cosmincatalin:
@n1ru4l, with #7745 in, can you look into merging this one in?

packages/services/server/src/supertokens-at-home.ts
```diff
@@ -1304,20 +1306,49 @@ export async function registerSupertokensAtHome(
...
+        ) {
+          // Azure Workload Identity: use client_assertion with the federated token file
+          const tokenFilePath = env.azureFederatedTokenFile;
+          if (!tokenFilePath) {
+            req.log.error('AZURE_FEDERATED_TOKEN_FILE environment variable is not set');
+            broadcastLog(
+              oidcIntegration.id,
+              'Federated credential is configured but AZURE_FEDERATED_TOKEN_FILE environment variable is not set. ' +
+                'Ensure the pod has Azure Workload Identity configured.',
+            );
```
n1ru4l, line 1329: This is an internal implementation detail that should not be leaked to the outside world.

packages/migrations/src/actions/2026.04.09T00-00-00.oidc-federated-credential.ts
```diff
@@ -0,0 +1,12 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2026.04.09T00-00-00.oidc-federated-credential.sql',
+  run: ({ psql }) => psql`
+    ALTER TABLE "oidc_integrations"
+      ADD COLUMN IF NOT EXISTS "use_federated_credential" BOOLEAN NOT NULL DEFAULT FALSE;
+
+    ALTER TABLE "oidc_integrations"
+      ALTER COLUMN "client_secret" DROP NOT NULL;
```
n1ru4l, line 10: NIT: this can be combined into a single statement

n1ru4l (commented):
Thank you for this contribution! I added some initial comments with an initial review, it would be great if you could address these. We cannot give an exact timeline yet, but we will try to test this on our cluster and see if it feasible to merge and support soon.

packages/migrations/src/actions/2026.04.09T00-00-00.oidc-federated-credential.ts
cosmincatalin: Hi @n1ru4l , I believe that your suggested changes are in place now. Let me know how it looks and if you want me to make some other changes.

I haven't marked your comments as _Resolved_ to make it easier for you to compare with your previous feedback.

n1ru4l (changes requested):
I have been thinking more about Federation Workload Identity and how we want to support it in the future as a Hive Console Cloud feature that we provide all users, which means at some point Hive Console comes with its own built in OIDC discovery and JWKS endpoint, with assigned subjects for each organization.

_____

However, I still see in value in this solution as it solves a problem today and is much simpler for self-hosters, without the complexity that the above direction requires.

I have the following feedback and changes I would like to see.

##  Set the organization for which the file can be used for via an environment variable

Instead of the current approach, where every org can opt-into using the token file, I would prefer this to be more explicit via required configuration.

To me it feels a bit unsettling, that any organization existing on the Hive Console deployment could use this identity file.

My proposal is that new environment variable `OIDC_WORKLOAD_FEDERATION_ORGANIZATION_ID`, would isolate which exact organization can use the Azure Workload Identity.

```
OIDC_WORKLOAD_FEDERATION_IDENTITY_PROVIDER=azure
OIDC_WORKLOAD_FEDERATION_ORGANIZATION_ID=<organization-uuid>
AZURE_FEDERATED_TOKEN_FILE=/var/run/secrets/azure/tokens/azure-identity-token
```

All these three values need to be set during startup.

Startup should fail when:
- The organization ID is malformed
- The file is unreadable
- The file is empty

## Remove the `use_federated_credential` column from pg

Instead whether a org uses federation credentials is read from the configuration.

## Remove the nullability for the `CreateOIDCIntegrationInput.clientSecret` (for now)

Since this is for now a very specific implementation for certain self-hosting environments, I do not (yet) want this to affect the existing API contract we have in Cloud (until we have a solution that also works for our Cloud deployment).

Instead, I think it is an okay tradeoff to require passing a dummy value (which is then never used).

_____

@cosmincatalin These changes would make it comfortable enough for me to accept this contribution. Let me know if you have additional thoughts or pushback on these.

cosmincatalin:
Based on your suggestions, @n1ru4l , I have greatly simplified the PR, to the point where it's way easier to review.

Instead of being able to provide a single _organization_ that can use federation workloads, I thought it's appropriate to be able to do it for any number, by providing a comma separated list of organizations UUIDs in `OIDC_WORKLOAD_FEDERATION_ORGANIZATION_IDS`.

Earlier migrations, UI code and other complexities have been removed.

Let me know if this is fine.

n1ru4l (approved):
Nice thank you for the patience and the implementation! 🙏

cosmincatalin:
> Nice thank you for the patience and the implementation! 🙏

Can you approve the workflows again? I've missed your approval and I had to update the branch, therefor not being able to merge anymore

n1ru4l:
@cosmincatalin Could you add a changeset? I tried to push it here, but I do not have permissions for your fork.

````
---
'hive': patch
---

Add Azure Workload Identity Federation support for OIDC SSO in self-hosted deployments. Enabled
organizations use the projected Azure token as a client assertion instead of an OIDC client secret.

For example, enable federation for an organization with:

```env
OIDC_WORKLOAD_FEDERATION_IDENTITY_PROVIDER=azure
AZURE_FEDERATED_TOKEN_FILE=/var/run/secrets/azure/tokens/azure-identity-token
OIDC_WORKLOAD_FEDERATION_ORGANIZATION_IDS=00000000-0000-4000-8000-000000000000
```

This will effectively override the client secret configured for that organization to use the OIDC provider.
````

cosmincatalin:
Done, @n1ru4l . Never used this changeset approach before. I like it.

PR #7970, opened 2026-04-10
fix(app-deployments): auto-inject MCP directive definitions during operation validation

adambenhassen:
### Summary

When pushing persisted operations that use MCP directives (`@mcpTool`, `@mcpDescription`, `@mcpHeader`) via app deployments, validation fails because the target schema doesn't define these directives. Users shouldn't need to manually add MCP directive definitions to their schema just to use them in operations, the gateway plugin handles them at runtime.

This fix auto-injects the canonical MCP directive definitions into the schema used for validation in the persisted document ingester. It strips any existing (potentially outdated) definitions first, then appends the current ones.

### Changes made
- Purely additive: collects existing directive and scalar names from the AST, then only injects definitions that are missing. User-defined `@mcpTool` (via `@composeDirective`, `@link`, etc.) is left untouched.
- No-op fast path: if all directives already exist, returns the original SDL string unchanged (no parse-print round-trip on the schema).
- `MCP_DIRECTIVES_DOC` parsed once at module load, avoids re-parsing the constant SDL on every batch.

### Tests added
- add documents to app deployment succeeds with MCP directives even when schema does not define them
- add documents to app deployment succeeds with MCP directives when schema already defines them

### DIFF
--- a/packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
+++ b/packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
@@ -47,11 +47,11 @@ const MCP_DIRECTIVES_DOC = parse(/* GraphQL */ `
     description: String
     title: String
     descriptionProvider: String
-    meta: JSON
+    meta: _HiveMCPJSON
   ) on QUERY | MUTATION
   directive @mcpDescription(provider: String!) on VARIABLE_DEFINITION | FIELD
   directive @mcpHeader(name: String!) on VARIABLE_DEFINITION
-  scalar JSON
+  scalar _HiveMCPJSON
 `);
 
 /**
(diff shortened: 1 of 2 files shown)


packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
```diff
@@ -30,6 +39,56 @@ const AppDeploymentOperationHashModel = z
...
+      existingNames.add(def.name.value);
+    }
+  }
+
+  const missing = MCP_DIRECTIVES_DOC.definitions.filter(def => {
+    if (def.kind === Kind.DIRECTIVE_DEFINITION || def.kind === Kind.SCALAR_TYPE_DEFINITION) {
+      return !existingNames.has(def.name.value);
+    }
+    return false;
+  });
```
n1ru4l: NIT: instead of adding all scalars and types to the set, we just track if we found the types we are looking for, bit more memory efficient - but probably not a big deal 😆

n1ru4l (commented):
legit, added a nit comment but this is fine as is for me

n1ru4l (approved):

packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
adambenhassen: That's true.. hmm.. I'll invert it instead

PR #7972, opened 2026-04-13
fix sentry sourcemaps: point the inject and upload commands at dist/client instead of dist

jonathanawesome:
This PR fixes Sentry sourcemap resolution for the web app by adopting `@sentry/vite-plugin` for client-side sourcemaps. Previously, `sentry-cli sourcemaps inject` ran after the Docker build, so deployed JS files had no debug IDs and the fallback URL matching failed because files were uploaded under `~/client/assets/`... while the browser requests `~/assets/`... The Vite plugin embeds debug IDs at build time (before Docker copies the files), making URL-based matching unnecessary. This PR also adds a separate server sourcemap upload with `--dist app` to match the server SDK init, which was never working before.

### DIFF
--- a/packages/web/app/vite.config.ts
+++ b/packages/web/app/vite.config.ts
@@ -3,6 +3,7 @@ import type { Plugin, UserConfig } from 'vite';
 import monacoEditor from 'vite-plugin-monaco-editor';
 import tsconfigPaths from 'vite-tsconfig-paths';
 import viteFastify from '@fastify/vite/plugin';
+import { sentryVitePlugin } from '@sentry/vite-plugin';
 import tailwindcss from '@tailwindcss/vite';
 import react from '@vitejs/plugin-react';
 
@@ -31,6 +32,11 @@ export default {
     react(),
     tailwindcss(),
     reactScanPlugin,
+    sentryVitePlugin({
+      org: process.env.SENTRY_ORG,
+      project: process.env.SENTRY_PROJECT,
+      release: { name: process.env.RELEASE, dist: 'webapp' },
+    }),
     // @ts-expect-error temp
     monacoEditor.default({
       languageWorkers: ['json', 'typescript', 'editorWorkerService'],
--- a/scripts/upload-sourcemaps.sh
+++ b/scripts/upload-sourcemaps.sh
@@ -15,7 +15,7 @@ for dir in packages/services/*/dist; do
   fi
 done
 
-pnpm sentry-cli sourcemaps inject packages/web/app/dist
-pnpm sentry-cli sourcemaps upload --release=$SENTRY_RELEASE packages/web/app/dist --dist webapp --url-prefix ~/
+pnpm sentry-cli sourcemaps inject packages/web/app/dist/client
+pnpm sentry-cli sourcemaps upload --release=$SENTRY_RELEASE packages/web/app/dist/client --dist webapp --url-prefix ~/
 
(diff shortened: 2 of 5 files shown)


scripts/upload-sourcemaps.sh
jdolle, line 1: There's also a plugin that we could potentially use to simplify the configuration -- https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/vite/

But I think you're right that there's a mismatch in the path. However, the source map is located in `packages/web/app/dist/`

I _think_ rather than adjusted the upload path, you want to use the command argument `--strip-prefix` to remove the `client` prefix from the local file path so that the url we serve from matches

scripts/upload-sourcemaps.sh
```diff
@@ -15,7 +15,9 @@ for dir in packages/services/*/dist; do
   fi
 done
 
-pnpm sentry-cli sourcemaps inject packages/web/app/dist
-pnpm sentry-cli sourcemaps upload --release=$SENTRY_RELEASE packages/web/app/dist --dist webapp --url-prefix ~/
+# Client sourcemaps handled by @sentry/vite-plugin during build
+
+# Server sourcemaps (dist: app, matching backend.ts SDK init)
+pnpm sentry-cli sourcemaps upload --release=$SENTRY_RELEASE packages/web/app/dist --dist app --url-prefix /usr/src/app/\@hive/app --ignore client
```
jdolle, line 21: So this uploads all the other source maps, but no longer web-app (aka `client`). Nice :)

jdolle (commented):
This looks reasonable to me. Maybe ideally we switch to using the plugin on all our services as well. But for now, let's be sure to confirm:

1. local development and local building is still fine and doesn't error due to not being able to upload because of a bad key or anything
2. deploying works -- you can specify the branch sha in the deployment repo to try it out before merging

jonathanawesome:
> This looks reasonable to me. Maybe ideally we switch to using the plugin on all our services as well. But for now, let's be sure to confirm:
> 
> 1. local development and local building is still fine and doesn't error due to not being able to upload because of a bad key or anything
> 2. deploying works -- you can specify the branch sha in the deployment repo to try it out before merging

Confirmed that local dev and build behave as expected. The plugin only runs on build and silently fails if there's no `SENTRY_AUTH_TOKEN`.

Deploy was successful as well.

Worth exploring changes to services at some point, but that's a bit of an effort because they're not on vite.

jdolle (approved):

packages/web/app/vite.config.ts
```diff
@@ -31,6 +32,11 @@ export default {
     react(),
     tailwindcss(),
     reactScanPlugin,
+    sentryVitePlugin({
+      org: process.env.SENTRY_ORG,
+      project: process.env.SENTRY_PROJECT,
+      release: { name: process.env.RELEASE, dist: 'webapp' },
+    }),
```
n1ru4l, line 39: This only affects build or also dev mode? If so does that potentially slow down the dev experience?

n1ru4l (approved):
Approved with additional question

packages/web/app/vite.config.ts
jonathanawesome: Two things. The plugin 1) checks for `SENTRY_AUTH_TOKEN` and fails silently without it and 2) uses `vite build`-only hooks that don't get called during `dev`.

PR #7978, opened 2026-04-14
Support type extensions with unreachable types policy rule

jdolle:
### Background

The policy schema is a merged version of the subgraph schemas. It contains directives and type extensions as it is not yet converted to an API schema... (Confusing, right?)

Since this can contain type extensions, and linting runs against it, the linting rules must handle type extensions

Internal ref: https://linear.app/the-guild/issue/GW-594/investigate-unreachable-types-policy-issue-in-hive

### Description

This change adds support for type extensions with the unreachable types rule by ensuring it visits these extension nodes.

### Checklist

- [X] Testing


n1ru4l (approved):
add a changeset before merging 👍

PR #7979, opened 2026-04-15
chore: remove apollo-router-hive-fork from this repo

dotansimha:
Apollo-Router fork has been moved to this repository: https://github.com/graphql-hive/router/pull/723 , so we can remove it from Console repo.

PR #8011, opened 2026-04-25
Fix insights stat grid displaying incorrectly

jonathanawesome:
This PR fixes an overly-broad `target` rule in `.gitignore` (no leading slash) that was matching `packages/web/app/src/components/target/`, causing Tailwind's Oxide scanner to silently skip 57 files in that directory. 

Classes only used in those files (like `lg:grid-cols-4`) stopped landing in the production CSS, breaking the insights stats grid at ≥1024px. Git itself ignored the rule for already-tracked files, so git status looked clean and nobody noticed. Tailwind doesn't have that exemption.

The bad rule came from PR #7979 , which replaced `/target` with `target` while removing apollo-router. The duplicate Cargo.lock lines below it are appended on every release by a workaround in `release-stable.yaml` that's no longer needed (the Rust crates that justified it are gone).


.gitignore
```diff
@@ -140,8 +140,3 @@ resolvers.generated.ts
 docker/docker-compose.override.yml
 
 test-results/
-
-target
```
jdolle, line 144: The `/target` contains numerous built rust files.
We want to still ignore that directory.

```suggestion
/target
```

scripts/compress/src/main.rs
```diff
@@ -1,39 +0,0 @@
-extern crate flate2;
```
jdolle, line 1: why delete this?

.gitignore
jonathanawesome: #7979 removed the only code building rust. @dotansimha , can you confirm?

scripts/compress/src/main.rs
jonathanawesome: This code was solely in support of the apollo-router bits that were removed in #7979. I also see a `cargo:fix` script in `package.json` now that I believe we can also remove. Once again, awaiting review from @dotansimha here.

.gitignore
dotansimha: yeah i don't think it's in use, LGTM

scripts/compress/src/main.rs
dotansimha: yeah not needed

dotansimha (approved):

n1ru4l (approved):

PR #8014, opened 2026-04-27
fix: show correct error code for introspection failure in dev command

### LINKED ISSUE #8006: [CLI] Apollo server introspection failure yields wrong status code

n1ru4l:
### Background

Closes https://github.com/graphql-hive/console/issues/8006

### Description

The dev command did fail with API error instead of an introspection error in case a service can not be introspected. This resolves that scenario, improves the error message to include the service name and adds tests to avoid regressions

### DIFF
--- a/integration-tests/tests/cli/introspect.spec.ts
+++ b/integration-tests/tests/cli/introspect.spec.ts
@@ -2,10 +2,11 @@ import { AddressInfo } from 'node:net';
 import { parse } from 'graphql';
 import { createSchema, createYoga } from 'graphql-yoga';
 import { buildSubgraphSchema } from '@apollo/subgraph';
+import { useDisableIntrospection } from '@graphql-yoga/plugin-disable-introspection';
 import { createServer } from '@hive/service-common';
 import { introspect } from '../../testkit/cli';
 
-async function createHTTPGraphQLServer() {
+export async function createHTTPGraphQLServer() {
   const server = await createServer({
     sentryErrorHandler: false,
     log: {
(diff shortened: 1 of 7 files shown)


integration-tests/tests/cli/introspect.spec.ts
```diff
@@ -259,3 +288,13 @@ test.concurrent('can introspect protected federation with header', async ({ expe
...
 });
+
+test.concurrent('error handling on server with no introspection enabled', async ({ expect }) => {
+  const server = await createHTTPGraphQLServer();
+  const command = introspect([server.url + '/graphql-federation-no-introspection']);
+  await expect(command).rejects.toThrowError(
+    'Could not get introspection result from the service.',
+  );
+  await expect(command).rejects.toThrow('[116]');
+  await expect(command).rejects.not.toThrow('[115]');
```
jdolle, line 299: seems redundant after the previous check

packages/libraries/cli/src/helpers/graphql-request.ts
```diff
@@ -93,6 +94,9 @@ export function graphqlRequest(config: {
         if (jsonData.errors[0].message === 'Invalid token provided') {
           throw new InvalidRegistryTokenError();
         }
+        if (isIntrospectionDisabledError(jsonData.errors[0])) {
+          throw new IntrospectionError();
```
jdolle, line 98: The IntrospectionError message: `Make sure introspection is enabled by the server.` makes it seem like this could be the issue. But here we've been able to identify that introspection is definitely disabled. Could we maybe make this message more clear? Or do we feel it's clear enough from the logging of the original error?

jdolle (approved):

packages/libraries/cli/src/helpers/graphql-request.ts
n1ru4l: Hmm, I think the original error should make that clear.

PR #8015, opened 2026-04-27
feat: write app deployment manifest to CDN

n1ru4l:
### Background

By knowing the app deployment name and version it is currently impossible to retrieve a list of all documents that belong to the version via the CDN.

- [RFC: App Deployment Manifest](https://guild-oss.slack.com/docs/TAYJ1FSUA/F0ASURR7CQ1)

### Description

Write a app deployment manifest to the CDN that contains all the document hashes and whether the app deployment is enabled or disabled.

The endpoint `/apps/:appname/:appVersion` returns a JSON manifest with the following information:

```json5
{
  id: "some uuid",
  isActive: false,
  appName: "app-name",
  appVersion: "app-version",
  // list of all hashes this app version contains
  documentHashes: ["aaa", "bbb", "ccc", "ddd"],
}
```

This enabled routers to pre-load documents of app deployments and the router MCP plugin to load all available tools for an app deployment version.

### Out of Scope

This pull request does not introduce **backfilling app deployment manifests** for existing published app deployments.

- We can introduce this later on before we start adjusting gateway/router to leverage this endpoint
- For now this manifest is only needed in the context of the MCP gateway plugin, where this trade-off is acceptable

### Checklist

- [x] Input validation
- [x] Output encoding
- [x] Access control
- [x] Testing

### DIFF
--- a/integration-tests/tests/api/app-deployments.spec.ts
+++ b/integration-tests/tests/api/app-deployments.spec.ts
@@ -2217,6 +2217,18 @@ test('app deployment manifest is written to and accessible via CDN', async () =>
   }).then(res => res.expectNoGraphQLErrors());
   expect(addDocumentsToAppDeployment.error).toBeNull();
 
+  const cdnAccess = await createCdnAccess();
+  const persistedOperationUrl = `${cdnAccess.cdnUrl}/apps/app-name/app-version`;
+
+  let response = await fetch(persistedOperationUrl, {
+    method: 'GET',
+    headers: {
+      'X-Hive-CDN-Key': cdnAccess.secretAccessToken,
+    },
+  });
+  // before the app deployment is activated it shall not exist.
+  expect(response.status).toEqual(404);
+
   const { activateAppDeployment } = await execute({
     document: ActivateAppDeployment,
     variables: {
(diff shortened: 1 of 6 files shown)


packages/services/cdn-worker/src/artifact-handler.ts
```diff
@@ -461,9 +467,6 @@ export const createArtifactRequestHandler = (deps: ArtifactRequestHandler) => {
         const response = createResponse(
           analytics,
           result.body,
-          // We're using here a public location, because we expose the Location to the end user and
-          // the public S3 endpoint may differ from the internal S3 endpoint. E.g. within a docker network.
-          // If they are the same, private and public locations will be the same.
```
n1ru4l, line 466: no idea what this comment means 😅

dotansimha (approved):
Nice, LGTM! 

One question: this will only be available for app deployment that will be activated after deploying this change, right? 
Do you see a value in back-filling manifests or it's not needed?

n1ru4l:
@dotansimha see the description: 😅 

image

I think we can be fine with shipping this now and then re-evaluate based on actual usage (all current app deployment versions could be already retired by then) whether we need to backfill once we update router/gateway to use this new manifest endpoint.

integration-tests/tests/api/app-deployments.spec.ts
```diff
@@ -2157,6 +2157,125 @@ test('app deployment usage reporting', async () => {
...
+        appName: 'app-name',
+        appVersion: 'app-version',
+      },
+    },
+    authToken: token.secret,
+  }).then(res => res.expectNoGraphQLErrors());
+  expect(activateAppDeployment.error).toBeNull();
+
+  const cdnAccess = await createCdnAccess();
+  const persistedOperationUrl = `${cdnAccess.cdnUrl}/apps/app-name/app-version`;
```
jdolle: can you fetch before activating?

jdolle (approved):
I'd like to see a test covering what happens before activating the app deployment but otherwise looks great

n1ru4l:
@jdolle The manifest does not exist before that. BUt it makes sense to add a test for it!

Issue #8017, opened 2026-04-28
[CLI] Improve hive dev errors


n1ru4l:
The error for the `hive dev` and `hive introspect` commands can be misleading.

Both commands partially rely on introspecting federation subgraphs (via the `Query._service` field) or normal GraphQ introspection (via the GraphQL introspection query).

There is many error cases that could occur:
- GraphQL introspection is disabled
- The target is expected to be a federation subgraph but is not (`Query._service` is missing)
- Network errors etc.

We need to revise the whole flow and errors and need to make sure it is obvious whether an error originates from HTTP calls/responses to the subgraph or our API (in case of the `dev` command).

Right now we have mixed logic and error classes and it can happen that something that an error originating from a subgraph is displayed as a Hive Console API error which is misleading.

In order to solve this technically, we first need to analyze which things can go wrong for each flow and define all the errors for those flows.

PR #8018, opened 2026-04-28
fix: federation subgraph introspection with disabled GraphQL introspection

n1ru4l:
### Background

This change allows introspecting Apollo Federation subgraphs for the `dev` and `introspect` command, even if GraphQL introspection is disabled on the server.

### Description

For the `introspect` command the type must be explicitly passed e.g. `hive introspect --type federation`. Otherwise, we attempt to check whether the subgraph contains the `_Service` via GraphQL introspection to figure out whether we can do "Federation-style" introspection.

For the `dev` command we always assume that the target services are Federation.

In both cases an error is being raised if the target service is not a federation subgraph (but expected to be one).

**Note**: This is a quick solution to unblock one of our customers. However, we should revise the whole flow as stated over here: https://github.com/graphql-hive/console/issues/8017

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [x] Testing


packages/libraries/cli/src/helpers/schema.ts
```diff
@@ -217,66 +218,84 @@ class FederationSubgraphUrlLoader implements Loader {
...
+
+    this.logger?.debug?.('Identify whether the service is a Federation subgraph.');
+
+    this.logger?.debug?.('Attempt "_Service" type lookup via "Query.__type".');
+
+    // We can check if the schema is a subgraph by looking for the `_Service` type.
+    const isSubgraph = await client.request({
+      operation: parse(/* GraphQL */ `
+        query ${'LookupService'} {
+          __type(name: "_Service") ${' '}{
```
jdolle: The point of this loader is that it performs both the federation subgraph lookup and then, if that fails because the service is not a federated subgraph, it falls back to introspection.
For the same reasons the _type check was removed from the federation loader, we should not query _type here.

Instead, we should catch the introspection error and fallback

jonathanawesome (approved):

Issue #7821, opened 2026-03-11
schema variants/ graph feature flags


n1ru4l:
# Problem Statement

Developers create pull requests with service and schema changes and deploy these services ad-hoc to their environment for internal testing/staging purposes. These are never exposed to actual customers/users of their product.

They want their Federation Gateway being able to serve a variant of the default supergraph that is composed with the service(s) schema(s) from one or multiple PR(s) instead of the production service schema(s).

Today, we offer no such feature that allows ad-hoc composition and serving this ad-hoc supergraph to the router for resolution.

The client that wants to use the ad-hoc version of the supergraph will usually indicate so via a header or cookie within the request to the gateway (e.g. Cookie: feature_flag=pr-69).

As a workaround customers build their own layer into the gateway for ad-hoc pulling graphql schemas from the hive schema registry and composition in memory on the gateway for serving the schema based on such a header.

# Proposed Solution

## Hive Console: Workflow for using CLI/API for creating Ad-Hoc Deployments

The user publishes a variant subgraph to the schema registry. This variant subgraph publish is a write that does not mark the subgraph as the latest subgraph version; and it does not trigger a supergraph composition and a resulting new schema version.

This can be done via a CLI command:

```bash
hive schema:publish-service \
  --target the-guild/hive/production \
  --service.name "<service_name>" \
  --service.tag "<service_tag>" \ # unique identifier for the service "version"
  --service.url "<service_url>" \
  --service.schema "./schema.graphqls" \
```

Then the user defines a feature flag where he overrides some overriden services based on the previously published variant subgraphs.

```bash
hive feature-flag:create \
  --target the-guild/hive/production \
  # the services are referenced by the previous publish command
  --service "<service_name>@<service_tag>" \
  --service "<service_name>@<service_tag>" \
  "<flag_name>"
```

After the feature flag was defined, this results in a schema composition that updates the CDN with the information:

- After the feature flag was created, we start composing a new supergraph variant.
    
    The input for these composition is:
    
    - The main registry state (all services)
    - feature flag service overrides
- Each publish to the main schema registry, will result in also attempting a composition of the feature flag supergraph
- Similar to the main supergraph, we publish the (successful) composition result to the CDN at the following key:
    - `https://cdn.graphql-hive.com/artifacts/v1/<target_id>/feature-flags/<flag_name>/supergraph` (and other artifacts sdl; services etc.)
- Upon the first successful composition the feature flag is also added to the list of active feature flags on the CDN:
[...]

Issue #4021, opened 2024-02-21
Promote schema version from one target to another target


n1ru4l:
Right now you need to run `schema:publish <pointer-to-graphql-schema>`, for every target.
As an alternative we are thinking about just having the option of promoting the schema version from one target to the schema version of another target.

e.g. `hive schema:promote <id>`.

This would have the benefit of not requiring access to the graphql schema when mirroring a stage in a multi-environment deployment. E.g. when progressing a docker image from one one stage to another and the flow is not 1:1 backed by a version control system such as Git.

We need to figure out whether the `<id>` is something a Hive users pass along as a `hive schema:publish` parameter that needs to be unique across all publishes within a project, or whether this is something we provide as an output when running `schema:publish`, that the Hive user needs to store to later on pass it as a parameter for `hive schema:promote <id>`.

**Example:**

| Git Branch | Hive Target |
| ---------- | ----------- |
| main       | development |
| staging    | staging     |
| production | production  |

New features are developed on a branch that targets the `main` branch. The `hive schema:check`
command is run against the `development` target. conditional breaking changes are configured on
`staging` and `production` targets.

When the feature is ready for QA, the branch is merged into `main`. The `hive schema:publish`
command is triggered by CD workflow, and the schema is published to the `development` target.

When the feature is ready for staging, the `main` branch is merged into `staging`. The
`hive schema:promote` command is triggered by the CD workflow, and the schema version is published to the
`staging` target.

When the feature is ready for production, the `staging` branch is merged into `production`. The
`hive schema:promote` command is triggered by the CD workflow, and the schema published to the
`production` target.

srdjan-bestic-ananas:
Hi, in lack of this functionality, is there any other way to copy over a schema from target to target? For example, when bootstrapping a new or temp environment it would be super useful to be able to clone a schema

jdolle:
If I'm understanding it correctly, it sounds like there are two use cases:
1. clone the entire target to another in order to duplicate the entire environment
2. provide a cleaner approach to promoting a single subgraph through the different stages/targets of a deployment pipeline

#### Promoting a Target

One complication I see is that a registry token is bound to a specific target. So I think we're forced to set the promotion target via the target's settings in the UI. Otherwise, if we allow promoting to any target via the CLI then one registry token would be able to update every other target's schema and that seems like a security concern.

For the command, I was thinking:
```
HIVE_TOKEN=*** hive schema:promote <service> <commitId>
```

[...]

PR #8029, opened 2026-05-07
chore: move all schema version crud logic from legacy storage module to dedicated class

n1ru4l:
### Background

Storage module toooo big; we want to add more logic due to graph variants (https://github.com/graphql-hive/console/issues/7821) and promotion (https://github.com/graphql-hive/console/issues/4021); thus we break it down

### Description

Introduces a new class `SchemaVersionStore` and moves all schema version related crud operations from the legacy `Storage` class over there.


packages/services/api/src/modules/organization/providers/resource-selector.ts
```diff
@@ -163,7 +165,9 @@ export class ResourceSelector {
     if (target.type === GraphQLSchema.ProjectType.SINGLE) {
       return null;
     }
-    const latest = await this.storage.getMaybeLatestValidVersion({ targetId: target.targetId });
+    const latest = await this.schemaVersions.getMaybeLatestSchemaVersionForTargetId(
+      target.targetId,
+    );
```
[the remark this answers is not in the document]
n1ru4l: any latest is fine here, we only want to get the services associated with the target to give something to select to the user.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -1493,15 +1495,11 @@ export class SchemaPublisher {
         if (deleteResult.conclusion === SchemaDeleteConclusion.Accept) {
           this.logger.debug('Delete accepted');
           if (input.dryRun !== true) {
-            const schemaVersion = await this.storage.deleteSchema({
-              organizationId: selector.organizationId,
-              projectId: selector.projectId,
-              targetId: selector.targetId,
+            const schemaVersion = await this.schemaVersions.deleteSubgraphFromTarget(target, {
```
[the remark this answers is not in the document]
n1ru4l: this was never used and is a relict of the past

packages/internal/postgres/src/psql.ts
```diff
@@ -18,6 +18,20 @@ function psqlFn(template: TemplateStringsArray, ...values: ValueExpression[]) {
   return tag.type(z.unknown())(template, ...values);
 }
 
-Object.assign(psqlFn, createSqlTag());
```
jdolle, line 21: We don't need to assign `createSqlTag` anymore?

packages/services/api/src/modules/schema/providers/schema-manager.ts
```diff
@@ -336,17 +329,20 @@ export class SchemaManager {
...
 
-    const result = await this.storage.getMaybeVersion(selector);
+    const version = await this.schemaVersions.getSchemaVersionById(selector.versionId);
+
+    if (!version) {
+      return null;
+    }
 
-    if (!result) {
+    if (version.targetId !== selector.targetId) {
```
jdolle, line 338: My understanding here is that we're moving the ownership validation out of the storage function, and into the provider/manager. I assume this is to let us to fetch resources more easily without passing org, project, target into every storage function. However, if we need to update a resource, we'll still need to pass in this data to avoid updating resources that don't belong to the org/etc. So we may not be getting the benefit that we intended here.

Are there other reasons for this design change?

packages/services/api/src/modules/schema/providers/schema-version-store.ts
```diff
@@ -0,0 +1,1117 @@
...
+            Array<{ name: string; content: string; source: string | null }>
+          >;
+          metadataAttributes: null | Record<string, string[]>;
+        }
+    ),
+  ): Promise<SchemaVersion> {
+    const url = args.url ?? null;
+    const service = args.service ?? null;
+
+    const output = await this.pg.transaction('createVersion', async trx => {
```
jdolle: minor: prefer "createSchemaVersion"

packages/services/api/src/modules/schema/providers/schema-version-store.ts
```diff
@@ -0,0 +1,1117 @@
...
+          metadataAttributes: null | Record<string, string[]>;
+        }
+    ),
+  ) {
+    return this.pg.transaction('deleteSubgraphFromTarget', async trx => {
+      // fetch the latest version
+      const latestVersion = await trx
+        .maybeOne(
+          psql`/* findLatestSchemaVersion */
+          SELECT sv.id, sv.base_schema
```
jdolle: minor: `as baseSchema` for consistency?

packages/services/api/src/modules/schema/providers/schema-version-store.ts
```diff
@@ -0,0 +1,1117 @@
...
+      return { serviceName: after.service_name, after: after.sdl, before: null };
+    }
+
+    const before = await this.pg
+      .maybeOne(
+        psql`/* getMatchingServiceSchemaOfVersions */
+        SELECT sl.sdl
+        FROM schema_version_to_log as svtl
+        LEFT JOIN schema_log as sl ON svtl.action_id = sl.id
+        WHERE svtl.version_id = ${versions.before} AND sl.service_name = ${after.service_name}
```
jdolle, line 956: should maybe use lowercase for `service_name` to be safe

jdolle (approved):

packages/services/api/src/modules/schema/providers/schema-version-store.ts
n1ru4l:
```suggestion
    const output = await this.pg.transaction('createSchemaVersion', async trx => {
```

packages/internal/postgres/src/psql.ts
n1ru4l: I realised we created it an additional time where it is not needed. We can just re-use the one we already created in L5 😅

packages/services/api/src/modules/schema/providers/schema-version-store.ts
n1ru4l: fixed!

packages/services/api/src/modules/schema/providers/schema-manager.ts
n1ru4l: > I assume this is to let us to fetch resources more easily without passing org, project, target into every storage function

Yup, in my follow-up PR we only care if a schema version is within a project not whether it belongs to a specific target.

> However, if we need to update a resource, we'll still need to pass in this data to avoid updating resources that don't belong to the org/etc.

We never update a schema version, we only create it so it's a bit of a bad example here. 😓

_____

Aside from that I am still not 100% confident how we should structure data store fetching logic and business logic - this is what I currently think makes sense:

Having a "database store" class and a wrapper "manager" class that handles authorization, caching and houses "entity-to-graphql-type-methods" (where needed) is the way to go currently.

GraphQLLayer ---> SchemaManagerClass ---> DatabaseStore

However, this is also not a fits all solution for cases where we need to do multiple actions across database store as part of a transaction etc.

GraphQLLayer ---> SchemaPublisher ---> SchemaManagerClass ---> DatabaseStore

I am happy to talk about these kind of things in general - these usually go lost in "ship the thing" mode. 😆

Issue #7780, opened 2026-03-05
IAM authentication support for Redis, Kafka, S3 and PG


n1ru4l:
Enterprise self-hosters have the requirement of using IAM authentication for the following services:
- Redis
- Kafka
- S3
- Postgres

The configuration for these options should be done via the existing environment variables prefixes, e.g. `POSTGRES_*`, `REDIS_*`, `KAFKA_*` and `S3_*`.
___

Resources:
- [ElastiCache (Redis)](https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/auth-iam.html)
- [S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/security-iam.html)
- [MSK (Kafka)](https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html)
- [RDS (Postgres)](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html)

mish-elle:
Hi team! We’re looking into this now, and here are our current thoughts on the implementation approach.

1. S3 - We plan to use the AWS SDK credential provider chain for S3 credentials. The aws4fetch client already supports session tokens.
2. Postgres - We plan to generate RDS IAM authentication tokens and use them as the database password. We will also implement token refresh so long-running deployments do not require restarts.
3. Redis - We plan to follow the same pattern as Postgres, including generating credentials/tokens as required and handling refresh.
4. MSK - We plan to use KafkaJS's oauthbearer SASL mechanism with aws-msk-iam-sasl-signer-js to generate short-lived SigV4-signed tokens. KafkaJS re-invokes the token provider on each broker connection, so refresh should be automatic.

We can implement the shared AWS/IAM auth mechanisms as a shared package, per your recommendation. Would you prefer we extend the existing service-commons package for this, or would you rather we introduce a new shared package (for example internal/aws, similar to internal/postgres) and have the relevant services depend on it?

Other than S3, the following env vars will be introduced

1. If AWS_REGION is set (set by AWS), then IAM Auth is enabled.
2. Each service will have <SERVICE>_AWS_REGION, to allow users to override the aws region for a service.
3. Each service will have <SERVICE>_IAM_AUTH_ENABLED, which is true by default, so that users can turn off IAM AUTH for specific services when deployed on AWS.

[...]

PR #8049, opened 2026-05-14
Support for MSK AWS IAM auth

mish-elle:
### Background

Self-hosters running Hive on AWS with Amazon MSK (Managed Streaming for Apache Kafka) currently have no way to use IAM-based authentication for Kafka connections. This forces them to use static credentials.

This PR adds opt-in MSK AWS IAM authentication support for the services that communicates to Kafka/MSK: `usage` and `usage-ingestor`. In addition, the KAFKA_BROKER previously only accepted a single broker address. This PR also updates it to accept a comma-separated list of brokers. This functionality is already supported in the KafkaJS library.

### Description

The MSK token generation logic happens in `service-common/src/iam-msk.ts`'s `createMskIamTokenProvider()` function. This generates a short-lived SigV4 token using `aws-msk-iam-sasl-signer-js`.  The services use this provider to authenticate to Kafka via the OAUTHBEARER SASL mechanism when IAM auth is enabled. KafkaJS supports an oauthBearerProvider to take a function for dynamic passwords so it will handle the token refreshes.  

### New environmental variables introduced
| Variable | Required | Description |
|---|---|---|
| `AWS_REGION` | No | Default AWS region for the service used for all AWS connections |
| `KAFKA_AWS_IAM_AUTH_ENABLED` | No | Set to `1` to enable IAM authentication |
| `KAFKA_AWS_REGION` | No | Optional override for the Kafka broker region (defaults to `AWS_REGION`) |

### Credential Flow for MSK
```mermaid
---
title: MSK / Kafka (Long-lived Connection)
---
flowchart TD
    A[Resolve Kafka SASL] --> B{AWS_REGION set?}
    B -->|Yes| C{KAFKA_AWS_IAM_AUTH_ENABLED = true?}
    C -->|Yes| D[✅ OAUTHBEARER + SigV4 token<br/>Region: KAFKA_AWS_REGION ?? AWS_REGION]
    C -->|No / not set| E{Explicit SASL mechanism?<br/>plain / scram-sha-*}
    B -->|No| E
    E -->|Yes| F[✅ Use username + password]
    E -->|No| G[✅ No SASL]
```

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/service-common/package.json
```diff
@@ -13,6 +13,7 @@
     "@trpc/server": "10.45.3"
   },
   "devDependencies": {
+    "aws-msk-iam-sasl-signer-js": "1.0.0",
```
dotansimha: It's probably better to already upgrade to latest? (`1.0.3`)

packages/services/usage-ingestor/src/environment.ts
```diff
@@ -26,6 +26,7 @@ const EnvironmentModel = zod.object({
   ENVIRONMENT: emptyString(zod.string().optional()),
   RELEASE: emptyString(zod.string().optional()),
   HEARTBEAT_ENDPOINT: emptyString(zod.string().url().optional()),
+  AWS_REGION: emptyString(zod.string().optional()),
```
dotansimha, line 33: What's the incentive to add multiple region environment variables? 

I mean, `AWS_REGION` is usually the standard one, also when using injected properties into EC2 containers. 
While I understand from the description that this adds an option to override the env var for Kafka <> AWS only, I don't think it's used for anything else? So if the region is used only for Kafka in that service, I argue that we might not need two env vars.

dotansimha: Also, if we want to allow users to have full control over the region, and not to depend on the default `AWS_REGION` provided, we can enforce this by having only a single env var, `KAFKA_AWS_REGION`

packages/services/usage-ingestor/src/ingestor.ts
```diff
@@ -51,7 +52,7 @@ export function createIngestor(config: {
 
   const kafka = new Kafka({
     clientId: 'usage-ingestor',
-    brokers: [config.kafka.connection.broker],
+    brokers: config.kafka.connection.broker.split(','),
```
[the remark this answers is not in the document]
dotansimha: That's a good note from Gemini, cc @mish-elle . 
Also in the other services

packages/services/usage-ingestor/src/environment.ts
```diff
@@ -189,13 +194,18 @@ export const env = {
             : true
           : false,
       sasl:
-        kafka.KAFKA_SASL_MECHANISM != null
+        base.AWS_REGION != null && kafka.KAFKA_AWS_IAM_AUTH_ENABLED === '1'
```
dotansimha: This condition could potentially silently fail when `KAFKA_AWS_IAM_AUTH_ENABLED` is enabled, but `AWS_REGION` is not specified. Regardless of the option to override with `KAFKA_AWS_REGION` (becuase that's used and checked only inside the `if`)

packages/services/usage-ingestor/src/ingestor.ts
```diff
@@ -51,7 +52,7 @@ export function createIngestor(config: {
...
     ssl: config.kafka.connection.ssl,
```
dotansimha, line 56: As far as I know, when AWS IAM is used with external services, it will enforce using SSL, and it can't be turned off (TLS is enforced). 

In that case, user will have to specifyy `KAFKA_SSL=1`. 

Should we just override it to be `KAFKA_SSL=1` when AWS IAM is used? or maybe add a validation to fail the config when it's not set/set to `0`?

packages/services/service-common/src/iam-msk.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Creates an OAuth bearer token provider for AWS MSK IAM authentication
+ * @param region AWS region (e.g., 'us-east-1')
+ * @returns Async function that returns { value: token }
+ */
+export function createMskIamTokenProvider(region: string) {
+  return (async () => {
+    const token = await generateAuthToken({ region });
+    return { value: token.token };
+  }) as () => Promise<{ value: string }>;
```
dotansimha: why the cast `as` here? can't it just be `async` and infer the type?

packages/services/service-common/src/iam-msk.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { createMskIamTokenProvider } from './iam-msk';
+
+vi.mock('aws-msk-iam-sasl-signer-js', () => ({
+  generateAuthToken: vi.fn(async () => ({ token: 'abc' })),
+}));
+
+test('returns kafkajs-compatible oauth bearer provider', async () => {
```
dotansimha, line 10: While this test makes sense, it doesn't cover all edge cases. 

Can you please add the non-happy-path tests to this PR as well?

.changeset/silly-flowers-yell.md
```diff
@@ -0,0 +1,21 @@
+---
+'hive': minor
+---
+
+Added opt-in AWS IAM authentication on MSK for self-hosters deployed on AWS. When enabled, the
+`usage` and `usage-ingestor` services authenticate to Kafka using AWS IAM (SigV4) via the
+OAUTHBEARER SASL mechanism ??? no static username/password required.
```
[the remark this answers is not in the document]
dotansimha: Gemini is right here

dotansimha (changes requested):

packages/services/usage-ingestor/src/environment.ts
mish-elle: `AWS_REGION` will be reused by the upcoming PRs for Redis, S3, and RDS IAM auth (#7780). For example, the usage service connects to MSK, RDS, and Redis, a shared `AWS_REGION` lets the user configure the region once, with `<SERVICE>_AWS_REGION` available as a per-service override when resources live in different regions.

If we went with only `KAFKA_AWS_REGION`, users would end up setting `KAFKA_AWS_REGION`, `REDIS_AWS_REGION`, `RDS_AWS_REGION` all to the same value, which is the common case. The two-tier pattern (`AWS_REGION` + optional override) avoids that repetition while still giving full control. We expanded this logic to all services to keep behavior consistent even if the service only connects to one AWS Managed Service. 

So either
1. `AWS_REGION` as a shared default with per-service overrides (`KAFKA_AWS_REGION`, `REDIS_AWS_REGION`, etc.)
2. Only per-service vars (`KAFKA_AWS_REGION`, `REDIS_AWS_REGION`, etc.), no shared default, fully explicit.

Happy to go with either approach!

packages/services/usage-ingestor/src/ingestor.ts
mish-elle: I think we should add a fail fast for this instead of overriding it for the user. Since silently overriding might be confusing when debugging connection issues. For example, if they see `KAFKA_SSL=0` in their environment but SSL is actually on. 

So we can add a fail fast, if the user enables `KAFKA_AWS_IAM_AUTH_ENABLED=1` but forgets `KAFKA_SSL=1` we can have a startup error message like `KAFKA_SSL must be '1' when KAFKA_AWS_IAM_AUTH_ENABLED is enabled`.

dotansimha: Yes. I agree this should fail fast as this behavior is going to fail any.

packages/services/usage-ingestor/src/environment.ts
dotansimha: I see, thank you for the clarification! <br>Now that I understand the need to use this value for other purposes, I understand the initial approach you took. <br><br>Personally, I tend to prefer the 2nd one. Worse case user will need to re-configure the same env var twice, and that's not the worse thing to do. <br>But - I will let @laurin decide on this one, as I'm not sure what's his plans with the other PRs.

mish-elle:
Also the CI pipelines fail because of the pnpm-lock file. We're downgrading dependencies and not building a few packages, due to constraints in our environment. Is it possible for a code maintainer to help us generate the pnpm-lock file?

packages/services/usage/src/environment.ts
```diff
@@ -191,13 +209,18 @@ export const env = {
...
           : false,
       sasl:
-        kafka.KAFKA_SASL_MECHANISM != null
+        kafka.KAFKA_AWS_IAM_AUTH_ENABLED === '1'
           ? {
-              mechanism: kafka.KAFKA_SASL_MECHANISM,
-              username: kafka.KAFKA_SASL_USERNAME,
-              password: kafka.KAFKA_SASL_PASSWORD,
+              mechanism: 'aws-iam' as const,
+              region: (kafka.KAFKA_AWS_REGION ?? base.AWS_REGION)!,
```
n1ru4l: Please do not use the `!` operator, it can introduce subtle unnoticed bugs.

Instead do something like the following:

```ts
function raiseInvariant(reason: string) {
  throw new Error(reason)
}

kafka.KAFKA_AWS_REGION ?? base.AWS_REGION ?? raiseInvariant("This should have been asserted by now.")
```

packages/services/service-common/src/iam-msk.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+  mockedGenerateAuthToken.mockRejectedValueOnce(new Error('credentials not found'));
+  const provider = createMskIamTokenProvider('us-east-1');
+  await expect(provider()).rejects.toThrow('credentials not found');
+});
+
+test('throws when generateAuthToken returns undefined token', async () => {
+  mockedGenerateAuthToken.mockResolvedValueOnce({ token: undefined } as any);
+  const provider = createMskIamTokenProvider('us-east-1');
+  await expect(provider()).resolves.toEqual({ value: undefined });
+});
```
n1ru4l: The test name is not consistent with what it actually does

packages/services/usage-ingestor/src/environment.ts
n1ru4l: I don't have any concerns against the current "inherit" approach, it seems like a convenience feature.
@mish-elle I am curious Do you have an actual use-cases for overwriting single service regions?

n1ru4l (changes requested):
i added a few comments that should be addressed + CI and linting is failing currently. Can you please address these?

packages/services/usage-ingestor/src/environment.ts
mish-elle: We do not, but it's a possible scenario.

n1ru4l (approved):
Legit! Thank you for helping out!

dotansimha (approved):

PR #8050, opened 2026-05-14
Feat/migration from cypress to playwright

mskorokhodov:
### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


n1ru4l (changes requested):
tests and typechecks are failing, can you please resolve these? Please also make sure in the future to resolve these before making the pull request ready for review. See https://www.notion.so/theguildoss/Code-Review-Guide-2c6b6b71848a8003b8e8efa2af52e038?source=copy_link

integration-tests/testkit/seed.ts
```diff
@@ -164,28 +209,11 @@ export function initSeed() {
...
-        new NoopLogger(),
-      );
-
-      await redis.set(key, JSON.stringify(challenge));
-      await redis.disconnect();
+      await pool.query(psql`
+        UPDATE "oidc_integration_domains"
+        SET "verified_at" = NOW()
+        WHERE "id" = ${domainChallengeId}
+      `);
```
n1ru4l: Why are we remove the domain challenge verification forging? Without this we do not have confidence that the DNS lookup actually works.

e2e/fixtures.ts
```diff
@@ -0,0 +1,134 @@
...
+  app: AppHelper;
+  oidc: OIDCHelper;
+  laboratory: LaboratoryHelper;
+  usage: UsageHelper;
+};
+
+const execFileAsync = promisify(execFile);
+const rootDir = new URL('..', import.meta.url).pathname;
+
+async function loadSeedHelper(): Promise<SeedHelper> {
```
n1ru4l: Do we need to run the seeds as a separate process? Is it not possible to call these directly from the tests like we do within out integration tests? Are the test specs run on the node.js thread or within the browser?

n1ru4l: Oh already flagged here: https://github.com/graphql-hive/console/pull/8050/changes#r3243960999

pnpm-lock.yaml
n1ru4l, line 1: seeing this getting slimmer is great 😻

n1ru4l (changes requested):
I flagged some things:
- We invoke the seed as a CLI, when the tests run on Node.js. We can just invoke the testkit directly as in the integration tests
- Tests are failing and some CI checks
- The test for verifying the DNS lookup works properly was altered and creating the redis record was removed fromt he functionality; this should be reverted

n1ru4l (approved):

PR #8031, opened 2026-05-10
feat: schema version promotion

n1ru4l:
## Description

Introduces functionality for [promoting a schema version](https://github.com/graphql-hive/console/issues/4021) from one target to another target (or the same target), making it the latest version within that target.

This enables workflows such as:
- Promote schema version from one environment to another environment (staging to production)
- Quick rollbacks to a specific previous schema version within one environment

> A video walkthrough over this feature and implementation details exists [here](https://guild-oss.slack.com/archives/C01E8ATBQGN/p1779895882698669).

What should be reviewed?
- I want Feedback on the UI
- I am pretty confident with the database changes and API changes as they are covered with quite a few tests and I also manually confirmed all the behaviours many times 😄 Any comments on that are appreciated though

For either of those feel free to hit me up if you want to walk me through things.

### CLI

A new CLI command `schema:promote` is introduced for promoting schema Version

```bash
# Promote specific version to target
hive schema:promote \
  --version 33172112-788d-469b-b06f-d5992b4f1ec7 \
  --to the-guild/graphql-hive/production

# Promote from target to target
hive schema:promote \
  --from the-guild/graphql-hive/staging \
  --to the-guild/graphql-hive/production
````

### Permissions

A new permission `schemaVersion:promote` is introduced to restrict to which targets a schema version can be promoted.

### New History UI

Alongside a new CLI command for performing these promotions this change also comes along a new UI for schema versions with more insights into the individual changes of the subgraphs.

image

> A demo with some [seeded history has been created on our dev environment](https://guild-oss.slack.com/archives/C01E8ATBQGN/p1779973862010349?thread_ts=1779895882.698669&cid=C01E8ATBQGN).

<details>

<summary>More UI Screenshots</summary>

image

image

image

image

</details>

### Database

The database was altered with new columns for storing the new data that is collected for subgraphs.
Refer to the migration file for a breakdown of all new introduced columns.

The database schema is fully backwards compatible. If any major issues occur on the business logic of the API, we can easily roll it back as all the changes introduced are additional.

### GraphQL API

- New fields have been added to determine the origin of a schema version (publish; delete; promotion)
- A new mutation for promoting a schema version was introduced

## Other

Closes https://github.com/graphql-hive/console/issues/4021

PR #8054, opened 2026-05-17
feat: make schema log lookup target_id independent

n1ru4l:
### Background

This is a prerequisite for rolling out https://github.com/graphql-hive/console/pull/8031 in a way that allows easy rollbacks.

### Description

The current implementation for resolving the GraphQL `SchemaVersion.log` field filters the result by the `target_id` column on the `schema_log` table.

Since https://github.com/graphql-hive/console/pull/8031 will introduce a way to associate a log from one target with another targets schema version, this would cause a runtime exception in case we need to revert and rollback the implementation as the `target_id` mismatch would prevent a successful lookup.

By making the lookup less "strict" rolling back https://github.com/graphql-hive/console/pull/8031 becomes straight-forward and requires no manual code changes.

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-version-store.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-version-store.ts
@@ -881,14 +881,14 @@ export class SchemaVersionStore {
       FROM
         "schema_versions"
       WHERE
-        "action_id" = (
+        "target_id" = ${target.id}
+        AND "action_id" = (
           SELECT
             "id"
           FROM
             "schema_log"
           WHERE
             "schema_log"."project_id" = ${target.projectId}
-            AND "schema_log"."target_id" = ${target.id}
             AND "schema_log"."commit" = ${commit}
           ORDER BY "schema_log"."created_at" DESC
           LIMIT 1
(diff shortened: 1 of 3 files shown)


packages/services/api/src/modules/schema/providers/schema-version-store.ts
```diff
@@ -888,7 +888,6 @@ export class SchemaVersionStore {
             "schema_log"
           WHERE
             "schema_log"."project_id" = ${target.projectId}
-            AND "schema_log"."target_id" = ${target.id}
             AND "schema_log"."commit" = ${commit}
           ORDER BY "schema_log"."created_at" DESC
```
n1ru4l, line 893: This query uses the existing index for a scan

```
CREATE INDEX CONCURRENTLY IF NOT EXISTS "schema_log_by_commit_ordered" ON "schema_log"(
  "project_id"
  , "target_id"
  , "commit"
  , "created_at" DESC
)
```

If we remove the `AND "schema_log"."target_id" = ${target.id}` condition it still uses the index but is slightly less efficient.

It is an okayish trade-off (unless someone has strong opinions about it) and we can adjust the index in the future if needed. This feature for looking up a schema version by commit is not frequently used anyways (based on production traffic).
- https://app.graphql-hive.com/the-guild/graphql-hive/production/insights/schema-coordinate/Query.schemaVersionByCommit?search=schemaVersionByCommit&from=now-90d&to=now
- https://app.graphql-hive.com/the-guild/graphql-hive/production/insights/schema-coordinate/Query.schemaVersionForActionId?search=schemaVersionForActionId&from=now-90d&to=now

PR #8057, opened 2026-05-18
fix(api): `CreateOrganizationAccessTokenInput.description` nullability

### LINKED ISSUE #8056: `CreateOrganizationAccessTokenInput.description` marked as nullable `String`, but it is required in the DB
```graphql
  """
  Input for creating a new organization-scoped access token.
  """
  input CreateOrganizationAccessTokenInput {

   # .... 

    """
    Optional description providing additional context about the purpose of the access token.
    """
    description: String @tag(name: "public")
    #             ^ 

    # ... 
}

```

But in the DB: 

```sql
 CREATE TABLE IF NOT EXISTS "organization_access_tokens" (
      "id" UUID PRIMARY KEY NOT NULL DEFAULT uuid_generate_v4()
      , "organization_id" UUID NOT NULL REFERENCES "organizations" ("id") ON DELETE CASCADE
      , "created_at" timestamptz NOT NULL DEFAULT now()
      , "title" text NOT NULL
      , "description" text NOT NULL
 #                            ^ 

} 
```

dotansimha:
Fixes https://github.com/graphql-hive/console/issues/8056


n1ru4l (changes requested):
On the GraphQL type the description is nullable:

Image

This I think the correct approach is to instead fix the zod model to allow a nullable description property.

Image

~~Also this fix would not solve existing database records as the AI comment points out.~~ <- Should have read the issue first. We should fix the table definition removing the non null constraint.

n1ru4l (approved):

packages/migrations/src/actions/2026.03.25T00-00-00.access-token-expiration.ts
```diff
@@ -10,7 +10,7 @@ export default {
    */
   run: ({ psql }) => psql`
     ALTER TABLE IF EXISTS organization_access_tokens
-    ADD COLUMN IF NOT EXISTS "expires_at" TIMESTAMP WITH TIME ZONE
+      ADD COLUMN IF NOT EXISTS "expires_at" TIMESTAMP WITH TIME ZONE
```
dotansimha, line 13: oopsi?

PR #8062, opened 2026-05-22
Field level usage metrics with errors

jdolle:
### Background

Part of the subgraph visibility initiative. This PR adds coordinate level resolution count tracking, and tracks errors for coordinates by error code.

This can be broken down into several components:
1. Clickhouse table schema, which sets up the structure of the data in our database for our UI's time periods.
2. A new gateway plugin that uses subgraph calls to extract field and error counts
3. Modifies usage data pipeline to accept and process additional data on the v2 usage data.
4. And renders this new data on our UI on the explorer and coordinate insights pages.

### Description

The tables and materialized views are in Clickhouse. These support two query patterns -- only centered around the coordinates and another that is hash (operation) specific.

This supports our two current usage patterns -- the first being in use today on the explorer page, and the other being a proposed new feature to show the success/error results by field inside an operation. In either case, we want to support filtering by other metrics such as the error code.

The Hive client has been adjusted to add a subgraph call between starting and sending the request data to be batched. Field counts are added to the existing operations data and a new structure is being submitted for errors.

Rather than use a materialized view from the source table for every time period, I've introduced cascading updates. This can impact insert time, but these materialized views are inexpensive (no joins or costly functions), so I do not anticipate an issue. Regardless, it may be best to enable async inserts in the future if we've not already done so.

Screenshot 2026-06-11 at 7 43 58 PM
Screenshot 2026-06-08 at 8 41 40 PM
Screenshot 2026-06-08 at 8 32 26 PM

#### Gateway Benchmark Comparison

To ensure gateway performance is not impacted too heavily, a benchmark was ran. This was ran locally, in constant mode, with 10 cpu. Here are the results:

No Usage Reporting

```
     ✓ response code was 200
     ✓ no graphql errors
     ✓ valid response structure

     checks.........................: 100.00% ✓ 1266159     ✗ 0
     data_received..................: 37 GB   615 MB/s
     data_sent......................: 491 MB  8.1 MB/s
     http_req_blocked...............: avg=1.8µs   min=0s     med=1µs    max=7.46ms   p(90)=2µs    p(95)=2µs    p(99.9)=51.84µs
     http_req_connecting............: avg=520ns   min=0s     med=0s     max=7.44ms   p(90)=0s     p(95)=0s     p(99.9)=0s
     http_req_duration..............: avg=6.98ms  min=2.07ms med=6.46ms max=150.65ms p(90)=7.96ms p(95)=8.83ms p(99.9)=48.86ms
       { expected_response:true }...: avg=6.98ms  min=2.07ms med=6.46ms max=150.65ms p(90)=7.96ms p(95)=8.83ms p(99.9)=48.86ms
     http_req_failed................: 0.00%   ✓ 0           ✗ 422153
     http_req_receiving.............: avg=51.36µs min=9µs    med=24µs   max=39.03ms  p(90)=68µs   p(95)=147µs  p(99.9)=1.95ms
     http_req_sending...............: avg=6.75µs  min=1µs    med=3µs    max=29.54ms  p(90)=6µs    p(95)=16µs   p(99.9)=504.84µs
     http_req_tls_handshaking.......: avg=0s      min=0s     med=0s     max=0s       p(90)=0s     p(95)=0s     p(99.9)=0s
     http_req_waiting...............: avg=6.92ms  min=2.04ms med=6.4ms  max=150.61ms p(90)=7.87ms p(95)=8.73ms p(99.9)=48.76ms
     http_reqs......................: 422153  7006.314119/s
     iteration_duration.............: avg=7.1ms   min=2.39ms med=6.59ms max=150.81ms p(90)=8.1ms  p(95)=8.98ms p(99.9)=49.03ms
     iterations.....................: 422053  7004.654456/s
     success_rate...................: 100.00% ✓ 422053      ✗ 0
     vus............................: 50      min=50        max=50
     vus_max........................: 50      min=50        max=50
```

With Usage Reporting

```
     ✓ response code was 200
     ✓ no graphql errors
     ✓ valid response structure

     checks.........................: 100.00% ✓ 1162125     ✗ 0
     data_received..................: 34 GB   563 MB/s
     data_sent......................: 451 MB  7.5 MB/s
     http_req_blocked...............: avg=1.57µs  min=0s     med=1µs    max=3.7ms    p(90)=2µs    p(95)=2µs     p(99.9)=65µs
     http_req_connecting............: avg=179ns   min=0s     med=0s     max=2.22ms   p(90)=0s     p(95)=0s      p(99.9)=0s
     http_req_duration..............: avg=7.61ms  min=2.38ms med=6.96ms max=132.13ms p(90)=9.07ms p(95)=10.55ms p(99.9)=52.4ms
       { expected_response:true }...: avg=7.61ms  min=2.38ms med=6.96ms max=132.13ms p(90)=9.07ms p(95)=10.55ms p(99.9)=52.4ms
     http_req_failed................: 0.00%   ✓ 0           ✗ 387475
     http_req_receiving.............: avg=58.88µs min=9µs    med=25µs   max=24.17ms  p(90)=76.6µs p(95)=170µs   p(99.9)=2.32ms
     http_req_sending...............: avg=7.63µs  min=1µs    med=3µs    max=19.33ms  p(90)=6µs    p(95)=18µs    p(99.9)=619.05µs
     http_req_tls_handshaking.......: avg=0s      min=0s     med=0s     max=0s       p(90)=0s     p(95)=0s      p(99.9)=0s
     http_req_waiting...............: avg=7.55ms  min=2.34ms med=6.9ms  max=131.72ms p(90)=8.97ms p(95)=10.43ms p(99.9)=52.32ms
     http_reqs......................: 387475  6408.88252/s
     iteration_duration.............: avg=7.73ms  min=3.17ms med=7.07ms max=132.38ms p(90)=9.21ms p(95)=10.71ms p(99.9)=52.58ms
     iterations.....................: 387375  6407.228508/s
     success_rate...................: 100.00% ✓ 387375      ✗ 0
     vus............................: 50      min=50        max=50
     vus_max........................: 50      min=50        max=50
```

With Gateway Plugin Usage Reporting
```
     scenarios: (100.00%) 1 scenario, 50 max VUs, 1m30s max duration (incl. graceful stop):
              * default: 50 looping VUs for 1m0s (gracefulStop: 30s)

     ✓ response code was 200
     ✓ no graphql errors
     ✓ valid response structure

     checks.........................: 100.00% ✓ 590064      ✗ 0     
     data_received..................: 17 GB   287 MB/s
     data_sent......................: 229 MB  3.8 MB/s
     http_req_blocked...............: avg=1.97µs  min=0s     med=1µs     max=3.54ms  p(90)=2µs     p(95)=2µs     p(99.9)=80.21µs 
     http_req_connecting............: avg=575ns   min=0s     med=0s      max=3.25ms  p(90)=0s      p(95)=0s      p(99.9)=0s      
     http_req_duration..............: avg=15.11ms min=2.07ms med=14.6ms  max=65.99ms p(90)=16.35ms p(95)=17.73ms p(99.9)=54.58ms 
       { expected_response:true }...: avg=15.11ms min=2.07ms med=14.6ms  max=65.99ms p(90)=16.35ms p(95)=17.73ms p(99.9)=54.58ms 
     http_req_failed................: 0.00%   ✓ 0           ✗ 196788
     http_req_receiving.............: avg=65.33µs min=9µs    med=26µs    max=13.32ms p(90)=96µs    p(95)=206µs   p(99.9)=2.45ms  
     http_req_sending...............: avg=8.15µs  min=1µs    med=3µs     max=4.15ms  p(90)=7µs     p(95)=25µs    p(99.9)=571.63µs
     http_req_tls_handshaking.......: avg=0s      min=0s     med=0s      max=0s      p(90)=0s      p(95)=0s      p(99.9)=0s      
     http_req_waiting...............: avg=15.04ms min=2.04ms med=14.54ms max=65.79ms p(90)=16.21ms p(95)=17.6ms  p(99.9)=54.48ms 
     http_reqs......................: 196788  3263.204376/s
     iteration_duration.............: avg=15.24ms min=2.62ms med=14.71ms max=66.09ms p(90)=16.5ms  p(95)=17.9ms  p(99.9)=54.73ms 
     iterations.....................: 196688  3261.546142/s
     success_rate...................: 100.00% ✓ 196688      ✗ 0     
     vus............................: 50      min=50        max=50  
```


packages/migrations/src/clickhouse-actions/021-usage-coordinate-errors.ts
```diff
@@ -0,0 +1,344 @@
...
+
+  /** Minutely metric table */
+  await exec(`
+    CREATE TABLE IF NOT EXISTS default.coordinate_errors_minutely
+    (
+      target UUID
+      , hash FixedString(16) CODEC(LZ4)
+      , timestamp DateTime('UTC') CODEC(DoubleDelta, ZSTD(1))
+      , expires_at DateTime('UTC') CODEC(DoubleDelta, ZSTD(1))
+      , coordinate String CODEC(ZSTD(1))
```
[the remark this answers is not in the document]
jdolle: there can be a significant number of coordinates.

packages/libraries/gateway-usage/package.json
```diff
@@ -0,0 +1,65 @@
+{
+  "name": "@graphql-hive/gateway-usage",
```
jdolle: I don't like this name but I don't have a better one atm. It's hard since everything is "hive".

packages/libraries/core/src/client/types.ts
```diff
@@ -11,17 +11,53 @@ type HeadersObject = {
...
+};
+
 export interface HiveClient {
   [hiveClientSymbol]: true;
   [autoDisposeSymbol]: boolean | NodeJS.Signals[];
   info(): void | Promise<void>;
   reportSchema: SchemaReporter['report'];
   /** Collect usage for Query and Mutation operations */
-  collectUsage(): CollectUsageCallback;
+  collectUsage(): CollectUsage;
```
n1ru4l, line 57: This is a  breaking API change, correct? Should we remain the object still callable for backwards compatibility?

packages/services/storage/src/db/types.ts
```diff
@@ -282,6 +282,16 @@ export interface projects {
...
 
+export interface proposal_approved_changes {
+  change: any;
+  hash: string;
+  id: string;
+  proposal_id: string;
+  schema_version_id: string | null;
+  service: string | null;
+  target_id: string;
+}
```
n1ru4l: this seems wrong here? :D

integration-tests/tests/gateway-plugin-console-sdk/plugin.spec.ts
n1ru4l, line 1: Should we have some tests for more than one subgraph? ~~Is there some specific federtion features that we should test for specifically (e.g. `@interfaceObject`)~~?

packages/libraries/gateway-usage/package.json
n1ru4l: Maybe something like this?

```suggestion
  "name": "@graphql-hive/gateway-plugin-console-sdk",
```

packages/libraries/gateway-plugin-console-sdk/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "check:build": "bob check"
+  },
+  "peerDependencies": {
+    "@envelop/types": "^5.0.0",
+    "@graphql-hive/gateway-runtime": "^1.0.0 || ^2.0.0",
+    "graphql": "^0.13.0 || ^14.0.0 || ^15.0.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "@graphql-hive/core": "workspace:*",
+    "tiny-lru": "11.4.7"
```
n1ru4l: this should be a range for packages we publish for others to use.

packages/services/api/src/modules/schema/resolvers/Target.ts
```diff
@@ -123,4 +125,9 @@ export const Target: Pick<
       organizationId: target.orgId,
     });
   },
+  hasFieldLevelMetrics: async (target, _, { injector }) => {
+    const org = await injector.get(Storage).getOrganization({ organizationId: target.orgId });
+    // @TODO also check if any metrics have been sent with the new format.
+    return org.featureFlags.subgraphVisibility;
```
n1ru4l: TODO

packages/services/usage/src/usage-processor-2.ts
```diff
@@ -262,6 +280,55 @@ const OperationMapRecordSchema = tb.Object(
...
+    /** Number of times the field has been requested. Regardless of success or failure */
+    fields: tb.Record(tb.String(), tb.Number()),
+
+    /** Error code for a coordinate, with a code returned from the graphql extensions */
+    errors: tb.Optional(
+      tb.Array(tb.Object({ coordinate: tb.String(), code: tb.Optional(tb.String()) })),
+    ),
+
+    /** Which subgraph resolved this path */
+    subgraph: tb.String(),
```
n1ru4l, line 314: What do monolithic schemas report here?

jdolle: empty string

packages/libraries/gateway-plugin-console-sdk/src/index.ts
```diff
@@ -0,0 +1,183 @@
...
+      for (const signal of signals) {
+        process.once(signal, () => hive.dispose());
+      }
+    } else {
+      console.error(
+        'It seems that GraphQL Hive is not being executed in Node.js. ' +
+          'Please attempt manual client disposal and use autoDispose: false option.',
+      );
+    }
+  }
```
enisdenjo: hive gateway plugins implement the explicit resource management, instead of this you can add a property if `Signal.asyncDispose` and do cleanup there. see:

https://github.com/ardatan/whatwg-node/blob/60d9a88040f8b0e91b70067db3f982cc2c5cb229/packages/server/src/plugins/types.ts#L32-L49

enisdenjo (commented):
the gateway plugin is quite straightforward, nothing pops up that I see could be a performance bottleneck on the spot there - I dont think the plugin itself is the problem.

I didnt run the benchmarks myself, but you should profile the process and see whats taking up most of the time compared to without the field-level reporting

jdolle:
By a large margin, the logic taking the most time is the extracting of field coordinate usage counts. This is what we'd expect, since it's iterating over the entire result payload and operation to determine the coordinates for each of these values.

I of course wanted this to be more efficient, but ultimately I think this is fine. Most observability tools note that they reduce performance, and this is exactly why. It's simply expensive to iterate over the entire response.

deployment/services/graphql.ts
```diff
@@ -141,6 +141,7 @@ export function deployGraphQL({
           HIVE_USAGE: '1',
           HIVE_TARGET: hiveConfig.require('target'),
           HIVE_USAGE_ENDPOINT: serviceLocalEndpoint(usage.service),
+          HIVE_FIELD_USAGE_ENABLED: '0',
```
dotansimha, line 144: Any reason to disable it here? I mean, we are experiment it with our own API to see how it works, before rolling it out.

packages/libraries/core/src/client/subrequests/path-to-coordinate.ts
```diff
@@ -0,0 +1,43 @@
...
+  type GraphQLFieldMap,
+  type GraphQLSchema,
+} from 'graphql';
+
+export function pathToCoordinate(
+  schema: GraphQLSchema,
+  errorPath: readonly (string | number)[],
+  operationType: 'mutation' | 'subscription' | 'query' = 'query',
+): string | undefined {
+  let currentType;
```
dotansimha: nit: this could be much more readable if it would just be `let currentType = resolveRootType(operationType)`

packages/libraries/core/src/client/subrequests/path-to-coordinate.ts
```diff
@@ -0,0 +1,43 @@
+import {
+  getNamedType,
+  isInterfaceType,
+  isObjectType,
+  type GraphQLFieldMap,
+  type GraphQLSchema,
+} from 'graphql';
+
+export function pathToCoordinate(
```
dotansimha: I assume this is used only for errors? should it be named `errorPathToCoordinate`?

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
```diff
@@ -0,0 +1,316 @@
...
+  fieldName: string;
+  childTypeName?: string;
+  leafTypeName?: string;
+}
+
+type TypePlan = Map<string, CompiledField>;
+
+type ExecutionPlan = Map<string, TypePlan>;
+
+const RETENTION_CACHE_TTL_IN_SECONDS = 120;
```
dotansimha: why `120`?

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
```diff
@@ -0,0 +1,316 @@
...
+
+const documentPlanCache = new WeakMap<DocumentNode, ExecutionPlan>();
+const hashPlanCache = new MemoryCache({
+  max: 1_000,
```
dotansimha: why `1000`?

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
```diff
@@ -0,0 +1,316 @@
...
+  queryHash?: string;
+}
+
+export function extractCoordinates({
+  schema,
+  document,
+  resultData,
+  queryHash,
+}: ExtractCoordinatesArgs): Record<string, number> {
+  if (!resultData) return Object.create(null);
```
dotansimha: Any reason to use `Object.create(null)` over just `{}`

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
```diff
@@ -0,0 +1,316 @@
...
+const schemaCoordinateExpansionCache = new WeakMap<GraphQLSchema, Map<string, string[]>>();
+
+export interface ExtractCoordinatesArgs {
+  schema: GraphQLSchema;
+  document: DocumentNode;
+  resultData: any;
+  queryHash?: string;
+}
+
+export function extractCoordinates({
```
dotansimha, line 43: can you maybe add a high level overview of what this flow does? 
I mean, it is split to smaller functions, but they feel a bit complex.

packages/libraries/core/src/client/types.ts
```diff
@@ -150,6 +186,15 @@ export interface HiveUsagePluginOptions {
    * Default: false
    */
   processVariables?: boolean;
+
+  /**
+   * Report usage data for resolved coordinates. This counts the number of times a coordinate is resolved and reports
+   * graphql errors and their codes to Hive. Before enabling this, be aware that this is CPU intensive since it must
```
dotansimha, line 192: Do we know how much CPU-intensive this is? 
I mean, how is it different than other traversal over the response? Do we know what's the estimated overhead for small, medium, and large operations?

Did we try to optimize the way it's being done? (I assume that why the cache in extract-coordinated? how much impact did it have?)

packages/libraries/core/src/client/types.ts
```diff
@@ -12,17 +12,53 @@ type HeadersObject = {
...
+
+export type FinishSubRequestCallback = (args: {
+  /** HTTP Status Code */
+  status: number;
+
+  /** Used to calculate error code for a coordinate, with a code returned from the graphql extensions */
+  result?: GraphQLResult;
+
+  /** The GraphQL schema being accessed. Used to calculate coordinate from error path and the coordinate for field counts */
+  subgraphSchema: GraphQLSchema;
```
dotansimha, line 40: If we are dealing with Federation, does the collected really needs to have a full `GraphQLSchema` of the subgraph? do we really have it? (cc @enisdenjo )

packages/libraries/core/src/client/usage.ts
```diff
@@ -224,18 +269,36 @@ export function createUsage(pluginOptions: HiveInternalPluginOptions): UsageColl
...
           schema: args.args.schema,
           max: options.max ?? 1000,
           ttl: options.ttl,
           processVariables: options.processVariables ?? false,
         });
 
+        let fetches = args.fetches;
+        if (!fetches?.length && options.fieldLevelMetricsEnabled) {
+          /**
+           * No subgraph requests, so this must be a monolith.
```
dotansimha, line 280: I wonder if this should be addressed better? If this is a monolith, should we just have a separate way of handling this? 

Also, the fake-fetch assumes the response is 200?

deployment/services/graphql.ts
jdolle: I want to roll out the API/DBs before sending the new usage format.

If all rolled out at once, some usage data may be the new format and therefore would error and be retried until the new usage service went live. I'm okay with this, but if something unanticipated happened, then it could complicate a rollback.

Additionally, since the plugin's field tracking can impact performance, I want to roll this out separately to monitor more closely.

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
jdolle: It matched our other default caching limits. It was also high enough that for most gateways, a majority of operations would be cached, but not so large as to have a negative impact on memory.

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
jdolle: I was trying to eke out performance everywhere. It didnt do anything though and is probably not worth it.

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
jdolle: The best way to understand this function is through the tests.
The goal is to count coordinate resolutions/executions. This is relatively simple for types, fields, and scalars, but abstract types and fragments make this complex.

At a high level, this must:
1. **Determine the coordinate**. This is determined by iterating over the payload and document simultaneously, and it expects a __typename, which is added to the operation automatically by the plugin
2. **Determine count**. If the value is null then it does not count the return type as used, but it does count the field as used.

packages/libraries/core/src/client/types.ts
jdolle: We do have it. We could use the gateway's schema but knowing that the subgraph's schema would be much smaller, I was thinking that this would be more efficient.

packages/libraries/core/src/client/usage.ts
jdolle: I considered a few options here. The 3 extra bits of data this sends for a monolith (subgraph name, status 200, and request time) are minimal and keeping the format the same goes a long way to simplify data ingestion.
I dont see the 200 as an issue. It represents a successful request -- which is guaranteed since it's hitting the the local schema.

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
jdolle: I've adjusted the caching to take as much as necessary now using weakmaps. This is because I am trying to get every bit of performance possible

packages/libraries/core/src/client/subrequests/path-to-coordinate.ts
jdolle: It is only used for errors but it could technically be used for any path. I'd be fine renaming though

packages/libraries/core/src/client/types.ts
jdolle: I've revise my approach and increased the amount of caching. The latest numbers (running locally):

```
     scenarios: (100.00%) 1 scenario, 50 max VUs, 1m30s max duration (incl. graceful stop):
              * default: 50 looping VUs for 1m0s (gracefulStop: 30s)

     ✓ response code was 200
     ✓ no graphql errors
     ✓ valid response structure

     checks.........................: 100.00% ✓ 810489      ✗ 0     
     data_received..................: 24 GB   394 MB/s
     data_sent......................: 314 MB  5.2 MB/s
     http_req_blocked...............: avg=1.75µs  min=0s     med=1µs     max=3.68ms  p(90)=2µs     p(95)=2µs     p(99.9)=82µs    
     http_req_connecting............: avg=336ns   min=0s     med=0s      max=2.57ms  p(90)=0s      p(95)=0s      p(99.9)=0s      
     http_req_duration..............: avg=10.97ms min=2.15ms med=10.22ms max=92.42ms p(90)=12.63ms p(95)=14.51ms p(99.9)=59.02ms 
       { expected_response:true }...: avg=10.97ms min=2.15ms med=10.22ms max=92.42ms p(90)=12.63ms p(95)=14.51ms p(99.9)=59.02ms 
     http_req_failed................: 0.00%   ✓ 0           ✗ 270263
     http_req_receiving.............: avg=62.51µs min=9µs    med=25µs    max=11.31ms p(90)=88µs    p(95)=198µs   p(99.9)=2.19ms  
     http_req_sending...............: avg=7.82µs  min=1µs    med=3µs     max=6.27ms  p(90)=7µs     p(95)=21µs    p(99.9)=646.73µs
     http_req_tls_handshaking.......: avg=0s      min=0s     med=0s      max=0s      p(90)=0s      p(95)=0s      p(99.9)=0s      
     http_req_waiting...............: avg=10.9ms  min=2ms    med=10.16ms max=92.37ms p(90)=12.51ms p(95)=14.4ms  p(99.9)=58.83ms 
     http_reqs......................: 270263  4481.625118/s
     iteration_duration.............: avg=11.09ms min=2.26ms med=10.33ms max=92.53ms p(90)=12.77ms p(95)=14.68ms p(99.9)=59.57ms 
     iterations.....................: 270163  4479.966872/s
     success_rate...................: 100.00% ✓ 270163      ✗ 0     
     vus............................: 50      min=50        max=50  
```

This represents a 37% increase in throughput (iterations) compared to my original version.

This is now 65% of the iterations as compared to our existing usage plugin. I think this is an acceptable starting point, especially since the benchmark may not be the best representation of a production system.

packages/libraries/core/src/client/subrequests/extract-coordinates.ts
jdolle: Ive adjusted the caching. I initially chose this due to other similar plugins using this default, and that 2min seemed appropriate for reducing work during high-traffic periods without using memory unnecessarily otherwise.

packages/web/app/src/lib/utils.ts
```diff
@@ -23,6 +23,27 @@ function hslToHex(h: number, s: number, l: number): string {
...
+
+  return `hsl(${hue}, ${saturation}%, ${lightness}%)`;
+}
+
+// export function foregroundColor(str: string, theme: 'light' | 'dark') {
+//   const h = stringToHue(str);
+//   return theme === 'light'
+//     ? { backgroundColor: `hsl(${h}, 75%, 72%)`, color: '#f2f2f2' }
+//     : { backgroundColor: `hsl(${h}, 60%, 90%)`, color: '#4f4f4f' };
+// }
```
n1ru4l: NIT: why not delete it?

packages/services/server/src/environment.ts
```diff
@@ -166,6 +166,7 @@ const HiveModel = zod.intersection(
       HIVE_USAGE_ENDPOINT: zod.string().url().optional(),
       HIVE_TARGET: zod.string(),
       HIVE_ACCESS_TOKEN: zod.string(),
+      HIVE_FIELD_USAGE_ENABLED: zod.boolean().optional().default(false),
```
n1ru4l: The value can never be a boolean, it must be a string

n1ru4l:
```suggestion
      HIVE_FIELD_USAGE_ENABLED: emptyString(z.union([zod.literal('0'), zod.literal('1')]).optional())
```

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -1138,6 +1151,16 @@ export default gql`
...
+
+    """
+    The total amount of resolutions of the schema coordinate within the contextual period.
+    """
+    totalResolutions: Float
+
+    """
+    The total amount of errors of the schema coordinate within the contextual period.
+    """
+    errorTotal: Float
```
n1ru4l: NIT: instead of Float we can use SafeInt (which we already have in the project). https://the-guild.dev/graphql/scalars/docs/scalars/safe-int

packages/libraries/yoga/src/index.ts
```diff
@@ -86,6 +105,28 @@ export function useHive(clientOrOptions: HiveClient | HivePluginOptions): Plugin
...
+            const query = parseCtx.params.source;
+            const cachedDocument = operationCache.get(query);
+            if (cachedDocument) {
+              // If "true" is cached, then this operation doesn't need stored because it's identical to the original.
+              // Else, the document hash been modified and cached
+              if (cachedDocument !== true) {
+                parseCtx.setParsedDocument(cachedDocument);
+              }
+            } else if (latestSchema) {
+              const modifiedDocument = addTypenames(ctx.result, latestSchema);
```
n1ru4l: Does this lead to the client receiving a response containing fields (`__typename`) it did not ask for?
Does this affect the hash that is generated and reported to Hive Console?

integration-tests/tests/gateway-plugin-console-sdk/plugin.spec.ts
```diff
@@ -0,0 +1,528 @@
...
+          }
+        `),
+        resolvers: {
+          Query: {
+            users: () => [{ id: 2 }],
+          },
+          User: {
+            name: () => {
+              const err = new Error('Something went wrong');
+              Object.assign(err, { code: 'OOPSIE' });
```
n1ru4l: Can we add a test for an error code containing special characters that could potentially break clickhouse ingestion? E.g. `"`

n1ru4l: Also I see not tests for error codes (extensions?)

packages/web/app/src/components/target/explorer/common.tsx
```diff
@@ -50,26 +53,76 @@ export function SchemaExplorerUsageStats(props: {
...
-          <div
-            title={`${toDecimal(percentage)}% of all requests`}
-            className="bg-neutral-2/20 relative z-0 mt-1 w-full min-w-[25px] overflow-hidden rounded-sm"
-            style={{ width: 50, height: 5 }}
-          >
-            <div className="bg-neutral-2 z-0 h-full" style={{ width: `${percentage}%` }} />
-          </div>
+          <div className="min-w-[25px] text-center">{formatNumber(usage.total)}</div>
         </div>
+        {availability ? (
```
n1ru4l:
```suggestion
        {availability !== null ? (
```

The availability could be `0%` if the field always fails?

packages/libraries/core/src/client/subrequests/path-to-coordinate.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (isObjectType(currentType) || isInterfaceType(currentType)) {
+      const fields: GraphQLFieldMap<any, any> = currentType.getFields();
+      const field = fields[segment];
+
+      if (!field) {
+        console.warn(
+          `Hive Usage Client: Field '${segment}' not found on type '${currentType.name}'. Was this aliased? Error ignored.`,
+        );
+        return;
+      }
```
n1ru4l: Why does it not work with aliasing? It should work for aliasing?

jdolle: We originally skipped alias support because it's an edge case.

I've now added support, but putting all that logic in the gateway is proving complicated and resource-heavy. It may be worth reevaluating placing all the logic in the gateway.

integration-tests/tests/gateway-plugin-console-sdk/plugin.spec.ts
jdolle: Added support for `\"` and a test case

packages/libraries/yoga/src/index.ts
jdolle: > Does this affect the hash that is generated and reported to Hive Console?

It would. The hash is generated in `packages/libraries/core/src/client/utils.ts` `cacheDocumentKey`, which is called from the `usage.finish()` call after execution.

jdolle: I've reviewed the code in detail again, manually tested, and added clarifying comments. I'm actually using the original parsed document in the `usage.finish` call, so it would not modify the cacheKey or the reported operation sdl.

.changeset/tricky-items-sit.md
```diff
@@ -0,0 +1,8 @@
+---
+'@graphql-hive/core': minor
+'@graphql-hive/yoga': minor
+'hive': minor
+---
+
+Add a new experimental config option that allows sending additional metrics (errors and resolution counts) to
+Hive.
```
n1ru4l:
```suggestion
Hive Console.
```

packages/libraries/core/src/client/add-hive-typenames.ts
n1ru4l, line 1: Nit: lets renamde this `addHiveTypenames` and `add-hive-typenames.ts`

packages/libraries/core/src/client/usage.ts
```diff
@@ -224,18 +267,36 @@ export function createUsage(pluginOptions: HiveInternalPluginOptions): UsageColl
...
+           * We still want to track the field metrics, so create an artificial
+           * fetch that represents the local lookup.
+           */
+
+          fetches = [
+            {
+              document: args.args.document,
+              duration: args.duration,
+              start: 0,
+              status: 200,
```
n1ru4l: lets make this optional, because a monolith does not actually do a http call

n1ru4l (approved):

PR #8063, opened 2026-05-22
fix: set Redis maxmemory eviction policy to allkeys-lru

saihaj:
## What

Add `--maxmemory-policy allkeys-lru` to the Redis runtime args in `deployment/utils/redis.ts`.

## Why

Redis had no `maxmemory-policy` configured, so it defaulted to `noeviction`. On reaching `maxmemory` (already set to 90% of the container limit), Redis rejects every write with `OOM command not allowed` instead of freeing space.

With `allkeys-lru`, Redis evicts the least-recently-used keys to make room for new writes. Everything stored in this Redis is a cache / lock / rate-limit entry with a TTL, so dropping cold keys is safe and LRU keeps hot keys resident.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/brave-laws-cut.md
+++ b/.changeset/brave-laws-cut.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+add eviction policy to redis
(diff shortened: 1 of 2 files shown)


dotansimha (approved):

dotansimha:
We've tested this as a hotfix and it seems stable and improves perf and overall results. Thanks @saihaj

PR #8064, opened 2026-05-22
Extend Hive CLI timeout for schema:check from 20s to 55s 

jdolle:
### Background

We're seeing clients timeout on checks for large, complex schemas while the server eventually completes. This causes an unnecessary retry. The Hive CLI should have a timeout value that aligns closer with the server's timeout to avoid this situation.

### Description

Sets timeout to 60s for schema:check

### DIFF
--- a/packages/libraries/cli/src/commands/schema/check.ts
+++ b/packages/libraries/cli/src/commands/schema/check.ts
@@ -291,7 +291,7 @@ export default class SchemaCheck extends Command<typeof SchemaCheck> {
           },
         },
         /** Gateway timeout is 60 seconds. */
-        timeout: 60_000,
+        timeout: 55_000,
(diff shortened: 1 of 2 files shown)


saihaj (approved):

packages/libraries/cli/src/commands/schema/check.ts
```diff
@@ -290,6 +290,8 @@ export default class SchemaCheck extends Command<typeof SchemaCheck> {
             schemaProposalId: flags.schemaProposalId,
           },
         },
+        /** Gateway timeout is 60 seconds. */
+        timeout: 60_000,
```
n1ru4l:
```suggestion
        timeout: 55_000,
```

PR #8066, opened 2026-05-23
Switch paginated queries to use psql tuple comparison to improve planner performance

jdolle:
### Background

Performance improvements

### Description

This query is the same as the OR but is easier on the planner

### DIFF
--- a/packages/services/api/src/modules/schema/providers/contracts.ts
+++ b/packages/services/api/src/modules/schema/providers/contracts.ts
@@ -315,17 +315,7 @@ export class Contracts {
         "target_id" = ${args.targetId}
         ${args.onlyActive ? psql`AND "is_disabled" = false` : psql``}
         ${
-          cursor
-            ? psql`
-                AND (
-                  (
-                    c."created_at" = ${cursor.createdAt}
-                    AND c."id" < ${cursor.id}
-                  )
-                  OR c."created_at" < ${cursor.createdAt}
-                )
-              `
-            : psql``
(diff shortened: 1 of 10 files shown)


packages/services/api/src/modules/schema/providers/contracts.ts
```diff
@@ -315,17 +315,7 @@ export class Contracts {
         "target_id" = ${args.targetId}
         ${args.onlyActive ? psql`AND "is_disabled" = false` : psql``}
         ${
-          cursor
-            ? psql`
-                AND (
-                  (
-                    c."created_at" = ${cursor.createdAt}
```
jdolle, line 322: This was a bug before. I guess we werent ever using this?

n1ru4l: What is the bug that was here?

n1ru4l (changes requested):
AFAIK these two approaches are functional equivalent. I personally use the AND OR over the tuple approach as it is more explicit to me. But I can be convinced to use these instead if it provides some benefit.

I will still reject this unless I am convinced to either
1. why this improves performance
2. why this is more readable and understandable (without having to open the documentation)

packages/services/api/src/modules/schema/providers/contracts.ts
jdolle: It's referencing `c.` but nothing is being aliased to `c`.

jdolle:
I'm fine closing this if we're not comfortable with the syntax. Personally I find it much easier to read since it's very compact. But if we prefer the verbose form, it's functionally equivalent.

PR #8070, opened 2026-05-24
fix(server): reproduce and fix memory leak with graphql-modules

dotansimha:
This PR reproduce and fix (by updating) a memory leak in GraphQL-Modules.

# The leak

The scenario is simple: an unexpected failure during `schema:check` call to the GraphQL API, causes the checked SDL to retain in memory forever, causing a memory leak. 
The unexpected failures can be any failing HTTP request within the execution context: either to an internal service (schema-service), or to external (clickhouse). 

Under the hood, the reason is because of NodeJS attaches a `kAsyncContextFrame` symbol when a Promise is rejected (in case of unexpected failure of schema check), this is also true in case of a global timer that uses data from within the execution scope/closure. This eventually captures the `AsyncLocalStorage` scope. 
GraphQL-Modules uses `AsyncLocalStorage` to track per-operation context. 
This combination leads to a full capture of the user-context in GraphQL-Modules, in our case, this context holds the GraphQL variables and the schema that was checked. If the checked scheam is 5MB, then 5MB will be retained and never released. 

The fix in GraphQL-Modules (https://github.com/graphql-hive/graphql-modules/pull/2681) invovles a better cleanup process, and uses referneces to point to the user-context, so cleanup process is cleaner and more safe, ensuring user-context is never captured, and is released when the operation completes.  

# How it was tested? 

To reproduce the leak, I've added an integration test that does the following:

1. Creates a full environment for testing 
2. Publishes a simple schema
3. `schema:check` a simple schema to warm up caches related to the target 
4. Creates a 5MB schema
5. Measures memory (`heapUsed`) of graphql-api service
6. Tries to `schema:check` the huge schema - this fails because the schema service worker will reject it due to size. This is the key point: in graphql-api, this will cause a rejection of the `fetch` promise. 
7. Measures memory (`heapUsed`) of graphql-api service
8. Diff and confirm that data is not retained. 

> Also, added a new CI config to run these memory leak tests in full isolation so other tests won't run against at the same time. This ensures the results are stable.

### DIFF
--- a/integration-tests/docker-compose.integration.yaml
+++ b/integration-tests/docker-compose.integration.yaml
@@ -3,7 +3,6 @@
 # This file also includes services that only exists in Hive's Cloud version, and are not available in the self-hosted version.
 # Please refer to TESTING.md for more information.
 
-version: '3.8'
 name: 'hive-tests'
 
 networks:
@@ -142,6 +141,8 @@ services:
   # Overrides only to `docker-compose.community.yaml` from now on:
   server:
     environment:
+      NODEJS_FLAGS: '--expose-gc'
+      EXPOSE_MEMORY_UTILS: '1'
       CDN_CF: '1'
       CDN_CF_BASE_PATH: http://local_cdn:3004
       CDN_CF_ACCOUNT_ID: 103df45224310d669213971ce28b5b70
--- a/packages/services/server/turbo.json
+++ b/packages/services/server/turbo.json
@@ -5,7 +5,7 @@
     "build": {
       "dependsOn": ["^build"],
       "outputs": ["dist/**"],
-      "inputs": ["../../web/app/src/gql/persisted-documents.json"]
+      "inputs": ["$TURBO_DEFAULT$", "../../web/app/src/gql/persisted-documents.json"]
     },
     "dev": {
       "persistent": true,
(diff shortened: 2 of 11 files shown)


integration-tests/docker-compose.integration.yaml
```diff
@@ -3,7 +3,6 @@
 # This file also includes services that only exists in Hive's Cloud version, and are not available in the self-hosted version.
 # Please refer to TESTING.md for more information.
 
-version: '3.8'
```
dotansimha, line 6: removed because Docker warns about it, it's no longer needed to be specified.

packages/services/server/turbo.json
```diff
@@ -5,7 +5,7 @@
     "build": {
       "dependsOn": ["^build"],
       "outputs": ["dist/**"],
-      "inputs": ["../../web/app/src/gql/persisted-documents.json"]
+      "inputs": ["$TURBO_DEFAULT$", "../../web/app/src/gql/persisted-documents.json"]
```
dotansimha, line 8: This one was missing here and was causing Turbo not to spot changes between runs of `intergration:prepare`. I think other services might need it as well. 

cc @n1ru4l @jonathanawesome

pnpm-workspace.yaml
```diff
@@ -22,3 +22,4 @@ minimumReleaseAgeExclude:
   - '@graphql-inspector/*'
   - '@graphql-tools/*'
   - '@theguild/federation-composition'
+  - graphql-modules
```
dotansimha, line 25: needed because otherwise i can't use the new version or alpha.

n1ru4l (approved):

PR #8078, opened 2026-05-26
Support for Redis AWS IAM auth

mish-elle:
Closes https://github.com/graphql-hive/console/pull/8177 

### Background

Self-hosters running Hive on AWS with AWS ElastiCache Redis currently have no way to use IAM-based authentication for Redis connections, which forces them to use static passwords.

This PR adds opt-in AWS IAM authentication support for ElastiCache Redis across all services that communicate with Redis: `api`, `schema`, `server`, `tokens`, `usage`, and `workflows`. It also adds Redis Cluster mode support. 

### Description

The ElastiCache IAM token generation logic lives in two new modules inside `service-common`:

- `service-common/src/iam-aws.ts`: A generic AWS SigV4 pre-signed token generation and a reusable periodic token refresh timer with retry/backoff logic. This is necessary since Elasticache does not have a dedicated signer like MSK or RDS.
- `service-common/src/iam-redis.ts`: ElastiCache-specific helpers built on top of `iam-aws`. Handles token generation (`generateIamAuthToken`), in-place re-authentication for both standalone and cluster connections (`refreshIamAuth`), periodic token rotation (`startIamTokenRefresh`), and credential resolution (`resolveRedisCredentials`).

When Redis IAM auth is enabled:
1. Generates an initial token at startup using `resolveRedisCredentials()`
2. Connects to Redis using the token as the password and the configured Redis username
3. Starts a background refresh timer `startIamTokenRefresh()` that rotates the token every ~12 minutes (15-min SigV4 TTL minus 3-min buffer), with jitter to prevent thundering-herd refreshes across instances
4. On each refresh, issues `AUTH` commands to re-authenticate active connections (including per-node AUTH for cluster mode)

Redis Cluster mode is added as an opt-in feature `REDIS_CLUSTER_MODE_ENABLED=1`. When enabled, services connect using `Redis.Cluster` from ioredis with `dnsLookup` configured to pass addresses through directly (required for ElastiCache's DNS-based endpoint resolution). 

Note that the underlying ioredis library does not support dynamic passwords, so using a token refresh timer is the workaround for now until ioredis supports dynamic passwords. 

- https://github.com/redis/ioredis/pull/2039

### New environment variables introduced

| Variable                     | Required | Description                                                                                     |
| ---------------------------- | -------- | ----------------------------------------------------------------------------------------------- |
| `AWS_REGION`                 | No       | Default AWS region for the service. Used as the fallback for all AWS connections.               |
| `REDIS_USERNAME`             | No       | Redis Access Control List username.                                                             |
| `REDIS_CLUSTER_MODE_ENABLED` | No       | Set to `1` to connect using Redis Cluster mode.                                                 |
| `REDIS_AWS_IAM_AUTH_ENABLED` | No       | Set to `1` to enable IAM authentication for Redis.                                              |
| `REDIS_AWS_REGION`           | No       | Optional override for the Redis region (defaults to `AWS_REGION`).                              |
| `REDIS_AWS_IAM_CACHE_NAME`   | No       | ElastiCache cache name used as the hostname for the SigV4 signer. Required when IAM is enabled. |

### Environment Variable Validation

When `REDIS_AWS_IAM_AUTH_ENABLED=1`, the environment validation enforces:
- `REDIS_TLS_ENABLED=1` (ElastiCache IAM requires TLS)
- `REDIS_AWS_IAM_CACHE_NAME` must be set
- `REDIS_AWS_REGION` or `AWS_REGION` must be set

### Pnpm-lock file generation

Like MSK, the CI pipelines will fail because of the pnpm-lock file. We're downgrading dependencies and not building a few packages, due to constraints in our environment. Is it possible for a code maintainer to help us generate the pnpm-lock file?

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/tokens/src/environment.ts
```diff
@@ -107,6 +117,21 @@ for (const config of Object.values(configs)) {
...
+  if (!configs.redis.data.REDIS_AWS_IAM_CACHE_NAME)
+    missingRedisIamVars.push('REDIS_AWS_IAM_CACHE_NAME');
+  if (!configs.redis.data.REDIS_AWS_REGION && !configs.base.data?.AWS_REGION)
+    missingRedisIamVars.push('REDIS_AWS_REGION or AWS_REGION');
+  if (missingRedisIamVars.length > 0) {
+    environmentErrors.push(
+      `REDIS_AWS_IAM_AUTH_ENABLED is enabled but the following required variables are missing or invalid: ${missingRedisIamVars.join(', ')}`,
+    );
+  }
+}
```
n1ru4l, line 117: This validation logic is added identical in a lot of services. It make sense to move the whole parsing building of the redis configuration to the service-common module.

I kind of did not flag this within https://github.com/graphql-hive/console/pull/8049, but now in this PR it gets more obvious to me.

Can you please move the environment variable parsing logic to a function within service-common and have it return a discriminated union with either the redis config object a list of error?

```typescript
type RedisConfig = {
  host: string;
  port: number;
  // etc
  clusterModeEnabled: boolean;
}


function parseRedisConfigFromEnvironment(env: typeof process.env, awsRegion?: string): {
  type: "error";
  errors: Array<string>;
} | {
  type "ok";
  config: RedisConfig;
} {}
```

packages/services/schema/src/index.ts
```diff
@@ -71,22 +73,66 @@ async function main() {
...
+  }
+
+  if (redisIamConfig) {
+    startIamTokenRefresh(
+      redis,
+      redisIamConfig,
+      env.redis.clusterModeEnabled,
+      server.log.child({ source: 'RedisIamTokenRefresh' }),
+    );
+  }
```
n1ru4l: Can you please abstract the redis client creation into service-common and share that in all services?

n1ru4l (changes requested):
Thank you for the contribution, could you please address the comments I pointed out? Things are getting a bit too much duplicated and I would prefer a central place where we manage these.

packages/services/schema/src/index.ts
mish-elle: Hi @n1ru4l, thanks for the review. I'm working on abstracting the Redis client creation into `service-common`, 
so all services now go through one shared factory instead of duplicating the connection/cluster/IAM logic.

Looking for some clarification on how far you'd like me to take the centralization:

**Option A: Abstract creation only**

Keeps `ioredis` listed as a dependency in the individual service `package.json` files. Services still
import the `Redis` type directly from `ioredis` for type annotations (e.g. constructor params,
injection tokens), but all client creation goes through the new redis client factory in `service-common`.

- Smaller diff, fewer files touched
- `ioredis` stays scattered across 7 `package.json` files as a type-only dep
  - For example, usage's `authn.ts` will still use `import type Redis from 'ioredis';`

**Option B: Abstract creation & remove all direct `ioredis` imports**

`service-common` re-exports `type Redis` from `ioredis`, and all consumer packages import the type
from `@hive/service-common` instead of directly from `ioredis`. Then remove `ioredis` from every
`package.json` except `service-common` (the single owner) and `pubsub` (which has a runtime import
for pub/sub channels).

- Additional files updated (type import changes only, no runtime behavior changes)
  - For example, usage's `authn.ts` will use `import type { Redis } from '@hive/service-common';` 
- `ioredis` lives in only 2 packages: `service-common` and `pubsub`
- But will prevent someone from accidentally importing the `Redis` constructor directly and bypassing the
  shared factory

Which would you prefer?

n1ru4l: @mish-elle I prefer Option B, since we already introduced this approach for our postgres client.

mish-elle:
Hey @n1ru4l sorry about the delay, 

- All redis client creation is abstracted to `service-commons/redis-client.ts` `createRedisClient()` function.
- Redis validations happens on `service-commons/redis-config-validation.ts`
- Additionally, while I was removing all direct `ioredis` imports there's a ioredis version mismatch between the ioredis bentocache is expecting,  `5.10.1`, and the version that the repo uses, `5.8.2`, which causes issues during runtime connection timeouts on bentocache falling back to localhost. For now I've added an pnpm override since I wasn't sure if you wanted to handle the ioredis upgrade in another git issue/PR.

packages/services/service-common/src/redis-config-validation.ts
```diff
@@ -0,0 +1,93 @@
...
+  | {
+      type: 'ok';
+      config: RedisConfig;
+    };
+
+/**
+ * Validates Redis IAM requirements and returns a normalized Redis config object.
+ */
+export function parseRedisConfigFromEnvironment(args: {
+  redis: RedisEnvironment;
```
n1ru4l: This file could contain configuration for the redis environment (so it does not need to be duplicated in each service) and then `process.env` could be the input argument for this function instead?

```ts
const RedisModel = zod.object({
  REDIS_HOST: zod.string(),
  REDIS_PORT: NumberFromString,
  REDIS_PASSWORD: emptyString(zod.string().optional()),
  REDIS_USERNAME: emptyString(zod.string().optional()),
  REDIS_TLS_ENABLED: emptyString(zod.union([zod.literal('1'), zod.literal('0')]).optional()),
  REDIS_CLUSTER_MODE_ENABLED: emptyString(
    zod.union([zod.literal('0'), zod.literal('1')]).optional(),
  ),
  REDIS_AWS_REGION: emptyString(zod.string().optional()),
  REDIS_AWS_IAM_AUTH_ENABLED: emptyString(
    zod.union([zod.literal('0'), zod.literal('1')]).optional(),
  ),
  REDIS_AWS_IAM_CACHE_NAME: emptyString(zod.string().optional()),
});
```

See what I proposed originally here: https://github.com/graphql-hive/console/pull/8078#discussion_r3322973628

Unless I am missing something obvious I would prefer this approach.

packages/services/workflows/src/index.ts
```diff
@@ -112,15 +112,19 @@ const server = await createServer({
...
 
 const pubSub = createHivePubSub({
-  publisher: redis,
-  subscriber: createRedisClient(
-    'subscriber',
-    env.redis,
-    server.log.child({ source: 'RedisSubscribe' }),
-  ),
+  publisher: redis as any,
+  subscriber: redisSubscriber as any,
```
n1ru4l: Why the `as any` it seems unneeded? 

```suggestion
  publisher: redis,
  subscriber: redisSubscriber ,
```

packages/services/service-common/src/redis-client.ts
```diff
@@ -0,0 +1,180 @@
...
+      clusterModeEnabled: config.clusterModeEnabled,
+      maxRetriesPerRequest: options.maxRetriesPerRequest,
+    },
+    options.logger,
+  );
+
+  // Start IAM token refresh if enabled
+  // Each client gets its own independent token lifecycle
+  if (redisIamConfig) {
+    const iamLogger = options.iamTokenRefreshLogger ?? options.logger;
```
n1ru4l: NIT: Instead of requiring to pass in `options.iamTokenRefreshLogger`, wouldn't it be a bit cleaner to just extend the `options.logger` like the following?

```suggestion
    const iamLogger = options.logger.extend({ source: "IAM"});
```

packages/services/service-common/src/iam-redis.ts
```diff
@@ -0,0 +1,202 @@
...
+ * @param logger - Logger with `debug` and `warn` methods.
+ */
+export async function refreshIamAuth(
+  redis: RedisInstance,
+  token: string,
+  username: string,
+  isCluster: boolean,
+  logger: { debug(...args: unknown[]): void; warn(...args: unknown[]): void },
+): Promise<void> {
+  if (isCluster) {
```
n1ru4l: Woudn't it be possible to instead do a `redis instanceof Cluster` check here to avoid the `as any` cast?

packages/services/service-common/src/redis-client.ts
```diff
@@ -0,0 +1,180 @@
...
+    startIamTokenRefresh(redis, redisIamConfig, config.clusterModeEnabled ?? false, iamLogger);
```
n1ru4l: ~~It seems like the timer this creates is not correctly stopped when the redis instance is disposed of?~~

packages/services/service-common/src/iam-aws.spec.ts
n1ru4l, line 1: @mish-elle any idea why this is a binary file? Can you fix the file type?

packages/services/service-common/src/iam-aws.ts
n1ru4l, line 1: @mish-elle any idea why this is a binary file? Can you fix the file type?

packages/services/service-common/src/iam-redis.spec.ts
n1ru4l, line 1: @mish-elle any idea why this is a binary file? Can you fix the file type?

packages/services/service-common/src/iam-redis.ts
```diff
@@ -0,0 +1,202 @@
...
+export function startIamTokenRefresh(
+  redis: RedisInstance,
+  config: IamRedisConfig,
+  isCluster: boolean,
+  logger: {
+    debug(...args: unknown[]): void;
+    warn(...args: unknown[]): void;
+    error(...args: unknown[]): void;
+  },
+): ReturnType<typeof setInterval> {
```
n1ru4l: NIT: instead of returning the raw timer return a cleanup function, that can then just be called, that way the outer API surface does not need to know about the underlying implementation.


```suggestion
): ReturnType<() => void> {


/// ...


return () => clearInterval(timer)
```

packages/services/service-common/src/iam-aws.ts
n1ru4l: Since I cannot comment the contents of this file directly due to the binary type. 😓 

I have some concerns about the implementation of the `startTokenRefreshTimer` implementation:



- The implementation may trigger a new refresh while a previous `refreshFn` (including retries) is still running, potentially causing concurrent refreshes and race conditions.
- Failures are not logged or surfaced, even after all retries are exhausted, which makes production debugging difficult.
- Random jitter is computed once at timer creation, meaning all subsequent refresh cycles use the same offset (may be unintended depending on goal).
- ~~First refresh only occurs after `intervalMs + jitterMs` - There is no initial one - do I understand that correctly?~~ Okay, i Just realised that the initial resolving happens sync within the redis client creation 👍 

I would strongly encourage you to replace the `setInterval` with a recursive `setTimeout` loop or adding an “in-flight” guard to prevent overlapping executions.

n1ru4l:
@mish-elle Please feel free to bump the `ioredis` version to `5.10.1` instead of using the override.

n1ru4l (changes requested):
Hey @mish-elle, I have some more feedback for you that I would like to see implemented. Aside from these points the implementation looks solid to me. 🙏

packages/services/service-common/src/iam-aws.ts
n1ru4l: I just realized, that no actual HTTP request is performed, correct? There is just a token generated with the HTTP presigned url SDK?

In that case please ignore this comment - no further action required here!

n1ru4l: Second take: Since we are actually also modifying the redis instances within the callback passed to `startTokenRefreshTimer` and running some commands on them, there is a potential race condition here when we shutdown, but still run commands on the redis instance? Could this become an issue, and should we maybe short-circuit the retries in the callback in that case? 🤔

packages/services/service-common/src/redis-config-validation.ts
mish-elle: So move the RedisModel into the shared service-commons? So the helper functions like `NumberFromString` and `emptyString` are currently duplicated in all service environment.ts, so we would either need to duplicate those  into the redis-config.ts or abstract it out as well?

packages/services/service-common/src/iam-aws.ts
mish-elle: > I just realized, that no actual HTTP request is performed, correct? There is just a token generated with the HTTP presigned url SDK?
> 
> In that case please ignore this comment - no further action required here!

Yes, token generation does not have actual HTTP requests performed! Just the HTTP presigned url. 

> Second take: Since we are actually also modifying the redis instances within the callback passed to `startTokenRefreshTimer` and running some commands on them, there is a potential race condition here when we shutdown, but still run commands on the redis instance? Could this become an issue, and should we maybe short-circuit the retries in the callback in that case? 🤔

Yeah, I agree, I can add this.

mish-elle: Actually, taking another pass on this, so `startIamTokenRefresh` in `iam-redis.ts` implements `refreshfn` which has a network call when redis does AUTH in `refreshIamAuth` in `iam-redis.ts`. So I'll switch to `setTimeout` recursive loop to prevent overlapping executions.

packages/services/service-common/src/redis-config-validation.ts
n1ru4l: duplicating these helpers is fine with me for now, it seems better than duplicating all the redis config

mish-elle:
> Hey @mish-elle, I have some more feedback for you that I would like to see implemented. Aside from these points the implementation looks solid to me. 🙏

Thanks for the detailed feedback, @n1ru4l! I believed I addressed everything mentioned, please let me know if we need additional changes.

n1ru4l:
@mish-elle Thank you, we are close to getting this in! 🥳 Could you please have a stab at the typescript and linting issues?

packages/services/service-common/src/iam-redis.ts
```diff
@@ -0,0 +1,200 @@
...
+      'Refreshing IAM token (service=elasticache) — re-authenticating %s cluster node(s)',
+      nodes.length,
+    );
+    for (const node of nodes) {
+      node.options.password = token;
+      try {
+        await node.call('AUTH', username, token);
+      } catch (err) {
+        logger.warn('Failed to re-AUTH cluster node (service=elasticache, error=%s)', err);
+      }
```
n1ru4l, line 162: @mish-elle Isn't this swallowing the errors and would result in no retries? Should the error be thrown instead?

packages/services/service-common/src/iam-aws.ts
```diff
@@ -0,0 +1,160 @@
...
+      void (async () => {
+        for (let attempt = 1; attempt <= maxRetries; attempt++) {
+          if (isShuttingDown) return;
+
+          try {
+            await refreshFn(attempt, maxRetries);
+            break;
+          } catch {
+            if (attempt < maxRetries) {
+              await new Promise(r => setTimeout(r, attempt * retryBackoffMs));
```
n1ru4l: This would delay the graceful shutdown in case of a retry.

I have an alternative proposal of this that is a bit more elegant and does not have this issue: https://github.com/graphql-hive/console/pull/8134

@mish-elle Let me know what you think, feel free to just cherry-pick the commit into this branch

n1ru4l:
@mish-elle I added two more comments:
- one for cluster error handling
- one alternative proposal for the abort logic

Once these two are addressed, we are good to merge 👍

mish-elle:
Actually there seems to be a minor issue with the subscriber reconnecting with a stale token, causing WRONG user/pass pair error and the container crashes, looking into it 🤔

mish-elle:
> Actually there seems to be a minor issue with the subscriber reconnecting with a stale token, causing WRONG user/pass pair error and the container crashes, looking into it 🤔

[`Cluster.nodes('all')`](https://github.com/redis/ioredis/blob/v5.10.1/lib/cluster/index.ts#L451-L457) delegates to [`ConnectionPool.getNodes()`](https://github.com/redis/ioredis/blob/v5.10.1/lib/cluster/ConnectionPool.ts#L20-L23), which only returns pool nodes discovered via CLUSTER SLOTS, it never includes the [`ClusterSubscriber`](https://github.com/redis/ioredis/blob/v5.10.1/lib/cluster/ClusterSubscriber.ts#L78-L100)'s internal Redis, which is created separately with `password: options.password` (a value copy) and stored in [`this.subscriber`](https://github.com/redis/ioredis/blob/v5.10.1/lib/cluster/index.ts#L73). 

This means `refreshIamAuth`'s loop over `nodes('all')` never updates the subscriber's password, so after a disconnect it reconnects with the stale startup token -> `WRONGPASS` -> crash. We'll add  `getClusterSubscriberInstance()` to access the internal subscriber and update its `options.password` directly (since AUTH can't be sent to a connection in SUBSCRIBE mode).

dotansimha (approved):

dotansimha:
Thanks @mish-elle ! I approved the PR, it seems like it needs a slight rebase and then we can run ci and merge :)

dotansimha:
Thanks @mish-elle , I ran the full ci checks in https://github.com/graphql-hive/console/pull/8177 , once cleared, i'll merge this one :)

PR #8079, opened 2026-05-26
Support for S3 AWS IAM auth

mish-elle:
closes https://github.com/graphql-hive/console/pull/8211

### Background

Self-hosters running Hive on AWS with AWS S3 currently have no way to use IAM-based authentication for S3 connections, which forces them to manage and rotate static access keys `S3_ACCESS_KEY_ID` / `S3_SECRET_ACCESS_KEY`.

This PR adds opt-in AWS IAM authentication support for S3 connections. When IAM auth is enabled, services authenticate to S3 using the AWS SDK default credential chain (IRSA, EKS Pod Identity, instance profile, etc.) instead of static credentials.

### Description

Since S3 connections are HTTP requests, there is no need for a token refresher. Instead, tokens are generated per request via a credential provider.

Additionally, since the console already has a custom AWS S3 client (aws4fetch) which already supports a session token, we'll build on top of that and use a credential provider interface to avoid deeply modifying the underlying pre-existing S3 client.

We created a new `createDefaultCredentialProvider()` function in `api/src/modules/cdn/providers/aws.ts` which creates an `AwsCredentialProvider` that:

- When IAM auth is enabled, uses `fromNodeProviderChain()` from `@aws-sdk/credential-providers`
  (the SDK handles credential caching and automatic refresh before expiry)
- When IAM auth is disabled, wraps the static credentials in an async provider

The `AwsClient` class is updated to accept a `credentialProvider` instead of raw credential strings. On every `sign()` call, it resolves fresh credentials from the provider.

Since the `cdn-worker` package targets Cloudflare Workers and cannot import from the AWS SDK. To maintain compatibility, the `AwsCredentialProvider` interface is duplicated in `cdn-worker/src/aws.ts`. So this supports S3 IAM Auth on the CDN when the server service has `CDN_API=1`. 

Deploying the CDN to AWS Lambda does not support passwordless connections to S3 at this time.

This PR also fixes a bug where the audit log S3 client was incorrectly using the primary S3 bucket's endpoint and bucket name instead of its own. The `AuditLogS3Config` now correctly uses `s3AuditLogs.endpoint` and `s3AuditLogs.bucketName`. Additionally, the audit log CSV upload now includes `aws: { signQuery: true }` to ensure the S3 PUT request is signed correctly.

### New environment variables

| Variable                            | Required | Description                                                          |
| ----------------------------------- | -------- | -------------------------------------------------------------------- |
| `S3_AWS_IAM_AUTH_ENABLED`           | No       | Set to `1` to enable IAM authentication for the primary S3 bucket.   |
| `S3_MIRROR_AWS_IAM_AUTH_ENABLED`    | No       | Set to `1` to enable IAM authentication for the S3 mirror bucket.    |
| `S3_AUDIT_LOG_AWS_IAM_AUTH_ENABLED` | No       | Set to `1` to enable IAM authentication for the S3 audit log bucket. |

### Environment Variable Validation

When IAM auth is **not** enabled for a given S3 bucket, the environment validation enforces that static credentials (`S3*_ACCESS_KEY_ID` / `S3*_SECRET_ACCESS_KEY`) are provided. Previously these were always required; now they are only required when IAM is disabled.

When IAM auth is enabled:
- `S3_ENDPOINT` should be set to the AWS S3 regional endpoint (e.g.
  `https://s3.us-east-1.amazonaws.com`)
- `S3_BUCKET_NAME` must be set to the AWS S3 bucket name
- The pod/instance must have AWS credentials available (e.g. IRSA, EKS Pod Identity, instance profile)

### Pnpm-lock file generation
Like MSK, the CI pipelines will fail because of the pnpm-lock file. We're downgrading dependencies and not building a few packages, due to constraints in our environment. Is it possible for a code maintainer to help us generate the pnpm-lock file?

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/app-deployments/worker/persisted-documents-worker.ts
```diff
@@ -21,20 +22,12 @@ export function createWorker(
     s3: {
       readonly bucketName: string;
       readonly endpoint: string;
-      readonly credentials: {
-        readonly accessKeyId: string;
-        readonly secretAccessKey: string;
-        readonly sessionToken: string | undefined;
-      };
+      readonly credentialProvider: AwsCredentialProvider;
```
dotansimha: Now that we've unified the S3 auth configuration under a single `type`, maybe it should be named `S3CredentialProvider`? 

The main motivation for this, is because we use this for both `s3` and for the `s3Mirror`, and S3 became a standard - it doesn't have to be used through AWS (for example: we are using it via CloudFlare R2)

packages/services/api/src/modules/cdn/providers/aws.ts
```diff
@@ -1,4 +1,127 @@
...
+ */
+export interface AwsCredentialProvider {
+  getCredentials(): Promise<AwsCredentials>;
+}
+
+/**
+ * Options for creating a credential provider via the default chain.
+ */
+export interface DefaultCredentialChainOptions {
+  /** Static credentials as fallback when IAM auth is unavailable or disabled. */
```
dotansimha: I think the comment here is a bit misleading: it's the default way of using creds, just static ones.

packages/services/api/src/modules/cdn/providers/aws.ts
```diff
@@ -1,4 +1,127 @@
...
+}
+
+/**
+ * A provider that returns AWS credentials, potentially fetching them
+ * from an external source (ECS task role, EKS Pod Identity, etc.).
+ *
+ * NOTE: This interface is also defined (structurally) in the cdn-worker package,
+ * which targets Cloudflare Workers and cannot import from the AWS SDK.
+ */
+export interface AwsCredentialProvider {
```
dotansimha: Please see my comment above regarding the naming of this interface.

packages/services/api/src/modules/audit-logs/providers/audit-logs-manager.ts
```diff
@@ -171,6 +171,9 @@ export class AuditLogManager {
           'Content-Type': 'text/csv',
         },
         body: csvData,
+        aws: {
```
dotansimha, line 174: We did not have it before - any reason to add it now? Is it something only AWS supports? or is it supported also with other vendors providing S3 services?

packages/services/api/src/modules/cdn/providers/aws.ts
```diff
@@ -1,4 +1,127 @@
...
+  const logger = options?.logger;
+  const prefix = options?.label ? `[${options.label}] ` : '';
+  const awsIamAuthEnabled = options?.awsIamAuthEnabled ?? false;
+  const { accessKeyId, secretAccessKey } = options?.staticCredentials ?? {};
+
+  let sdkProvider: () => Promise<{
+    accessKeyId: string;
+    secretAccessKey: string;
+    sessionToken?: string;
+  }>;
```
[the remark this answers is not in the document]
dotansimha: @mish-elle yeah, this needs a fix

dotansimha (dismissed):
Thanks @mish-elle , please see my comments above. Also some rebasing is needed.

dotansimha:
Hi @mish-elle , just wanted to check if you had the chance to evaluate my review feedback ^ 

thanks

packages/services/api/src/modules/audit-logs/providers/audit-logs-manager.ts
mish-elle: Hey! Without this addition we were unable to export audit logs to S3. We noticed that this aws sign happens for all S3 calls in the rest of the repo, but was missing in audit logs. 

https://github.com/graphql-hive/console/blob/main/packages/services/api/src/modules/schema/providers/artifact-storage-writer.ts#L115-L118 
https://github.com/graphql-hive/console/blob/main/packages/services/api/src/modules/app-deployments/providers/app-deployments.ts#L480-L484

mish-elle:
> Hi @mish-elle , just wanted to check if you had the chance to evaluate my review feedback ^
> 
> thanks

Hey @dotansimha, sorry about the delay! Thanks for the feedback! I believe I got everything.

n1ru4l (changes requested):
hey @mish-elle, I pushed the branch to our origin and updated the lockfile for CI tests to run. Some unit tests are currently failing. Can you please have a look at these and fix the implementation/tests?

mish-elle:
Updated! Thanks, @n1ru4l

n1ru4l (approved):
Thank you @mish-elle!

PR #8085, opened 2026-05-27
chore: temporary disable github check and publish

n1ru4l:
### Background

Until we figured out the actual issue. Let's not have CI block us.

### DIFF
--- a/.github/workflows/graphql-schema-check.yaml
+++ b/.github/workflows/graphql-schema-check.yaml
@@ -41,5 +41,5 @@ jobs:
         run: |
           curl -sSL https://graphql-hive.com/install.sh | sh
           hive schema:check "schema.graphql" ${{steps.pr-label-check.outputs.safeFlag }} \
-            --github \
+            # --github \
             --service "graphql"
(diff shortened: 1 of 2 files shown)


.github/workflows/graphql-schema-check.yaml
```diff
@@ -41,5 +41,5 @@ jobs:
         run: |
           curl -sSL https://graphql-hive.com/install.sh | sh
           hive schema:check "schema.graphql" ${{steps.pr-label-check.outputs.safeFlag }} \
-            --github \
+            # --github \
```
n1ru4l:
```suggestion
```

.github/workflows/graphql-schema-publish.yaml
```diff
@@ -24,5 +24,5 @@ jobs:
         run: |
           curl -sSL https://graphql-hive.com/install.sh | sh
           hive schema:publish "schema.graphql" \
-          --github \
+          # --github \
```
n1ru4l:
```suggestion
```

PR #8093, opened 2026-05-29
Feat/new projects screen

mskorokhodov:
### Background

Recent changes into org/project screen caused huge latency increasing
### Description

Batched clickhouse and postgres calls to provide better optimization

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


packages/services/api/src/modules/operations/providers/operations-reader.ts
n1ru4l, line 1: It would be good if we use zod for I/O decoding here instead of relying on type annotation casts

packages/services/api/src/modules/project/module.graphql.ts
```diff
@@ -83,8 +83,34 @@ export default gql`
     SINGLE @tag(name: "public")
   }
 
+  enum ProjectsSortField {
```
n1ru4l: The enum field names do not follow our convention for the public graphql schema: https://www.notion.so/theguildoss/Hive-Console-GraphQL-API-Design-Patterns-18ab6b71848a80678939c669699fe8e6?source=copy_link#1c7b6b71848a80fa91f3fd25cdb50c84

packages/services/api/src/modules/target/providers/target-stats.ts
```diff
@@ -0,0 +1,97 @@
...
+    organizationId: string;
+    projectId: string;
+    targetIds: readonly string[];
+    period: DateRange | null;
+  }): Promise<Map<string, number>> {
+    const result = new Map<string, number>();
+
+    for (const targetId of args.targetIds) {
+      result.set(targetId, 0);
+    }
```
n1ru4l: This seems not needed as the usage below does `counts.get(item.targetId) ?? 0` anyways?

packages/services/api/src/modules/target/providers/target-stats.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (args.targetIds.length === 0) {
+      return result;
+    }
+
+    const rows = await this.pool
+      .any(
+        psql`/* TargetStats.countSchemaVersionsByTargetIds */
+          SELECT
+            sv.target_id as target_id,
+            COUNT(*)::int as total
```
n1ru4l:
```suggestion
            COUNT(*) as total
```

NIT: The `::int` cast is not really needed

packages/services/api/src/modules/target/providers/target-stats.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (args.targetIds.length === 0) {
+      return result;
+    }
+
+    const rows = await this.pool
+      .any(
+        psql`/* TargetStats.countSchemaVersionsByTargetIds */
+          SELECT
+            sv.target_id as target_id,
```
n1ru4l: NIT: if possible we prefer snake case in application code


```suggestion
            sv.target_id as targetId,
```

packages/services/api/src/modules/target/module.graphql.ts
```diff
@@ -257,10 +263,36 @@ export default gql`
   }
 
   extend type Project {
-    targets: TargetConnection! @tag(name: "public")
+    targets(
+      first: Int @tag(name: "public")
+      after: String @tag(name: "public")
+      search: String @tag(name: "public")
+      sort: TargetsSortInput @tag(name: "public")
```
n1ru4l: NIT: We don't necessarily need to make these public for now.

packages/services/api/src/modules/project/module.graphql.ts
n1ru4l: If we don't make them public I am fine with keeping the name like this for now - but it seems easy to change to not do it tbh.

packages/services/api/src/modules/operations/providers/operations-reader.ts
```diff
@@ -2109,6 +2203,141 @@ export class OperationsReader {
...
+    });
+
+    const result = await this.clickHouse.query<{
+      date: string;
+      target: string;
+      total: number;
+      totalOk: number;
+      average: number;
+      percentiles: [number, number, number, number];
+    }>(query);
```
n1ru4l: Can you please use zod to decode the database result I/O?

packages/services/api/src/modules/operations/providers/operations-reader.ts
```diff
@@ -2109,6 +2203,141 @@ export class OperationsReader {
...
+      const targetRows = collection.get(row.target) ?? [];
+
+      targetRows.push({
+        date: toUnixTimestamp(row.date),
+        total: ensureNumber(row.total),
+        totalOk: ensureNumber(row.totalOk),
+        duration: toDurationMetrics(row.percentiles, row.average),
+      });
+
+      collection.set(row.target, targetRows);
```
n1ru4l: NIT: I would write it as the following, however, I doubt the current way would ever be a bottle neck anyways.

```suggestion
      let targetRows = collection.get(row.target);
      if (!targetRows) {
        targetRows = [];
        collection.set(row.target, targetRows);
      }

      targetRows.push({
        date: toUnixTimestamp(row.date),
        total: ensureNumber(row.total),
        totalOk: ensureNumber(row.totalOk),
        duration: toDurationMetrics(row.percentiles, row.average),
      });
```

packages/services/api/src/modules/target/providers/target-manager.ts
```diff
@@ -213,6 +236,274 @@ export class TargetManager {
...
+        const startIndex = targets.findIndex(target =>
+          isDesc
+            ? target.slug < cursor.slug || (target.slug === cursor.slug && target.id < cursor.id)
+            : target.slug > cursor.slug || (target.slug === cursor.slug && target.id > cursor.id),
+        );
+
+        return startIndex === -1 ? targets.length : startIndex;
+      } catch {
+        return 0;
+      }
```
n1ru4l: If we just swallow the error it might get unnoticed. 🤔

I think it is okay to raise an exception if the api consumer provides an invalid cursor.

packages/services/api/src/modules/operations/providers/operations-manager.ts
```diff
@@ -52,6 +52,27 @@ interface ReadFieldStatsOutput {
   percentage: number;
 }
 
+export type DurationAndCountOverTime = Array<{
+  date: any;
```
n1ru4l: Can this be a more conrete type? `any` is not so great and we generally want to avoid using it.

packages/services/api/src/modules/operations/providers/operations-reader.ts
```diff
@@ -2109,6 +2203,141 @@ export class OperationsReader {
...
+    clients?: readonly string[];
+    clientVersionFilters?: readonly { clientName: string; versions: readonly string[] | null }[];
+    excludeOperations?: boolean;
+    excludeClientVersionFilters?: boolean;
+  }) {
+    if (targets.length === 0) {
+      return new Map<
+        string,
+        Array<{
+          date: any;
```
n1ru4l: Can this be a more concrete type?

packages/services/api/src/modules/project/providers/project-manager.ts
```diff
@@ -278,6 +300,317 @@ export class ProjectManager {
...
+    organization: Organization,
+    args: {
+      first: number | null;
+      after: string | null;
+      search: string | null;
+      sort: ProjectsSortArgs | null;
+    },
+  ) {
+    if (!args.sort?.period) {
+      throw new GraphQLError(
```
n1ru4l: We have `HiveError` that we use throughout the codebase

n1ru4l (approved):
I added some comments, the NITs are good to have to me but not really necessary.

PR #8094, opened 2026-05-29
Stitching sdl directives

jdolle:
### Background

Stitching projects don't include custom directives in the composed SDL

### Description

We should include the directives so that the stitched gateway can operate on it


n1ru4l (changes requested):
Current tests are failing and we need an additional test that confirms this behaviour

jdolle:
I adjusted the printing logic to remove the schema definition node if it's a default definition, and added a test to show that the logic works.

n1ru4l (approved):

Issue #2993, opened 2023-10-03
sync users to Hive Console via SCIM 


n1ru4l:
# Overview

Support syncing users from external identity providers (e.g. Okta/Microsoft Entra ID/etc.) to Hive Console.

## The Problem

Manage user identities across many different applications and services is difficult. In organizations, users need access to lots of apps like Slack, Grafana, Hive Console (this is us!!!), etc.

*  User accounts must be created manually in each app 
*  Updates (name, role, department) must be synced manually
*  When someone leaves, accounts must be disabled everywhere manually

SSO (Single Sign On) only solves the login/authentication part of this story. It let’s the user log in once and then access multiple apps without having to re-entering credentials for each single one.
Also the user only starts existing in the app (Hive Console) after the first login, the only way to assign permissions before hand is to leverage the invite by email feature to assign a role/resources.
Organization administrators want to automate the provisioning and onboarding process for new organization members at scale, without leaving their identity provider interface.

## Why SCIM?

SCIM (System for Cross-domain Identity Management) is a REST(-style) API standard for automating user provisioning and deprovisioning between systems.
This standard is supported by popular identify providers like Okta and EntraID. 
In short:

* Create a new user on identity provider (Okta), REST endpoint is called on the app (Hive Console) end for creating the new user
* User is updated on identity provider (Okta), REST endpoint on the app is called for updating the users attributes.
* User is removed on identity provider (Okta), REST endpoint is called for removing the user
* User is added to a group on identity provider (Okta), REST endpoint is called for adding the user to the group
* User is removed from a group on identity provider (Okta), REST endpoint is called for removing the user from the group

In SCIM we have two types of entities, users and groups.
**SCIM Users**: These map 1:1 to users on our end. A SCIM user has a list of SCIM Groups that we need to translate to our permission system.
**SCIM Groups**: These are used for organizating users. One SCIM user can have multiple groups assigned.
It is the organization administrators responsibility to manage a mapping between SCIM Groups and role assignments on Hive Console side.

## Mapping SCIM Groups to the Hive Console permission system (Introducing Groups)

A common pattern in SaaS apps is that administrators maintain a mapping of SCIM group to the actual app permission system (instead of relying on matching to roles based on SCIM group attributes or names). Thus, it makes sense to not try to reinvent what people are familiar with.
We can introduce groups and groups can be assigned to users (or users assigned to groups).

[...]

PR #8096, opened 2026-06-01
feat: scim group and user provisioning

n1ru4l:
### Description

This PR implements https://github.com/graphql-hive/console/issues/2993, which adds:
- Syncing groups and users from a identity provider (such as Okta or Entra) to Hive Console using SCIM 2.0
- Assigning role + resource mappings to provisioned groups to grant users permissions based on assigned groups
- Configure a custom claim for the OIDC identity to SCIM provisioned user identity mapping

What is still missing to get this released?
- [x] Manual test with all the identity providers our customers will use
- [x] Ensure users personal access token permissions are restricted on groups
- [x] Ensure provisioned user is jailed to the organization (business logic + tests)
- [x] Ensure provisioned user can no be modified/altered outside SCIM (business logic + tests)
- [x] UI for configuring user id claim
- [x] Polish the UI around users and groups
- [ ] ~~I want to cleanup the logic for creating users which is right now split into our old storage module and also the new users store I introduced~~

####  Testing the Changes

- In order to use the SCIM endpoints, the organization needs to be set up to an OIDC provider.
- Once that is done you can use the base endpoint `http://localhost:3001/scim/v2/` for provisioning users and groups.
- For Authorization, use a organization access token with the permissions to Modify/Manage organization members (`scim:provision`). Send it via Bearer authentication (Header: `Authorization: Bearer ACCESS_TOKEN`).
- Use your identity provider to provision users and groups
- Within the members settings new "groups" section on the Hive dashboard you can assign role mappings to grant permissions to groups

#### Notes on the SCIM Endpoints

The SCIM endpoints have been field tested against a Okta custom integration (Note: Okta also has integration network integration, which have different behaviour. [More info here](https://developer.okta.com/docs/api/openapi/okta-scim/guides/scim-20)) and Microsoft Entra.

A quite extensive integration test suite exists for the endpoints (based on example payloads shared both in the documentation provided by Okta and Entra), though we can only gain 100% confidence by actually running a manual provisioning test against each identity provider as the protocol implementations and usages are quite a bit messy.

The endpoints are implemented as defensive as possible in an attempt to support a large variety of quirks the identity providers could have when calling the endpoints.

#### Notes on Groups

- Groups can currently only be managed (created, updated and deleted) via the SCIM endpoints.
- While groups have an `externalId` in SCIM, it seems that both Okta and Entra do not use an `externalId` and instead map the group via the `displayName` instead, thus in our implementation providing a `externalId` is optional and the `displayName` will be used as the unique property for detecting conflicts if not provided.

### Resource Links

- https://developer.okta.com/docs/api/openapi/okta-scim/guides/scim-20
- https://learn.microsoft.com/en-us/entra/identity/app-provisioning/use-scim-to-provision-users-and-groups
- https://datatracker.ietf.org/doc/html/rfc7644
- https://scim.cloud/

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


iftika1:
👋 Hi — really appreciate this SCIM work; it's been solid to integrate against. While validating against the SCIM 2.0 spec (RFC 7643/7644) we came across a few upstream observations we thought would be useful to share:

#### SCIM 2.0 conformance
- **`PATCH` isn't atomic** (RFC 7644 §3.5.2) — in `PATCH /Groups`, the property-update / member-remove / member-add steps run as separate un-wrapped statements, so a partial failure isn't rolled back (only full-replace is transactional). The spec requires all-or-nothing.
- **`totalResults` returns the page size, not the total** (RFC 7644 §3.4.2) — it's set to `resources.length`, so paginated clients under-collect / stop after the first page. Needs a `COUNT(*)` for the query.
- **`meta` is missing `location` / `created` / `lastModified`** (RFC 7643 §3.1) — user and group resources return only `meta.resourceType`.
- **User resource has no `groups[]`** (RFC 7643 §4.1.2) — the read-only `groups` attribute (the groups a user belongs to) isn't returned on `GET /Users`.

#### Personal access tokens don't honor provisioned-user group permissions
The session correctly resolves a provisioned user's permissions from groups (`loadPolicyStatementsForOrganization` → `if (user.provisionedByOrganizationId) …`), but the PAT surfaces in `organization-access-tokens.ts` all resolve from the **base** role (`membership.assignedRole.role.allPermissions` / `.resources`) with no group branch: `createPersonalAccessTokenForViewer`, `getAvailablePermissionGroupsForMembership`, `getPermissionsForAccessToken`, `computeAuthorizationStatements`, `getGraphQLResolvedResourcePermissionGroupForAccessToken`. So a provisioned user (base role `Viewer`) can't put a group-granted permission on a PAT — it's stripped at creation and again on every use (cache recompute). 
Since a provisioned user's PAT can't carry their group-granted permissions, there's no straightforward way to issue a scoped-down token for that user and exercise the group-based permissions end-to-end.

Also, question: Is single-org the intended model for SCIM-provisioned users, or is cross-org membership something you're planning to support?

mish-elle:
To add some context on Aisha's last point, our main use case for multi-org SCIM users is for Platform Operators: the team that manages all organizations on the platform i.e. setting up OIDC, etc. We could work around this by keeping Platform Operators as SSO-only users outside SCIM, but if we're externalizing all user management to SCIM then ideally Platform Operators should be managed through the same automated provisioning/deprovisioning path.

n1ru4l:
@mish-elle @iftika1 This PR focuses on provisioning users for organizations, not platform operators (which is out of scope here). It is intentional that all provisioned users are restricted/jailed to the provisioned organization.

@iftika1 I addressed the `PATCH` operation and `totalResult` concerns, thanks for reporting these!

I still feel hesitant in including the `User.groups` attribute as this adds a lot of additional DB reads for list endpoints. Both Okta and Entra work without these. Is this just a concern regarding the correctness of specification implementation or an actual blocker?

 Could you succeed testing the integration against your SCIM provider, or were the issues you described (missing `User.groups` attribute and invalid `totalResults`) preventing a successful test?

iftika1:
@n1ru4l Appreciate the quick followup. Unfortunately, `User.groups` is a real requirement for us, not just spec correctness. Our SCIM provider runs a periodic "collect entitlements" reconciliation that reads each user's current group/entitlement set from the User resource itself, so without `groups` it can't reconcile what a user holds.

n1ru4l:
@iftika1 I did  a lot of improvements to the SCIM implementation, including the groups property. Could you please re-evaluate whether the endpoints now correctly work with your provider?

integration-tests/tests/api/access-tokens/personal.spec.ts
n1ru4l, line 1: just refactoring to use the new flow test helper instead of the local mutation.

integration-tests/tests/api/auth/scim.spec.ts
n1ru4l, line 1: extensive test suite of all the scim endpoints and scim user restrictions (jail, ownership transfer etc.)

integration-tests/tests/api/oidc-integrations/crud.spec.ts
n1ru4l, line 1: just refactoring to use the new flow test helper instead of the local mutation.

integration-tests/tests/api/organization/groups.spec.ts
n1ru4l, line 1: testing all the crud logic around groups and assigning group mappings to them.

integration-tests/tests/api/user.spec.ts
n1ru4l, line 1: just refactoring to use the new flow test helper instead of the local mutation.

packages/internal/postgres/src/postgres-database-pool.ts
n1ru4l, line 1: I added support for nested transactions, these are only virtual transactions that makes it easier to use a transaction in the context of another transaction, which makes writing some code easier.

packages/migrations/src/actions/2026.06.01T00-00-00.scim-user-group-provisioning.ts
```diff
@@ -0,0 +1,91 @@
...
+export default {
+  name: '2026.05.18T00-00-00.scim-group-provisioning.ts',
+  run: ({ psql }) => psql`
+    CREATE OR REPLACE FUNCTION update_last_updated_at_column()
+    RETURNS TRIGGER AS $$
+    BEGIN
+        NEW."last_updated_at" = NOW();
+        RETURN NEW;
+    END;
+    $$ LANGUAGE plpgsql;
```
n1ru4l, line 12: It seemed to be the easiest way to consistently update the last "updated_at" column is to introduce a trigger.

packages/services/api/src/modules/auth/resolvers/Mutation/updateMe.ts
n1ru4l, line 1: moved all business logic to the `AuthManager` class.

packages/services/api/src/modules/shared/providers/storage.ts
```diff
@@ -58,7 +58,6 @@ export interface Storage {
     email: string;
     oidcIntegration: null | {
       id: string;
-      defaultScopes: Array<OrganizationAccessScope | ProjectAccessScope | TargetAccessScope>;
```
n1ru4l, line 61: unused since 1999.

packages/services/api/src/shared/helpers.ts
```diff
@@ -187,7 +188,7 @@ export function batchBy<TItem, TResult>(
   /** Function to determine the batch group. */
   buildBatchKey: (arg: TItem) => unknown,
   /** Loader for each batch group. */
-  loader: (args: TItem[]) => Promise<Promise<TResult>[]>,
+  loader: (args: TItem[]) => Promise<MaybePromise<TResult>[]>,
```
n1ru4l: I was tired of writing `ids.map(async id => records.get(id))`. Instead i wanted to just do `ids.map(id => records.get(id))` 👍

packages/services/api/package.json
```diff
@@ -13,7 +13,10 @@
     "./modules/shared/providers/logger": "./src/modules/shared/providers/logger.ts",
     "./modules/shared/providers/redis": "./src/modules/shared/providers/redis.ts",
     "./modules/schema/providers/schema-version-store": "./src/modules/schema/providers/schema-version-store.ts",
-    "./modules/token/providers/target-token-storage": "./src/modules/token/providers/target-token-storage.ts"
+    "./modules/token/providers/target-token-storage": "./src/modules/token/providers/target-token-storage.ts",
+    "./modules/organization/providers/group-store": "./src/modules/organization/providers/group-store.ts",
+    "./modules/organization/providers/group-member-store": "./src/modules/organization/providers/group-member-store.ts",
+    "./modules/organization/providers/users-store": "./src/modules/organization/providers/users-store.ts"
```
n1ru4l: vitest requires these entries

packages/services/server/src/api.ts
n1ru4l, line 1: some old logic that is very outdated

jonathanawesome (approved):

Issue #7702, opened 2026-02-17
Invitation Link not correctly send when dot is at wrong place in url


stefan-geiger-mw:
Hi team, 
we are very happy to use a self hosted hive console for our graphql. Some weeks ago, we changed our domain from 

```
https://graphql-hive-dataservicelayer-prod.prod.zzzzzzz.aa.bbbbbb.cccc/xxxxxx
```

something like 

```
https://graphql-hive-dataservicelayer-prod.proddsl.zzzzzzz.aa.bbbbbb.cccc/xxxxxx
```

(I need to mask the real domain because of sensitive data, but the length of the domain and subdomains is correct)

When I now invite colleagues, the invitation link is correctly generated https://graphql-hive-dataservicelayer-prod.proddsl.zzzzzzz.aa.bbbbbb.cccc/join/3f6f16d1c but in the email the dot between proddsl and zzzzzzz is missing and it looks like https://graphql-hive-dataservicelayer-prod.proddslzzzzzzz.aa.bbbbbb.cccc/join/3f6f16d1c

I looked at the logs of the email component and the line breaks in the logs are exactly at this place, so maybe this is related to the missing dot in the email

                 href=3D"https://graphql-hive-dataservicelayer-prod.proddsl=
.zzzzzzz.aa.bbbbbb.cccc/join/3f6f16d1c" style=3D"display:inline-block;backg=
round:#245850;color:#ffffff;font-family:Ubuntu, Helvetica, Arial, =
sans-serif;font-size:16px;font-weight:normal;line-height:120%;margin:0;text=
-decoration:none;text-transform:none;padding:10px 25px;mso-padding-alt:0px;=
border-radius:3px;" target=3D"_blank"

I would be happy to hear from you about this, it seems/feels like a bug in the hive component.

Best Stefan

stefan-geiger-mw:
I suggest the following change to fix the issue: https://github.com/graphql-hive/console/pull/7864
Today I wasn't able to test it, if someone of you have the local setup for testing it, please do so (needs some time for me to set up everything for testing and finding a suitable domain to reproduce the issue)

stefan-geiger-mw:
Second try to fix the issue here: https://github.com/graphql-hive/console/pull/8103

stefan-geiger-mw:
Unfortunately, can't find a fix for the issue.

PR #8103, opened 2026-06-02
fix: ISSUE-7702 use beautify=false for rendering html

stefan-geiger-mw:
### Background

See [ISSUE-7702](https://github.com/graphql-hive/console/issues/7702): After lengthening the deployment domain (...prod.proddsl.zzzzzzz...), invitation and password-reset emails rendered the link with a missing dot (proddslzzzzzzz), breaking the URL. A separate regression in providers.ts also caused all emails to fail with ECONNREFUSED 127.0.0.1:587.

### Description

The fix explicitly disable beautify in the mjml2html call:

```ts
  const rendered = mjml2html(body, {
    beautify: false,
    minify: false,
    minifyOptions: undefined,
  });
```

With beautify: false, MJML won't pass the HTML through js-beautify, so it won't wrap the long href attribute and the dot stays intact.

Deeper investigation: 
- Template (renderPasswordResetEmail) just injects the URL → fine.
- button() just interpolates ${input.url} into href="..." → fine.
- email() compiles via mjml2html with no beautify: false → MJML beautifies → js-beautify wraps the long line → dot split.
- The QP = soft break in the raw log is QP simply preserving/re-wrapping HTML that already contained the awkward line break.

Impact / side effects of beautify: false

- HTML is no longer pretty-printed — output becomes a single, dense block with minimal whitespace/indentation. This is purely cosmetic for the raw source; it does not affect how the email renders for recipients.
- Smaller payload — less whitespace means a slightly smaller email body.
- No more line-wrapping inside attributes — the intended fix; long URLs stay intact.
- Debugging the raw HTML is harder — the source is no longer human-readable/indented. If readability is ever needed, view it through a formatter rather than re-enabling beautify.
- No functional/visual change to the email itself — MJML's actual layout output is identical; only formatting of the source string changes.

### Checklist

- [x] Input validation
- [x] Output encoding
- [x] Session management
- [x] Access control
- [x] Cryptographic practices
- [x] Data protection
- [x] Communication security
- [x] System configuration
- [x] Database security
- [x] File management
- [x] Memory management
- [ ] Testing -> done in my DEV environment after docker build is successful

### DIFF
--- a/packages/services/workflows/__tests__/components.spec.ts
+++ b/packages/services/workflows/__tests__/components.spec.ts
@@ -2,7 +2,7 @@ import { button, email, mjml } from '../src/lib/emails/components.js';
 
 
 describe('email', () => {
-  it('preserves button URLs without modification', () => {
+  it('preserves button to click url', () => {
     const url =
       'https://graphql-hive-dataservicelayer-dev01.vvvvvv.wwwwwww.xx.yyyyyy.zzzz/auth/reset-password?rid=thirdpartyemailpassword&tenantId=public&token=xxxxxxxxxxxxxxxxxxxxxxxxxxx';
 
@@ -14,7 +14,7 @@ describe('email', () => {
     expect(html).toContain(url);
   });
 
-  it('preserves button URLs with special characters', () => {
+  it('preserves the input url itself', () => {
     const url =
       'https://graphql-hive-dataservicelayer-dev01.vvvvvv.wwwwwww.xx.yyyyyy.zzzz/auth/reset-password?rid=thirdpartyemailpassword&tenantId=public&token=xxxxxxxxxxxxxxxxxxxxxxxxxxx';
 
--- a/packages/services/workflows/__tests__/components.spec.ts
+++ b/packages/services/workflows/__tests__/components.spec.ts
@@ -1,0 +1,28 @@
+import { button, email, mjml } from '../src/lib/emails/components.js';
+
+
+describe('email', () => {
+  it('preserves button URLs without modification', () => {
+    const url =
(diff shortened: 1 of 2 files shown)


stefan-geiger-mw:
@n1ru4l I started a second try to fix the issue, since the old PR was not working for me and I was getting stuck on it. I investigated again (with the help of AI) and found the following configuration in the HTML template rendering to be changed to be able to avoid these line breaks of the url. From my perspective this is worth a try to test out, so if you could trigger the docker build again, then I can test on our setup.
Best Stefan

n1ru4l:
@stefan-geiger-mw Can please you create a unit test for the `email` function to confirm this behaviour works as expected and the url is not broken?

stefan-geiger-mw:
> @stefan-geiger-mw Can please you create a unit test for the `email` function to confirm this behaviour works as expected and the url is not broken?

Done

n1ru4l (changes requested):
I cloned the code locally and set `beautify` to `true` and could not see any failing test.

In fact, it seems like `beautify: true` even prints the following to the console:

```
email > preserves button to click url
"minify" option is not available
``` 

Thus, I do not think this is causing your issue.

stefan-geiger-mw:
> I cloned the code locally and set `beautify` to `true` and could not see any failing test.
> 
> In fact, it seems like `beautify: true` even prints the following to the console:
> 
> ```
> email > preserves button to click url
> "minify" option is not available
> ```
> 
> Thus, I do not think this is causing your issue.

You are correct. The issue is still there and I am out of ideas and time how to solve it. For now, we will live with this little bug.

PR #8117, opened 2026-06-04
App deployments CLI goodies

enisdenjo:
- `app:create --version` flag is optional, generates a random 7-char alphanumerical version
- `app:create --publish` flag to publish after creating
- `app:create ./dir` or `app:create **/*.graphql`command to generate an operations manifest from operations recursievely in a dir or glob

### TODO

- [x] docs https://github.com/graphql-hive/docs/pull/128


packages/libraries/cli/src/commands/app/genops.ts
```diff
@@ -0,0 +1,111 @@
...
+      if (!operation) {
+        this.warn(`Skipping empty operation in file "${relative(process.cwd(), file)}".`);
+        continue;
+      }
+      const hash = createHash('sha256').update(operation).digest('hex');
+      if (hash in manifest) {
+        this.warn(
+          `Hash collision detected for file "${relative(process.cwd(), file)}". The operation is identical to another operation already in the manifest. Skipping.`,
+        );
+        continue;
```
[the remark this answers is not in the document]
enisdenjo: ehh, overoptimization

.changeset/spotty-tools-say.md
n1ru4l, line 1: Regarding the `genops` name. Maybe something that includes `manifest` makes more sense? This is already a name that is used by other tooling?

- https://www.npmjs.com/package/@apollo/generate-persisted-query-manifest
- https://gql-tada.0no.co/guides/persisted-documents#generating-persisted-manifests

Some alternative naming proposal ideas:

```
hive app:manifest
hive app:create-manifest
hive app:createManifest
hive app:build
hive app:build-manifest
hive app:buildManifest
```

packages/libraries/cli/src/commands/app/genops.ts
```diff
@@ -0,0 +1,113 @@
...
+      let operation: string;
+      try {
+        // print+parse will remove comments
+        operation = print(parse(raw))
+          // remove new lines
+          .replace('\n', ' ')
+          // remove extra spaces
+          .replace(/\s+/g, ' ')
+          // trim leading and trailing spaces
+          .trim();
```
n1ru4l: If the user creates the manifest using this command it implies that he has to map application embedded operation to the hash at runtime in the network layer if he has no solution to embed the hashes directly within the code (as for example relay and codegen do). 🤔

One of the easiest solutions for that would be to embed the JSON manifest file in the built app and flip the manifest from `Record<hash, operation_contents>` to a `Record<operation_contents, hash>` map.

If we normalize the operation here that would no longer be possible and the application developer would need look at this source code of this command in order to pre-process the operation the same way we minify it here in order to yield the correct hash that is then allowed .

Thus, my proposal to not modify and take the operation document string for this command.

packages/libraries/cli/src/commands/app/genops.ts
```diff
@@ -0,0 +1,113 @@
...
+      'utf-8',
+    );
+
+    this.logSuccess(
+      `Generated persisted operations manifest with ${Object.keys(manifest).length} operation(s) to "${relative(process.cwd(), flags.out)}".`,
+    );
+  }
+}
+
+async function findGraphQLFiles(dir: string): Promise<string[]> {
```
n1ru4l: I know this is not really a concern for MCP tools - but kinda for app developers.

What about operations that are in code files? Should we use `@graphql-tools/graphql-file-loader` here?

n1ru4l (approved):
🟢 for the publish quality of life features

I flagged some things regarding the operation manifest generation command from the viewpoint of a non-MCP tool user. Please let me know what you think

packages/libraries/cli/src/commands/app/create.ts
n1ru4l, line 1: Especially for the mcp tools, i think it would be valuable if you could just pass a folder or glob instead of a .json file to the create command and that would result in a single command you can execute instead of first the manifest building command and then the create command 🤔

packages/libraries/cli/src/commands/app/genops.ts
enisdenjo: Intentionally didnt add the `@graphql-tools/graphql-file-loader` here, felt too simple to add a new dependency. Would you like me to integrate it? (Note that I have to install `@graphql-tools/load` then too)

.changeset/spotty-tools-say.md
enisdenjo: I am fine with any of them, the only preference is shorter commands haha. Maybe `manifest` or `build`?

packages/libraries/cli/src/commands/app/create.ts
enisdenjo: >  genops is intentionally a separate command because you might first want to validate the manifest before creating an app deployment, agreed?

But I am fine with merging them. Or having both? I think having both is good, yeah?

packages/libraries/cli/src/commands/app/genops.ts
enisdenjo: I dont really think that this manifest will be a replacement to any form of graphql codegen. I see it more of a tool to test out / develop faster using app deployments. Like for MCP's case, you really dont care about the hashes at all (you'll be pulling the manifest and then loading all documents on runtime).

I am trying to not overengineer this because the intention is not to replace relay or codegen. What is done now is _closest_ to codegen and matches the expectations of one using it.

If we were to use the document string as the key, what would the console UI look like when creating the app deployment displaying all those long keys?

I dont think we have to overthink it - but if you disagree, I'll do the changes. No biggie.

packages/libraries/cli/src/commands/app/genops.ts
n1ru4l: Both `@graphql-tools/graphql-file-loader` and `@graphql-tools/load` are already dependencies!

packages/libraries/cli/src/commands/app/create.ts
n1ru4l: yeah both sounds good

packages/libraries/cli/src/commands/app/genops.ts
enisdenjo: 🤦‍♂️

packages/libraries/cli/src/commands/app/genops.ts
n1ru4l: > If we were to use the document string as the key, what would the console UI look like when creating the app deployment displaying all those long keys?

I can not follow the implications of displaying long keys on Console. I simply said that the application developer that uses this command to create the manifest file that is then shared between Console and their App would:
- publish the manifest (`Record<hash, operation_contents>`) to Hive Console to register the operations
- would embed the inverted manifest within their app, to map the raw operation string to the sha hash (document_id) within the network layer to send the persisted document id instead of the raw document to the gateway/server.

For the latter, if the manifest operations are stripped of empty characters, the network layer would need to apply the same transform of stripping the whitespace to succeed in looking up the correct document id.

For codegen and relay specifically the minifaction before hashing approach works, because the hashes are embedded within the document objects that are passed to the network layer, so there is no lookup map.

______

If the sore purpose of this command is to create mcp manifest, then I think directly embedding the functionality into the `hive app:create` step is and dropping this command makes sense.

If this should be a general purpose manifest generation tool for clients that do not have a relay/codegen like integration - then I think we should not minify the operations so that the operation to sha mapping is as frictionless as possible.

enisdenjo:
ok now using the GQL file loader and everything's under the `app:create` command. Point it to a directory or give it a glob and it will find ops and create a manifest out of it.

n1ru4l (approved):
If you could add some basic integration fixture test for the glob and the other new config options on the CLI that would be great. 🙏

PR #8118, opened 2026-06-04
fix(api): validate metric alert UUID inputs

dnsnljw:
## Summary
- validate metric alert rule, channel, and saved-filter IDs before they reach Postgres UUID parameters
- return structured mutation errors for malformed metric alert IDs instead of surfacing unexpected database errors
- return null for malformed Target.metricAlertRule(id:) queries

## Tests
- added integration coverage for malformed metric alert UUID inputs in add/update/delete/query paths

Note: I could not run the full integration suite locally in this environment because pnpm was not available and package-manager download was blocked.

### DIFF
--- a/integration-tests/testkit/flow.ts
+++ b/integration-tests/testkit/flow.ts
@@ -810,6 +810,30 @@ export function readTargetInfo(
   });
 }
 
+export function readMetricAlertRule(
+  target: GraphQLSchema.TargetReferenceInput,
+  ruleId: string,
+  authToken: string,
+) {
+  return execute({
+    document: graphql(`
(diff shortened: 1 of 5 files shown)


jonathanawesome:
Thanks for this! The idea is right and matches how we validate IDs elsewhere, and it closes a real gap. A couple of notes before it can merge. 

The import is broken: we already have an `isUUID()` helper at `packages/services/api/src/shared/is-uuid.ts`, but this imports it from `../../shared/is-uuid`, which resolves to a file that doesn't exist. It just needs the correct path, `../../../shared/is-uuid` (what `id-translator` uses). 

Gemini's note about the empty-string case is a good one and worth folding in, though I'd apply the strict null check consistently across the `id` fields rather than just `savedFilterId`.

PR #8121, opened 2026-06-05
fix: apply lower case to all email invitations

n1ru4l:
### Description

When someone used a non-lowercase email for sending an invite, we stored it as such within our database. Since we always lower case the emails of signed up users, this created a conflict where an invited users email does not match the signed up user email.

By always lower-casing the invitees user email, we avoid this from happening in the first place.

### DIFF
--- a/packages/services/api/src/modules/auth/providers/email-verification.ts
+++ b/packages/services/api/src/modules/auth/providers/email-verification.ts
@@ -84,7 +84,7 @@ export class EmailVerification {
           FROM "email_verifications" "ev"
           WHERE
             "ev"."user_identity_id" = ${input.userIdentityId}
-            AND "ev"."email" = lower(${input.email})
+            AND lower("ev"."email") = lower(${input.email})
         `,
       )
       .then(v => EmailVerificationModel.nullable().parse(v));
@@ -143,7 +143,7 @@ export class EmailVerification {
           FROM "email_verifications" "ev"
           WHERE
             "ev"."user_identity_id" = ${input.userIdentityId}
-            AND "ev"."email" = lower(${superTokensUser.email})
+            AND lower("ev"."email") = lower(${superTokensUser.email})
         `,
       )
       .then(v => EmailVerificationModel.nullable().parse(v));
--- a/packages/services/api/src/modules/auth/providers/email-verification.ts
+++ b/packages/services/api/src/modules/auth/providers/email-verification.ts
@@ -216,7 +216,7 @@ export class EmailVerification {
           FROM "email_verifications" "ev"
           WHERE
             "user_identity_id" = ${input.userIdentityId}
-            AND "email" = lower(${input.email})
+            AND lower("email") = lower(${input.email})
             AND "expires_at" IS NOT NULL
             AND "verified_at" IS NULL
         `,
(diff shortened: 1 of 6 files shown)


packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -84,7 +84,7 @@ export class EmailVerification {
           FROM "email_verifications" "ev"
           WHERE
             "ev"."user_identity_id" = ${input.userIdentityId}
-            AND "ev"."email" = ${input.email}
+            AND lower("ev"."email") = lower(${input.email})
```
n1ru4l, line 87: This is lookup is okay. For each user-identity we do not have a high variety of emails

packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -143,7 +143,7 @@ export class EmailVerification {
           FROM "email_verifications" "ev"
           WHERE
             "ev"."user_identity_id" = ${input.userIdentityId}
-            AND "ev"."email" = ${superTokensUser.email}
+            AND lower("ev"."email") = lower(${superTokensUser.email})
```
n1ru4l, line 146: Same as https://github.com/graphql-hive/console/pull/8121/changes#r3363671245

packages/services/api/src/modules/auth/providers/email-verification.ts
```diff
@@ -216,7 +216,7 @@ export class EmailVerification {
           FROM "email_verifications" "ev"
           WHERE
             "user_identity_id" = ${input.userIdentityId}
-            AND "email" = ${input.email}
+            AND lower("email") = lower(${input.email})
```
n1ru4l, line 219: same as https://github.com/graphql-hive/console/pull/8121/changes#r3363671245

packages/services/api/src/modules/auth/providers/supertokens-store.ts
```diff
@@ -387,7 +387,7 @@ export class SuperTokensStore {
         , ${args.thirdPartyId}
         , ${args.thirdPartyUserId}
         , ${userId}
-        , ${args.email}
+        , lower(${args.email})
```
n1ru4l, line 390: Prevents this from happening for new users.

packages/services/api/src/modules/auth/providers/supertokens-store.ts
```diff
@@ -560,7 +560,7 @@ export class SuperTokensStore {
        'public'
        , ${args.user.userId}
        , ${args.token}
-       , ${args.user.email}
+       , lower(${args.user.email})
```
n1ru4l, line 563: Prevents this from happening for new users.

packages/services/api/src/modules/auth/providers/supertokens-store.ts
```diff
@@ -329,7 +329,7 @@ export class SuperTokensStore {
       UPDATE
         "supertokens_thirdparty_users"
       SET
-        "email" = ${args.newEmail}
+        "email" = lower(${args.newEmail})
```
n1ru4l, line 332: Prevents this from happening for new email changes.

PR #8123, opened 2026-06-05
Metric alerts saved filter wiring

jonathanawesome:
This PR wires saved filters into metric alerts and fixes the alert detail "Status transitions" bar misreporting state.

Saved filters: alert rules can now be scoped by a shared saved filter, which is applied during headless evaluation so a rule fires only on the slice of traffic the filter selects. Attachment is restricted to shared filters, the alert form lists only shared filters, and the manage-filters UI surfaces which filters are in use by an alert and disables deleting an in-use one (with server/DB guards behind the disabled UI as a backstop). The operations filter SQL builder is extracted into a shared @hive/clickhouse package so evaluation and the dashboard build the same WHERE conditions.

Status-transitions bar fix: the bar only received transitions inside the selected window, so a window with no transitions defaulted to green even when the rule was firing the entire time (e.g. the 6h view showed green while the 24h view showed red for the same period). A new MetricAlertRule.stateAt(timestamp) returns the state the rule carried into the window (the last transition before it, NORMAL if none) and the bar uses it to seed the leading edge, so the two views now agree.


packages/migrations/src/actions/2026.06.05T00-00-00.metric-alert-filter-shared-only.ts
```diff
@@ -0,0 +1,28 @@
...
+    {
+      name: 'metric_alert_rules.saved_filter_id FK -> ON DELETE RESTRICT',
+      query: psql`
+        ALTER TABLE "metric_alert_rules"
+          DROP CONSTRAINT IF EXISTS "metric_alert_rules_saved_filter_id_fkey"
+          , ADD CONSTRAINT "metric_alert_rules_saved_filter_id_fkey"
+            FOREIGN KEY ("saved_filter_id")
+            REFERENCES "saved_filters"("id")
+            ON DELETE RESTRICT
+      `,
```
n1ru4l, line 21: Doesn't this prevent project deletion from happening as the cascade delete will fail?

packages/migrations/src/actions/2026.06.05T00-00-00.metric-alert-filter-shared-only.ts
```diff
@@ -0,0 +1,28 @@
...
+  name: '2026.06.05T00-00-00.metric-alert-filter-shared-only.ts',
+  run: ({ psql }) => [
+    {
+      name: 'detach alert rules that reference private saved filters',
+      query: psql`
+        UPDATE "metric_alert_rules"
+        SET "saved_filter_id" = NULL
+        WHERE "saved_filter_id" IN (
+          SELECT "id" FROM "saved_filters" WHERE "visibility" = 'private'
+        )
```
n1ru4l: Is this really a concern? Nobody used the feature 😓

n1ru4l (approved):

PR #8129, opened 2026-06-09
unlock metric alerts across orgs for admins

jonathanawesome:
This PR ensures that the metric alerts feature is unlocked across orgs for Hive admins.


n1ru4l (approved):
Nice!

PR #8131, opened 2026-06-09
Improve clickhouse query in metric alert evaluator

jonathanawesome:
This PR speeds up the metric-alert evaluator's ClickHouse query for unfiltered rules by reading from new target-keyed rollups instead of the full per-hash/client tables.

The evaluator's window query sums across all operations for a target over a time range. The existing `operations_minutely` / `operations_hourly` tables are ordered (target, hash, client_name, client_version, timestamp) (timestamp is last) and `operations_minutely` is `PARTITION BY tuple()`, so that query can't prune by time and ends up scanning the target's entire slice (up to 24h of per-hash/client rows for a 5-minute alert).

This adds two tables fed by materialized views, `operations_minutely_by_target` and `operations_hourly_by_target`, that re-aggregate default.operations keyed on (target, timestamp) only, partitioned by hour (minutely) / by day (hourly), ordered (target, timestamp).

Routing is keyed on `filterConditions.length === 0`...rules with a saved filter keep using the existing tables, where the hash/client predicate already exploits the sort-key prefix and the rollups (which drop those dimensions) wouldn't be usable.

### DIFF
--- a/packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
+++ b/packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
@@ -1,0 +1,89 @@
+import type { Action } from '../clickhouse';
+
+// Target-keyed rollups for the metric-alert evaluator. The existing
+// `operations_minutely` / `operations_hourly` tables are ordered
+// `(target, hash, client_name, client_version, timestamp)`, so an unfiltered
+// alert window query (sum across all hashes/clients for a target over a time
+// range) can't use timestamp for index/partition pruning...`operations_minutely`
+// in particular is `PARTITION BY tuple()`, so a 5-minute alert scans the
+// target's whole 24h slice.
+//
+// These two views re-aggregate the same source (`default.operations`) keyed on
+// `(target, timestamp)` only, day-partitioned, so the same window query reads
+// just the window's granules. The alert evaluator routes unfiltered rules here
+// behind a feature flag; filtered rules keep using the existing tables (where
+// the hash/client predicate already exploits the sort-key prefix).
+//
+// The SELECTs intentionally mirror `createSelectStatementForOperationsMinutely`
+// / `...Hourly` in 004-version-2 (same aggregate-state functions) minus the
+// hash/client dimensions, so `avgMerge` / `quantilesMerge` at read time stay
+// correct.
+const createByTargetSelect = (bucket: 'toStartOfMinute' | 'toStartOfHour', where: string) => `
+  SELECT
+    target,
+    ${bucket}(timestamp) AS timestamp,
+    count() AS total,
+    sum(ok) AS total_ok,
+    avgState(duration) AS duration_avg,
+    quantilesState(0.75, 0.9, 0.95, 0.99)(duration) AS duration_quantiles
+  FROM default.operations
+  ${where}
+  GROUP BY
+    target,
+    timestamp
+`;
+
+export const action: Action = async (exec, _query, hiveCloudEnvironment) => {
+  // No historical backfill (start-fresh): the views capture inserts going
+  // forward. In prod, begin aggregating from the next UTC day boundary so the
+  // view starts on a clean partition rather than mid-minute during deploy
+  // ...same convention as 007-james-bond-aggregates-hourly-and-minutely.
+  let where = '';
+
+  if (hiveCloudEnvironment === 'prod') {
+    const tomorrow = new Date();
+    tomorrow.setUTCDate(tomorrow.getUTCDate() + 1);
+    const startOfTomorrow = tomorrow.toISOString().split('T')[0];
+
+    where = `WHERE toDate(timestamp) >= toDate('${startOfTomorrow}')`;
+  }
+
+  await Promise.all([
+    exec(`
+      CREATE MATERIALIZED VIEW IF NOT EXISTS default.operations_minutely_by_target
+      (
+        target LowCardinality(String) CODEC(ZSTD(1)),
+        timestamp DateTime('UTC') CODEC(DoubleDelta, LZ4),
+        total UInt32 CODEC(T64, ZSTD(1)),
+        total_ok UInt32 CODEC(T64, ZSTD(1)),
+        duration_avg AggregateFunction(avg, UInt64) CODEC(ZSTD(1)),
+        duration_quantiles AggregateFunction(quantiles(0.75, 0.9, 0.95, 0.99), UInt64) CODEC(ZSTD(1))
+      )
+      ENGINE = SummingMergeTree
+      PARTITION BY toYYYYMMDD(timestamp)
+      PRIMARY KEY (target, timestamp)
+      ORDER BY (target, timestamp)
+      TTL timestamp + INTERVAL 24 HOUR
+      SETTINGS index_granularity = 8192 AS
+      ${createByTargetSelect('toStartOfMinute', where)}
+    `),
+    exec(`
+      CREATE MATERIALIZED VIEW IF NOT EXISTS default.operations_hourly_by_target
+      (
+        target LowCardinality(String) CODEC(ZSTD(1)),
+        timestamp DateTime('UTC') CODEC(DoubleDelta, LZ4),
+        total UInt32 CODEC(T64, ZSTD(1)),
+        total_ok UInt32 CODEC(T64, ZSTD(1)),
+        duration_avg AggregateFunction(avg, UInt64) CODEC(ZSTD(1)),
+        duration_quantiles AggregateFunction(quantiles(0.75, 0.9, 0.95, 0.99), UInt64) CODEC(ZSTD(1))
+      )
+      ENGINE = SummingMergeTree
+      PARTITION BY toYYYYMMDD(timestamp)
+      PRIMARY KEY (target, timestamp)
+      ORDER BY (target, timestamp)
+      TTL timestamp + INTERVAL 30 DAY
+      SETTINGS index_granularity = 8192 AS
+      ${createByTargetSelect('toStartOfHour', where)}
+    `),
+  ]);
+};
--- a/packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
+++ b/packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
@@ -19,14 +19,18 @@ import type { Action } from '../clickhouse';
 // the older operations_* tables use. Splitting them lets the view and the table
 // be migrated independently.
 
-// Data table columns, with storage codecs.
+// Data table columns, with storage codecs. Percentiles use `quantilesTDigest`
+// rather than the older `quantiles` the legacy operations_* rollups use:
+// t-digest is deterministic and more accurate in the tails (what P95/P99 latency
+// alerting needs) and merges better across the MV's partial states. The reader
+// must match it with `quantilesTDigestMerge` (see `queryClickHouseWindows`).
 const tableColumns = `
   target LowCardinality(String) CODEC(ZSTD(1)),
   timestamp DateTime('UTC') CODEC(DoubleDelta, LZ4),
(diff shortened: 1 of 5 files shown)


packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,76 @@
...
+      CREATE MATERIALIZED VIEW IF NOT EXISTS default.operations_minutely_by_target
+      (
+        target LowCardinality(String) CODEC(ZSTD(1)),
+        timestamp DateTime('UTC') CODEC(DoubleDelta, LZ4),
+        total UInt32 CODEC(T64, ZSTD(1)),
+        total_ok UInt32 CODEC(T64, ZSTD(1)),
+        duration_avg AggregateFunction(avg, UInt64) CODEC(ZSTD(1)),
+        duration_quantiles AggregateFunction(quantiles(0.75, 0.9, 0.95, 0.99), UInt64) CODEC(ZSTD(1))
+      )
+      ENGINE = SummingMergeTree
```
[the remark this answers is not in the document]
jonathanawesome: https://clickhouse.com/docs/engines/table-engines/mergetree-family/summingmergetree#common-rules-for-summation

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,76 @@
...
+      CREATE MATERIALIZED VIEW IF NOT EXISTS default.operations_hourly_by_target
+      (
+        target LowCardinality(String) CODEC(ZSTD(1)),
+        timestamp DateTime('UTC') CODEC(DoubleDelta, LZ4),
+        total UInt32 CODEC(T64, ZSTD(1)),
+        total_ok UInt32 CODEC(T64, ZSTD(1)),
+        duration_avg AggregateFunction(avg, UInt64) CODEC(ZSTD(1)),
+        duration_quantiles AggregateFunction(quantiles(0.75, 0.9, 0.95, 0.99), UInt64) CODEC(ZSTD(1))
+      )
+      ENGINE = SummingMergeTree
```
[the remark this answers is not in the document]
jonathanawesome: Same as the minutely view above...this engine choice is correct. Per the ClickHouse SummingMergeTree docs: "For columns of AggregateFunction type ClickHouse behaves as AggregatingMergeTree engine aggregating according to the function."

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -266,7 +272,13 @@ export async function queryClickHouseWindows(
...
   const previousWindowStart = new Date(anchorMs - offsetMs - 2 * windowMs);
 
-  const tableName = timeWindowMinutes <= 360 ? 'operations_minutely' : 'operations_hourly';
+  // Pick the resolution table by window size, then optionally its target-keyed
+  // rollup variant. Both rollups carry the same total/total_ok/duration_*
+  // aggregate columns the SELECT below reads, so only the table name changes...
+  // and `_by_target` is day-partitioned and ordered (target, timestamp), so this
+  // window query prunes by time instead of scanning the target's whole slice.
+  const baseTable = timeWindowMinutes <= 360 ? 'operations_minutely' : 'operations_hourly';
+  const tableName = useTargetRollup ? `${baseTable}_by_target` : baseTable;
```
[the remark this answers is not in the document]
jonathanawesome: Fixed at the root rather than with the guard...removed the useTargetRollup parameter entirely and now derive the routing inside queryClickHouseWindows from `filterConditions.length === 0`.

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,76 @@
...
+      PARTITION BY toYYYYMMDD(timestamp)
+      PRIMARY KEY (target, timestamp)
+      ORDER BY (target, timestamp)
+      TTL timestamp + INTERVAL 24 HOUR
+      SETTINGS index_granularity = 8192 AS
+      ${createByTargetSelect('toStartOfMinute')}
```
jdolle: It's considered best practice to split the table creation from the materialized view. This allows for independent migrations and is therefore much easier to maintain in a production environment.

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,76 @@
...
+    quantilesState(0.75, 0.9, 0.95, 0.99)(duration) AS duration_quantiles
+  FROM default.operations
+  GROUP BY
+    target,
+    timestamp
+`;
+
+export const action: Action = async exec => {
+  // No historical backfill (start-fresh): the views capture inserts going
+  // forward, starting the moment they're created.
```
jdolle: Since this feature is new and behind a flag, this seems reasonable
But consider notifying any org using this feature about this restriction.

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,76 @@
...
+      (
+        target LowCardinality(String) CODEC(ZSTD(1)),
+        timestamp DateTime('UTC') CODEC(DoubleDelta, LZ4),
+        total UInt32 CODEC(T64, ZSTD(1)),
+        total_ok UInt32 CODEC(T64, ZSTD(1)),
+        duration_avg AggregateFunction(avg, UInt64) CODEC(ZSTD(1)),
+        duration_quantiles AggregateFunction(quantiles(0.75, 0.9, 0.95, 0.99), UInt64) CODEC(ZSTD(1))
+      )
+      ENGINE = SummingMergeTree
+      PARTITION BY toYYYYMMDD(timestamp)
```
jdolle: For minutely I recommend using `PARTITION BY toStartOfHour(timestamp)`
This gives a maximum of 25 partitions given the TTL, because one will be queued for cleanup at the start of each hour. Otherwise, you have just a single partition and the TTL cleanup will be less efficient.

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,76 @@
...
+      PRIMARY KEY (target, timestamp)
+      ORDER BY (target, timestamp)
+      TTL timestamp + INTERVAL 24 HOUR
+      SETTINGS index_granularity = 8192 AS
```
jdolle: Again for TTL efficiency, use the setting `ttl_only_drop_parts = 1`. This makes it so that it drops the entire partition at once rather than dropping chunks of records.
Since our partitions align with the TTL, this works nicely.

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -266,7 +266,18 @@ export async function queryClickHouseWindows(
...
+  // so the query prunes by time instead of scanning the target's whole slice.
+  // Filtered queries MUST stay on the legacy tables...the rollups dropped the
+  // hash/client dimensions a filter predicates on, so routing one there would
+  // reference columns that don't exist. Deriving this from `filterConditions`
+  // (rather than a separate flag/param) makes the filtered-query-on-rollup
+  // combination unrepresentable. Both rollups carry the same total/total_ok/
+  // duration_* columns the SELECT below reads, so only the table name changes.
+  const useTargetRollup = filterConditions.length === 0;
+  const baseTable = timeWindowMinutes <= 360 ? 'operations_minutely' : 'operations_hourly';
+  const tableName = useTargetRollup ? `${baseTable}_by_target` : baseTable;
```
jdolle: This code is likely fine, but if you want to adjust the name so that the time window is a postfix, then you could use `pickAggregationByPeriod` (in OperationsReader class).
Although you then might need to adjust the logic to exclude daily aggregate, since there is no `operations_daily_by_target`

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,76 @@
...
+    `),
+    exec(`
+      CREATE MATERIALIZED VIEW IF NOT EXISTS default.operations_hourly_by_target
+      (
+        target LowCardinality(String) CODEC(ZSTD(1)),
+        timestamp DateTime('UTC') CODEC(DoubleDelta, LZ4),
+        total UInt32 CODEC(T64, ZSTD(1)),
+        total_ok UInt32 CODEC(T64, ZSTD(1)),
+        duration_avg AggregateFunction(avg, UInt64) CODEC(ZSTD(1)),
+        duration_quantiles AggregateFunction(quantiles(0.75, 0.9, 0.95, 0.99), UInt64) CODEC(ZSTD(1))
```
jdolle: Per the research from @kamilkisiela here: https://app.notion.com/p/ClickHouse-quantiles-improvements-c9ae591ce4bb4e82af0b4daaeeda9621?assetsVersion=23.13.20260603.2311
we should use `quantileTDigest`

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,114 @@
...
+    PRIMARY KEY (target, timestamp)
+    ORDER BY (target, timestamp)
+    TTL timestamp + INTERVAL ${ttlInterval}
+    SETTINGS index_granularity = 8192, ttl_only_drop_parts = 1
+  `);
+
+  await exec(`
+    CREATE MATERIALIZED VIEW IF NOT EXISTS default.${table}_mv TO default.${table}
+    (
+      ${mvColumns}
```
jdolle: minor: I dont think you need the mvColumns.
If necessary, you can cast values in the select to the correct format.
E.g. `CAST(sum(total_errors) AS UInt32) as total_errors`

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -266,7 +266,24 @@ export async function queryClickHouseWindows(
...
+
+  // The two table families store the percentile column with different aggregate
+  // functions: the `_by_target` rollups use quantilesTDigest (more accurate in
+  // the tails, merges better), the legacy tables the older quantiles. A merge
+  // function must match the state function it reads, so pick it alongside the
+  // table. `total`/`total_ok`/`duration_avg` are identical across both.
+  const percentilesMerge = useTargetRollup ? 'quantilesTDigestMerge' : 'quantilesMerge';
```
jdolle, line 288: Good catch.

jdolle (approved):
Let's get another set of eyes on this but it's looking good to me.

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
n1ru4l, line 1: Are all these AI comments here needed?

packages/migrations/src/clickhouse-actions/018-metric-alert-target-rollups.ts
```diff
@@ -0,0 +1,114 @@
...
+  // forward, starting the moment they're created.
+  //
+  // Partition granularity is matched to each TTL, and `ttl_only_drop_parts = 1`
+  // (set in createRollup) makes cleanup drop whole expired partitions instead of
+  // rewriting parts to strip rows: the 24h minutely rollup partitions by hour
+  // (~25 partitions, one ages out each hour), the 30-day hourly rollup by day
+  // (~31 partitions, one ages out each day, same as operations_hourly).
+  await createRollup(
+    exec,
+    'operations_minutely_by_target',
```
n1ru4l: NIT: I would name these the following way so the prefix is more consistent

```
operations_by_target_hourly
operations_by_target_minutely
```

n1ru4l (approved):
Looks good with me. Added two comments

PR #8134, opened 2026-06-10
use abort controller for stopping pending timers

n1ru4l:
- Use an abort controller to stop a retry timer halting graceful process exit.
- Lift timer cleanup and handle it gracefully instead of using `unref`

### DIFF
--- a/packages/services/workflows/src/index.ts
+++ b/packages/services/workflows/src/index.ts
@@ -215,7 +215,7 @@ registerShutdown({
       await pg.end();
       logger.info('Shutdown postgres connection successful.');
       logger.info('Shutdown redis connection.');
-      redis.disconnect(false);
+      await Promise.all([redis.quit(), await redisSubscriber.quit()]);
       if (shutdownMetrics) {
         logger.info('Stopping prometheus endpoint');
         await shutdownMetrics();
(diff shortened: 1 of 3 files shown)


packages/services/workflows/src/index.ts
```diff
@@ -215,7 +215,7 @@ registerShutdown({
       await pg.end();
       logger.info('Shutdown postgres connection successful.');
       logger.info('Shutdown redis connection.');
-      redis.disconnect(false);
+      await Promise.all([redis.quit(), await redisSubscriber.quit()]);
```
n1ru4l:
```suggestion
      await Promise.all([redis.quit(), redisSubscriber.quit()]);
```

packages/services/workflows/src/index.ts
```diff
@@ -215,7 +215,7 @@ registerShutdown({
       await pg.end();
       logger.info('Shutdown postgres connection successful.');
       logger.info('Shutdown redis connection.');
-      redis.disconnect(false);
+      await Promise.all([redis.quit(), redisSubscriber.quit()]);
```
n1ru4l, line 218: unrelated the the actual changes 😓 - i just noticed we are not cleaning up the subscriber client

PR #8138, opened 2026-06-12
Various UI tweaks for metric alerts screens

jonathanawesome:
This PR completes a number of non-critical UI updates to the alerts screens:

- Smooths live polling so charts/tables update in place instead of flashing loading states
- Removes the brief "not found" and error-screen flashes when opening a rule
- Recomposes the detail page: presentational chart fed by one combined query, so chart + status bar update in lockstep.
- Stops the rule-form preview from polling on every keystroke; defaults range to 1 hour.
- Replaces the rules list's "Events" column with a lifetime "Incidents" count (new incidentCount field) and drops its polling.
- Adds a Destination tooltip (webhook URL / Slack channel) and a loading spinner to the rules list.
- Adds a relative-time ("Age") column to the activity/events tables.
- Keeps table pagination from resetting to page 1 on each poll.
- Tints the chart's threshold line to the rule's severity; improves Slack severity colors.
- Seeds a dummy Slack channel so the form preview has data.


n1ru4l (approved):
💅

PR #8143, opened 2026-06-12
reduce parallelism in e2e tests for ci runs

jdolle:
### Background

e2e tests can timeout randomly, causing failures. Locally, I don't experience this. I suspect the GH runner is being overwhelmed.

### Description

This change limits execution to one worker to prevent hardware overloading (and therefore test stalling), and slightly increases job timeout duration to account for the reduced parallelism.

Standard GitHub runner has 2 vCPUs.

### DIFF
--- a/.github/workflows/tests-e2e.yaml
+++ b/.github/workflows/tests-e2e.yaml
@@ -48,7 +48,7 @@ jobs:
             up -d --wait
 
       - name: playwright
-        timeout-minutes: 12
+        timeout-minutes: 15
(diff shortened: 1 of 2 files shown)


jonathanawesome (approved):

jdolle:
Another thing we can consider if the tests now take too long is to split them. e.g.
```
npx playwright test --shard=1/3
npx playwright test --shard=2/3
npx playwright test --shard=3/3
```

Issue #3010, opened 2023-10-06
use json schema for hive.json


n1ru4l:
We can declare a JSON schema for the hive.json file and also expose it via HTTP so that people can reference it via the [`$schema` property](https://json-schema.org/understanding-json-schema/reference/schema#dollarschema), for better IDE support.

CLI docs: https://the-guild.dev/graphql/hive/docs/api-reference/cli
CLI commands that reference `hive.json` keys: https://github.com/kamilkisiela/graphql-hive/blob/5bef13b15e05b639c76ea7d847045d017a12c8a7/packages/libraries/cli/src/commands
Example config file: https://github.com/kamilkisiela/graphql-hive/blob/7fafb0774fab3feefbf40c06855d8a9041b0e55a/packages/libraries/cli/hive.json

peterphitran:
Hello, I worked on a previous documentation pr would like to work on this issue, going to take a look at it!

PR #8154, opened 2026-06-16
feat(cli): add JSON Schema for hive.json config file

peterphitran:
### Background

The Hive CLI reads its settings from a `hive.json` file, but there is no
machine-readable description of that file. Editors cannot autocomplete its keys
or flag typos, and an invalid config is only surfaced at runtime as CLI error
100 ("Invalid Config Error"). 

Issue #3010 asks for a JSON Schema for `hive.json`, served over HTTP, so the file
can be referenced via the `$schema` property to get autocompletion, inline
documentation, and validation while editing.

### Description

This change affects the CLI `@graphql-hive/cli` only. No services, APIs, or the
client/agent are touched.

- `packages/libraries/cli/hive-config.schema.json` (new): a draft-07 JSON Schema.
  It is an `anyOf` union of the three shapes the `Config` class actually accepts:
  - modern: `registry` / `cdn` objects with `endpoint` + `accessToken`
  - deprecated flat: `registry` as a string plus `token`
  - namespaced: a `default` key, or a named space selected by `HIVE_SPACE`,
    wrapping either of the above
  `additionalProperties: false` on the modern and legacy branches gives typo
  detection (for example `registery` or `accesToken`).
- `packages/libraries/cli/package.json`: added the schema to the `files`
  allowlist so it ships with the published package, and added `ajv` as a
  devDependency for the test.
- `packages/libraries/cli/README.md`: documented the `$schema` line in the
  "Config file (`hive.json`)" section.
- `packages/libraries/cli/src/helpers/config.ts`: a short comment above
  `ConfigModel` / `LegacyConfigModel` reminding contributors to keep the schema
  in sync. The schema is hand-written to match the existing
  `usage-report-v2.schema.json` precedent and to avoid a `zod-to-json-schema`
  dependency.
- `packages/libraries/cli/hive.json`: added a `$schema` line to the dev config as
  a usage example.
- `.changeset/`: `@graphql-hive/cli` minor.

**Usage**

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/graphql-hive/console/main/packages/libraries/cli/hive-config.schema.json",
  "registry": {
    "endpoint": "https://app.graphql-hive.com/graphql",
    "accessToken": "..."
  }
}
```

**Testing**

- `packages/libraries/cli/tests/hive-config.schema.spec.ts` (new): compiles the
  shipped schema with ajv and runs 13 cases. It accepts all three valid shapes
  (each with and without a `$schema` line) plus an empty object, and rejects key
  typos and wrong value types.
- I also ran the same fixtures through the real `Config` class and confirmed the
  schema accepts what the CLI honors for modern and namespaced configs, and that
  adding `$schema` does not break parsing (zod strips the unknown top-level key).

**Open questions for maintainers**

1. `$schema` hosting URL: it currently points at the raw GitHub `main` file. If
   you prefer a stable or branded host (app domain, docs, or a CDN), it is a
   one-line change in three places (schema `$id`, README example, dev
   `hive.json`). I can also register `hive.json` with SchemaStore as a follow-up.
2. While testing I noticed legacy flat configs
   (`{ "registry": "<url>", "token": "..." }`) resolve to no endpoint in the
   current CLI. In `Config.read()` the legacy cache is set, then
   `ConfigModel.safeParse` runs unconditionally and throws on the string
   `registry`, and the `catch` wipes the cache. This looks like an accidental
   regression from `7fafb0774`. I kept `legacyConfig` in the schema as deprecated
   but intent-correct, and I am happy to fix `read()` in this PR or a separate one.
3. Namespaced branch: keep the `HIVE_SPACE` / `default` support in the schema, or
   trim to the documented modern shape only?
4. Dev `hive.json`: happy to drop the `$schema` line if you would rather not
   touch that file.

### Checklist

- [X] Input validation
- [X] Testing

### DIFF
--- a/packages/libraries/cli/src/helpers/config.ts
+++ b/packages/libraries/cli/src/helpers/config.ts
@@ -2,6 +2,8 @@ import fs from 'fs';
 import path from 'path';
 import { z } from 'zod';
 
+// NOTE: When changing these models, update the public JSON Schema at
+// packages/libraries/cli/hive-config.schema.json (and the SchemaStore catalog entry) to match.
 const LegacyConfigModel = z.object({
   registry: z.string().optional(),
   token: z.string().optional(),
--- a/packages/libraries/cli/src/helpers/config.ts
+++ b/packages/libraries/cli/src/helpers/config.ts
@@ -2,25 +2,56 @@ import fs from 'fs';
 import path from 'path';
 import { z } from 'zod';
 
-// NOTE: When changing these models, update the public JSON Schema at
-// packages/libraries/cli/hive-config.schema.json (and the SchemaStore catalog entry) to match.
-const LegacyConfigModel = z.object({
-  registry: z.string().optional(),
-  token: z.string().optional(),
+// These models are the single source of truth for the public JSON Schema at
+// packages/libraries/cli/hive-config.schema.json. After changing them, regenerate the schema with
+// `pnpm --filter @graphql-hive/cli generate:schema` (CI fails if the committed schema is stale).
+export const LegacyConfigModel = z.object({
+  registry: z
+    .string()
+    .describe('Deprecated. Registry endpoint URL. Use `registry.endpoint` instead.')
+    .optional(),
+  token: z
+    .string()
+    .describe('Deprecated. Registry access token. Use `registry.accessToken` instead.')
+    .optional(),
 });
 
-const ConfigModel = z.object({
+export const ConfigModel = z.object({
   registry: z
     .object({
-      endpoint: z.string().url().optional(),
-      accessToken: z.string().optional(),
+      endpoint: z
+        .string()
+        .url()
+        .describe(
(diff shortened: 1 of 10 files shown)


packages/libraries/cli/src/helpers/config.ts
```diff
@@ -2,6 +2,8 @@ import fs from 'fs';
 import path from 'path';
 import { z } from 'zod';
 
+// NOTE: When changing these models, update the public JSON Schema at
+// packages/libraries/cli/hive-config.schema.json (and the SchemaStore catalog entry) to match.
```
n1ru4l: Would it be possible to share the config definition, so it is not needed to update two places for introducing new configuration?

Could the `hive-jconfig.schema.json` be an artifact of that?

peterphitran:
Reworked the schema to be generated from the ConfigModel/LegacyConfigModel Zod
definitions (via `pnpm --filter @graphql-hive/cli generate:schema`) instead of
hand-maintained, so there's a single source of truth. A drift test fails CI if the
committed hive-config.schema.json is ever out of date with the models.

PR #8155, opened 2026-06-17
Console 2039 make metric alert notification fan out at least once with

jonathanawesome:
This PR limits metric alert notification retries and warns when a channel has stopped receiving alerts.

- Cap delivery retries at `max_attempts = 5` (was inheriting graphile-worker's default of 25, which retries a dead channel for days).
- New `metric_alert_channel_health` table: the send task records a `(rule, channel)` pair when it exhausts all retries for an event, and clears it on the next successful delivery.
- Expose `MetricAlertRule.degradedChannels`
- UI: a warning banner on the rule detail page and a warning icon + tooltip on the Destination field in the conditions panel and rules list. Non-blocking...the channel stays attached and future alerts still fire.

Screenshot 2026-06-16 at 8 14 46 PM
Screenshot 2026-06-16 at 8 15 47 PM

### DIFF
--- a/packages/migrations/src/actions/2026.06.16T00-00-00.metric-alert-channel-health.ts
+++ b/packages/migrations/src/actions/2026.06.16T00-00-00.metric-alert-channel-health.ts
@@ -1,0 +1,20 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2026.06.16T00-00-00.metric-alert-channel-health.ts',
+  run: ({ psql }) => psql`
+-- Per-(rule, channel) record of a dropped delivery: upserted when a channel
+-- exhausts its retries for an event, deleted on the next success. A sibling
+-- table (not a column on "metric_alert_rule_channels") so it survives that
+-- table's DELETE+INSERT churn when a rule's channels are edited.
+CREATE TABLE "metric_alert_channel_health" (
+  "metric_alert_rule_id" uuid NOT NULL REFERENCES "metric_alert_rules"("id") ON DELETE CASCADE,
+  "alert_channel_id"     uuid NOT NULL REFERENCES "alert_channels"("id")    ON DELETE CASCADE,
+  "degraded_at"          timestamptz NOT NULL DEFAULT now(),
+  "last_error"           text,
+  PRIMARY KEY ("metric_alert_rule_id", "alert_channel_id")
+);
+
+CREATE INDEX "idx_metric_alert_channel_health_channel" ON "metric_alert_channel_health" ("alert_channel_id"); -- FK-cascade index
+`,
+} satisfies MigrationExecutor;
--- a/packages/services/api/src/modules/alerts/providers/metric-alert-rules-storage.ts
+++ b/packages/services/api/src/modules/alerts/providers/metric-alert-rules-storage.ts
@@ -149,6 +149,18 @@ const AlertChannelRowSchema = zod.object({
   webhookEndpoint: zod.string().nullable(),
 });
 
+export type DegradedChannel = {
+  channel: AlertChannel;
+  degradedAt: string;
+  lastError: string | null;
+};
+
+const DegradedChannelRowSchema = AlertChannelRowSchema.extend({
+  ruleId: zod.string(),
+  degradedAt: zod.string(),
+  lastError: zod.string().nullable(),
+});
+
 const METRIC_ALERT_STATE_LOG_SELECT = psql`
   "id"
   , "metric_alert_rule_id" as "metricAlertRuleId"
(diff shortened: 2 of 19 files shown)


packages/migrations/src/actions/2026.06.16T00-00-00.metric-alert-channel-health.ts
```diff
@@ -0,0 +1,20 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2026.06.16T00-00-00.metric-alert-channel-health.ts',
+  run: ({ psql }) => psql`
+-- Per-(rule, channel) record of a dropped delivery: upserted when a channel
+-- exhausts its retries for an event, deleted on the next success. A sibling
```
jdolle, line 7: I'm not sure I agree with deleting this record on the next success.
Are there other ways of indicating to the user that this alert was delayed due to an internal issue? I think this is valuable info if they think they should have been alerted, but weren't. It provides confidence and accountability.

packages/services/api/src/modules/alerts/providers/metric-alert-rules-storage.ts
```diff
@@ -225,6 +237,47 @@ export class MetricAlertRulesStorage {
...
+          , to_json(ac."created_at") as "createdAt"
+          , ac."slack_channel" as "slackChannel"
+          , ac."webhook_endpoint" as "webhookEndpoint"
+        FROM "metric_alert_channel_health" h
+        INNER JOIN "metric_alert_rule_channels" rc
+          ON rc."metric_alert_rule_id" = h."metric_alert_rule_id"
+          AND rc."alert_channel_id" = h."alert_channel_id"
+        INNER JOIN "alert_channels" ac ON ac."id" = h."alert_channel_id"
+        WHERE h."metric_alert_rule_id" = ANY(${psql.array([...ruleIds], 'uuid')})
+          AND h."degraded_at" > now() - interval '24 hours'
```
jdolle, line 262: why limit to last 24hr?

packages/services/api/src/modules/alerts/resolvers/MetricAlertChannelDelivery.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Note: This object type is generated because "MetricAlertChannelDeliveryMapper" is declared. This is to ensure runtime safety.
+ *
+ * When a mapper is used, it is possible to hit runtime errors in some scenarios:
+ * - given a field name, the schema type's field type does not match mapper's field type
+ * - or a schema type's field does not exist in the mapper's fields
+ *
+ * If you want to skip this file generation, remove the mapper or update the pattern in the `resolverGeneration.object` config.
+ */
+export const MetricAlertChannelDelivery: MetricAlertChannelDeliveryResolvers = {
+  /* Implement MetricAlertChannelDelivery resolver logic here */
```
jdolle, line 13: should this be implemented?

packages/services/workflows/src/tasks/send-metric-alert-channel-notification.ts
```diff
@@ -226,6 +232,31 @@ export const task = implementTask(SendMetricAlertChannelNotificationTask, async
...
         span.setAttribute('error.type', err instanceof Error ? err.name : 'unknown');
+
+        // Final attempt failed → the delivery is dropped. Record the (rule,
+        // channel) pair as degraded so the rule UI can warn. Best-effort: a
+        // failure here must never mask the original send error.
+        if (helpers.job.attempts >= helpers.job.max_attempts) {
+          outcome = 'degraded';
+          try {
+            const lastError = (err instanceof Error ? err.message : String(err)).slice(0, 500);
+            await context.pg.query(psql`
```
jdolle, line 243: i wonder if instead of setting the retry limit on the queue to handle limiting a requests to dead channels, if this should just handle this case here and return a success. The distinction being that if this query fails, then the job would not be retried in the current configuration. But if you keep the queue retry count at the default, and handle the logic here instead, that if this fails it will be retried up to 20 more times so ensure we've recorded the degradation.

packages/web/app/src/pages/target-alerts-detail.tsx
```diff
@@ -67,6 +68,15 @@ const TargetAlertsDetailPage_RuleConfigQuery = graphql(`
           name
           type
         }
+        degradedChannels {
```
jdolle, line 71: instead of a new type, should this be part of the AlertChannel type? e.g.

```
interface AlertChannel {
  id: ID!
  name: String!
  type: AlertChannelType!
  degradedAt: DateTime
  lastError: String
}
```

packages/migrations/src/actions/2026.06.16T00-00-00.metric-alert-channel-health.ts
jonathanawesome: What's the alternative here? Keeping the record but marking it as non-degraded after the next successful delivery?

packages/services/api/src/modules/alerts/providers/metric-alert-rules-storage.ts
jonathanawesome: It's a backstop against rules that don't fire often. But now that I think about it...this should indicate that the _channel_ is degraded rather than the rule + channel combination.

packages/services/api/src/modules/alerts/resolvers/MetricAlertChannelDelivery.ts
jonathanawesome: This is generated code because of the mapper.

PR #8174, opened 2026-06-24
Console 2045 UI for enabling/disabling a metric alert rule

jonathanawesome:
This PR adds a toggle to enable/disable a metric alert rule, on both the alert rules list and the rule detail page. 

- adds a new base-ui sourced Switch component
- optimistic update for the toggle

https://github.com/user-attachments/assets/8d94312b-82ff-4f84-b6b0-0aff7b5a4b59


packages/web/app/src/pages/target-alerts-rules.tsx
```diff
@@ -154,6 +159,21 @@ function SortableHeader({ column, label }: { column: Column<RuleRow, unknown>; l
 const columnHelper = createColumnHelper<RuleRow>();
 
 const RULE_COLUMNS: ColumnDef<RuleRow, any>[] = [
+  columnHelper.display({
+    id: 'enabled',
+    header: 'Enabled',
+    enableSorting: false,
+    cell: ctx => (
+      <span onClick={e => e.stopPropagation()}>
```
jdolle: minor: should AlertRuleEnabledToggle stop the propagation instead to save on having yet another hook?

packages/web/app/src/pages/target-alerts-rules.tsx
```diff
@@ -332,7 +357,7 @@ export function TargetAlertsRulesPage(props: {
         }
       />
 
-      {result.fetching && !data ? (
+      {!data ? (
```
jdolle: if this errored and therefore doesn't have data, but is done fetching, then this will still show the spinner.

I think you want to keep the fetching check

packages/web/app/src/components/target/alerts/alert-conditions-panel.tsx
```diff
@@ -278,6 +280,21 @@ export function AlertConditionsPanel({
         ]}
       />
 
+      <div className="border-neutral-5 flex items-center justify-between border-y py-4">
+        <span className="flex flex-col gap-0.5">
+          <span className="text-neutral-12 text-sm font-medium">Alert status</span>
+          <span className="text-neutral-10 text-xs">
+            {rule.enabled ? 'Notifications are active' : 'Notifications are paused'}
```
jdolle: This mentions notifications, but the "enabled" boolean is on the rule itself.
It's unclear from this UI if this only impacts the alert, or also stops the evaluation. It's a subtle difference, but I'd expect if I'm disabling the alert, that I'd still have an history on the alert for when it evaluated to true, but it didn't notify anyone.

packages/web/app/src/components/target/alerts/alert-rule-enabled-toggle.tsx
```diff
@@ -0,0 +1,64 @@
...
+              },
+            },
+            ruleId: props.ruleId,
+            enabled: checked,
+          },
+        }).then(result => {
+          const message =
+            result.error?.message ?? result.data?.updateMetricAlertRule.error?.message;
+          if (message) {
+            toast({
```
jdolle, line 55: In general we can do a better job handling errors. Toasts are cheap and easy but aren't great. Is there a way to integrate the error message into the UI better?

packages/web/app/src/pages/target-alerts-rules.tsx
```diff
@@ -154,6 +159,21 @@ function SortableHeader({ column, label }: { column: Column<RuleRow, unknown>; l
 const columnHelper = createColumnHelper<RuleRow>();
 
 const RULE_COLUMNS: ColumnDef<RuleRow, any>[] = [
+  columnHelper.display({
```
jdolle: Personally I don't like the toggle on this page. I think it's too easy to click and I don't see a scenario where people need to toggle so many at once.
I think it's better to have the toggle only on the individual alert page.

packages/web/app/src/components/target/alerts/alert-conditions-panel.tsx
jonathanawesome: The toggle is for evaluation _and_ notifications...I'll clear up this copy.

 If a user wants to disable notifications but keep the rule evaluating and reporting via the UI, they're able to do so by removing destinations.

packages/web/app/src/pages/target-alerts-rules.tsx
jonathanawesome: This is a fair point. I'll remove the column but add a badge next to the rule name if it's disabled.

packages/web/app/src/components/target/alerts/alert-rule-enabled-toggle.tsx
jonathanawesome: It's a good point but, until we have a better common pattern here, I think the toast is appropriate.

jdolle (approved):

PR #8182, opened 2026-06-29
feat: show criticalityReason context for safe schema changes based on usage

### LINKED ISSUE #6892: consider giving more context for safe schema change based on usage
If a schema change is safe because of the client inclusion list we could give that information in the schema version/schema check view.

jenishasubedi05:
### Background

Closes #6892

When a schema change is marked as "safe" due to conditional breaking change settings (usage threshold or client exclusion), the schema check view gives no explanation of why. Developers have to manually cross-reference the target's Conditional Breaking Changes configuration to understand why a change passed.

### Description

This PR surfaces the existing `criticalityReason` field through to the frontend UI.

**Changes:**
- `packages/services/api/src/modules/schema/providers/inspector.ts` — Added `reason: change.criticality.reason` to pass the criticality reason through the model
- `packages/web/app/src/components/target/history/errors-and-changes.tsx` — Added `criticalityReason` to both GraphQL fragments and displays it inline next to "Safe based on usage data" in the schema check results view

### Checklist

- [ ] Tests added/updated
- [x] `pnpm graphql:generate` run after GraphQL changes
- [x] No breaking changes


jenishasubedi05:
Hi! I've addressed all the review feedback from the bot:
- Removed the temporary bundled .mjs build artifacts
- Reverted the prebuild script change in package.json
- Fixed indentation in errors-and-changes.tsx
- Removed schema-v1.graphql test file

Ready for another look when you have time!

jonathanawesome:
Thanks for the contribution, and for addressing the bot's earlier feedback! Good feature idea. A few things to sort out before it can go in.

Blocker: the import line in `errors-and-changes.tsx` reads AccordionItem,2 (stray 2).

Use `severityReason` instead. `criticalityReason` is deprecated in favor of `severityReason`, and both resolve to the same backend value. This component already fetches and renders `severityReason`, so you can build the same UI from it without adding a new fragment field. That should shrink the change down to roughly just the new `<span>`.

A question on the other changes: five of the six files aren't related to this feature, and they look like Windows/Docker workarounds: the ClickHouse named-volume swap in `docker-compose.dev.yml`, plus the path-separator and fileURLToPath tweaks in `vite.config.ts`, `runify.ts`, `build-example.mjs`, and `server/index.ts`. Are you on Windows and running into issues getting Console working locally? If so, these fixes could be genuinely useful, but we'd like them in a separate PR dedicated to Windows dev support so we can review them properly. Could you pull them out of this one?

Minor: clean up the leftover blank lines after the new `<span>` and in the fragment. Also, the description references `inspector.ts`, which isn't in the diff.

Since this is a UI change, could you also add before/after screenshots to the PR description? That makes it much easier to review the visual result.

Happy to re-review once the scope is trimmed. Thanks again!

PR #8183, opened 2026-06-30
Maintain scroll pos on explorer

jdolle:
### Background

Due to how elements are loaded on our target explorer page, the scroll position gets lost when navigating back from another page. I.e. if I explore insights on a type then press the back button, then the scroll position is lost

### Description

This implements a custom hook to store the scroll position on this page. Tanstack has its own way of managing scroll position but nothing I did was letting me hook into their state. Also, we disabled tanstack's auto handling of scroll restoration.

This hook was the most straight forward and least brittle solution.


packages/web/app/src/pages/target-explorer.tsx
```diff
@@ -220,7 +225,7 @@ function ExplorerPageContent(props: {
           )}
         </div>
       </div>
-      {!query.fetching && !query.stale ? (
+      {!query.fetching ? (
```
jonathanawesome, line 228: why remove the stale check here?

packages/web/app/src/components/target/explorer/scroll-restoration.ts
```diff
@@ -0,0 +1,70 @@
...
+    let timeoutId: number;
+    const handleScroll = () => {
+      window.clearTimeout(timeoutId);
+      timeoutId = window.setTimeout(() => {
+        const cacheKey = `__hive_${currentPath}`;
+        safeSessionStorage.setItem(cacheKey, window.scrollY.toString());
+      }, 150);
+    };
+
+    window.addEventListener('scroll', handleScroll, { passive: true });
```
jonathanawesome: We only need this position when we leave the page, right? Do we even need this listener running?

Pseudo-code to lean out this hook:

```export function useScrollRestoration(isReady: boolean) {
  const { pathname } = useLocation();
  const key = `__hive_scroll_${pathname}`;
  const hasRestored = useRef(false);

  // save scroll position when leaving the page (unmount or path change).
  useEffect(() => {
    hasRestored.current = false;
    return () => {
      try {
        sessionStorage.setItem(key, String(window.scrollY));
      } catch {}
    };
  }, [key]);

  // restore once, pre-paint
  useEffect(() => {
    if (!isReady || hasRestored.current) {
      return;
    }
    hasRestored.current = true;

    const frame = requestAnimationFrame(() => {
      let saved: string | null = null;
      try {
        saved = sessionStorage.getItem(key);
      } catch {}
      window.scrollTo({ top: saved ? parseInt(saved, 10) : 0, left: 0, behavior: 'instant' });
    });

    return () => cancelAnimationFrame(frame);
  }, [isReady, key]);
}

jonathanawesome:
Ideally we'd be using the Tanstack built-in for this, but that requires moving to using the route loader which is a bigger pattern refactor.

jonathanawesome (changes requested):

packages/web/app/src/pages/target-explorer.tsx
jdolle: The stale check prevents us from rendering the cached data for this page. As you can imagine, this impacts the time to load significantly if you're going back from a page.

I thought initially this was the cause of the position getting lost, which is why I removed it, but there were other causes as well. I opted to still remove this check so that the page loads quicker using the response cache.

jonathanawesome (approved):

Issue #130, opened 2022-05-31
Support Discord as a first-class notification/alert target


n1ru4l:
Currently, it is only possible to integrate notifications/alerts into Slack by using a webhook proxy. It might make sense to have a first-class integration for Discord.

kmcgill88:
I have a need for Microsoft Teams. I'm happy to contribute; would you be open to PR's?

kamilkisiela:
we're always open for PRs, yes.

capaj:
@kmcgill88 PR for MS Teams is opened, should land soon

ben-m-klein:
Hey all!
Would love to tackle this as my first issue in this codebase. Let me know if you have any pointers or if there are any gotchas to be aware of. I'll be sure to reference the other notification/alerts and the contributing guide as I work.

jonathanawesome:
Completed in #8184...thanks @ben-m-klein!

PR #8184, opened 2026-06-30
Add Discord webhook alert channel support

ben-m-klein:
### Background

This is closing issue #130.

Hive supports generic webhooks and Slack/MS Teams alerts, but Discord alerts currently don't work with the generic webhook. This PR adds a Discord compatible webhook option.

### Description

The implementation pattern matches the existing MS Teams alerts:

- Adds `DISCORD_WEBHOOK` channel type
- Uses discord specific webhook formatting for confirmations and alert notifications
- UI drop down option added in alerts section
- Adapter tests included.

### Checklist

- [x] Input validation
- [x] Output encoding
- [x] Testing

Screenshot 2026-06-30 134159

Screenshot 2026-06-30 134315

Screenshot 2026-06-30 134354


packages/services/api/src/modules/alerts/providers/adapters/discord.ts
```diff
@@ -0,0 +1,274 @@
...
+                inline: true,
+              },
+              {
+                name: 'Target',
+                value: input.event.target.name || input.event.target.slug,
+                inline: true,
+              },
+              {
+                name: 'Commit',
+                value: input.event.schema.commit,
```
jonathanawesome: Unlike the Project and Target fields just above, `Commit` has no fallback. Discord rejects an embed whose field `value` is an empty string with a 400 ("Cannot send an empty message" / invalid form body), which is stricter than the Teams MessageCard path that tolerates empty facts. `commit` is almost always populated for a schema notification, so this is low-probability...but a `|| '—'` (or dropping the field when empty) would make it bulletproof and match the fallback pattern used for Project/Target.

jonathanawesome (changes requested):
Thanks for the contribution...this is clean and follows the existing MS Teams pattern nicely. 🐝

One gap to flag: Discord is now selectable as a metric-alert destination, a new feature that's currently hidden behind a feature flag, but the live notification preview in the rule form wasn't updated for it...AlertPreview in `alert-notification-preview.tsx` only handles SLACK / WEBHOOK / MSTEAMS_WEBHOOK and falls through to `default: return null`, so picking a Discord channel renders a blank preview. A DiscordPreview branch mirroring the TeamsPreview (title + severity color bar + Type/Severity/Target fields) is needed.

To repro/test locally, `scripts/seed-alerts-live/README.md` walks through enabling the metric-alerts feature flag and seeding a live alert (you won't need to ).

packages/services/api/src/modules/alerts/providers/adapters/discord.ts
ben-m-klein: Awesome, thanks for taking a look! I've added a metric alert preview and a fallback (|| '—') in the latest commit, let me know if you notice anything else missing.

alert preview

packages/migrations/src/actions/2026.06.22T20-10-00.discord-webhook.ts
```diff
@@ -0,0 +1,8 @@
+import { type MigrationExecutor } from '../pg-migrator';
+
+export default {
+  name: '2026.06.22T20-10-00.discord-webhook.ts',
+  run: ({ psql }) => psql`
+    ALTER TYPE alert_channel_type ADD VALUE 'DISCORD_WEBHOOK';
```
jdolle:
```suggestion
    ALTER TYPE alert_channel_type ADD VALUE IF NOT EXISTS 'DISCORD_WEBHOOK';
```
minor: This allows us to force rerun migrations without side effects

jdolle: minor: I think just `DISCORD` would be better everywhere this is used.

ben-m-klein: Thanks for the review! I've updated the enum name to 'DISCORD'.

jonathanawesome: @ben-m-klein this hasn't been resolved

ben-m-klein: @jonathanawesome Fixed now, I've added 'IF NOT EXISTS'.

ben-m-klein:
There was a CI issue earlier:
```
 usage-ingestor Error manifest unknown
...
Error response from daemon: manifest unknown
Error: Process completed with exit code 1.
```
It looks like an issue with the container image.
I just wanted to check to see if there is anything I need to do on my end to resolve it, or if this is to be expected?

jonathanawesome (approved):

PR #8198, opened 2026-07-07
Usage estimator performance

jdolle:
### Background

I've noticed several spikes in IOWait duration in Clickhouse. This could be caused by a number of things, but one of them is if the read amount is excessive.

Usage estimator polls for usage data using our `operations_hourly` table. This table has an unoptimized order for querying a target's operation count without any filters.

### Description

This adjusts the operation to use the new table that is ideal for this query.

This results in significantly fewer rows being read and a much faster response.

**Before:**
usage-estimator-before

**After:**
usage-estimator-after


packages/services/commerce/src/usage-estimator/estimator.ts
```diff
@@ -50,6 +51,11 @@ export function createEstimator(config: {
         },
       });
 
+      const tableName =
+        input.startTime > env.hiveServices.operationsByTargetTableCreatedAt
```
jdolle: we could also query the table to get the first timestamp. That way this behavior would be automatic but I like the simplicity and speed of this check the way it's implemented

packages/services/commerce/src/environment.ts
```diff
@@ -82,6 +82,7 @@ const PostgresModel = zod.object({
 
 const HiveServicesModel = zod.object({
   WEB_APP_URL: emptyString(zod.string().url().optional()),
+  OPERATIONS_BY_TARGET_TABLE_CREATED_AT: zod.coerce.date().optional().default(new Date()),
```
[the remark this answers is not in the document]
jdolle: the goal of using `new Date()` is to avoid any issues with self hosters who may not have this table's data populated yet.

kamilkisiela:
I think we could use `monthly_overview` instead.

The query you're optimizing is from `estimateOperationsForAllTargets` and it's called only by tRPC's `estimateOperationsForAllTargets` procedure and finally by:
https://github.com/graphql-hive/console/blob/27a58e53c2c37998c2351c697288858a56f73287/packages/services/commerce/src/rate-limit/limiter.ts#L80

The time window there looks like this:
https://github.com/graphql-hive/console/blob/27a58e53c2c37998c2351c697288858a56f73287/packages/services/commerce/src/rate-limit/limiter.ts#L70-L73

And we sum the targets by organization:
https://github.com/graphql-hive/console/blob/27a58e53c2c37998c2351c697288858a56f73287/packages/services/commerce/src/rate-limit/limiter.ts#L118

so yeah, we could do:
```sql
SELECT
    sum(total) as total,
    organization
FROM
    monthly_overview PREWHERE date = '2026-07-01'
GROUP BY
    organization
# Elapsed: 0.053s Read: 2,650 rows (38.85 KB)
```

jdolle:
Nice. Even better.

Screenshot 2026-07-09 at 1 21 49 PM

packages/services/commerce/src/usage-estimator/index.ts
```diff
@@ -17,8 +17,8 @@ export const usageEstimatorRouter = router({
       const now = new Date();
       const rangeValidation = z
         .number()
-        .min(now.getFullYear() - 1)
-        .max(new Date(now.setMonth(now.getMonth() + 1)).getFullYear());
+        .min(now.getUTCFullYear() - 1)
```
jdolle, line 20: We should always use UTC to avoid timezone issues.

packages/services/commerce/src/rate-limit/limiter.ts
```diff
@@ -113,10 +116,7 @@ export function createRateLimiter(config: {
       }
 
       const orgRecord = newCachedResult.get(record.organization)!;
-
-      for (const target of record.targets) {
-        orgRecord.operations.current += operations[target] || 0;
-      }
+      orgRecord.operations.current = operations[record.organization];
```
n1ru4l: I think it makes sense to retain the fallback value to not compare against `undefined` below?

```suggestion
      orgRecord.operations.current = operations[record.organization] ?? 0;
```

n1ru4l (approved):
Approved with one comment (disabled auto-merge): https://github.com/graphql-hive/console/pull/8198/changes#r3593487373

PR #8201, opened 2026-07-09
Additional metric alert clickhouse performance improvements

jonathanawesome:
This PR flattens the climbing ClickHouse query duration and total run time of the `evaluateMetricAlertRules` cron by cutting how often long-window rules run, how many buckets each query scans, and how much work each query does. It also adds Grafana dash panels to keep watch on the number of rules with filters attached (expensive).

- **Cadence gating**: long-window rules evaluate less often (30-day rules ~every 30 min instead of every 60s) via `last_evaluated_at`, keeping full 1-min resolution during PENDING/RECOVERING so confirmation dwell is unaffected.
- **Daily rollup + routing**: new `operations_by_target_daily` ClickHouse rollup; windows ≥7 days read a daily table (~60 buckets for a 30-day query instead of ~1,440), fixing both per-query cost and the 30-day hourly-retention cliff for long windows.
- **Traffic accuracy**: TRAFFIC-count rules stay on the hourly table, because a rolling window's edges don't fall on day boundaries, so a daily aggregate rounds them to whole days and adds or drops up to a full edge day of traffic (a real error for an absolute count threshold, though it barely moves a latency percentile); groups split by rollup tier so a co-located latency rule still gets the daily win.
- **Skip the previous window** for absolute-only groups (1× scan instead of 2×); FIXED_VALUE rules persist a `null` previous value consistently regardless of group-mates so the UI responds accordingly.
- **Skip unused duration columns** for error/traffic groups, so ClickHouse never reads the heavy percentile blob when no latency rule needs it.
- **Concurrency pool** replaces the batch barrier so one slow group can't stall cheaper ones (including across orgs).
- **Whole-day validation**: windows ≥7 days must be a whole number of days, guarding direct API callers so daily buckets never silently round a window.
- **Fail loud** if a latency rule ever reads a duration column its query didn't select, instead of silently reading 0 (which renders as 0ms and never fires). Not reachable today, but the column-skipping optimization above is what makes a select/read mismatch possible, so this guards against a future desync.
- **Observability**: added data to the evaluate-group spans (metric/threshold types, whether a filter is applied) so slow traces are self-describing in TraceQL, plus filtered-vs-unfiltered rule and group Prometheus gauges with matching Metric-Alerts dashboard panels.
- **Deploy ordering**: the workflows service now deploys after DB migrations so the daily table exists before the new route goes live.


packages/migrations/src/clickhouse-actions/019-metric-alert-target-daily-rollup.ts
```diff
@@ -0,0 +1,49 @@
...
+        target,
+        toStartOfDay(timestamp) AS timestamp,
+        CAST(count() AS UInt32) AS total,
+        CAST(sum(ok) AS UInt32) AS total_ok,
+        avgState(duration) AS duration_avg,
+        quantilesTDigestState(0.75, 0.9, 0.95, 0.99)(duration) AS duration_quantiles
+      FROM default.operations
+      GROUP BY
+        target,
+        timestamp
```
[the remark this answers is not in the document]
jdolle: It's fine because the timestamp is aliased in the select (toStartOfDay(timestamp) AS timestamp,) and any references to timestamp elsewhere refer back to the aliased value

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -77,24 +133,38 @@ function extractMetricValue(row: ClickHouseWindowRow, rule: MetricAlertRuleRow):
...
-      // so any non-trivial latency trips the rule.
-      return (rule.metric ? metricMap[rule.metric] : 0) / NS_TO_MS;
+      // Read only the column this metric needs (the other may be a NULL placeholder),
+      // and require it to be present (see requireColumn).
+      const ns =
+        rule.metric === 'AVG'
+          ? requireColumn(row.average, 'duration_avg', rule)
+          : requireColumn(row.percentiles, 'duration_quantiles', rule)[
+              percentileIndex(rule.metric)
+            ];
```
jdolle, line 143: Minor: I think what this is doing is clear and no need for the long comments.

For requireColumn -- is this if for example, there's no data for this window? Or is it for if there's an issue in the code and this column wasnt fetched for some reason?
Could we get away with a null check and return null if it's missing?

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -196,6 +267,31 @@ export function groupRulesByQuery(
...
+}
+
+// PENDING/RECOVERING rules stay at 1-min resolution so the confirmationMinutes
+// dwell is sampled every tick. A never-evaluated rule is always due.
+export function isRuleDue(
+  rule: Pick<MetricAlertRuleRow, 'timeWindowMinutes' | 'lastEvaluatedAt' | 'state'>,
+  evaluationTime: Date,
+): boolean {
+  if (rule.state === 'PENDING' || rule.state === 'RECOVERING') return true;
+  if (!rule.lastEvaluatedAt) return true;
```
jdolle, line 287: can combine this condition with the others to return true.
If you want them separate for clarity on what they are checking, then you could assign a new variable, i.e. `const isInPriorityState = rule.state === 'PENDING' || rule.state == 'RECOVERING';`

packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -247,17 +343,60 @@ export function buildSavedFilterConditions(rawFilters: unknown, logger: Logger):
...
+  // counts, which daily buckets would skew by up to a full edge day (buckets snap
+  // to day boundaries, the rolling window doesn't), so a group with any TRAFFIC
+  // rule stays on hourly. Latency/error-rate tolerate the rounding.
+  allowDailyRollup: boolean;
+};
+
+// All rules in a group share a saved filter (it's part of the group key), so the
+// filter is built once from the representative. A malformed filter yields no
+// conditions (evaluates unfiltered) and is logged, isolating the failure here.
+export function deriveGroupNeeds(groupRules: MetricAlertRuleRow[], logger: Logger): GroupNeeds {
```
jdolle, line 370: This function was unexpected. I anticipated more concrete query building depending on the rule.

E.g. `buildLatencyAlertQuery(rule: MetricAlertRuleRow)`

I think generic can be good, but as we add more rules i worry this will become spaghetti.

packages/services/workflows/src/tasks/evaluate-metric-alert-rules.ts
```diff
@@ -56,6 +59,26 @@ export const task = implementTask(EvaluateMetricAlertRulesTask, async args => {
...
+  // retains a stale reading.
+  let filteredGroups = 0;
+  let filteredRules = 0;
+  for (const group of groups.values()) {
+    if (deriveGroupNeeds(group, logger).filterConditions.length > 0) {
+      filteredGroups += 1;
+      filteredRules += group.length;
+    }
+  }
+  metricAlertRuleGroups.set({ filtered: 'true' }, filteredGroups);
```
jdolle, line 78: I'm not following this, but it's late. I'll try to take another look tomorrow.

kamilkisiela:
This PR combines quite a few pretty much independent changes. Could we split work like this into smaller, focused PRs? Reviewing all of these changes together makes it difficult to understand each decision and verify performance.
Smaller PRs would be much easier to review, test, discuss, and roll back independently.

kamilkisiela:
There are known things that can improve performance, that we can verify in production quickly:
- Queue instead of Promise.all(5 items)
- query only what's needed (error rate metrics does not need duration quantiles etc)

jonathanawesome:
Superseded by split PRs.

PR #8205, opened 2026-07-10
Laboratory preflight clickable log links

jonathanawesome:
This PR makes `http(s)://` URLs in the Laboratory preflight Logs pane clickable (open in a new tab), so device-code OAuth scripts can log a sign-in link instead of making users copy a URL out of a log line.

Adds a pure `tokenizeUrls` helper in `lib/utils.ts` and renders URL tokens as anchors in `preflight.tsx`. Render-only: no worker, protocol, or `lab` API changes. Only `https?://` is linkified, links use `rel="noopener noreferrer"`, and non-URL logs render as before.


packages/libraries/laboratory/src/lib/utils.ts
jdolle, line 1: Let's add some tests for each of these cases for regression

dotansimha (approved):

PR #8207, opened 2026-07-10
another attempt at improving e2e flakiness

jonathanawesome:
This PR fixes the flaky usage e2e tests. They flaked because the usage-ingestor writes to ClickHouse fire-and-forget (`wait_for_async_insert=0`), so a reported operation isn't queryable in Insights until ClickHouse's async-insert buffer flushes, and that timing is non-deterministic under CI load (it got worse with the 24.8 to 26.3 image bump, #8120). The fix makes the e2e ingestor's write synchronous: it adds an optional `CLICKHOUSE_WAIT_FOR_ASYNC_INSERT` env var (default `0`, so prod and self-hosters are unchanged) and, for the e2e ingestor only, sets it to `1` together with a low `CLICKHOUSE_ASYNC_INSERT_BUSY_TIMEOUT_MS` (100ms). Both are required: waiting for the flush makes the insert block until the row is queryable, and the low busy-timeout makes that flush happen in ~100ms instead of the default 30s. The row is then queryable the moment the insert returns, deterministically.

**What we tried first, and why it wasn't enough:**
- Bigger poll and per-test timeouts: treated the symptom; visibility sometimes exceeded 90s, so this just postponed the failure.
- Lowering the flush timer alone (`async_insert_busy_timeout_ms`, `max_data_size`): with fire-and-forget writes nothing waits for the flush to finish, so under CI load it still lagged; it only helps combined with waiting for the flush (as above).
- `SYSTEM FLUSH ASYNC INSERT QUEUE`: proven to work in local isolation against ClickHouse 26.3.12.3, but made no difference in CI under the loaded parallel stack.

### DIFF
--- a/docker/docker-compose.end2end.yml
+++ b/docker/docker-compose.end2end.yml
@@ -47,12 +47,12 @@ services:
     ports:
       - '6379:6379'
 
-  clickhouse:
-    # Expose the HTTP interface so the usage e2e helper can force a synchronous flush of the
-    # async insert queue (SYSTEM FLUSH ASYNC INSERT QUEUE), making ingested reports queryable
-    # immediately instead of racing the non-deterministic async-insert flush timer.
-    ports:
-      - '8123:8123'
+  usage-ingestor:
+    environment:
+      # Wait for the async insert to flush before acking, so a reported operation is queryable in
+      # Insights immediately instead of racing ClickHouse's non-deterministic async-insert timer
+      # (the source of usage e2e flakiness). Prod keeps the default fire-and-forget behavior.
+      CLICKHOUSE_WAIT_FOR_ASYNC_INSERT: '1'
 
   server:
     environment:
--- a/docker/docker-compose.end2end.yml
+++ b/docker/docker-compose.end2end.yml
@@ -47,6 +47,12 @@ services:
     ports:
       - '6379:6379'
 
+  usage-ingestor:
+    environment:
+      # Flush ClickHouse async inserts fast so a report is visible in Insights within the
+      # e2e poll window (prod keeps the default 30s batching).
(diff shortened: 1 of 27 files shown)


docker/docker-compose.end2end.yml
```diff
@@ -47,6 +47,12 @@ services:
     ports:
       - '6379:6379'
 
+  usage-ingestor:
+    environment:
+      # Flush ClickHouse async inserts fast so a report is visible in Insights within the
+      # e2e poll window (prod keeps the default 30s batching).
+      CLICKHOUSE_ASYNC_INSERT_BUSY_TIMEOUT_MS: '1000'
```
jdolle: great find! I think let's go even lower. We dont need to worry about the performance of these async inserts for our tests. 100ms?

e2e/helpers/usage.ts
```diff
@@ -90,7 +90,7 @@ export function createUsageHelper(
 
             return page.getByRole('link').filter({ hasText: operationName }).count();
           },
-          { timeout: 60_000, intervals: [2_000] },
+          { timeout: 90_000, intervals: [2_000] },
```
jdolle: Why so long? Even 60s seems generous

jonathanawesome: Dropped the poll to 30s and brought the per-test timeout back down to 120s. There _could_ still be hiccups with the runner, so these seem reasonable.

jdolle (approved):

e2e/helpers/usage.ts
```diff
@@ -86,23 +111,25 @@ export function createUsageHelper(
       await expect
         .poll(
           async () => {
+            await flushAsyncInserts();
```
jdolle: Minor: This could probably be moved to just after the insert to avoid calling it on every poll for this condition.

jdolle (approved):

PR #8213, opened 2026-07-13
Evaluate metric-alert groups with a concurrency pool

jonathanawesome:
This PR replaces the fixed-batch barrier in the `evaluateMetricAlertRules` cron with a `p-limit` concurrency pool, so metric-alert groups are evaluated with continuous bounded parallelism instead of synchronized batches of 5. 

Before, a batch couldn't start until the previous one fully drained, so one slow group (like a timed-out ClickHouse query) idled the other slots and delayed every later group, including cheap ones from other orgs; the pool keeps up to 5 groups in flight at all times, starting the next as soon as one finishes. 

Behavior is otherwise unchanged: same cap (still below the Postgres pool size), same `allSettled` failure isolation, and the same batched `last_evaluated_at` update once all groups drain. `p-limit` is the limiter already used by the usage and migrations packages.

### DIFF
--- a/.changeset/giant-cups-matter.md
+++ b/.changeset/giant-cups-matter.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+Runs metric alerts cron with a concurrency pool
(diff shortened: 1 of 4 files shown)


jdolle (approved):
please add a changeset is all

kamilkisiela:
LGTM

PR #8217, opened 2026-07-14
Query only the duration columns each alert group needs

jonathanawesome:
This PR makes the metric-alert evaluator select only the ClickHouse duration columns each group actually reads, instead of always computing both `avgMerge` and the percentile merge. 

Error-rate and traffic groups need neither, and a latency group needs only its own metric's column, so the query emits a `NULL` placeholder for anything unused; since ClickHouse is columnar, not selecting the large `duration_quantiles` blob means not reading it, cutting CPU and bytes scanned. The window row's duration fields become nullable to match, and `extractMetricValue` throws if a latency rule reads a column that wasn't selected rather than silently treating it as 0ms.


packages/services/workflows/src/lib/metric-alert-evaluator.ts
```diff
@@ -257,6 +272,9 @@ export async function queryClickHouseWindows(
   // this job up late, using wall-clock would shift the queried window
   // forward and could miss the spike that should have fired the alert.
   evaluationTime: Date,
+  // False selects `NULL as <col>` so ClickHouse skips reading that duration column.
+  needsAverage: boolean = true,
+  needsPercentiles: boolean = true,
```
kamilkisiela, line 277: Can we tell based on the metric which p* it needs? If so, we could squeeze even more performance when it needs only p99 or p90 etc

kamilkisiela: 2 more PRs and you will either love me or hate me :))

jonathanawesome: `duration_quantiles` stores one t-digest per row, so the cost is reading and merging those states...the same whether we pull one quantile or all four. Extracting just `p95` from the merged digest is basically free, so narrowing doesn't cut bytes read or CPU.

kamilkisiela: In theory yes, but theory sometimes is far from reality, but it's fine to merge as is, no issues with that

jdolle (approved):
Great job on the test coverage.
Let's add a changeset is all

PR #8218, opened 2026-07-15
Gate metric-alert evaluation cadence by rule window

jonathanawesome:
This PR makes the `evaluateMetricAlertRules` cron skip re-evaluating long-window rules every minute, since a multi-hour or multi-day rolling metric barely moves tick to tick. 

Each rule gets a minimum interval derived from its window (1h→every minute unchanged, up to 30-day→every 30 min), gated by the existing `last_evaluated_at` column. 

Gating is per-group: a group evaluates when any member is due, and the batched `last_evaluated_at` update keeps its members aligned so the shared ClickHouse query still fires once per interval. Rules in PENDING or RECOVERING stay at full 1-minute resolution so `confirmationMinutes` dwell is still sampled every tick; only steady-state detection latency is coarsened, bounded by the interval.


packages/services/workflows/src/lib/metric-alert-evaluator.spec.ts
```diff
@@ -68,6 +71,76 @@ describe('groupRulesByQuery', () => {
...
+    expect(evaluationIntervalMinutes(361)).toBe(15);
+    expect(evaluationIntervalMinutes(1440)).toBe(15);
+    expect(evaluationIntervalMinutes(1441)).toBe(30);
+    expect(evaluationIntervalMinutes(10080)).toBe(30);
+    expect(evaluationIntervalMinutes(43200)).toBe(30);
+  });
+
+  test('interval never exceeds the window, so the dwell stays sampled', () => {
+    for (const window of [1, 60, 61, 360, 361, 1440, 1441, 10080, 43200]) {
+      expect(evaluationIntervalMinutes(window)).toBeLessThanOrEqual(window);
```
jdolle: this test feels redundant with the last one

packages/services/workflows/src/tasks/evaluate-metric-alert-rules.ts
```diff
@@ -65,8 +66,21 @@ export const task = implementTask(EvaluateMetricAlertRulesTask, async args => {
 
   logger.info({ count: rules.length }, 'Evaluating metric alert rules');
 
+  // Evaluate only groups with at least one due member. Members of a group share
+  // a window (so one cadence), and the batched UPDATE below keeps their
+  // last_evaluated_at aligned, so a group is due as a unit.
   const groups = groupRulesByQuery(rules);
-  const groupList = [...groups.values()];
+  const dueGroupList = [...groups.values()].filter(group =>
+    group.some(rule => isRuleDue(rule, evaluationTime)),
```
jdolle, line 74: When we have many rules, would it make more sense to apply this condition to the `fetchEnabledRules` query?
This can be a followup based on when/if we encounter performance issues

jdolle (approved):
Looks good. A few minor comments but nothing blocking

PR #8219, opened 2026-07-15
Log resolved ClickHouse writer settings and operations insert duration

jonathanawesome:
This PR adds two debug-level logs to the usage-ingestor's ClickHouse writer to diagnose an intermittent usage e2e flake where reported operations take too long to become queryable in Insights. We previously landed a fix (setting `wait_for_async_insert=1` plus a low `busy_timeout` for the e2e ingestor) that was supposed to make the write synchronous, but the flake still happens, and every static check says the config is correct, so we need to see what the ingestor actually does at runtime. On startup it now logs the resolved async-insert settings (`wait_for_async_insert`, `async_insert_busy_timeout_ms`, `async_insert_max_data_size`), and per operations INSERT it logs how long the write blocked. Together with the existing retry log, a single failing e2e run will show whether the ingestor resolved `wait_for_async_insert=1` at runtime and whether the insert actually blocked for the flush (synchronous) or returned immediately (fire-and-forget). Both are at `debug`, so they appear in the e2e stack (which runs at debug) but stay out of prod logs (info).


packages/services/usage-ingestor/src/writer.ts
```diff
@@ -74,6 +82,13 @@ export function createWriter({
         logger,
         3,
       );
+      // With wait_for_async_insert=1 this blocks until the row is flushed (queryable);
+      // fire-and-forget returns in a few ms. Logged at debug so it stays out of prod (info).
+      logger.debug(
+        'operations INSERT completed in %sms (operations=%s)',
+        Math.round(performance.now() - startedAt),
```
jdolle: shouldnt need to round

jdolle (approved):

PR #8220, opened 2026-07-15
Reduce replica count for staging schema and graphqlApi services

jdolle:
### Background

Staging deployed once with the increased replica count but is failing to allocate the necessary memory now. We could either scale our containers or reduce replica count on staging.

### Description

Reduces staging replicas that were previously increased in https://github.com/graphql-hive/console/pull/8185
Keeps production higher

### DIFF
--- a/.changeset/huge-emus-call.md
+++ b/.changeset/huge-emus-call.md
@@ -1,5 +1,0 @@
----
-'hive': patch
----
-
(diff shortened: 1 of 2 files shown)


jdolle:
Synced w/Dotan via slack and we're good to do this to unblock. He will investigate what the permanent change should be soon

PR #8221, opened 2026-07-15
Skip the previous window for absolute-only alert groups

jonathanawesome:
This PR skips fetching the previous comparison window for metric-alert groups that contain only absolute (`FIXED_VALUE`) rules, since those rules only look at the current window. The window query now scans 1× the window instead of 2× for those groups, and persists a null previous value. Any `PERCENTAGE_CHANGE` rule in a group keeps both windows. Absolute rules also persist null consistently even when they share a group with a percentage rule that did fetch the previous window, so a rule's stored value and its history display don't depend on its group-mates. Notifications, the webhook payload, and the payload's JSON schema are updated to accept a null previous value.


packages/web/app/src/components/target/alerts/alert-notification-preview.tsx
```diff
@@ -17,7 +17,6 @@ const WEBHOOK_JSON_SCHEMA = {
     'state',
     'alert',
     'currentValue',
-    'previousValue',
```
n1ru4l, line 20: NIT: we always include this value anyways even if it is `null`, so we could keep it in the list of required properties

n1ru4l (approved):
Approved with one minor NIT. 🥳

PR #8226, opened 2026-07-16
fix: clarify safe/approved breaking changes in github check run

n1ru4l:
### Background

This clarifies the message we show in case there a breaking changes, but the schema check passed.


dotansimha (approved):

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -3143,9 +3143,9 @@ export class SchemaPublisher {
           summary = 'No changes detected';
           shortSummaryFallback = summary;
         } else {
-          title = 'No breaking changes';
-          summary = this.changesToMarkdown(changes);
-          shortSummaryFallback = this.changesToMarkdown(changes, false);
+          title = 'No blocking changes';
```
dotansimha, line 3146: nit: I don't remember we used this terminology in other places

PR #8230, opened 2026-07-17
feat: schema delete retries

n1ru4l:
### Background

For busy projects it is possible that the registry lock acquiring times out. For the schema publishing process we already introduced automatic retries. This also introduces the changes for the schema delete process.

### Description

Implement the same behaviour as for schema publishes for retrying from the client side if the registry lock can not be acquired.

**Note**: I did not add tests as these would require running for a long time (and we also do not have them for the schema publishing counterpart).

I manually tested this be introducing a sleep statement into the schema publish/delete process.

### DIFF
--- a/packages/libraries/cli/src/commands/schema/delete.ts
+++ b/packages/libraries/cli/src/commands/schema/delete.ts
@@ -1,6 +1,6 @@
 import { Args, Errors, Flags, ux } from '@oclif/core';
 import Command from '../../base-command';
-import { graphql, useFragment } from '../../gql';
+import { DocumentType, graphql, useFragment } from '../../gql';
 import * as GraphQLSchema from '../../gql/graphql';
 import { graphqlEndpoint } from '../../helpers/config';
 import {
@@ -36,6 +36,9 @@ const schemaDeleteMutation = graphql(/* GraphQL */ `
         errors {
           ...RenderErrors_SchemaErrorConnectionFragment
         }
+      }
+      ... on SchemaDeleteRetry {
+        reason
       }
     }
   }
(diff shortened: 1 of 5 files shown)


packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l, line 1: The noise here is mostly indent changing. Relevant changes start at `L1670`

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -887,6 +894,10 @@ export default gql`
     target: TargetReferenceInput
     serviceName: ID!
     dryRun: Boolean
+    """
+    Whether the client supports retries in case the registry is busy and the operation could not be performed.
+    """
+    supportsRetry: Boolean = False
```
n1ru4l: New clients need to specify that, as old clients do not gracefully handle the `SchemaDeleteRetry` type member.

packages/libraries/cli/src/commands/schema/delete.ts
```diff
@@ -145,60 +148,68 @@ export default class SchemaDelete extends Command<typeof SchemaDelete> {
...
+        });
 
-        if (errors) {
-          const unmaskedErrors = useFragment(RenderErrors_SchemaErrorConnectionFragment, errors);
-          if (unmaskedErrors.edges.length > 0) {
-            this.log(renderErrors(errors));
+        if (result.schemaDelete.__typename === 'SchemaDeleteRetry') {
+          this.log(result.schemaDelete.reason);
+          this.log('Waiting for other schema registry actions to complete...');
+          result = null;
```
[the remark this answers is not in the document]
n1ru4l: I think it is not needed as the server lock acquiring will only time out after 20 seconds or so.

packages/libraries/cli/src/commands/schema/delete.ts
```diff
@@ -145,60 +148,68 @@ export default class SchemaDelete extends Command<typeof SchemaDelete> {
...
+          const errors = result.schemaDelete.errors;
 
-      if (errors) {
-        throw new APIError(renderErrors(errors));
-      }
+          if (errors) {
+            throw new APIError(renderErrors(errors));
+          }
+        }
+      } while (result === null);
```
jdolle, line 212: Should we add throttling, exponential delays, and/or a maximum retry count?

packages/services/api/src/modules/schema/resolvers/Mutation/schemaDelete.ts
```diff
@@ -11,10 +11,15 @@ export const schemaDelete: NonNullable<MutationResolvers['schemaDelete']> = asyn
       dryRun: input.dryRun,
       serviceName: input.serviceName.toLowerCase(),
       target: input.target,
+      supportsRetry: input.supportsRetry,
     },
     request.signal,
   );
 
+  if (result.__typename === 'SchemaDeleteRetry') {
```
jdolle, line 19: We're blurring the separation of concerns when the provider returns the resolver's response with the __typename and everything. But we're doing this already so I don't think that needs to be addressed in this PR.

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -429,7 +429,14 @@ export default gql`
...
     | GitHubSchemaCheckError
 
-  union SchemaDeleteResult = SchemaDeleteSuccess | SchemaDeleteError
+  union SchemaDeleteResult = SchemaDeleteSuccess | SchemaDeleteError | SchemaDeleteRetry
+
+  type SchemaDeleteRetry {
+    """
+    The reason for the retry.
+    """
+    reason: String!
```
jdolle, line 438: If we want to push the retry logic to the client, I think we should provide useful information about the retry. E.g. is there a retry limit?
If we just need a field to exist to be able to return this type, then this is fine

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -1407,264 +1407,288 @@ export class SchemaPublisher {
...
-      },
-    );
+          return {
+            __typename: 'SchemaDeleteError',
+            valid: false,
+            changes: [],
+            errors: [
+              {
+                message:
+                  'Another schema operation is currently in progress. Please retry the delete.',
```
jdolle: The point in having `SchemaDeleteRetry` and `SchemaDeleteError` as separate types is to allow clients to handle them differently. `"Please retry the delete."` is a red flag to me since it indicates you want both errors to be handled the same way.

packages/libraries/cli/src/commands/schema/delete.ts
n1ru4l: Why? We want to acquire the lock as soon as possible so we can perform the action. Can you elaborate your reasoning behind throttling and a maximum retry count?

packages/services/api/src/modules/schema/module.graphql.ts
n1ru4l: The field mostly exists because it is required to have a field. There is currently no retry limit (and I see no reason on why there should be as at some point the lock should be able to be acquired unless someone is doing a schema publish every few seconds, which is another issue imho).

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: This is what is returned if the client (aka previous CLI versions) does not support retries (and does not gracefully handle the `SchemaDeleteRetry` type). In that case the user needs to manually re-run the command.

packages/libraries/cli/src/commands/schema/delete.ts
jdolle: I forgot that we wait and retry for the lock on the backend also. So given the duration of that request, throttling and delaying isnt necessary. But a maximum retry count would be to avoid retrying indefinitely if the system is very busy. Basically so that an org's CI usage doesnt spike because it's waiting on us.

packages/services/api/src/modules/schema/providers/schema-publisher.ts
jdolle:
```suggestion
                  `Another schema operation is currently in progress.${input.supportsRetry ? '' : ' Please retry the delete.'}`,
```

jdolle (approved):

packages/services/api/src/modules/schema/providers/schema-publisher.ts
n1ru4l: In [L1671](https://github.com/graphql-hive/console/pull/8230/changes#diff-559e8e3f328b69f95f8d3ab16c03be3319e7b647c297fe9ceb7b32d87a6238e9R1671), if the client supports retries, it will return `SchemaDeleteRetry`, this condition would never be met.

packages/libraries/cli/src/commands/schema/delete.ts
n1ru4l: I don't think people run the delete command from CI, but manually on their own machine when a subgraph is deleted.

If your ultimate goal is to get the subgraph deleted, why should there be a maximum attempt time? You would then re-run the command again.

Happy to continue the discussion around this elsewhere, for now let's release this and eliminate the unhandled exception noise.

PR #8242, opened 2026-07-22
Reduce composition memory for stitching projects

jdolle:
### Background

We're seeing Out of memory errors during schema stitching

### Description

Swaps a `.map` for a `for` loop which allows us to free parsed ASTs a bit sooner.

I had thought the AST was used by the GraphQLSchema and therefore this wouldn't have much impact, but in testing, several megabytes of memory (~15MB with the large test schemas I was using) are freed by this change.


scripts/extract_schemas.sh
jdolle, line 1: This was used for extracting problem schemas from our environment(s). We can opt to not include it, but since it was a fairly short script and potentially useful in the future, I kept it here.

packages/services/schema/src/composition/stitching.ts
```diff
@@ -68,23 +68,32 @@ export type ComposeStitchingArgs = {
...
+    const parsed = parse(schema.raw);
+    const stitchedErrors = validateStitchedSchema(parsed);
+    if (stitchedErrors.length) {
+      errors.push(...stitchedErrors);
+    } else {
+      subschemas.push(
+        buildASTSchema(trimDescriptions(parsed), {
+          assumeValid: true,
+          assumeValidSDL: true,
+        }),
```
n1ru4l, line 85: This is now no longer happening within the try catch block, could that be an issue?

packages/services/schema/src/composition/stitching.ts
```diff
@@ -68,23 +68,32 @@ export type ComposeStitchingArgs = {
...
+  const errors: Array<CompositionErrorType> = [];
+  const subschemas: GraphQLSchema[] = [];
+  for (const schema of args.schemas) {
+    // parse inside a loop instead of using map to avoid holding on to the parsed schema
+    // in memory. This matters for large schemas.
+    const parsed = parse(schema.raw);
+    const stitchedErrors = validateStitchedSchema(parsed);
+    if (stitchedErrors.length) {
+      errors.push(...stitchedErrors);
+    } else {
```
n1ru4l, line 80: If there is already some errors from previous schemas, we can skip subsequent building of the AST schema in the first place?

.changeset/odd-areas-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'hive': patch
+---
+
+Reduce memory usage during stitching composition; upgrade stitching package
```
n1ru4l: The stitching package is not upgraded as part of this PR

n1ru4l (approved):
I added some comments.

As a general feedback for performance improvements PRs there should be a reproducible benchmark - otherwise it is tedious to verify and approve these changes.

packages/services/schema/src/composition/stitching.ts
jdolle: We could technically, but we want to be sure to capture and return all errors to the API. This matches the current functionality

packages/services/schema/src/composition/stitching.ts
jdolle: `buildASTSchema` won't throw because the configuration settings:
```
assumeValid: true,
assumeValidSDL: true,
```

`stitchSchemas` is the function that we need inside the try/catch block to catch composition errors.

.changeset/odd-areas-yell.md
jdolle:
```suggestion
Reduce memory usage during stitching composition
```

jdolle: :+1: I was planning to upgrade here but will submit a separate PR

packages/services/schema/src/composition/stitching.ts
n1ru4l: The `subschemas` arrays is never used if there is at least one composition error. Thus all the overhead of building the ast schema and trimming descriptions can be skipped. In case of validation errors occurring we now kind of allocate more memory than before as we unnecessarily build the ast schema of the subgraphs that raise no errors?

PR #8243, opened 2026-07-22
fix: monolithic schema version promotion

n1ru4l:
### Background

This fixes schema version promotion for single-schema projects.

Previously, promoting a version could fail because the composition result only accepted composite schema logs.

The fix adds support for single-schema logs and includes an integration test covering promotion of multiple monolith schema versions.

### DIFF
--- a/packages/services/api/src/modules/schema/providers/schema-version-helper.ts
+++ b/packages/services/api/src/modules/schema/providers/schema-version-helper.ts
@@ -409,6 +409,9 @@ export class SchemaVersionHelper {
         } satisfies ResolversUnionTypes<any>['SubgraphDiff'];
       }
       if (edge.type === 'added') {
+        invariant(!!edge.node.service_name, 'node of edge cannot be null');
+        invariant(!!edge.node.service_url, 'url of edge cannot be null');
+
         return {
           __typename: 'SubgraphDiffAdded',
           subgraphVersion: {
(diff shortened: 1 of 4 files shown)


packages/services/api/src/modules/schema/providers/schema-version-helper.ts
```diff
@@ -409,6 +409,9 @@ export class SchemaVersionHelper {
         } satisfies ResolversUnionTypes<any>['SubgraphDiff'];
       }
       if (edge.type === 'added') {
+        invariant(!!edge.node.service_name, 'node of edge cannot be null');
+        invariant(!!edge.node.service_url, 'url of edge cannot be null');
```
n1ru4l, line 413: This cannot happen as this code is never invoked for single schema projects ([see the guard in L361](https://github.com/graphql-hive/console/pull/8243/changes#diff-8bbfbb4cee2af10625d161a9498def42432f0e88bf8db2c3412ee06178d41095R361)) and would indicate invalid database state.

We can potentially remove these assertions by doing a bigger refactoring of the code. This seemed like the easiest solution to resolve the typing issue for now.

jdolle: Thanks for this comment. I was going to ask about this

jdolle (approved):

PR #8246, opened 2026-07-22
Fix target nav on history route

jonathanawesome:
This PR fixes a bug where switching target while viewing target/history kept the previous target version UUID in the URL, so the new target landed on "Schema Version not found." instead of its own history. 

The target switcher now drops resource ids that are scoped to the old target and stays on the same section under the new target, and the history route resolves its own default (the latest version) rather than inheriting the old id.

The "go to latest version" redirect was also moved out of a render-time `useEffect` and into the history route's TanStack `beforeLoad`. As a render effect it raced with urql's data settling after a target switch (urql briefly returns the previous target's data) and with `useParams`, which could spin into a navigate loop that froze the tab; running it in `beforeLoad` means it fires once, before render, and structurally can't loop. Looked through the rest of the routes for any other _high-risk_ effects with redirects...none found.

It also silences an urql graphcache "invalid key" warning flood on the history page by declaring the keyless `SchemaVersion` value types (meta, origin, github metadata) as keyless in the cache config.


n1ru4l (approved):

n1ru4l:
If you think it makes sense, let's add a changeset before merging.

jonathanawesome:
> If you think it makes sense, let's add a changeset before merging.

These changes are internal, so I'll skip the changeset here.

PR #8247, opened 2026-07-22
Test memory limits for composition worker

jdolle:
### Background

Our composition worker has a strict memory limit. This test is a regression to ensure we don't make changes that exceed the memory limit without knowing it.

### Description

I did manual testing to determine that the memory usage is roughly a multiplier of the schema size.

The test generates a schema given a size requirement, and creates a worker based on a multiplier of that schema size.

This adds two checks, both with the same schema size, but with different multipliers. One passes because it's 150x the schema size (5mb schema -> 750mb worker), and the other check errors because the worker is 735mb.

The 150x multiplier is surprising. It seems very high. But based on testing it seems accurate.


jdolle:
We added a way to track memory usage and this test is too flaky. Closing

PR #8256, opened 2026-07-24
Track composition worker memory

jdolle:
### Background

The composition worker can crash when the heap memory limit is exceeded. We see occasional crashes and can track those back to targets from the traced request, but it's currently not easy to observe how memory is being impacted over time for a target.

### Description

This adds a new metric for composition worker memory, and displays it on the schema service dashboard

### DIFF
--- a/packages/services/schema/src/composition-worker.ts
+++ b/packages/services/schema/src/composition-worker.ts
@@ -8,6 +8,13 @@ import { createComposeFederation, type ComposeFederationArgs } from './compositi
 import { composeSingle, type ComposeSingleArgs } from './composition/single';
 import { composeStitching, type ComposeStitchingArgs } from './composition/stitching';
 import type { env } from './environment';
+import { compositionWorkerMemoryUsedBytes } from './metrics';
+
+type ErrorResultEvent = {
+  event: 'error';
+  id: string;
+  err: unknown;
+};
 
 export function createCompositionWorker(args: {
   baseLogger: Logger;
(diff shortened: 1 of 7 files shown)


packages/services/schema/src/composition-worker.ts
```diff
@@ -33,6 +40,16 @@ export function createCompositionWorker(args: {
       event: message.event,
     });
     logger.debug('processing message');
+    const postResult = (result: CompositionResultEvent | ErrorResultEvent) => {
+      args.port.postMessage(result);
+
+      if (args.env.compositionWorker.trackMemoryUsage) {
+        compositionWorkerMemoryUsedBytes.set(
+          { target: message.targetId, type: 'federation' },
```
n1ru4l: This is not federation but all kinds of compositions?

This call happens in a worker - is this metric actually then exposed on the main thread?

Doesn't the prometheus node.js plugin already gather some of those metrics that we could make accessible?

jdolle: The prometheus default metrics don't include workers because they run on separate V8 instances. We could create another prom-client instance on the worker and pass those metrics back via the the parent port:
```
client.collectDefaultMetrics();
const metrics = await client.register.getMetricsAsJSON();
```
but this seems like overkill to me when we only want the heap usage.

I've adjusted the code to pass back the metric to the main thread for reporting.

packages/services/schema/src/composition-worker.ts
```diff
@@ -33,6 +47,18 @@ export function createCompositionWorker(args: {
...
     });
     logger.debug('processing message');
+    const postResult = (result: CompositionResultEvent | ErrorResultEvent) => {
+      args.port.postMessage(result);
+
+      if (args.env.compositionWorker.trackMemoryUsage) {
+        // send metric back to main thread so it can be reported
+        args.port.postMessage({
+          event: 'metric',
+          type: 'heapUsed',
```
jdolle, line 56: kept this metric event structure generic in case we decide to send back other kinds of metrics as well

PR #8258, opened 2026-07-24
feat: make tracking OOM memory issues in schema worker easier

n1ru4l:
In case of an oom issue, write all the subgraphs as an artifact to s3 and add a reference to that on the trace.

I tested the flow locally by:

Setting low memory
image
Publishing a highe schema, causing an OOM for the worker thread
image

The trace is now populated with some reference:
image

And an artifact is written to our S3 bucket containing all the subgraphs involved in the composition: 
image
image

### DIFF
--- a/packages/services/api/src/modules/schema/providers/orchestrator/composition-orchestrator.ts
+++ b/packages/services/api/src/modules/schema/providers/orchestrator/composition-orchestrator.ts
@@ -1,7 +1,7 @@
 import { CONTEXT, Inject, Injectable, Scope } from 'graphql-modules';
 import { abortSignalAny } from '@graphql-hive/signal';
 import type { ContractsInputType, SchemaBuilderApi } from '@hive/schema';
-import { traceFn } from '@hive/service-common';
+import { trace, traceFn } from '@hive/service-common';
 import { createTRPCProxyClient, httpLink } from '@trpc/client';
 import {
   ComposeAndValidateResult,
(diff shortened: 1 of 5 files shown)


packages/services/api/src/modules/schema/providers/orchestrator/composition-orchestrator.ts
```diff
@@ -159,6 +161,21 @@ export class CompositionOrchestrator {
           signal: abortSignalAny([this.incomingRequestAbortSignal, timeoutAbortSignal]),
         },
       );
+
+      // If a composition error happens due to OOM we write the subgraphs to S3
+      // This makes it easier to reproduce and debug the issue
```
n1ru4l, line 166: Technically we could do that for every composition failure. But this would end up using a lot lot more s3 storage and we do not really have a cleanup strategy. Thus only storing the OOM ones for now seems kind  okay in my opinion.

packages/services/api/src/modules/schema/providers/orchestrator/composition-orchestrator.ts
```diff
@@ -159,6 +161,21 @@ export class CompositionOrchestrator {
...
+      if (result.errors.some(error => error.code === 'ERR_WORKER_OUT_OF_MEMORY')) {
+        const span = trace.getActiveSpan();
+        if (span) {
```
n1ru4l, line 169: This behaviour is only enabled if tracing is enabled for now.

packages/services/api/src/modules/schema/providers/orchestrator/composition-orchestrator.ts
n1ru4l: In the future we want to store SDL on s3 exclusively anyways - as it is eating up a lot of space on postgres and would be more cost effective over there. Then this would no longer be needed.

PR #8266, opened 2026-07-26
Fix various UI issues in laboratory

jonathanawesome:
This PR fixes the reported builder-collapse bug and works through a round of laboratory UI defects found alongside it, adding copy as cURL, a reload-schema button and an introspection polling toggle in the process.

The trigger was a report that expanded fields in the builder collapse "at a certain level" whenever an introspection response arrives. Other reports included: missing hovers, tofu glyphs in the editors, a response size stuck at `0KB`, an inert tree/list toggle, an undiscoverable query plan panel, and a steady stream of React DOM-nesting warnings.

**Fixes**

- The builder no longer collapses expanded fields while introspection polls. `introspectionFromSchema` returned a fresh object every poll, so an unchanged schema still produced a new `GraphQLSchema`, which reset expansion to the depth of the current document.
- Editor hovers work on mouse over again, so validation messages and schema docs are readable. Monaco installs a document-level `mousemove` fallback and decides the pointer left the editor with `viewDomNode.contains(event.target)`; inside a shadow root that target is retargeted to the host, so every move looked like a mouse leave and dismissed the hover as it opened.
- Monaco folding chevrons render as icons rather than empty squares, because `@font-face` rules now go on the document instead of inside the shadow root where browsers ignore them.
- Response size shows real units and is measured in UTF-8 bytes; it previously rounded anything under 512 bytes to `0KB`.
- `asChild` tooltips appear at all. `Button`, `TooltipTrigger` and `AlertDialogTrigger` were plain function components, so on React 18 the ref was dropped and the tooltip had no anchor to position against.
- The tree/list toggle is controlled, can no longer be deselected into an empty state, and is disabled with an explanation until a search is active.
- The Query Plan panel no longer throws during render when a response body is not JSON.
- `monaco-graphql` is initialized once and updated in place, so variables validation registers regardless of which editor mounts first and survives an endpoint change.
- Invalid nested buttons are gone from the builder and collection rows, which also makes the collection edit and delete actions keyboard reachable for the first time.
- The query plan visualization no longer sets state during render.

**Added**

- Copy as cURL in the operation toolbar, mirroring the request the lab actually sends: preflight and operation headers merged in the same precedence, `{{env}}` and plugin templates substituted, only the selected operation plus its fragments, and `useGETForQueries` honoured.
- A reload-schema button that introspects over the network even when the host supplied a schema, and spins while in flight.
- An `introspection.pollSchema` setting to turn off the 5 second poll and refresh only on demand.
- An `enableFullScreen` prop (default `true`) so hosts that already fill the viewport can hide the control.
- The Query Plan tab is always visible, with an empty state describing when plans appear.

**Dev tooling** (not shipped to consumers)

- A mock GraphQL endpoint mounted onto the Vite dev server, serving the repo's own `schema.graphql`, so `pnpm dev` gives the lab a working same-origin endpoint with no second process.
- Query plan fixtures covering all eight node kinds, selected with an `x-query-plan` header, plus seeded collections for the states that are otherwise tedious to reach by hand.


n1ru4l (approved):
Nice 😻

dotansimha (approved):

PR #8212, opened 2026-07-12
remove the tokens service; use the same strategy for target access tokens (bentocache) as with organization access tokens

n1ru4l:
**BEFORE DEPLOYING THIS**: The [first commit](https://github.com/graphql-hive/console/pull/8272
) in this branch should be deployed separately, to ensure zero downtime.

**Description:**

Nuke the tokens service and use the same strategy for reading target (legacy) access tokens from postgres/redis cache with a bentocache powered module shared between all services (usage, server) that need to read access tokens.

This removes quite a lot of code and I wanted to do this for a long time. 🙏

**BEFORE**
```mermaid
flowchart LR
    direction TB

    beforeApi["GraphQL API"]
    beforeUsage["Usage Service"]
    tokensTrpc["Tokens Service<br/>tRPC API"]
    oldL1["In-memory cache"]
    oldL2["Redis cache"]
    oldDb[("Postgres<br/>tokens table")]

    beforeApi -->|"tRPC request"| tokensTrpc
    beforeUsage -->|"tRPC request"| tokensTrpc
    tokensTrpc --> oldL1
    oldL1 -->|"miss"| oldL2
    oldL2 -->|"miss"| oldDb
```

**After**

```mermaid
flowchart LR
    direction TB

    afterApi["GraphQL API"]
    afterUsage["Usage Service"]
    apiCache["Local BentoCache<br/>L1 memory"]
    usageCache["Local BentoCache<br/>L1 memory"]
    sharedRedis[("Shared Redis<br/>L2 cache")]
    sharedDb[("Postgres<br/>tokens table")]

    afterApi -->|"validate token"| apiCache
    afterUsage -->|"validate token"| usageCache

    apiCache -->|"L1 miss"| sharedRedis
    usageCache -->|"L1 miss"| sharedRedis
    sharedRedis -->|"L2 miss"| sharedDb

    afterApi -->|"create / delete / list"| sharedDb
    afterApi -.->|"populate / purge cache"| sharedRedis
```

Closes https://github.com/graphql-hive/console/pull/8112

PR #8272, opened 2026-07-27
remove dependency of `server` and `usage` service on the `tokens` service by using a local L1/L2 bentocache

n1ru4l:
applies the first commit of https://github.com/graphql-hive/console/pull/8212, which allows us to roll out the change without a brief downtime, where the graphql/usage service talks to a tokens service that does no longer exist.

Introduces two tasks on the workflow service:
1. purge access tokens from l2 cache (to make sure retries succeed in case redis hicks up

2. batch update last used date, to reduce pg pressure

```mermaid
flowchart TD
    A["Cron runs every minute"] --> B["Load token sets from the last 5 minute buckets"]
    B --> C["Keep the newest bucket for each token"]
    C --> D["Batch-update token dates in PostgreSQL"]
    D --> E["Delete processed Redis buckets"]

### CODE REVIEW

File: packages/services/api/src/modules/token/providers/target-token-cache.ts [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+    return await this.cache.set({
+      key: record.token,
+      value: record,
+      ttl: '5min',
+      grace: '24h',
+    });
+  }
+
+  async purge(tokens: readonly string[]) {
+    await Promise.all(tokens.map(token => this.cache.delete({ key: token })));
```
jdolle:
```suggestion
    await this.cache.deleteMany({ keys: tokens });
```

packages/services/api/src/modules/token/providers/target-token-cache.ts
```diff
@@ -0,0 +1,78 @@
...
+      default: 'targetTokens',
+      plugins: prometheusConfig.isEnabled
+        ? [
+            prometheusPlugin({
+              prefix: 'bentocache_target_tokens',
+            }),
+          ]
+        : undefined,
+      stores: {
+        targetTokens: bentostore()
```
jdolle:
```suggestion
        targetTokens: bentostore({ prefix: 'bentocache:target-tokens' )
```
No need to duplicate prefix in drivers. It can be defined in the store.

packages/services/api/src/modules/token/providers/target-token-cache.ts
```diff
@@ -0,0 +1,78 @@
...
+export class TargetTokenCache {
+  private cache: BentoCache<{ store: ReturnType<typeof bentostore> }>;
+
+  constructor(
+    @Inject(REDIS_INSTANCE) redis: Redis,
+    private pool: PostgresDatabasePool,
+    prometheusConfig: PrometheusConfig,
+  ) {
+    this.cache = new BentoCache({
+      default: 'targetTokens',
```
jdolle, line 36:
```suggestion
      default: 'targetTokens',
      suppressL2Errors: false,
```
Removing the token from the cache is critical. I think we should explicitly handle errors and retry this and/or push it to a retry queue.

packages/services/api/src/modules/token/providers/target-token-cache.ts
```diff
@@ -0,0 +1,78 @@
...
+        : undefined,
+      stores: {
+        targetTokens: bentostore()
+          .useL1Layer(
+            memoryDriver({
+              maxItems: 10_000,
+              prefix: 'bentocache:target-tokens',
+            }),
+          )
+          .useL2Layer(redisDriver({ connection: redis, prefix: 'bentocache:target-tokens' })),
```
jdolle:
```suggestion
          .useL2Layer(redisDriver({ connection: redis, prefix: 'bentocache:target-tokens' }))
          .useBus(
            redisBusDriver({ 
              connection: redis 
            })
          ),
```
We're not currently syncing this across service instances. We have multiple APIs and usage service instances and so we need to synchronize the caches. (When one instance deletes, others need to delete from their L1 cache also)

packages/services/api/src/modules/token/providers/target-token-cache.ts
```diff
@@ -0,0 +1,78 @@
...
+      .getOrSet({
+        key: hashedToken,
+        factory: () => TargetTokenStorage.findByHash({ pool: this.pool })(hashedToken),
+        ttl: '5min',
+        grace: '24h',
+      })
+      .then(result => {
+        if (result) {
+          void TargetTokenStorage.touchTokenByHash({ pool: this.pool })(hashedToken).catch(
+            () => {},
```
jdolle: I'd like the error cases to be logged. This should be very uncommon and it may help us find the cause of odd auth issues in the future.

packages/services/api/src/modules/token/providers/token-storage.ts
```diff
@@ -75,59 +67,37 @@ export class TokenStorage {
...
-          }),
-      ),
-    );
-
-    return deletedIds;
+    const deletedTokens = await this.targetTokenStorage.deleteTokens({
+      targetId: input.targetId,
+      tokens: input.tokenIds,
+    });
+    await this.targetTokenCache.purge(deletedTokens);
```
jdolle, line 74: ^ if not retried inside purge then should be retried here. Or consider calling `this.invalidateTokens` and retrying there. Since the logic is "delete and invalidate"

jdolle (changes requested):

packages/services/api/src/modules/token/providers/target-token-cache.ts
n1ru4l: This PR aims to implement the same feature set as before. We have never used the bus driver and for now the 5 minute propagation delay (which we communicate to the user) seemed to be a good enough trade off.

Happy to tackle this as a follow up PR if you have concerns around the 5 minute delay.

packages/services/api/src/modules/token/providers/token-storage.ts
n1ru4l: In this case such an error would delay invalidation by 5 minutes (in worst case scenario that the access token was last used one second before the purge) - unless postgres goes down immediately after in which the grace period would take over again. I think it is an okay trade off.

packages/services/api/src/modules/token/providers/target-token-cache.ts
n1ru4l: it is now done async in the workflows service

jdolle (approved):

PR #8274, opened 2026-07-27
add test cases demonstrating implementation shortcomings

n1ru4l:


packages/libraries/core/src/client/add-typenames.ts
```diff
@@ -14,19 +14,24 @@ import {
...
  * type that implements an abstract type (i.e. it appears as a possible type
  * of some interface or union in the schema).
  *
  * Requires the schema so it can resolve field return types as it walks the
  * document tree.
+ *
+ * The requested typename is mapped to "__hive_typename__" to avoid conflicting
+ * with crazy aliases like `__typename: id
  */
 export function addTypenames(document: DocumentNode, schema: GraphQLSchema): DocumentNode {
```
jdolle, line 36: Should this be renamed to `addHiveTypenames` since this was adjusted?

jdolle:
I've addressed all the issues called out in these tests.
Once satisfied, we can merge this into the feature branch to prep a release.

packages/libraries/core/src/client/add-typenames.ts
n1ru4l: that makes sense to me

packages/libraries/yoga/src/index.ts
```diff
@@ -156,6 +158,10 @@ export function useHive(clientOrOptions: HiveClient | YogaPluginOptions): Plugin
           record.executionArgs = args;
 
           if (!isAsyncIterable(result)) {
+            if (result.data) {
+              hideInjectedTypenames(result.data);
+            }
```
n1ru4l: Do we also apply this even when `fieldLevelMetricsEnabled` is not set?

n1ru4l:
@jdolle feel free to merge this into your branch if needed. I just created this draft PR to quickly share examples

PR #8275, opened 2026-07-27
Upgrade federation-composition to support oneOf directive in the public SDL

jdolle:
### Background

oneOf is currently only being composed into the supergraph. The package is filtering out the directive from the public SDL

### Description

Upgrades federation-composition package to support outputting the directive definition in the public sdl

### Checklist

- [X] Testing

### DIFF
--- a/packages/services/schema/__tests__/federation.spec.ts
+++ b/packages/services/schema/__tests__/federation.spec.ts
@@ -1,5 +1,6 @@
 import { parse } from 'graphql';
 import { composeAndValidate } from '@apollo/federation';
+import { composeServices as hiveComposeAndValidate } from '@theguild/federation-composition';
 
 test('patch', () => {
   const result = composeAndValidate([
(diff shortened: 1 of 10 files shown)


packages/services/schema/__tests__/federation.spec.ts
```diff
@@ -52,3 +53,34 @@ test('patch', () => {
     expect.stringMatching('Unknown type "Review"'),
   );
 });
+
+test('oneOf directive', async () => {
```
jdolle, line 57: adds a happy path unit test to make sure this behavior is consistent

jonathanawesome (approved):

n1ru4l:
@jdolle This PR does not contain a test for the happy path (schema publish + schema check + maybe schema promotion), can you please create these tests to make sure that nothing on the GraphQL API layer strips out the `@oneOf` directive?

PR #8277, opened 2026-07-28
Track composition worker memory usage in trace

jdolle:
### Background

We want to track composition worker stats by target and composition type. This will allow us to proactively monitor and address issues before they the worker starts running out of memory.

### Description

Adds memory used in the composition worker as a percent to the trace


n1ru4l (changes requested):
What benefit does this bring us? We cannot correlate memory usage with a target id as the maximum memory usage depends on when garbage collection ran the last time.

The only meaningful thing I could think of would be to measure memory before and after the composition in order to have a ballpark on how much memory a single composition accumulates.

I am also a bit concerned of the prometheus cardinality this will introduce.

I see all kinds of measure for attempting to detect when we need to bump memory - yet we are running OOM today - which is still unaddressed. What I still haven't seen is a breakdown (flame graph) that gives us an understanding of what we can improve within stitching for not having that happen in the first place. This is where we should start.

jdolle:
Confirmed on dev
Screenshot 2026-07-29 at 2 59 17 PM

n1ru4l (approved):
Nice! 🥳

PR #8283, opened 2026-07-28
Federation oneof int test

jdolle:
### Background

https://github.com/graphql-hive/console/pull/8275 addresses feedback from

### Description

Adds integration tests for schema publish, check, and promotion to ensure oneOf directive is included in the resulting schemas.

### Checklist

- [X] Testing


n1ru4l (approved):
Some tests are failing - Instead of adding the fetching to all existing tests we could use the `getSchemaVersionWithAllDetails` helper for those cases where we are actually interested in the SDL.

PR #8287, opened 2026-07-29
Add link back to hive console in metric alert notification 

jonathanawesome:
This PR adds a link back to Hive Console in metric alert notifications. Slack and Teams messages now carry a "View alert in Hive" link to the rule detail page, and the webhook payload gains a top-level `url`.

The workflows service had no notion of the Console's base URL, so this adds an optional `WEB_APP_URL` to it. The env var is optional so self-hosted deployments keep booting without it, in which case notifications go out unchanged and `url` is `null`.


n1ru4l (approved):
🥳

PR #8291, opened 2026-07-30
Update to Node.js `24.18.1` and setup trusted publishing

n1ru4l:
Blocked by https://github.com/docker-library/official-images/pull/21951

I also got tired of replacing the version in to many files, so I changed the setup a bit so that .node-version  is the single source of truth for the whole repo.


dotansimha (approved):
thanks for also making is less hard-code

PR #8301, opened 2026-08-03
feat(workflows): add EMAIL_PROVIDER_SMTP_IGNORE_TLS

### LINKED ISSUE #1561: Email: Sending emails does not work if the server does not have a TLS.
I need to pass the **ignoreTLS** parameter [here](https://github.com/kamilkisiela/graphql-hive/blob/5e62d45f046bd895a8f88ed7f0116562dd9a1b84/packages/services/emails/src/providers.ts#L85)

Please friends add this feature.

overpod:
### Background

Resolves #1561.

Sending emails fails when the SMTP server advertises `STARTTLS` but cannot actually complete a TLS
handshake. The existing `EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED` variable only relaxes certificate
validation, which does not help when TLS itself is unusable — nodemailer still attempts the upgrade
and the connection fails.

### Description

Affected package: `@hive/workflows` (the `smtp` email provider).

Adds a new optional environment variable `EMAIL_PROVIDER_SMTP_IGNORE_TLS`, which maps to the
nodemailer [`ignoreTLS`](https://nodemailer.com/smtp/#tls-options) transport option.

- `packages/services/workflows/src/environment.ts` — added the variable to `SMTPEmailModel` (same
  `'0' | '1'` shape as the neighbouring `EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED`) and wired it into
  the SMTP provider config.
- `packages/services/workflows/src/lib/emails/providers.ts` — added `ignoreTLS` to
  `SMTPEmailProviderConfig` and passed it to `nodemailer.createTransport`.
- `packages/services/workflows/README.md` — documented the new variable, plus
  `EMAIL_PROVIDER_SMTP_REJECT_UNAUTHORIZED` which was previously missing from the table.

Notes on behaviour:

- The variable is **opt-in** and defaults to `0`. Unlike `REJECT_UNAUTHORIZED` (which defaults to
  enabled via `!== '0'`), this one is checked with `=== '1'`, so deployments that do not set it keep
  their current behaviour and STARTTLS stays enabled.
- nodemailer only honours `ignoreTLS` when `secure` is `false`, so setting
  `EMAIL_PROVIDER_SMTP_PROTOCOL=smtps` is unaffected by this option.

### Checklist

- [x] Input validation
- [x] Communication security
- [x] System configuration


n1ru4l (approved):
Thank you for this contribution!

PR #8305, opened 2026-08-03
decrypt the Slack token when dispatching metric alert notifications

jonathanawesome:
This PR fixes metric alert Slack notifications failing with `invalid_auth`. `organizations.slack_token` is written encrypted by the API, but the workflows service read the column raw and handed the ciphertext to Slack as a bearer token. Every org that connected Slack after encryption was introduced is affected.

To fix it, the encryptor moves out of the API's `CryptoProvider` into `@hive/service-common` so services without graphql-modules can decrypt, with a frozen ciphertext vector pinning the wire format. The workflows service gains an `ENCRYPTION_SECRET` variable (Pulumi, compose, env template), kept optional so an existing self-hosted deployment keeps booting on upgrade: without it Slack notifications are skipped, rather than the whole service, including all transactional email and crons, failing to start. The dispatcher then decrypts the token before constructing the Slack client, using `possiblyRaw` so legacy plaintext tokens keep working, and returns `failed-config` instead of throwing when the secret is missing or wrong, so an unfixable error cannot burn all 25 graphile-worker retries.

While tracing this, two paths turned out to record `outcome = 'sent'` while delivering nothing: a channel missing its settings, and a webhook whose retry budget was already spent. Both now report `skipped-config` and `gave-up`, the dedupe row is only written on a real delivery, and the Grafana panel documents the new outcomes. This is why the dashboard showed no failures for the misconfigured cases.

Note that the fix is not retroactive: already-failed jobs sit at `attempts = 25` and are never retried, so alerts from the broken window are lost and notifications resume on the next state transition. Self-hosters running custom manifests need to set `ENCRYPTION_SECRET` on the workflows service to the same value as the API.


packages/services/service-common/src/crypto.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+/**
+ * Shared by every service that touches an encrypted column (`organizations.slack_token`,
+ * OIDC client secrets, external composition secrets). The wire format is load-bearing:
+ * values encrypted by one service are decrypted by another, so changing the algorithm or
+ * the output shape silently bricks every stored secret. `crypto.spec.ts` pins it with a
+ * frozen ciphertext vector.
+ */
+export function createEncryptor(encryptionSecret: string): Encryptor {
```
n1ru4l: You can just make this a class and remove the `crypto.ts` file in `modules/shared` .

Then just provide a new global instance for DI similar to how we provide the `PostgresDatabasePool`: https://github.com/graphql-hive/console/blob/6b5c0de68629560329644d33b67aca452892437c/packages/services/api/src/create.ts#L302-L306

n1ru4l (approved):

PR #8308, opened 2026-08-03
Fix usage collector error catching

jdolle:
### Background

The usage client is capable of crashing a node application due to an unhandled error if collectSchemaCoordinates throws.

### Description

Captures this error and reports it as intended.

### Checklist

- [X] Testing

### DIFF
--- a/packages/libraries/core/tests/client/agent.spec.ts
+++ b/packages/libraries/core/tests/client/agent.spec.ts
@@ -1,0 +1,41 @@
+import { createHiveTestingLogger } from 'test-utils.js';
+import { createAgent } from '../../src/client/agent.js';
+
+describe('createAgent', () => {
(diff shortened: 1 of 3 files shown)


jonathanawesome (approved):

packages/libraries/core/tests/client/agent.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { createHiveTestingLogger } from 'test-utils.js';
+import { createAgent } from '../../src/client/agent.js';
+
+describe('createAgent', () => {
```
n1ru4l, line 4: Could we create a more real world scenario for the specific error that happened at runtime (involving an actual GraphQL schema)? E.g. if there is no subscription type and a subscription operation is attempted to be expected, we should probably never raise an exception in the first place? Or is this something that can only happen how the SDK is used within Hive Gateway?

PR #8322, opened 2026-08-04
Add error case tests for usage collector

jdolle:
### Background

Our collector throws when it encounters unexpected conditions I.e. missing types or fields required to resolve the coordinate.

### Description

This PR adds tests to verify this is happening and to document this case.
There is some debate on whether or not this function should throw or handle these cases silently. I believe throwing is still correct because it allows the caller to decide what to do about this case. It's feasible that this function could be used in cases where we don't want to swallow the error.

I believe the recent change to the capture function handles the collector issue correctly and this test is all that we should add.

### Checklist

- [X] Testing


n1ru4l (approved):

n1ru4l:
I figured out what actually causes the issue we were facing over here: https://github.com/graphql-hive/console/pull/8380

Issue #8098, opened 2026-06-01
UI notification/alert for projects on legacy composition


n1ru4l:
Add a clear indicator when a project is still using the legacy federation composition workflow, along with guidance encouraging migration to the native federation composition approach.

This applies only to existing projects created before native federation composition became the default, as all new projects already use the native workflow.

Providing this visibility would help prevent confusion during project reviews or re-evaluations, where customers may encounter issues caused by older projects still relying on the legacy composition method.

ben-m-klein:
I'll work on a PR for this notification soon.

PR #8323, opened 2026-08-04
Feature/legacy composition UI alert

ben-m-klein:
### Background

Closing #8098

Older projects using the legacy federation composition workflow are currently lacking a warning, which might result in some confusion.

### Description

This is a small change, adding a warning banner when a federation project is on Legacy Federation v1 composition. It follows the same pattern as the RateLimitWarn banner (Callout type="warning"). The banner links to the composition settings and the announcement for native Federation v2: https://the-guild.dev/graphql/hive/product-updates/2023-10-10-native-federation-2

image


jdolle (approved):
Thank you for this contribution. This is very helpful for making teams aware that they're using this legacy composer and nudging them to upgrade.

PR #8328, opened 2026-08-05
Fix link to composition compatibility report

jdolle:
### Background

Compatibility report link doesn't work

### Description

Fixes link for compatibility report.
Also addresses an issue with specifically the local cdn urls. This uses the default docker host local url to reference the machine's service rather than using localhost, which in docker uses the docker container's localhost.

### DIFF
--- a/packages/web/app/src/components/project/settings/native-composition.tsx
+++ b/packages/web/app/src/components/project/settings/native-composition.tsx
@@ -319,7 +319,8 @@ export function NativeCompositionSettings(props: {
                   To determine the cause of this issue, view the{' '}
                   <Link
                     variant="primary"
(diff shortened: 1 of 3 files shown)


packages/web/app/src/components/project/settings/native-composition.tsx
```diff
@@ -319,7 +319,8 @@ export function NativeCompositionSettings(props: {
                   To determine the cause of this issue, view the{' '}
                   <Link
                     variant="primary"
-                    href={`/native-composition-compatibility-report/${projectQuery.data.project.id}`}
```
jdolle, line 322: I dislike that the type for this element includes the href attribute... such a bait lol

jonathanawesome (approved):

PR #8343, opened 2026-08-07
Print all SDL inserted into store and version store for consistent formatting

jdolle:
### Background

Description and directive whitespacing can differ in graphs. This causes issues when viewing the diff in console.
Screenshot 2026-08-10 at 10 01 40 PM

### Description

This change formats all SDL immediately before inserting into our stores to guarantee the formatting will be consistent.

However, this is yet another parse and print that needs to happen.

### Checklist

- [ ] Testing -- needs an integration test to make sure SDL string matches when published and when checked.


n1ru4l (changes requested):
Before fixing an (unconfirmed) behaviour it would make sense to first have a failing test

n1ru4l:
Closed via https://github.com/graphql-hive/console/pull/8361

PR #8345, opened 2026-08-07
card component consolidation

jonathanawesome:
This PR consolidates duplicated component families down to one implementation each, with cards as the main thread. Four components are deleted outright and every call site moves onto a `base/` equivalent, which is where the bulk of the diff comes from: 112 files, +5203/-4106.

## Cards

`@/components/ui/card` and `@/components/v2/card` are both gone. Every card in the app now renders through `base/card`.

Absorbing the two of them meant giving `base/card` variants for the things the old components expressed through `className`, since `base/card` deliberately takes none:

- `onSurface` — `base` (border only) or `raised` (filled, one step off the page). `raised` matches ui/card's default fill and the `bg-neutral-2/50` that most call sites were setting by hand, so both converge on the same value.
- `titleSize` — `default` (`text-sm`), `large` (`text-lg`, ui/card's `CardTitle`) and `xlarge` (`text-2xl`, the auth card).
- `bodyPadding` — `default` or `none`, for content that runs edge to edge such as the project card's full-bleed sparkline.
- `interactive` — hover treatment for a card that is itself a link.

Layout that used to sit on the card root — grid placement, `h-full`, margins, min-heights — moved to wrapper elements at the call sites. Where a card needed to fill a row, the wrapper is `grid` rather than `flex flex-col`, because grid items stretch on the block axis and flex column items do not.

Two structural patterns were resolved at the call sites instead of by adding API: `CardFooter` folded into `children` as a trailing row, and a card that had two `header`/`content` pairs became one header with the second caption inside the body.

## Stat cards

19 near-identical stat tiles across the four insights files are now one `base/stat-card`, with `tone` covering the success/failure colouring and an optional info-icon `hint`. In `insights/stats.tsx` this replaced seven separate wrapper components whose only real differences were an icon, a label and a formatting hook.

The insights pages also had six panels that never imported a Card at all — hand-rolled `<div className="border-neutral-5 bg-neutral-2/50 rounded-md border p-5">` with `Section.Title` inside. Those are now `base/card` too, so they were invisible to the migration inventory but shipped as cards.

## Not-found

Four components collapsed into one `base/not-found`. Two of them were both exported as `NotFound`, differing only in a card wrapper and a back button, and named the same two props differently (`title`/`description` vs `heading`/`subheading`). The other two were the route-level 404 and a resource-access variant with its own illustration and horizontal layout, both now variants: `layout`, `illustration`, `fullScreen` and an optional `bigHeading` for the "404" display line. The route-level Sentry `captureMessage` moved into `router.tsx`, where the other routing concerns already live.

## Project and target cards

`ProjectCard` and `TargetCard` were ~172 lines each with only 49 differing — the ~70-line ECharts config, the data math and both tooltips were byte-identical. They now share `common/resource-card`, which takes the link as a render prop so each call site keeps TanStack's route and param checking on its own literal route.

## Radio groups

`ui/radio-group.tsx` and `v2/radio-group.tsx` are deleted, with all call sites on `base/radio-group`.

## Previews

The foundry preview environment now renders real app components, not just design-system primitives, so a base-component change can be judged against the compositions that actually ship. New previews cover Card, StatCard, NotFound, RadioGroup and BillingPlanPicker. The temporary per-call-site migration inventories that tracked ui/card and v2/card have been deleted now that both are gone.

## Other fixes on this branch

- Fix `DropdownMenuSubContent` overflow
- Add a shared scrollbar style
- Fix the insights-and-alerts seed: write the plan's real usage limits (org creation always applies HOBBY's column defaults whatever plan you pick) and clamp the backfill to that plan's retention
- Raise `max_partitions_per_insert_block` to 10000 for local dev ClickHouse only...the minutely tables partition by hour, so a 30-day backfill exceeds the default of 100 and puts the usage ingestor into a replay loop
- Refactor `SchemaEditor` to contain theme resolving
- Simplify `SubPageLayout` and replace its Card usage
- Remove unused `DocsNote`
- Clean up `DocsLink` and `ProductUpdatesLink`
- CSS updates: scroll styling, basic semantic colors, update Tailwind's `@variant`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docker/configs/clickhouse-dev/insert-partitions.xml
```diff
@@ -0,0 +1,13 @@
+<clickhouse>
+  <profiles>
+    <default>
+      <!--
+        The seed scripts backfill weeks of history, the usage service coalesces many reports into
+        one Kafka message, and the minutely rollups partition by hour, so a single INSERT can carry
+        far more than the default 100 partitions. A rejected insert wedges the usage ingestor in a
+        replay loop. See scripts/seed-insights-and-alerts/README.md.
+      -->
+      <max_partitions_per_insert_block>10000</max_partitions_per_insert_block>
```
jdolle, line 10: Prefer setting this in your insert script to avoid impacting standard behavior.
```
SET max_partitions_per_insert_block = 500;
INSERT INTO ...
```

packages/web/app/src/components/base/button/button.tsx
```diff
@@ -120,6 +120,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(function Button
           ? '' // segmented/icon buttons handle their own padding
           : 'h-9 px-4 gap-1.5 text-[13px]';
 
+  console.log('sizClass', { sizeClass, size });
```
jdolle:
```suggestion
```
remove log

jdolle (dismissed):
Some of the headings under the various settings pages become gray instead of white.

This happened previously on the schema checks details page as well, and it makes it more difficult to scan due to the less obvious visual separation.

Image

packages/web/app/src/components/base/floating/select/select.tsx
```diff
@@ -80,6 +80,7 @@ export function Select({
           trigger ??
           ((
             <Button
+              size="default"
```
jdolle: Unnecessary. Default is default

packages/web/app/src/components/base/not-found/not-found.tsx
```diff
@@ -0,0 +1,120 @@
...
+const illustrations = {
+  ghost,
+  connection,
+};
+
+/**
+ * The copy the project and target layouts and the alert-rule page all show for a resource that is
+ * missing or not visible to you. Shared rather than repeated, since all three are word for word
+ * identical and it used to be hardcoded inside `ResourceNotFoundComponent`.
+ */
```
jdolle: I think we can ditch the majority of the comments across all files.
These specific comments are more about where these options are used rather than the intended purpose of these options, which is better explained by the simple names, like "fullscreen".

packages/web/app/src/components/base/stat-card/stat-card.tsx
```diff
@@ -0,0 +1,57 @@
...
+    },
+  },
+  defaultVariants: {
+    tone: 'default',
+  },
+});
+
+type StatCardProps = {
+  title: string;
+  /** A node rather than a string, so a call site can suffix the figure (`1,204,880 (3,120 errors)`). */
```
jdolle, line 23: It makes sense for all inputs to layout elements to be ReactNode. This is the most flexible since it can be an element or string or number, etc... but technically a string can also be add a suffix.

I think in the case of errors it's because we want to color the errors a red color though?

packages/web/app/src/components/common/GraphQLSDLBlock.tsx
```diff
@@ -11,7 +11,6 @@ export function GraphQLHighlight({
   const pretty = usePrettify(code);
   return (
     <SchemaEditor
-      theme="vs-dark"
```
jdolle, line 14: can this infer the desired theme based on the system?

docker/configs/clickhouse-dev/insert-partitions.xml
jonathanawesome: This is a local/dev-only change that unlocks the insights and alerts seed, which POSTS the usage reports over HTTP. The INSERT is within the usage-ingestor, which would be impacting and is untouched here.

packages/web/app/src/components/base/stat-card/stat-card.tsx
jonathanawesome: In the StatCard, yes, we use ReactNode because of this single-use situation where we show an error count in red:

Screenshot 2026-09-10 at 7 18 53 AM


I would make the case that we should pull resolution errors out into their own StatCard and use the variant that's used on the insights index, but that's a different discussion:

Screenshot 2026-09-10 at 7 27 48 AM


Limiting the title here to `string` is intentional. An implementor _can't_ add arbitrary content, they should instead use variants to match existing patterns.

packages/web/app/src/components/common/GraphQLSDLBlock.tsx
jonathanawesome: Yep, that is part of this update and allows us to remove the theme prop.

n1ru4l (approved):

PR #8346, opened 2026-08-09
Bump codegen packages

eddeee888:
### Background

This PR:
- bumps codegen packages to latest versions
- update related config
- update type referencing
- update `@graphql-codegen/schema-ast` patch


pnpm-workspace.yaml
```diff
@@ -25,3 +25,5 @@ minimumReleaseAgeExclude:
   - '@theguild/federation-composition'
   - graphql-modules
   - 'react-foundry'
+  - '@eddeee888/gcg-server-config@0.6.0' # TODO: eddeee888 remove when minimumReleaseAge passes
+  - '@eddeee888/gcg-typescript-resolver-files@0.18.0' # TODO: eddeee888 remove when minimumReleaseAge passes
```
eddeee888: TODO: remove this before merging.
Creating the PR early to ensure CI passes

codegen.mts
```diff
@@ -99,15 +97,13 @@ const config: CodegenConfig = {
       documents: ['./packages/libraries/core/src/client/**/*.ts'],
       config: {
         flattenGeneratedTypes: true,
-        onlyOperationTypes: true,
       },
-      plugins: ['typescript', 'typescript-operations'],
```
eddeee888, line 104: `onlyOperationTypes` is used by `typescript` plugin, but in the latest major version `typescript-operations` works by itself, so both the plugin and option can be removed

codegen.mts
```diff
@@ -119,4 +115,4 @@ const config: CodegenConfig = {
   },
 };
 
-module.exports = config;
+export default config;
```
eddeee888, line 124: Looks like this file had `.cjs` extension before, so we needed `module.exports`. This PR changes both [eslint config target](https://github.com/graphql-hive/console/pull/8346/changes#diff-e1ce717e1179a0db14e90ec5374768a206651ca807db58352991eb7895ed7c9cL56) and this export line to avoid errors being reported to IDE

packages/services/api/src/modules/organization/providers/organization-manager.ts
```diff
@@ -1,5 +1,4 @@
 import { Inject, Injectable, Scope } from 'graphql-modules';
-import { OrganizationReferenceInput } from 'packages/libraries/core/src/client/__generated__/types';
```
eddeee888, line 2: Other declarations in this file seem to use the local `GraphQLSchema.OrganizationReferenceInput` instead of the one from the client library. This helps with consistency and avoid cross-imports.

The client library also just generates used schema types for itself, in the latest version, so we can't depend on types from that library anymore.

There are a few other similar replacement in this PR.

codegen.mts
```diff
@@ -65,7 +65,11 @@ const config: CodegenConfig = {
       documents: ['./packages/web/app/src/(components|lib|pages|server)/**/*.ts(x)?'],
       preset: 'client',
       config: {
+        enumType: 'native',
         scalars: {
+          ID: 'string',
+          DateTime64: 'string',
+          JSONObject: 'Record<string,unknown>',
```
eddeee888, line 72: Previously, unknown scalars are generated as `any`.
Now, they are generated as `unknown` by default, so we need to add these to avoid type errors.

We could use `any` to achieve the same behaviour, but I think explicit types give better/safer DX

codegen.mts
```diff
@@ -65,7 +65,11 @@ const config: CodegenConfig = {
       documents: ['./packages/web/app/src/(components|lib|pages|server)/**/*.ts(x)?'],
       preset: 'client',
       config: {
+        enumType: 'native',
```
eddeee888, line 68: Previously, native enums are generated by default.
In the latest major version, `string-literal` is the default. 

We have already used native enums in some places so we need to ensure native enum is generated

codegen.mts
eddeee888: Actually `JSONObject` could be `any` so we don't have to do these  `as ...` in our code like [this](https://github.com/graphql-hive/console/pull/8346/changes#diff-f37f223c76a281374a7e9341a550cf6a5b5ab06841220c61ef368e084a301928R1640-R1648)

Do we have preferences? 🤔

packages/web/app/src/pages/target-trace.tsx
```diff
@@ -1659,7 +1668,7 @@ function SpanSheet(props: SpanSheetProps) {
                     fontSize: 10,
                     minimap: { enabled: false },
                   }}
-                  code={span.spanAttributes['graphql.document']}
+                  code={span.spanAttributes['graphql.document'] as string}
```
eddeee888, line 1671: I _think_ `span.spanAttributes['graphql.document']` could be `string | undefined`? 
But we didn't do `|| ''` here. Unless the server always guarantees the attribute is always there? 🤔

packages/libraries/cli/src/commands/schema/publish.ts
```diff
@@ -336,13 +336,19 @@ export default class SchemaPublish extends Command<typeof SchemaPublish> {
           this.log('Waiting for other schema publishes to complete...');
           result = null;
         } else if (result.schemaPublish.__typename === 'SchemaPublishMissingServiceError') {
-          throw new SchemaPublishMissingServiceError(result.schemaPublish.missingServiceError);
+          throw new SchemaPublishMissingServiceError(
+            result.schemaPublish.missingServiceError || 'Unknown schemaPublish.missingServiceError',
```
eddeee888, line 340: `missingServiceError` (and others below) could be undefined because of the conditional directive being applied on the fragment.

This change is needed because the previous type was not considering the conditional directives. Fixed [here](https://github.com/dotansimha/graphql-code-generator/pull/10918)

packages/libraries/cli/src/helpers/schema.ts
```diff
@@ -137,7 +137,13 @@ export const renderChanges = (maskedChanges: FragmentType<typeof RenderChanges_S
   return t.state.value;
 };
 
-export const renderWarnings = (warnings: SchemaWarningConnection) => {
```
eddeee888, line 140: `SchemaWarningConnection` is the full schema type but we only need a subset of it for this function to work.
I think inlining is fine, as this looks like a utility function, and it only needs a certain interface (which is a subset of the full schema SchemaWarningConnection) to work.

patches/@graphql-codegen__schema-ast@6.1.0.patch
eddeee888, line 1: [Previous patch](https://github.com/graphql-hive/console/pull/6679#discussion_r2022611794) stopped working because of minor differences in the file and hash.
This should apply the same patch to the current version.

jdolle (approved):

PR #8355, opened 2026-08-10
Laboratory: add built in docs component

jonathanawesome:
This PR adds a GraphiQL-style schema documentation pane to the Laboratory, opt-in via a new `enableDocs` prop.

A third rail icon opens docs in the same slot as Collections and History: browse root types, types and fields, search the whole type map, and read descriptions, deprecations and argument defaults. Builder rows get an "Open in Docs" context menu entry, and the editor hover gets an "Open in Docs" link (the Lab now serves that hover itself, rebuilt from the same `getHoverInformation` monaco-graphql uses, so the link sits in the existing card instead of a second one).

`renderLaboratory` enables the pane by default, so UMD embedders like Router need only a version bump.

**Removed:** the `introspection.schemaDescription` setting and its toggle. It was wired to graphql-js's `descriptions` option rather than `schemaDescription` and defaulted to `false` where graphql-js defaults to `true`, so nothing rendered descriptions and the toggle had no discoverable effect. Descriptions now follow `enableDocs`.

https://github.com/user-attachments/assets/00c4a3ab-c060-431b-aa27-40c9a485c31e

### DIFF
--- a/.changeset/large-seas-chew.md
+++ b/.changeset/large-seas-chew.md
@@ -4,3 +4,10 @@
 ---
 
 Lab: Add a schema documentation pane, opt-in via the new `enableDocs` prop.
+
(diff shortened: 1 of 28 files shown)


n1ru4l (approved):
nice, did a check on the dev server and everything works smooth 👍

PR #8357, opened 2026-08-11
Laboratory preflight updates

jonathanawesome:
This PR restores the preflight behaviour that was lost when the laboratory moved into its own package, and closes the "a script can wedge the lab" failure class around it.

The starting point was a bug report: `await lab.prompt()` worked from the Test button but silently did nothing when a preflight script ran as part of an operation. That turned out to be a regression...#7620 shipped the prompt wiring and its metadata on 2026-02-05, and #7697 dropped both when it extracted the lab into `@graphql-hive/laboratory` on 2026-03-06, so prompts have been dead on the operation path since March. Rather than fix that one line, this works from a full parity audit of the GraphiQL-era preflight implementation (still live behind the lab tab switcher) against the package.

- Prompts work during operation runs again, with `lab.prompt(title, defaultValue, { placeholder, description })` and a Cancel button restored.
- Runs are observable and stoppable: logs stream into the pane as a script runs (from every run, not just Test), plus Stop, Clear, and a 30 second execution limit that pauses while a prompt waits on the user.
- Workers are terminated and their object URLs revoked on every exit; `lab.environment.set()` rejects non-scalar values with a warning; logs and errors report the script line and column they came from.
- The editor now says the script is stored as plain text and readable by anyone who can open the lab.

One behavior change to flag for reviewers: the first argument to `lab.prompt()` is once again the field label rather than the input's placeholder, so scripts written since March will look different...pass `{ placeholder }` for the greyed-out hint.


.claude/plans/fix-laboratory-preflight-prompt-during-operation-run.md
n1ru4l, line 1: what is the point of committing this file?

n1ru4l (approved):
Thx 🥳 

Please remove the `.claude` markdown file.

NIT: The changeset seems a bit verbose. 

Regarding:

> One behavior change to flag for reviewers: the first argument to lab.prompt() is once again the field label rather than the input's placeholder, so scripts written since March will look different...pass { placeholder } for the greyed-out hint.

It seems like that could be properly handled in a backwards-compatible way. I will not hard push for it though - since it restores the original behaviour 😓

.claude/plans/fix-laboratory-preflight-prompt-during-operation-run.md
jonathanawesome: There isn't one...I was working through multiple plans at a time and the LLM moved this one into the local plan folder. Removed!

packages/libraries/laboratory/src/components/laboratory/preflight.tsx
```diff
@@ -28,42 +30,50 @@ export const Preflight = () => {
...
           </div>
+          {/*
+            Shown to readers as well as authors: the script runs in the browser of whoever
+            opens the lab, whether or not they can edit it.
+          */}
+          <div className="border-border text-muted-foreground border-b px-3 py-2 text-xs">
+            <TriangleAlertIcon className="mr-1.5 inline size-3.5 align-[-2px]" />
+            This script is stored as plain text and runs in the browser before every request. Anyone
+            who can open this lab can read it, so ask for secrets with{' '}
+            <code className="font-mono">lab.prompt()</code> rather than writing them in.
```
jdolle: True in our hosted environment. Is this also true in the gateway hosted version? Or is that all localStorage?

jdolle (approved):

PR #8361, opened 2026-08-11
Publishing then checking the same schema no longer can cause a diff

jdolle:
### Background

Related to https://github.com/graphql-hive/console/pull/8343 -- Took a long time to track down the cause of the difference.

Pretty printing before saving (as proposed in #8343) also solves this issue, but only because it shifted the pretty printing to the backend. So when the frontend pretty printed one version and not the other, it didnt matter since the front end pretty print was redundant.

### Description

This completely removes the pretty printing from the frontend, which was previously only called on the old schema and not the new schema due to a bug.

I've also added an integration test to put to rest the idea that the SDL being saved to the schema version and the SDL being saved to the schemacheck could differ. As well as an e2e test to guarantee that the frontend doesnt create a difference either.

Another option is to fix the `useEffect` logic to actually call the pretty print. But I'm in favor of removing the pretty print because it removes a parse+print call on the frontend, which can save valuable CPU on large schemas. If we need to format (like #8343) before write, then that is a better approach since we format it once.

There's still a slight discrepancy for the format of block comments between the subgraph's SDL and the supergraph and API schema. But since these are at least consistent for that type of SDL, then this is fine.

E.g. Subgraph shows `"""comment"""` and supergraph + API schema show:
```
"""
comment
"""
```

### Checklist

- [X] Testing

Before:
Screenshot 2026-08-11 at 2 42 45 PM

After:
Screenshot 2026-08-11 at 2 45 56 PM


packages/web/app/src/pages/target-checks-single.tsx
```diff
@@ -538,7 +543,12 @@ function DefaultSchemaView(props: {
       >
         <TabsList className="bg-neutral-5 dark:bg-neutral-3 border-neutral-5 dark:border-neutral-3 w-full justify-start rounded-none border-x border-b">
           {items.map(item => (
-            <TabsTrigger key={item.value} value={item.value} disabled={item.isDisabled}>
+            <TabsTrigger
+              data-testid={item.testId}
```
n1ru4l, line 547: Could the element instead be found via the elements innerText or value property?

n1ru4l (approved):
Good catch!

packages/web/app/src/pages/target-checks-single.tsx
n1ru4l: NIT: Aside form that that `data-testid` and `item.testId`, is a small inconsistency. I would expect `data-test-id` 🤣

PR #8373, opened 2026-08-14
fix: persisted documents hash allow colon

n1ru4l:
Latest graphql codegen introduced hashes being prefixed with `hash:`. In order to support that, we can just allow the colon character for the persisted document hash.

### DIFF
--- a/packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
+++ b/packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
@@ -35,8 +35,8 @@ const AppDeploymentOperationHashModel = z
   .min(1, 'Hash must be at least 1 characters long')
   .max(128, 'Hash must be at most 128 characters long')
   .regex(
-    /^([A-Za-z]|[0-9]|_|-)+$/,
-    "Operation hash can only contain letters, numbers, '_', and '-'",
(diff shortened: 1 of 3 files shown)


packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
```diff
@@ -35,8 +35,8 @@ const AppDeploymentOperationHashModel = z
   .min(1, 'Hash must be at least 1 characters long')
   .max(128, 'Hash must be at most 128 characters long')
   .regex(
-    /^([A-Za-z]|[0-9]|_|-)+$/,
-    "Operation hash can only contain letters, numbers, '_', and '-'",
+    /^([A-Za-z]|[0-9]|_|-|:)+$/,
```
eddeee888, line 38: May I know how you found out about this issue please? 🙂

I'm curious what else I should keep in mind the next time I make changes

n1ru4l: The publish of hive apps persisted documents to the staging hive instance failed 😓

dotansimha (approved):

packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
eddeee888: I see, thanks!

PR #8379, opened 2026-08-18
fix: correct status code for armor protection + do not reject typenames injected via response cache + hive sdk

n1ru4l:
- The armor rule errors are causing a 500 status code instead of a 400 status code -> adds integration tests and fixes the implementation
- The aliases injected by the hive sdk could cause the alias limit rule kick in -> Instead of injecting the typenames in `onParse` (and before `onValidate`), we now inject them in `onExecute` instead, making sure that it cannot be rejected.
- Adds a validation rule to our armor setup to avoid external API users using the aliases we use internally. This is technically breaking, but if people are sending it they are doing something weird :)

### DIFF
--- a/packages/services/server/src/use-armor.ts
+++ b/packages/services/server/src/use-armor.ts
@@ -40,7 +40,6 @@ export function useArmor<
       ctx.addValidationRule(
         maxAliasesRule({
           n: 20,
-          allowList: ['__responseCacheTypeName', '__responseCacheId', '__hive_typename__'],
           onReject: [
             (context, error) => {
               context?.reportError(error);
--- a/packages/services/server/src/use-armor.ts
+++ b/packages/services/server/src/use-armor.ts
@@ -1,4 +1,4 @@
-import type { ParseOptions, Source } from 'graphql';
+import type { ParseOptions, Source, ValidationRule } from 'graphql';
 import { createGraphQLError, type Plugin } from 'graphql-yoga';
 import promClient from 'prom-client';
 import { maxAliasesRule } from '@escape.tech/graphql-armor-max-aliases';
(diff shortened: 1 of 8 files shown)


packages/services/server/src/use-armor.ts
```diff
@@ -37,12 +53,15 @@ export function useArmor<
     onValidate(ctx) {
       const hiveClientVersion = getHiveClientVersion(ctx.context.request.headers.get('user-agent'));
 
+      ctx.addValidationRule(DisallowReservedAliasesRule);
+
       ctx.addValidationRule(
         maxAliasesRule({
           n: 20,
-          allowList: ['__responseCacheTypeName', '__responseCacheId'],
+          allowList: [],
```
n1ru4l, line 61: Not needed, we inject these post validation on our end. Setting an empty array, to not use the default value of `['__typename']`

packages/services/server/src/use-armor.ts
```diff
@@ -37,12 +53,15 @@ export function useArmor<
     onValidate(ctx) {
       const hiveClientVersion = getHiveClientVersion(ctx.context.request.headers.get('user-agent'));
 
+      ctx.addValidationRule(DisallowReservedAliasesRule);
```
n1ru4l, line 56: To make sure nobody uses reserved aliases and messes with us.

packages/services/server/src/use-armor.ts
```diff
@@ -59,13 +78,15 @@ export function useArmor<
               }
             },
           ],
+          propagateOnRejection: false,
```
n1ru4l, line 81: Might be worth figuring out why the default behaviour is to throw and maybe adjust the armor defaults. Not important though to resolve the issue on our end and move on.

packages/services/server/src/use-armor.ts
```diff
@@ -28,6 +28,22 @@ const getHiveClientVersion = (userAgent: string | null) => {
...
+  '__hive_typename__',
+  '__responseCacheTypeName',
+  '__responseCacheId',
+]);
+
+const DisallowReservedAliasesRule: ValidationRule = context => {
+  return {
+    Field(field) {
+      if (field.alias?.value && reservedTypenameAliases.has(field.alias.value)) {
+        context.reportError(createGraphQLError(`The alias "${field.alias.value}" cannot be used.`));
```
jdolle, line 41: I like this idea but i'm not sure if `useArmor` is the right place for it.

On one side, having it here means the check is more efficient. On the other hand, `useArmor` and other plugins should be able to be used independently. If we want to safeguard these fields then it should be in the directive where we use it.

For now I recommend removing this check at least for `__hive_typename__`

jdolle (approved):
Nice find.

packages/services/server/src/use-armor.ts
jdolle: Strange that this sets a default... 🤔

PR #8380, opened 2026-08-18
fix: skip subscription reporting without subscription type

n1ru4l:
### Background

This fixes the root cause that was mitigated in https://github.com/graphql-hive/console/pull/8308

### Description

The Yoga SDK attempts to do usage reporting for subscription operations even though the subscription failed due to the schema not having a subscription root type when using GraphQL.js 16.

This does not happen in GraphQL.js 17, as there is a new validation rule [`KnownOperationTypes`](https://github.com/graphql/graphql-js/pull/3601) that raises a validation error and stops `onSubscribe` being called if the schema has no subscription type.

### DIFF
--- a/packages/libraries/yoga/src/index.ts
+++ b/packages/libraries/yoga/src/index.ts
@@ -42,7 +42,7 @@ type CacheRecord = {
    */
   parsedDocument?: DocumentNode;
   /** persisted document id */
-  experimental__documentId?: string;
+  documentId?: string;
 };
 
 export type YogaPluginOptions = HivePluginOptions & {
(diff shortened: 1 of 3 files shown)


packages/libraries/yoga/src/index.ts
```diff
@@ -207,19 +207,22 @@ export function useHive(clientOrOptions: HiveClient | YogaPluginOptions): Plugin
       };
     },
     onSubscribe(context) {
+      if (!context.args.schema.getSubscriptionType()) {
```
jdolle, line 214: please add comment explaining why this is necessary

jdolle (approved):

packages/libraries/yoga/src/index.ts
n1ru4l: https://github.com/graphql-hive/console/pull/8380/changes/2ed2908fca37b193b847cb0dd2479c84fe08c4e4 🙏

PR #8383, opened 2026-08-19
Track errors thrown by the gateway separately in usage data

jdolle:
### Background

This adds support for errors thrown during the validation phase in the gateway. Previously, if subgraph requests were being made, then only those requests were used to collect the errors.

### Description

Adds a new `gatewayErrors` field to the payload, which will need documentation updated.

The rest of the gateway request is still appended to the `fetches` array as a single "fetch". So in this case, the errors are stripped from this faked "fetch" and added to the gatewayErrors array. This is to avoid double counting the errors, and to keep the gatewayErrors as a consistent place for, well, errors in the gateway.

This also modifies some of the operation building to more closely match the yoga plugin. E.g. using `context.waitUntil` and support for persisted documents.

### Checklist

- [X] Input validation
- [X] Output encoding
- [X] Testing

### DIFF
--- a/.changeset/tangy-ideas-join.md
+++ b/.changeset/tangy-ideas-join.md
@@ -3,4 +3,6 @@
 '@graphql-hive/core': patch
 ---
 
-Support errors thrown by the gateway. I.e. during validation
+Explicitly supports errors thrown in the gateway. I.e. authentication during execution.
+
+This adds a new field to the usage payload that specifically contains errors thrown within the gateway. This separates gateway errors from subgraph errors. All errors are currently ingested similarly into Clickhouse, but this change supports future improvements to show the source (subgraph name or gateway) of errors.
--- a/.changeset/tangy-ideas-join.md
+++ b/.changeset/tangy-ideas-join.md
@@ -1,0 +1,6 @@
+---
+'@graphql-hive/gateway-plugin-console-sdk': patch
+'@graphql-hive/core': patch
+---
+
+Support errors thrown by the gateway. I.e. during validation
--- a/integration-tests/testkit/flow.ts
+++ b/integration-tests/testkit/flow.ts
@@ -83,6 +83,7 @@ function pollInternal(
   /** In milliseconds */
   startTime: number = Date.now(),
 ) {
+  let lastError: unknown;
   setTimeout(
     async () => {
       try {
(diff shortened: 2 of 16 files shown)


integration-tests/testkit/flow.ts
```diff
@@ -92,19 +93,36 @@ function pollInternal(
         } else {
           const waited = Date.now() - startTime;
           if (waited > maxWait) {
-            reject(new Error(`Polling failed. Condition was not satisfied within ${maxWait}ms`));
+            reject(
+              new Error(`Polling failed. Condition was not satisfied within ${maxWait}ms`, {
```
jdolle: Helps us understand why integration tests failed.

integration-tests/testkit/flow.ts
```diff
@@ -92,19 +93,36 @@ function pollInternal(
...
+        lastError = e;
         reject(e);
       }
     },
     Math.round(pollFrequency + Math.random() * jitter),
   );
 }
 
+// Use this function to give our backend (clickhouse) time to insert into the materialized views' tables
+export const waitForExpectations = (expectation: () => Promise<void>) => {
```
jdolle, line 118: moved to the test utilities because this is easier to work with in tests than the pollFor API

integration-tests/testkit/seed.ts
```diff
@@ -291,8 +292,14 @@ export function initSeed() {
...
 
+          if (!orgResult.createOrganization.ok) {
+            throw new Error(
+              `Cannot create organization: ${JSON.stringify(orgResult.createOrganization.error)}`,
+            );
+          }
+
           const organization =
-            orgResult.createOrganization.ok!.createdOrganizationPayload.organization;
+            orgResult.createOrganization.ok.createdOrganizationPayload.organization;
```
jdolle, line 302: I had one test run locally fail because this was null.
I want to capture exactly why this failed if it fails.

.changeset/tangy-ideas-join.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphql-hive/gateway-plugin-console-sdk': patch
+'@graphql-hive/core': patch
+---
+
+Support errors thrown by the gateway. I.e. during validation
```
n1ru4l: This changeset is misleading

integration-tests/tests/gateway-plugin-console-sdk/plugin.spec.ts
```diff
@@ -618,6 +641,130 @@ describe.each(['js', 'rust'] as const)('GraphQL Hive Plugin (%s)', gatewayType =
...
+            typeInfo ??= new TypeInfo(args.schema);
+            const validationContext = new ValidationContext(
+              args.schema,
+              args.document,
+              typeInfo,
+              e => {
+                errors.push(e);
+              },
+            );
+            const visitor = visitInParallel([
```
n1ru4l: why not simply do the following:

```
return setResultAndStopExecution({ data: null, errors: [new GraphQLError("hm", { ... })] });
```

That would reduce the complexity of the test case with the same outcome

n1ru4l (approved):

PR #8385, opened 2026-08-19
update yoga

n1ru4l:
Updates to latest yoga version which brings us one stop closer to using graphql-js@17


packages/services/external-composition/federation-2/package.json
```diff
@@ -15,7 +15,7 @@
     "@whatwg-node/server": "0.10.17",
     "dotenv": "16.4.7",
     "graphql": "16.9.0",
-    "lru-cache": "^7.17.0",
+    "lru-cache": "11.5.2",
```
dotansimha, line 18: big bump

dotansimha (approved):

packages/services/external-composition/federation-2/package.json
n1ru4l: to avoid having 3 versions of the same one xD 👯

PR #8392, opened 2026-08-20
bump graphql-inspector

n1ru4l:
brings us one step closer to upgrading to graphql@17


pnpm-lock.yaml
n1ru4l, line 1: hell yeah nuking all these dupes

PR #8393, opened 2026-08-21
bump graphql-inspector

n1ru4l:
one step closer to graphql@17


pnpm-lock.yaml
n1ru4l, line 1: hell yeah nuking all these dupes

eddeee888 (approved):

PR #8394, opened 2026-08-21
feat: upgrade to GraphQL 17

n1ru4l:


packages/libraries/core/src/client/utils.ts
```diff
@@ -89,8 +90,44 @@ export function cache<R, A, K, V>(
...
+        ...value,
+        arguments: value.arguments ?? [],
+        directives: value.directives ?? [],
+      };
+    case 'Directive':
+      return { ...value, arguments: value.arguments ?? [] };
+    default:
+      return value;
+  }
+};
```
n1ru4l, line 125: as discussed in https://guild-oss.slack.com/archives/C09NPDY1XMH/p1787301570575969

packages/services/api/src/modules/shared/module.graphql.ts
```diff
@@ -33,6 +33,12 @@ export default gql`
   scalar JSONSchemaObject
   scalar SafeInt @tag(name: "public")
 
+  schema {
+    query: Query
+    mutation: Mutation
+    subscription: Subscription
+  }
+
   extend schema
```
eddeee888, line 42: The `extend schema @link` seems to override the implicit default schema declaration in v17 for some reason 🤔 
i.e. it sets the `schema` block as `schema {}`. This causes `Query`, `Mutation` and `Subscription` to be null.

When Codegen validates the schema on 2 generates blocks with `url-introspection` and `typescript-operaitons` plugins,  it sees Query is null, so it throws `Query root type must be provided.`

Let me try to trace back to graphql 17 upgrade guide to see if this is expected.

eddeee888: This should be fixed `@graphql-tools/merge@9.2.4` when this PR is merged: https://github.com/ardatan/graphql-tools/pull/8365

PR #8397, opened 2026-08-21
Deprecate reporting schemas from useHive

jdolle:
### Background

Schema reporting at runtime is an anti pattern that can cause a schema to be published multiple times since it reports on boot. Although we can detect that this schema is the same within hive, it can still take significant bandwidth to validate etc depending on the size of the schema.

One such situation is if this feature is enabled for lambdas, since these can scale rapidly.

### Description

To avoid potential issues with the system, rather than continue supporting this feature by adding a target ID, we've decided to deprecate it and it will eventually be removed.

The method of uploading a schema should be to use `@graphql-hive/cli` to publish the schema within a deployment flow.

### DIFF
--- a/.changeset/large-streets-remain.md
+++ b/.changeset/large-streets-remain.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/core': patch
+---
+
+Deprecate reporting schemas from useHive
--- a/.changeset/large-streets-remain.md
+++ b/.changeset/large-streets-remain.md
@@ -1,5 +1,11 @@
 ---
 '@graphql-hive/core': patch
+'@graphql-hive/apollo': patch
+'@graphql-hive/envelop': patch
+'@graphql-hive/gateway-plugin-console-sdk': patch
+'@graphql-hive/yoga': patch
 ---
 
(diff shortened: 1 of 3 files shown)


.changeset/large-streets-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphql-hive/core': patch
+---
```
n1ru4l, line 3: This should list all the plugins that use core, so they get a proper changelog entry aside from "dependency updated".

packages/libraries/core/src/client/reporting.ts
```diff
@@ -19,6 +19,10 @@ export interface SchemaReporter {
   dispose(): Promise<void>;
 }
 
+/**
+ *
+ * @deprecated Use @graphql-hive/cli to report schemas instead.
+ */
 export function createReporting(pluginOptions: HiveInternalPluginOptions): SchemaReporter {
```
n1ru4l, line 26: Should we add an annoying log message, so people are more aware of it?

n1ru4l (approved):
Approved with two comments

PR #8319, opened 2026-08-04
feat: improve federation supergraph formatting

n1ru4l:
Replaces the following with less whitespace:

```
schema
    @link(url: "https://specs.apollo.dev/link/v1.0")
    @link(url: "https://specs.apollo.dev/join/v0.3", for: EXECUTION)
    
    
    @link(url: "https://specs.apollo.dev/inaccessible/v0.2", for: SECURITY)
    
    
    
    @link(url: "https://docs.commercetools.com/api/feature/v1.0", import: ["@feature"])
```

**As a bonus** this also avoids `parse(print(schemaSdl))` cycles for the native composition.

PR #8398, opened 2026-08-22
Transform external composed sdl to public sdl

jdolle:
### Background

An issue was reported where suddenly a team's public SDL included federated directives

### Description

I was able to identify the cause as this change: https://github.com/graphql-hive/console/pull/8319/changes#diff-f912a95c08d02dee17df2da852f56f215ec961506f1d50eb792971fc09cb5b5aL247

Because the composition library we use automatically make this calls, it was mistakenly assumed that the external composer would also call this. Since this composer is external by definition, I dont think it's safe to make that assumption.

Long term, we should remove this extra transform from all composition logic. Before doing this, we must ensure all existing systems are compatible or we have a plan to migrate.

For now, I've added back the transform based on an environment variable. This will transform the external supergraph SDL result to match the previous behavior.

### Checklist

- [X] Testing

### DIFF
--- a/packages/services/schema/__tests__/external.spec.ts
+++ b/packages/services/schema/__tests__/external.spec.ts
@@ -1,0 +1,107 @@
+import { print } from 'graphql';
+import nock from 'nock';
+import { composeExternalFederation } from '../src/lib/compose';
+
+test('external composition sdl is transformed to a public schema if it includes supergraph SDL', async ({
+  expect,
+}) => {
+  const http = nock('http://localhost')
+    .post('/broker')
+    .once()
+    .reply((_, _body) => {
+      const supergraphSdl = `schema @link(url: "https://specs.apollo.dev/link/v1.0") @link(url: "https://specs.apollo.dev/join/v0.3", for: EXECUTION) {
+  query: Query
+}
+
+directive @join__enumValue(graph: join__Graph!) repeatable on ENUM_VALUE
+
+directive @join__graph(name: String!, url: String!) on ENUM_VALUE
+
+directive @join__field(
+  graph: join__Graph
+  requires: join__FieldSet
+  provides: join__FieldSet
+  type: String
+  external: Boolean
+  override: String
+  usedOverridden: Boolean
+) repeatable on FIELD_DEFINITION | INPUT_FIELD_DEFINITION
+
+directive @join__implements(graph: join__Graph!, interface: String!) repeatable on OBJECT | INTERFACE
+
+directive @join__type(
+  graph: join__Graph!
+  key: join__FieldSet
+  extension: Boolean! = false
+  resolvable: Boolean! = true
+  isInterfaceObject: Boolean! = false
+) repeatable on OBJECT | INTERFACE | UNION | ENUM | INPUT_OBJECT | SCALAR
+
+directive @join__unionMember(graph: join__Graph!, member: String!) repeatable on UNION
+
+scalar join__FieldSet
+
+directive @link(url: String, as: String, for: link__Purpose, import: [link__Import]) repeatable on SCHEMA
+
+scalar link__Import
+
+enum link__Purpose {
+  SECURITY
+  EXECUTION
+}
+
+enum join__Graph {
+  PRODUCTS @join__graph(name: "products", url: "https://products.example.com/graphql")
+}
+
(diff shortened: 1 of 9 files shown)


packages/services/schema/__tests__/external.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    requestId: '1',
+    requestTimeoutMs: 5000,
+    subgraphs: [
+      /** unimportant */
+    ],
+  });
+
+  expect(result.type).toBe('success');
+  expect(result.result.sdl).toMatchInlineSnapshot(`
+    scalar FieldSet
```
n1ru4l: Why is `FieldSet` still there 👁️👁️ ?

packages/services/schema/src/lib/compose.ts
```diff
@@ -28,7 +29,18 @@ const ExternalCompositionResultSuccessModel = z.object({
...
     supergraph: z.string(),
-    sdl: z.string(),
+    sdl: z.string().transform(sdl => {
+      try {
+        /**
+         * Ensure the externally composed schema doesn't include supergraph SDL
+         * This is done for us in the native composition library
+         * https://github.com/graphql-hive/federation-composition/blob/77d6b4ece2abacf94164beafb4e7f5961f726755/src/compose.ts#L228
+         */
+        return print(transformSupergraphToPublicSchema(parse(sdl)));
```
n1ru4l: shouldn't we run `transformSupergraphToPublicSchema` on the `supergraph` instead of the `sdl`?

I also flagged this internally - this restores previous behaviour - which is to not take the public sdl as it is provided by the external composition. I am not sure if we should do that. 🤔 I would argue it is the responsibility of user that uses the external composition to return the proper public schema sdl

n1ru4l: Is in the specific scenario the apollo composition returning an invalid public sdl schema that still contains supergraph specific SDL or is it a custom customer endpoint that is sending wrong stuff for the public sdl?

n1ru4l (changes requested):
The public sdl should be generated from the supergraph not the public sdl passed from the external composition: https://github.com/graphql-hive/console/pull/8398/changes#r3842626702 as seen in https://github.com/graphql-hive/console/pull/8319/changes#diff-f912a95c08d02dee17df2da852f56f215ec961506f1d50eb792971fc09cb5b5aL247

packages/services/schema/__tests__/external.spec.ts
jdolle: It was added to the supergraph sdl incorrectly.

packages/services/schema/src/lib/compose.ts
jdolle: The later. This is a custom external composition implementation that isnt converting the public sdl into a an API Schema

packages/services/schema/src/lib/compose.ts
```diff
@@ -258,14 +269,36 @@ export async function composeExternalFederation(args: {
...
+        sdlDocumentNode = parse(sdl);
+      }
+    } else {
+      sdlDocumentNode = parse(sdl);
+    }
     return {
       type: 'success',
       result: {
         supergraph,
         supergraphDocumentNode: parse(supergraph),
```
n1ru4l: Why parse the supergraph twice? Seems like it we could only do it once

packages/services/schema/src/lib/compose.ts
```diff
@@ -258,14 +269,36 @@ export async function composeExternalFederation(args: {
...
+        const publicAST = transformSupergraphToPublicSchema(ast);
+        sdl = print(publicAST);
+        sdlDocumentNode = publicAST;
+      } catch (e) {
+        // silently fail and revert to using the existing public sdl.
+        // This is the best option if the supergraph cannot be parsed.
+        // This is an extreme edge case because the supergraph should always be
+        // parseable.
+        sdlDocumentNode = parse(sdl);
+      }
```
n1ru4l: Shouldn't we at least log something in this case? It seems like it could otherwise lead to confusion if it happens for some reason

jdolle: I've removed the try/catch. This matches our existing logic and will raise the error.

n1ru4l (approved):

PR #8405, opened 2026-08-25
Improve Server Preset watcher time

eddeee888:
### Description

Previously, Server Preset time doesn't change between initial run vs subsequent runs.
This is because a TypeScript compiler is created between runs, so the previous TypeScript buildinfo and cache goes away.
With the upcoming version of Server Preset, the reference of the TypeScript compiler is kept between runs so it doesn't have to start from scratch (observed ~3s for TypeScript to compile the first time)

| Type | Old | New | Changes |
| -- | -- | -- | -- |
| Initial run, with all generates block | 7.33s | 7.33s | N/A |
| Subsequent runs, with all generates block | 7.33s | 5.06s | -30.9% |
| Subsequent runs, with only Server Preset | 4.53s | 3.02s | -33% |

I think there are space to optimise for this a bit more. I'm hopeful we can get to ~3s with all generates block

For comparison: 
- a run (initial or subsequent) of just client-preset takes ~1.4s
- a run (initial or subsequent) of everything but Server Preset takes ~2.2s

### DIFF
--- a/codegen.mts
+++ b/codegen.mts
@@ -5,6 +5,9 @@ import { addTypenameSelectionDocumentTransform } from '@graphql-codegen/client-p
 const config: CodegenConfig = {
   schema: './packages/services/api/src/modules/*/module.graphql.ts',
   emitLegacyCommonJSImports: true,
+  overwrite: {
+    removeStaleFiles: false,
+  },
   generates: {
     // API
     './packages/services/api/src': defineConfig(
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@0no-co/graphqlsp": "1.15.4",
     "@changesets/changelog-github": "0.7.0",
     "@changesets/cli": "2.31.1",
-    "@eddeee888/gcg-typescript-resolver-files": "0.0.0-pr474-run407-1-20260813135601",
+    "@eddeee888/gcg-typescript-resolver-files": "0.18.3",
     "@graphql-codegen/add": "7.1.0",
     "@graphql-codegen/cli": "7.3.1",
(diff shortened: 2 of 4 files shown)


codegen.mts
```diff
@@ -5,6 +5,9 @@ import { addTypenameSelectionDocumentTransform } from '@graphql-codegen/client-p
 const config: CodegenConfig = {
   schema: './packages/services/api/src/modules/*/module.graphql.ts',
   emitLegacyCommonJSImports: true,
+  overwrite: {
+    removeStaleFiles: false,
+  },
```
eddeee888, line 10: Ideally Server Preset's `defineConfig` should handle this without us having to set this globally like this.

However, the `overwrite` checks are not working for presets due to it using the actual filename and not the `generates` key. I'll fix this next

package.json
```diff
@@ -69,9 +69,9 @@
     "@0no-co/graphqlsp": "1.15.4",
     "@changesets/changelog-github": "0.7.0",
     "@changesets/cli": "2.31.1",
-    "@eddeee888/gcg-typescript-resolver-files": "0.18.0",
+    "@eddeee888/gcg-typescript-resolver-files": "0.0.0-pr474-run407-1-20260813135601",
```
eddeee888: Will be replaced with real version once [this is merged](https://github.com/eddeee888/graphql-code-generator-plugins/pull/479)

eddeee888:
Hi @n1ru4l could you try this new version to see if watch feels faster for you? 🙂

n1ru4l (approved):
Great 🥳

PR #8415, opened 2026-08-29
Bump Server Preset (the one with cache)

eddeee888:
### Description

This PR uses the [Server Preset version](https://github.com/eddeee888/graphql-code-generator-plugins/pull/487) that focuses on caching _schema type vs mappers comparison_ result (expensive calculation)

### Result

| Run type | Before | After | Changes |
| -- | -- | -- | -- |
| Cold run, all generates | 7.5s | 7.5s | N/A |
| Warm run, all generates, cache miss | 5.6s | 5.6s | N/A |
| Warm run, all generates, cache hit | 5.6s (no cache, same as above) | 2.6s | -53% |
| Cold run, only server preset generates | 5.2s | 5.2s | N/A |
| Warm run, only server preset generates, cache miss | 3.1s |3.1s | N/A |
| Warm run, only server preset generates, cache hit | 3.1s (no cache, same as above) | 0.6s | -80% |

This is useful in scenarios where schema and mapper files do not change:
- when we are working on the client side
- when we are writing resolver logic e.g. adding/removing resolver function. In those cases, the resolver codemod should happen quicker


package.json
```diff
@@ -69,7 +69,7 @@
     "@0no-co/graphqlsp": "1.15.4",
     "@changesets/changelog-github": "0.7.0",
     "@changesets/cli": "2.31.1",
-    "@eddeee888/gcg-typescript-resolver-files": "0.18.4",
+    "@eddeee888/gcg-typescript-resolver-files": "0.0.0-pr487-run427-1-20260829091237",
```
eddeee888: TODO: release stable version once we are happy with this preview

eddeee888:
Hi @n1ru4l , could you please try this version?
There'd be a few patches I need to do in Codegen and the preset for some edge cases, but this current alpha should have the main watch perf improvement

codegen.mts
```diff
@@ -5,9 +5,6 @@ import { addTypenameSelectionDocumentTransform } from '@graphql-codegen/client-p
 const config: CodegenConfig = {
   schema: './packages/services/api/src/modules/*/module.graphql.ts',
   emitLegacyCommonJSImports: true,
-  overwrite: {
-    removeStaleFiles: false,
-  },
```
eddeee888, line 10: Codegen now supports this being set on `generates` block, and Server Preset is doing that with `defineConfig`. So, this can go

n1ru4l (approved):
nice 🥳

PR #8420, opened 2026-08-31
feat: expose schema check, publish and promotion sources as prometheus metrics

n1ru4l:
Exposes total amount and error source for critical schema registry operations:
- schema publish
- schema check
- schema promotion
- app deployment create
- app deployment document upload
- app deployment publish

This will allow us overseeing the total amount of times each operation is performed and the source on why operation fails.

It can be used as a source for alerts so we are informed when object storage (r2/s3), clickhouse, pg, clickhouse or an internal service is causing issues.

### DIFF
--- a/packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
+++ b/packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
@@ -410,7 +410,7 @@ export class PersistedDocumentIngester {
 
             if (response.statusCode !== 200) {
               throw new Error(
-                `Failed to upload operation to S3: [${response.statusCode}] ${response.statusMessage}`,
+                `Failed to upload operation to S3 object storage (${s3.endpoint}/${s3.bucket}): [${response.statusCode}] ${response.statusMessage}`,
               );
             }
           }
--- a/packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
+++ b/packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
@@ -14,7 +14,7 @@ import PromiseQueue from 'p-queue';
 import { z } from 'zod';
 import { collectSchemaCoordinates, preprocessOperation } from '@graphql-hive/core';
 import { buildOperationS3BucketKey } from '@hive/cdn-script/artifact-storage-reader';
-import { ServiceLogger } from '@hive/service-common';
+import { ServiceLogger, setErrorSource } from '@hive/service-common';
 import { sql as c_sql, ClickHouse } from '../../operations/providers/clickhouse-client';
 import { S3Config } from '../../shared/providers/s3-config';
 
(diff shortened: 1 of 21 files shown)


packages/services/api/src/modules/cdn/providers/aws.ts
```diff
@@ -199,9 +200,10 @@ export class AwsClient {
     console.log(`Calling ${init.method} ${url}`);
     const startedAt = Date.now();
 
-    const signed = await this.sign(url, init);
+    const errorSource = `s3 ${new URL(url).origin}` as const;
```
n1ru4l, line 203: attaching origin, so we can easily distinguish if the errors source is r2 (very likely) or s3 (very unlikely).

packages/services/api/src/modules/schema/providers/schema-publisher.ts
```diff
@@ -1250,15 +1292,21 @@ export class SchemaPublisher {
...
-    });
+    return await this.internalCheck(input).then(
+      result => {
+        schemaCheckOutcomeCount.inc({ conclusion: 'success' });
+        return result;
+      },
+      error => {
+        if (error instanceof HiveError) {
+          schemaCheckOutcomeCount.inc({ conclusion: 'success' });
+        } else {
```
n1ru4l, line 1254: Error that is instance of HiveError is technically an expected error hence why I track them as a success. I want to `schemaCheckOutcomeCount` metric to total amount of performed operations.

packages/services/service-common/src/errors.ts
```diff
@@ -1,6 +1,41 @@
...
+  return sourcedError as ErrorWithSource;
+}
+
+export function getErrorSource(error: unknown): ErrorSource | null {
+  if (error instanceof Error && errorSourceProperty in error) {
+    return String(error[errorSourceProperty]);
+  }
+
+  return null;
+}
```
n1ru4l, line 37: Instead of having our own errors it felt more easy to introduce the error source this way for the wide range of errors we have 


- clickhouse client uses got
- s3 client uses got
- pg (slonik) uses their own errors and relies on subclassing; we use instanceof checks in order to track down specific errors
- trpc has their own errors

packages/services/api/src/modules/app-deployments/providers/persisted-document-ingester.ts
```diff
@@ -410,7 +410,7 @@ export class PersistedDocumentIngester {
 
             if (response.statusCode !== 200) {
               throw new Error(
-                `Failed to upload operation to S3: [${response.statusCode}] ${response.statusMessage}`,
+                `Failed to upload operation to S3 object storage (${s3.endpoint}/${s3.bucket}): [${response.statusCode}] ${response.statusMessage}`,
```
kamilkisiela: it won't leak to users right?

n1ru4l: hell nah

kamilkisiela (approved):

dotansimha (approved):

PR #8421, opened 2026-08-31
chore(dependencies): update `@theguild/federation-composition` to `v0.26.0`

kamilkisiela:
Support for `@context` and `@fromContext`


n1ru4l (approved):
worth to create a product update as well

PR #8423, opened 2026-09-01
chore: use correct metric for api operations graph

n1ru4l:
missed this distinction from local dev env and staging 👁️ 👁️


kamilkisiela (approved):
LGBT

PR #8427, opened 2026-09-02
Fix cli graphql request error handling edge case and log order

jdolle:
### Background

Noticed when testing the `--github` option locally using `act`.

### Description

If the graphql error response doesn't include the extension, then an unexpected error is thrown. If this error is thrown, then the error response json isnt logged even if the debug logs are enabled.

This PR moves the error debug logs ahead of the checks and adds an optional check to the extensions object in case it isnt returned.

### Checklist

### DIFF
--- a/.changeset/long-ants-worry.md
+++ b/.changeset/long-ants-worry.md
@@ -1,0 +1,7 @@
+---
+'@graphql-hive/cli': patch
+---
+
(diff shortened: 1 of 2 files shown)


n1ru4l (approved):
Please add a changeset before merging

PR #8439, opened 2026-09-04
Check schema validity after removing unaccessible types

jdolle:
### Background

https://github.com/graphql-hive/router/issues/1487

### Description

Our composition also validates the schema, but when we add inaccessible to unreachable types afterwards, it can cause the supergraph to become invalid.

We should determine if a supergraph is valid in the registry so that we don't expose an invalid graph to the gateways.

### Checklist

- [X] Testing

### DIFF
--- a/packages/services/schema/package.json
+++ b/packages/services/schema/package.json
@@ -10,6 +10,7 @@
   },
   "devDependencies": {
     "@apollo/federation": "0.38.1",
+    "@apollo/federation-internals": "2.14.0",
     "@graphql-tools/merge": "9.2.3",
     "@graphql-tools/stitch": "10.2.2",
     "@graphql-tools/stitching-directives": "4.0.27",
--- a/packages/services/schema/src/composition/federation.ts
+++ b/packages/services/schema/src/composition/federation.ts
@@ -6,6 +6,7 @@ import {
   type FieldDefinitionNode,
   type NameNode,
 } from 'graphql';
+import { Supergraph } from '@apollo/federation-internals';
 import type { ServiceLogger } from '@hive/service-common';
 import {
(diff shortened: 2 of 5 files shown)


packages/services/schema/src/composition/federation.ts
```diff
@@ -272,6 +273,33 @@ export const createComposeFederation = (deps: ComposeFederationDeps) =>
             publicDocumentNode,
           });
 
+          // Marking unreachable types as @inaccessible can produce an invalid supergraph.
+          // This verifies that the supergraph generated is valid.
+          try {
```
dotansimha, line 278: I think the invalid supergraph is a bad artifact we produced? sure, we can return an error, but isn't that a bug that we would need to fix first in composition/contracts?

dotansimha (changes requested):

packages/services/schema/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "devDependencies": {
     "@apollo/federation": "0.38.1",
+    "@apollo/federation-internals": "2.14.0",
```
n1ru4l, line 13: We cannot use this in our running service as Elastic-2.0 is restrictive:

You cannot offer the software itself as a managed/hosted service where customers get access to a substantial portion of its functionality.

n1ru4l (changes requested):
We cannot merge this for the following reasons: https://github.com/graphql-hive/console/pull/8439/changes#r3955582610

Instead we should fix the composition issue within https://github.com/graphql-hive/federation-composition (as @dotansimha already pointed out).

n1ru4l:
I created https://github.com/graphql-hive/federation-composition/pull/329 to resolve the root cause.

PR #8443, opened 2026-09-08
fix: schema promotion raising an exception due to invariants for legacy single schema version logs

n1ru4l:
Removes invalid schema version invariants for legacy single-schema log records.

These records can exist during schema promotion due to legacy backwards compatibility. The assertions caused legit promotions to fail.

These assertions were missed in a previous fix: https://github.com/graphql-hive/console/pull/8376.

Added an integration test covering this scenario.

### DIFF
--- a/.changeset/sixty-actors-post.md
+++ b/.changeset/sixty-actors-post.md
@@ -2,4 +2,4 @@
 'hive': patch
 ---
 
-Fix schema promotion raising an unexpected server error for legacy schema version logs
+Fix schema promotion raising an unexpected server error for legacy schema version logs.
--- a/.changeset/sixty-actors-post.md
+++ b/.changeset/sixty-actors-post.md
@@ -1,0 +1,5 @@
+---
+'hive': patch
+---
+
+Fix schema promotion raising an unexpected server error for legacy schema version logs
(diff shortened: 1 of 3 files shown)


packages/services/api/src/modules/schema/providers/schema-version-store.ts
```diff
@@ -1146,17 +1145,15 @@ export class SchemaVersionStore {
...
+        schemaVersion.actionId,
+        `The schema version '${schemaVersion.id}' without actionId should not have an edge without a type.`,
+      );
 
       // if the actionId does not match the node, we have a unchanged edge
       if (schemaVersion.actionId !== node.id) {
-        if (node.action === 'DELETE') {
-          throw new Error(`Invariant: The action can not be delete in this scenario.`);
-        }
+        invariant(node.action != 'DELETE', ` The action can not be delete in this scenario.`);
```
n1ru4l:
```suggestion
        invariant(node.action !== 'DELETE', ` The action can not be delete in this scenario.`);
```

.changeset/sixty-actors-post.md
```diff
@@ -0,0 +1,5 @@
+---
+'hive': patch
+---
+
+Fix schema promotion raising an unexpected server error for legacy schema version logs
```
n1ru4l:
```suggestion
Fix schema promotion raising an unexpected server error for legacy schema version logs.
```

dotansimha (approved):

PR #8444, opened 2026-09-08
measure s3 storage write latency and success/failure rate

n1ru4l:
Improve the observability story of the external S3 compatible storages (Cloudflare R2 and AWS S3).

Summary:
- Adds metrics for write latency and HTTP status codes, grouped by destination (R2 or S3) and operation (e.g. schema sdl, app deployment, etc)
- Adds Cloudflare Ray IDs to traces for failed R2 writes
- Centralizes object storage writes in a shared S3Writer class
- Propagates worker-thread metrics and R2 error details to the API process (so they can be scraped from the metrics endpoint)
- Adds a Grafana dashboard for monitoring write latency, volume, and failures

___

We now have some information about the r2 ray ids and status on the traces when writes to r2 fail (I enforced this failure by setting the wrong access key in a dev deployment):

image

Comes with a new dashboard for displaying the new metrics - this should help us make follow-up decissions.

image

### DIFF
--- a/deployment/grafana-dashboards/S3-R2-Write-Overview.json
+++ b/deployment/grafana-dashboards/S3-R2-Write-Overview.json
@@ -656,7 +656,7 @@
           "description": "",
           "id": 4,
           "links": [],
-          "title": "Overall write error rate (401s, all mirrors)",
+          "title": "Overall write error rate (all mirrors)",
           "vizConfig": {
             "group": "stat",
             "kind": "VizConfig",
@@ -1093,7 +1093,7 @@
           "description": "",
           "id": 9,
           "links": [],
-          "title": "Overall write error rate trend (401s, all mirrors)",
+          "title": "Overall write error rate trend (all mirrors)",
           "vizConfig": {
             "group": "timeseries",
             "kind": "VizConfig",
--- a/deployment/grafana-dashboards/S3-R2-Write-Overview.json
+++ b/deployment/grafana-dashboards/S3-R2-Write-Overview.json
@@ -37,14 +37,15 @@
                   "hidden": false,
                   "query": {
                     "datasource": {
-                      "name": "grafanacloud-prom"
+                      "name": "TEMPO_DATASOURCE_UID"
                     },
-                    "group": "prometheus",
+                    "group": "tempo",
                     "kind": "DataQuery",
                     "spec": {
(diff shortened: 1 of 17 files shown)


deployment/grafana-dashboards/S3-R2-Write-Overview.json
```diff
@@ -0,0 +1,1401 @@
+{
+  "apiVersion": "dashboard.grafana.app/v2",
```
n1ru4l, line 2: I used the new format for this one

packages/services/api/src/modules/cdn/providers/aws.ts
```diff
@@ -266,7 +266,7 @@ export class AwsClient {
       region: init?.aws?.region || this.region,
       cache: init?.aws?.cache || this.cache,
       datetime: init?.aws?.datetime,
-      signQuery: init?.aws?.signQuery,
+      signQuery: init?.aws?.signQuery ?? true,
```
n1ru4l, line 269: We pass `aws: { signQuery: true }` everywhere. It makes more sense to just set this as the default here.

packages/services/api/src/modules/cdn/providers/s3-iam.spec.ts
n1ru4l, line 1: The tests here started failing due to introducing `signQuery: true` as the default value.

The tests were introduced as part of https://github.com/graphql-hive/console/commit/2b6d22f5d36c46e9bdc2425451fc96c5d42a0dbb

However, they test for functionality that we don't use. We could probably even remove most of the test-cases here. I did not want to introduce too much noise in this PR, so I just made the tests pass again.

packages/services/api/src/modules/shared/providers/s3-config.ts
```diff
@@ -1,10 +1,18 @@
...
+};
 
-export const S3_CONFIG = new InjectionToken<S3Config>('S3_CONFIG');
+/**
+ * S3 bucket storage configurations for dual writes to a primary and secondary destination.
+ */
+@Injectable({ scope: Scope.Singleton, global: true })
+export class S3Config {
+  constructor(public destinations: AtLeastOneReadonlyArray<S3Destination>) {}
+}
```
n1ru4l, line 18: A class is nicer than an `InjectionToken` as it does not require an additional `@Inject(S3_CONFIG)` annotation. Less likely to cause miss-typings that way.

dotansimha (approved):

PR #8450, opened 2026-09-09
Fix latent bug in lab build scalar field

jonathanawesome:
This PR makes union and interface fields expandable in the Laboratory query builder, closing a customer report that a `content: [CmsContent!]!` field could not be expanded and returned a 400 when run.

Each abstract field now lists its possible types as a `... on Type` row with selectable fields beneath, and selecting an abstract field also selects `__typename`, so the builder can no longer write a field with no selection set. Builder search and the docs context menu now work below abstract types as well.

There's also a change to Monaco Editor that disables `stickyScroll`, which was never properly implemented and creating overlays/ghosting in the editors.

One behaviour change worth knowing: in a document containing a hand-written inline fragment, a field selected inside it now checks the row under its own type rather than the parent's, so two members declaring the same field name are separate selections. Named fragment spreads are still not managed by the builder and are left untouched.

### DIFF
--- a/.changeset/rare-donkeys-clap.md
+++ b/.changeset/rare-donkeys-clap.md
@@ -1,0 +1,5 @@
+---
+'@graphql-hive/laboratory': minor
+---
+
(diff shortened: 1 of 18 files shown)


n1ru4l (approved):
💅 nice

PR #8462, opened 2026-09-10
Consolidate menu components

jonathanawesome:
This PR replaces `@/components/ui/dropdown-menu` with `base/floating/menu`, reworking that component to a data-driven API in the process, and adds a `base/floating/tooltip` built on Base UI. It removes the last menu implementation from the app along with the `@radix-ui/react-dropdown-menu` dependency: 36 files, +1745/-1293.

**Stacked on #8345** (card consolidation).

### Menus are now described, not composed

`ui/dropdown-menu` was a Radix wrapper used by 12 files (12 menus, 95 elements). `base/menu` already existed and wrapped `@base-ui/react`, but composed its rows from JSX in a `sections` array. Both are gone in favour of one component whose `sections` prop takes data:

```tsx
<Menu
  trigger={<Button variant="ghost"><EllipsisIcon /></Button>}
  align="end"
  width="sm"
  sections={[
    { label: 'Options', items: [{ label: 'View details', onClick: view }] },
    [{ label: 'Delete', icon: Trash2, variant: 'destructiveAction', onClick: remove }],
  ]}
/>
```

There is no `MenuItem` to assemble by hand and no key to thread through. Each array in sections is a group and the component rules between them, so `DropdownMenuSeparator` becomes structure rather than an element. Falsy entries are dropped and an emptied section takes its separator with it, so a conditional row is just `cond && { label: … }`.

Seven entry kinds, each with its own preview so they are discoverable: action (the default, no kind), link, submenu, checkbox, toggle, radio and header.

### New: base/tooltip
Built on Base UI, taking trigger and content with delay, disabled, arrow and the usual side/align props. Deliberately smaller than the other floating panels at `text-xs` and `max-w-64`, since a tooltip explains the surface it sits over and previously rendered larger than the menu it was attached to. Adopted in `base/menu` and `base/stat-card`; the 105 `ui/tooltip` call sites elsewhere are untouched and will be migrated in their own PR.

### Fixes found along the way
- Submenus grew horizontal scrollbars. Once overflow-y is not visible, the spec computes overflow-x to auto, so the vertical scrollbar's own width was enough to bring one on. Now explicitly overflow-x-hidden.
- Long menus ran off screen rather than scrolling. Capped to the positioner's --available-height.
- The date range picker's quick-ranges list had an unstyled scrollbar; now uses thin-scrollbar.

### Smaller changes worth knowing
- `addOperation` in the laboratory collections plugin takes a `collectionId: string` instead of reading `e.currentTarget.dataset.collectionId` off the event. Both call sites already had the `id` in scope; the DOM round-trip existed only because the old API had nowhere else to put it.
- `FloatingProps.trigger` also accepts a render function, for the one case where another wrapper needs to own the same element. `GraphiQLTooltip` destructures only `{ children, align, side, sideOffset, label }` and forwards nothing, so passing it as the trigger would swallow the menu's props.


packages/web/app/src/components/base/floating/menu/menu.tsx
```diff
@@ -1,62 +1,461 @@
...
+ * A row that navigates to an href. For a typed app route use a `render` on an action row with a
+ * TanStack `Link`: its `to` constrains `params` and `search` through inference at the call site,
+ * which a data field cannot carry.
+ */
+type MenuLink = {
+  kind: 'link';
+  label: ReactNode;
+  href: string;
+  /** Opens in a new tab, with `rel="noreferrer"`. */
+  external?: boolean;
```
jdolle, line 142: could make this automatic based on the href, but having this be configurable is also not a big deal and is potentially safer. You can safely ignore this comment.

packages/web/app/src/pages/target-laboratory.tsx
```diff
@@ -167,85 +160,85 @@ function Save(props: {
...
+                  return;
+                }
+                toggleOperationModal();
+              },
+              attrs: { 'data-cy': 'save-operation-as' },
+            },
+          ],
+        ]}
+      />
+      {/* No longer a child of the menu: it is a modal, not a row. */}
```
jdolle:
```suggestion
```
no need to capture historic changes in comments.

jdolle (approved):

PR #8463, opened 2026-09-10
Add a Hive Gateway compatible schema fetcher function based on hive:dev

jdolle:
### Background

Currently, to automatically run a local gateway to test a local schema requires:
1. starting the subgraph in a way that exports the schema
2. running hive:dev to watch this exported file and compose it into a supergraph file
3. start hive gateway pointed at a local supergraph file

This requires knowledge about where all of these files exist and hive:dev feels like an unnecessary extra process that is slowing things down.

### Description

This change creates a `@graphql-hive/gateway` compatible schema fetcher that is based on `hive:dev`. You can then configure the local subgraphs in the gateway and use an environment variable or another conditional to enabled this schema fetcher.

### Checklist

- [X] Input validation
- [X] Testing

### DIFF
--- a/packages/libraries/cli/__tests__/dev.spec.ts
+++ b/packages/libraries/cli/__tests__/dev.spec.ts
@@ -1,9 +1,9 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 import {
-  InvalidRemoteCompositionResultError,
-  RegistryApiError,
-  RemoteCompositionError as CoreRemoteCompositionError,
-  SupergraphCompositionError,
+  InvalidSupergraphResultError,
+  LocalSupergraphCompositionError,
+  RemoteSupergraphCompositionError,
+  SupergraphRegistryApiError,
 } from '@graphql-hive/core';
 import {
   APIError,
@@ -41,10 +41,10 @@ beforeEach(() => {
 });
 
 describe('Dev.composeLocally', () => {
-  it('maps a SupergraphCompositionError to a LocalCompositionError', async () => {
+  it('maps a LocalSupergraphCompositionError to a LocalCompositionError', async () => {
     const compositionResult = { errors: [{ message: 'field conflict' }] } as any;
     mockComposeSupergraphLocally.mockRejectedValue(
-      new SupergraphCompositionError(compositionResult),
+      new LocalSupergraphCompositionError(compositionResult),
     );
 
     const dev = createDevInstance();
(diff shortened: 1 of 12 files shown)


packages/libraries/cli/src/dev-fetcher.ts
```diff
@@ -0,0 +1,115 @@
...
+  let target: GraphQLSchema.TargetReferenceInput | null = null;
+  if (options.target) {
+    const result = TargetInput.parse(options.target);
+    if (result.type === 'error') {
+      throw new InvalidTargetError();
+    }
+    target = result.data;
+  }
+
+  const config: UnifiedGraphConfig = async ({ log, fetch: gatewayFetch, cwd, cache }) => {
```
jdolle: The cache is a `KeyValueCache<any> | undefined` but I'm not sure when this cache is or isnt provided. Any clarity on this is appreciated.

n1ru4l (changes requested):
I don't think this belongs into the CLI package. Why not put this directly into the gateway or the gateway sdk package?

packages/libraries/cli/__tests__/dev.spec.ts
jdolle: Had tests from prior implementation so I opted to keep.

packages/libraries/core/src/client/dev-fetcher.ts
```diff
@@ -0,0 +1,410 @@
+import { readFile } from 'node:fs/promises';
+import { resolve as resolvePath } from 'node:path';
```
n1ru4l: This will potentially break non node.js runtime compatibility. Worth checking - I would recommend placing this utility in the gateway package in that case.

jdolle: It absolutely did break this.

I addressed the issue by forcing a dynamic important adding a safety check to verify if running in a node environment.

I prefer the fetcher living here since it shares code with hive dev and we have precedent for defining loaders in separate repos. We already expose the CDN fetcher from this repo.

packages/libraries/core/package.json
```diff
@@ -49,6 +49,7 @@
     "@graphql-hive/logger": "^1.1.0",
     "@graphql-hive/signal": "^2.0.0",
     "@graphql-tools/utils": "^12.0.0",
+    "@theguild/federation-composition": "^0.26.1",
```
n1ru4l, line 52: I am not sure whether we want to make this a dependency of the Hive SDK. Especially since it is only used by one API that is used by the gateway and SLI - but is completely irrelevant for the actual runtime SDKs for usage reporting etc.

jdolle:
moving to gateway repo

PR #8467, opened 2026-09-11
Fix Filter submenu sizing

jonathanawesome:
This PR primarily fixes an issue in our Menu/FilterMenu component where a submenu that is awaiting data computes width before the data arrives, causing a possible mismatch to the proper width.

- removes `stableWidth` from `Menu`
- restores submenu chevrons
- added close-submenus-on-scroll so the submenu doesn't scroll along
- adds foundry previews mimicking console usage of FilterMenu with simulated loading
- locks body scroll when menus are open


jdolle (approved):
There's a minor visual issue where the top element in a menu has a top margin but the bottom element doesnt have a bottom margin (see screenshot). Otherwise looks good

Image

PR #8469, opened 2026-09-11
feat: add `Target.schemaPublishCount` to the public GraphQL API schema

n1ru4l:
### Background

[Internal feature request](https://guild-oss.slack.com/archives/C077QTGB362/p1788183565689489).

### Description

This adds the ability to fetch the amount a (service) schema was published to a target via the public GraphQL API.

#### Time Range and Origin column

The implementation leverages the `origin` column, in order to be efficient and avoid any joins, which we shipped on Jun 1 2026.

Since we never intended to backfill this column, the `Target.schemaPublishCount` can only provide "accurate" results for time ranges after Jun 1 2026.

In order to also support records before that date, we have a slower implementation that joins `schema_versions` with `schema_log` for queries covering a time range before that date.

#### Supergraph Promotions

Subgraphs that are introduced/changed in a promotion are not tracked or counted. The workflow is that schemas are published in lower environments and then promoted to a higher environment. There should be no ad-hoc publishes to higher environments. Thus, users should only count the schema publishes on the lower environment.

#### Example Query Plan (based on `origin`)

Quite fast 😎 

```
GroupAggregate  (cost=0.43..5.85 rows=1 width=24) (actual time=0.940..0.942 rows=1 loops=1)
  Output: schema_versions.target_id, count(*)
  Buffers: shared hit=484
  ->  Index Scan using schema_versions_cursor_pagination on public.schema_versions  (cost=0.43..5.83 rows=1 width=16) (actual time=0.062..0.918 rows=44 loops=1)
        Output: schema_versions.id, schema_versions.created_at, schema_versions.is_composable, schema_versions.base_schema, schema_versions.target_id, schema_versions.action_id, schema_versions.has_persisted_schema_changes, schema_versions.schema_composition_errors, schema_versions.composite_schema_sdl, schema_versions.previous_schema_version_id, schema_versions.supergraph_sdl, schema_versions.github_repository, schema_versions.github_sha, schema_versions.diff_schema_version_id, schema_versions.record_version, schema_versions.tags, schema_versions.has_contract_composition_errors, schema_versions.conditional_breaking_change_metadata, schema_versions.schema_metadata, schema_versions.metadata_attributes, schema_versions.origin, schema_versions.supergraph_changes, schema_versions.meta
        Index Cond: ((schema_versions.target_id = 'redacted'::uuid) AND (schema_versions.created_at >= (now() - '360 days'::interval)))
        Filter: ((schema_versions.origin @> '{"type": "publish"}'::jsonb) AND (SubPlan 1))
        Rows Removed by Filter: 448
        Buffers: shared hit=484
        SubPlan 1
          ->  Function Scan on pg_catalog.jsonb_array_elements service  (cost=0.01..1.38 rows=1 width=0) (actual time=0.001..0.001 rows=0 loops=147)
                Function Call: jsonb_array_elements((schema_versions.origin -> 'services'::text))
                Filter: ((service.value ->> 'name'::text) = ANY ('{redacted}'::text[]))
                Rows Removed by Filter: 1
Settings: effective_cache_size = '6GB', jit = 'off', random_page_cost = '2'
Query Identifier: 7096207027361398681
Planning:
  Buffers: shared hit=59
Planning Time: 0.288 ms
Execution Time: 0.984 ms
```

### Example Query Plan (based on table join without origin)

```
Aggregate  (cost=8.91..8.92 rows=1 width=8) (actual time=5.819..5.821 rows=1 loops=1)
  Output: count(*)
  Buffers: shared hit=2447
  ->  Nested Loop  (cost=0.86..8.91 rows=1 width=0) (actual time=0.098..5.800 rows=80 loops=1)
        Inner Unique: true
        Buffers: shared hit=2447
        ->  Index Scan using schema_versions_cursor_pagination on public.schema_versions sv  (cost=0.43..4.45 rows=1 width=16) (actual time=0.023..0.788 rows=491 loops=1)
              Output: sv.id, sv.created_at, sv.is_composable, sv.base_schema, sv.target_id, sv.action_id, sv.has_persisted_schema_changes, sv.schema_composition_errors, sv.composite_schema_sdl, sv.previous_schema_version_id, sv.supergraph_sdl, sv.github_repository, sv.github_sha, sv.diff_schema_version_id, sv.record_version, sv.tags, sv.has_contract_composition_errors, sv.conditional_breaking_change_metadata, sv.schema_metadata, sv.metadata_attributes, sv.origin, sv.supergraph_changes, sv.meta
              Index Cond: ((sv.target_id = 'redacted'::uuid) AND (sv.created_at >= (now() - '360 days'::interval)))
              Buffers: shared hit=483
        ->  Index Scan using commits_pkey on public.schema_log sl  (cost=0.43..4.45 rows=1 width=16) (actual time=0.010..0.010 rows=0 loops=491)
              Output: sl.id
              Index Cond: (sl.id = sv.action_id)
              Filter: ((sl.action = 'PUSH'::text) AND (lower(sl.service_name) = ANY ('{redacted}'::text[])))
              Rows Removed by Filter: 1
              Buffers: shared hit=1964
Settings: effective_cache_size = '6GB', jit = 'off', random_page_cost = '2'
Query Identifier: -6303641494650499996
Planning:
  Buffers: shared hit=18
Planning Time: 0.422 ms
Execution Time: 5.879 ms
```


packages/services/api/src/modules/schema/providers/schema-version-store.ts
```diff
@@ -343,7 +343,7 @@ export class SchemaVersionStore {
           services: args.service
             ? [
                 {
-                  name: args.service.name,
+                  name: args.service.name.toLowerCase(),
```
jdolle, line 346: good catch. We've had issues before with these service names being inconsistent.

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -153,6 +153,17 @@ export default gql`
   }
 
   extend type Target {
+    """
+    Number of schema publishes within a time period, optionally filtered by subgraph names.
+    """
+    schemaPublishCount(
+      period: DateRangeInput! @tag(name: "public")
```
jdolle, line 160: Should the period be optional with the default returning for all time?

jdolle (changes requested):
One use case that could be improved is if we want to collect the publish count for every subgraph in our target.
Currently, the query would need to alias multiple times and we'd have to know the subgraphs ahead of time:

```graphql
{
  target(reference:  $ref) {
    fooCount: schemaPublishCount(
      period: $period
      subgraphNames: ["foo"]
    )
    barCount: schemaPublishCount(
      period: $period
      subgraphNames: ["bar"]
    )
  }
}
```

Another concern is that there's some overlap conceptually with the existing `Target.schemaVersionsCount` field. Although this is obviously intended to provide a more granular count for federated graphs, I'm not sure that the types reflect that distinction in a clear way.

I dont have a good solution for either of these points

packages/services/api/src/modules/schema/module.graphql.ts
n1ru4l: I thought about this, but I do not want to invite people to force a full table traversal.

packages/services/api/src/modules/schema/providers/schema-version-store.ts
n1ru4l: yeah, it is too easy to go unnoticed... 😭 Fortunately most people use lowercase

n1ru4l:
> One use case that could be improved is if we want to collect the publish count for every subgraph in our target.
We can solve that later one if we want to display it on our dashboard

> Another concern is that there's some overlap conceptually with the existing `Target.schemaVersionsCount` [...]

`Target.schemaVersionsCount` -> amount of schema versions
`Target.schemaPublishCount` ->  amount of schema publishes

They kind of answer different things right now, even though they are kind of similar...

Issue #8404, opened 2026-08-25
subgraph schema push/registration


n1ru4l:
## Goal

Allow users to check or push a named subgraph version once and later publish that exact version without locating and uploading the SDL again.

The intended workflow is:

```bash
# Check the SDL and make products@$GIT_SHA available for publication.
hive schema:push schema.graphql \
  --service "products@$GIT_SHA" \
  --project my-org/my-project

# Publish the exact checked SDL later without passing the file again.
hive schema:publish \
  --service "products@$GIT_SHA" \
  --target my-org/my-project/production
```

This separates producing a subgraph schema from publishing it into a target. The registry guarantees that a version always identifies the same SDL, while the normal publish pipeline still performs composition, policy, contract, and artifact processing against the destination target's current state.

### Push subgraph without checking/publishing

Users can explicitly push/register SDL to the schema registry:

```bash
hive schema:push schema.graphql \
  --service products@1.4.0 \
  --project my-org/my-project
```

Registration stores the SDL for later use but does not change any target's published schema.

### Publish a version

Users publish a registered version by omitting the schema file and including the version in the
service reference:

```bash
hive schema:publish \
  --service products@1.4.0 \
  --target my-org/my-project/production
```

The existing file-based publish remains supported:

```bash
hive schema:publish schema.graphql \
  --service products \
  --url https://products.example.com/graphql
```

Supplying both a schema file and `service@version` is ambiguous and rejected.

### Handle an existing version

Versions are immutable from the user's perspective:

- Registering the same service, version, and SDL again succeeds idempotently.
- Registering the same service and version with different SDL fails with a conflict.
- Correcting a version normally requires choosing a new version name.
- Mutable names such as `candidate` or `latest` are out of scope of this task

An example conflict is:

```text
Version products@1.4.0 already exists with a different schema.
Existing digest:  hive-sdl-v1:sha256:abc...
Submitted digest: hive-sdl-v1:sha256:def...
```

## Design Guarantees

- The registry assigns an immutable digest based on canonical SDL.
- A user-provided version is an immutable alias for one digest.
- Versions can be reused across targets in the same project.
- Registration does not modify published target state.
- Publishing by reference executes the existing publish pipeline.
- Published history records the resolved SDL revision and service URL.

PR #8490, opened 2026-09-15
feat: `hive schema:push` CLI command and API

n1ru4l:
### Background

https://github.com/graphql-hive/console/issues/8404; docs: https://github.com/the-guild-org/website/pull/2007

### Checklist

- [ ] Input validation
- [ ] Output encoding
- [ ] Session management
- [ ] Access control
- [ ] Cryptographic practices
- [ ] Data protection
- [ ] Communication security
- [ ] System configuration
- [ ] Database security
- [ ] File management
- [ ] Memory management
- [ ] Testing


.changeset/tidy-schema-revisions.md
```diff
@@ -0,0 +1,34 @@
...
+hive schema:push schema.graphql \
+  --target my-org/my-project/my-target \
+  --version "$VERSION"
+
+hive schema:publish \
+  --target my-org/my-project/my-target \
+  --version "$VERSION"
+```
+
+For a federated schema, provide the service when pushing and the service URL when publishing:
```
jdolle: the url should be optional for existing subgraphs like our other commands. Potentially worth clarifying.

integration-tests/tests/api/schema/push.spec.ts
```diff
@@ -0,0 +1,404 @@
+import 'reflect-metadata';
```
jdolle, line 1: why `reflect-metadata` here?

integration-tests/tests/api/schema/push.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+test.concurrent(
+  'pushes and publishes a federation revision, and rejects an unknown version',
+  async ({ expect }) => {
+    const { createOrg } = await initSeed().createOwner();
+    const { createProject } = await createOrg();
+    const { target, createTargetAccessToken } = await createProject(ProjectType.Federation);
+    const token = await createTargetAccessToken({ mode: 'readWrite' });
+    const targetReference = { byId: target.id } as const;
+
+    const push = await execute({
```
jdolle: minor: i dont think the initial push is required for this test. Trying to publish a version that doesnt exist is enough.

integration-tests/tests/api/schema/push.spec.ts
```diff
@@ -0,0 +1,404 @@
...
+  expect(conflict.schemaPush.ok).toBeNull();
+  expect(conflict.schemaPush.error?.message).toContain(
+    "Version 'products@v1' already exists with a different schema.",
+  );
+  expect(conflict.schemaPush.error?.message).toContain('Existing digest: hive-sdl-v1:sha256:');
+  expect(conflict.schemaPush.error?.message).toContain('Submitted digest: hive-sdl-v1:sha256:');
+});
+
+test.concurrent(
+  'queries the revision from a published monolith schema version',
```
jdolle, line 251: minor: I like that we use `each` in other places to run essentially the same test for different project types. Could do that here too.

packages/services/api/src/modules/auth/lib/target-access-token-strategy.ts
```diff
@@ -189,7 +189,7 @@ function transformAccessTokenLegacyScopes(args: {
               'appDeployment:retire',
               'schemaVersion:publish',
               'schemaVersion:deleteService',
-              'schemaVersion:publish',
```
jdolle, line 192: should this be additive rather than replacing?

packages/services/api/src/modules/schema/module.graphql.ts
```diff
@@ -372,7 +375,8 @@ export default gql`
     target: TargetReferenceInput
     service: ID
     url: String
-    sdl: String!
+    sdl: String
```
jdolle: should this be deprecated in favor of using `SchemaPublishSchemaInput.bySdl`?

Also i think `bySdl` is a bit odd. Just `sdl` makes more sense.

jdolle (approved):

packages/services/api/src/modules/auth/lib/target-access-token-strategy.ts
n1ru4l: look two lines up 👯

integration-tests/tests/api/schema/push.spec.ts
n1ru4l: I used some decorate classes directly in the tests before; missed removing them 😓

integration-tests/tests/api/schema/push.spec.ts
n1ru4l: There is a few different assertions per project type and inputs, for now I prefer to keep it like that to not add branches in the test cases

PR #8494, opened 2026-09-15
Fix usage buffer to redistribute reports before splitting internals, and stop silently dropping oversized reports

jdolle:
## Summary
- When a flush was too big, the buffer immediately split every report's internals in one shot and gave up permanently on any further overflow - so a batch of many reports would get its internals shredded even when simply redistributing the existing reports into smaller batches would have fixed it, and a report that still didn't fit was silently dropped with no visibility.
- The buffer now first tries redistributing existing reports into smaller/more batches, and only splits a report's own internals once it's alone in a batch and still too big - repeating until each piece fits or is irreducible (`isSplittable`), at which point it's dropped with a logged error, a Sentry event, and a dedicated metric instead of vanishing silently.
- Splitting a report's internals also no longer silently drops its `subscriptionOperations`, `errors`, or `appDeploymentUsageTimestamps`.

**Stack**: based on #8493. Followed by #8495 and #8496.


packages/services/usage/__tests__/usage.spec.ts
```diff
@@ -79,20 +103,129 @@ test('should split report based on operation map length', () => {
...
-  expect(reports[2].size).toEqual(2);
   expect(reports[2].operations[0].operationMapKey).toEqual('op3');
   expect(reports[2].operations[1].operationMapKey).toEqual('op3');
+  expect(reports[2].subscriptionOperations).toEqual([subscriptionOp3]);
+  expect(reports[2].errors).toBeUndefined();
+  expect(reports[2].size).toEqual(3);
+
+  // appDeploymentUsageTimestamps isn't keyed by operationMapKey, so it can't be
+  // partitioned - it's attached to exactly one (the last, guaranteed non-empty)
+  // chunk instead of duplicated into every chunk.
```
jdolle, line 132: the timestamp is used just to update the app deployment last used, so it's fine that it's only added to one of the reports.

packages/services/usage/src/usage.ts
```diff
@@ -91,9 +99,39 @@ export function splitReport(report: RawReport, numOfChunks: number) {
     reports[chunkIndex].size += 1;
   }
 
+  if (report.subscriptionOperations) {
+    for (const subscriptionOp of report.subscriptionOperations) {
+      const chunkIndex = keyReportIndexMap[subscriptionOp.operationMapKey];
+      const chunkReport = reports[chunkIndex];
+      (chunkReport.subscriptionOperations ??= []).push(subscriptionOp);
```
jdolle, line 106: i guess we just completely left out subscriptions from the split reports before?

packages/services/usage/src/usage.ts
```diff
@@ -206,14 +244,30 @@ export function createUsage(config: {
...
+    isSplittable,
     split(report, numOfChunks) {
-      logger.debug('Splitting report into %s (id=%s)', numOfChunks, report.id);
+      logger.debug('Splitting report into %s chunks (id=%s)', numOfChunks, report.id);
       return splitReport(report, numOfChunks);
     },
-    onRetry(reports) {
-      // Because we do a retry, we need to decrease the number of failures
+    onRetry() {
+      // No-op
```
jdolle: an oversized-payload retry should not increment `rawOperationFailures`

packages/services/usage/src/usage.ts
```diff
@@ -206,14 +244,30 @@ export function createUsage(config: {
...
+        0,
+      );
+      droppedOversizedOperations.inc(numOfOperations);
+      rawOperationFailures.inc(numOfOperations);
+      logger.error(
+        'Dropped %s operations (%s error entries) - report cannot be split any smaller and still exceeds the Kafka size limit',
+        numOfOperations,
+        numOfErrors,
+      );
+      Sentry.captureException(
```
jdolle, line 276: I know we're cautious of adding too many metrics since it's expensive, so as an alternative, we could opt only to send to sentry... But we should be very aware of if/when this happens.

PR #8495, opened 2026-09-15
Split operations below the map-key level instead of dropping them

jdolle:
## Summary
- A report already down to a single operation map key couldn't be split any further and got dropped outright if still oversized - which may happen whenever one query shape is called very often in a single flush interval, since one map key can carry an unbounded number of operations, subscriptionOperations, and errors. (This is unlikely, but possible especially since the flush interval is configurable)
- Once map-key division is exhausted, this now divides those arrays directly instead, duplicating the (single, cheap) map entry into each piece.
- Each error is routed into the same chunk as the operation it came from (matched by `operationMapKey` + `timestamp`, the only link between the two arrays), so a chunk that's later dropped or sent independently never separates an operation from the errors it produced.

**Stack**: based on #8494. Followed by #8496.


packages/services/usage/src/usage.ts
```diff
@@ -128,8 +134,98 @@ export function splitReport(report: RawReport, numOfChunks: number): RawReport[]
...
+  return reports;
+}
+
+export function calculateReportSize(report: RawReport): number {
+  const errorCount = report.errors?.reduce((sum, entry) => sum + entry.errors.length, 0) ?? 0;
+  return (
+    Object.keys(report.map).length +
+    report.operations.length +
+    (report.subscriptionOperations?.length ?? 0) +
+    errorCount
```
jdolle, line 217: now considers more than the operation key count!

PR #8496, opened 2026-09-15
Fix usage service health-status and failure-metric edge cases

jdolle:
## Summary
Three related correctness bugs in the Kafka producer/fallback path:
- The service came back Ready on reconnect even while the fallback queue still held undelivered messages, instead of waiting for it to drain - `shouldBecomeReady()` now gates on both conditions at every call site.
- A successful fallback-queue retry never decremented the failure count that was incremented when the message first failed and entered the queue, permanently over-counting `rawOperationFailures`.
- Kafka's own response parsing already throws on any non-zero error code before `producer.send()` can resolve, so the errorCode branch it was handled in was dead code that would have silently counted operations as failed with no way to recover them; it now throws instead, so if it's ever somehow reached, the operations are retried via the fallback queue.

**Stack**: based on #8495.


packages/services/usage/src/usage.ts
```diff
@@ -234,6 +238,10 @@ export function isSplittable(report: RawReport): boolean {
   );
 }
 
+export function shouldBecomeReady(isUnhealthy: boolean, fallbackQueueSize: number): boolean {
+  return isUnhealthy && fallbackQueueSize === 0;
```
jdolle, line 236: it's important to check the fallback queue size since we want to fully recover before we receive more requests

packages/services/usage/src/usage.ts
```diff
@@ -447,14 +456,15 @@ export function createUsage(config: {
           ],
         });
         rawOperationWrites.inc(numOfOperations);
-      } catch (error) {
-        rawOperationFailures.inc(numOfOperations);
-        throw error;
+        // These operations were counted as failing when they first entered
+        // the fallback queue. But a failed retry attempt isn't a new failure,
+        // so it must undo the increment here to indicate it's still pending.
```
jdolle, line 455: Previously, this would cause an incorrect operation count

jdolle:
Determined that my understanding of how the connect/disconnect work was wrong. Closing as this doesnt solve the problem it intended to solve

PR #8501, opened 2026-09-16
RFC graph manifest

n1ru4l:


docs/specs/graph-manifest.md
```diff
@@ -0,0 +1,318 @@
...
+feature-flag-a
+```
+
+The exact meaning of a graph name is determined by Hive Console. Routers should treat graph names as
+opaque identifiers. In particular, consumers should not infer special behavior from names such as
+`default/contract-a`.
+
+The router could use these names in order to route traffic based on a header or cookie.
+
+### `currentVersion`
```
n1ru4l, line 77: Might be overkill, but I choose `currentVersion` to allow introducing something like `previousVersion` in the future where we could gradually shift traffic over from one graph version to the next and react to errors etc.

dotansimha: That's good to have, yes.

dotansimha (approved):
I see that you already incorporated the feedback from Slack, so LGTM :) 

One question around change detection and the E-Tag:are we going to publish E-Tag also for the mainfest itself?

docs/specs/graph-manifest.md
```diff
@@ -0,0 +1,318 @@
+# Graph Manifest
+
+The Graph Manifest describes the set of GraphQL supergraphs or monolithic schemas that Hive Console
+manages and that should be served via a Hive Router/Gateway (in the following sections referred to
+as router for simplicity) instance.
+
+It is configuration file for a running router instance that is being polled continiously from the
```
jdolle, line 7:
```suggestion
It is configuration for a running router instance that is being polled continuously from the
```
either "a configuration file", just "configuration", or "configuration files"

docs/specs/graph-manifest.md
```diff
@@ -0,0 +1,318 @@
...
+        "artifactPath": "/my-target/versions/89780103-7ea3-490f-99ba-c60e90eed41b"
+      }
+    }
+  }
+}
+```
+
+### `graphs`
+
+graphs contains the complete set of graphs that should currently be served for the target. The
```
jdolle, line 58:
```suggestion
The `graphs` property contains the complete set of graphs, including contracts and variants, that should currently be served for the target. The
```

docs/specs/graph-manifest.md
```diff
@@ -0,0 +1,318 @@
...
+      "level": "debug"
+    },
+    "cache": {
+      "enabled": true
+    }
+  }
+}
+```
+
+An older router that only understands `graphs` should continue operating normally. This allows Hive
```
jdolle, line 316: I appreciate this callout

jdolle (approved):

PR #8504, opened 2026-09-16
Consolidate Table

jonathanawesome:
This PR migrates `ui/table`, `v2/table`, `v2/sortable` and raw `<table>` usages over to our base DataTable, with a new DataTableCell for the cell kinds, and the key/value tables over to DescriptionList.

The three old components are deleted along with the `Scale` helper and the inventory preview.

**Found along the way**
- Contracts, the three access token tables, the members list and alert activity had no loading or empty state (blank sections, a blank page between member pages, an empty-state flash on first paint). They use the table's now.
- The manage page sat inside a horizontal ScrollArea that swallowed wheel scrolling; removed, and its widest columns truncate instead.
- The permission overviews always showed a scrollbar track (`overflow-x-scroll`); it is `overflow-x-auto`.
- The alert form controls sat on a raised surface with base-surface styling; they take `onSurface="raised"`.
- The user menu trigger was a `div` with an onClick; it is a native button.
- `@/env/frontend` throws without `window.__ENV`, so foundry's config now injects the same values the app's index.html does and previews of components that reach it render.


jdolle (approved):
Added some recommendations here: https://github.com/graphql-hive/console/pull/8520

Else this looks good to me.

PR #8505, opened 2026-09-17
Fix trace attributes panel not fully scrollable

### LINKED ISSUE #8499: Hive Tracing attributes not fully scrollable
There are attributes that are overflowing but one cannot scroll to them on latest Chrome on macOS.

https://github.com/user-attachments/assets/d3995932-cb2f-430d-bc25-f00f5a185ba4

Abhist17:
Fixes #8499.

## What was broken

`TraceSheet`'s bottom panel (Attributes / Resource Attributes / Events / GraphQL Operation, in `packages/web/app/src/pages/target-trace.tsx`) wrapped its tab content in `<div className="h-full">` inside `<ScrollArea fill>`. `ScrollArea`'s own `Viewport` already applies `min-h-0 flex-1` + `overflow-y-auto` to size and scroll correctly (`packages/web/app/src/components/base/scroll-area/scroll-area.tsx`); the extra `h-full` on its direct child forced that child to exactly the visible viewport height, so content taller than the panel (a long attributes list) got clipped instead of becoming scrollable — you could see there was more, but couldn't scroll to it.

This is the same class of bug #8503 fixed one level up (the `TraceSheet` outer wrapper's `h-full` → `min-h-0 flex-1`) — same file, one panel down.

## What changed

- Removed the `<div className="h-full">` wrapper around the four tab views (span-attributes, resource-attributes, events, operation), so they're direct `ScrollArea` children.
- Removed the operation tab's now-redundant `<div className="absolute inset-y-0 w-full">` around `GraphQLHighlight`.

Both changes make this panel match the sibling span-level detail sheet in the same file (`SelectedSpanSheet`, ~line 1564), which renders the identical four tabs — including the same `GraphQLHighlight height="100%"` view — as direct `ScrollArea` children with no wrapper, and scrolls correctly today.

## How I verified it

- `tsc --noEmit` on `packages/web/app`: diffed the error list against the pre-fix file — identical set of pre-existing errors (mostly missing generated `@/gql` types from a codegen step this checkout doesn't run), just shifted by the line count change from removing the wrapper. Zero new errors.
- `eslint` on the changed file: same single pre-existing `import/extensions` warning on an unrelated line (44), present before this change too.
- Traced the fix by comparing directly against the already-merged #8503 diff (same anti-pattern, one level up) and the working span-level sheet in the same file, rather than guessing.

No test change: this is a Tailwind/CSS layout fix with no new logic branch to unit test, and the repo's Playwright/e2e suite isn't something I could run headless here.


Abhist17:
Gentle ping on https://github.com/graphql-hive/console/pull/8505 — it's been 3 days. Is there anything you'd like changed before it can be reviewed? Happy to rebase or split it if that helps.

n1ru4l (changes requested):
@Abhist17 can you pleases share a video that shows the issue before and how your fix addresses it afterwards?

PR #8507, opened 2026-09-17
Consolidate overlays

jonathanawesome:
This PR migrates `ui/dialog`, `ui/sheet`, `ui/alert-dialog` and `v2/modal` usages over to three new base components on Base UI: Dialog, Sheet and AlertDialog. Every overlay stays mounted and controls `open`, so all of them get their enter and exit transitions and reset once the close has finished; delete confirmations are AlertDialogs with a destructive confirm; outcome-only states (contract created or disabled, CDN token deleted) are toasts instead of a second dialog page; and every control inside an overlay sits on the raised surface.

The four old components are deleted along with `@radix-ui/react-dialog`, `@radix-ui/react-alert-dialog` and `@headlessui/react`, whose only use, the transfer-ownership member picker, is a searchable base Select now. The overlays inventory preview becomes per-component call-site previews under Base/Overlays.

**Found along the way**
- Updated base control style "default" to 14px
- The CDN token dialog's error state was titled "Delete CDN Access Token" while creating one.
- Labels in the CDN token, create channel, profile settings and transfer ownership forms pointed at ids no input carried, or at the wrong input; they target their fields now.
- The preflight prompt ignored Escape; it counts as Cancel.
- The transfer ownership, support ticket and leave organization outcomes went through react-toastify's dark-only toast in the bottom corner rather than the app toaster.


packages/web/app/src/components/base/overlays/dialog/dialog.spec.tsx
```diff
@@ -0,0 +1,48 @@
+// @vitest-environment jsdom
+import { fireEvent, render, screen } from '@testing-library/react';
+import { Dialog } from './dialog';
+
+describe('Dialog', () => {
+  it('renders the title, description, body and footer when open, on the width rung', () => {
```
jdolle: what's a "width rung"?
It it that it's aligned on the x-axis?

packages/web/app/src/components/organization/members/list.tsx
```diff
@@ -390,14 +376,17 @@ function MemberActionsCell(props: {
...
+  // Bumped once the picker has closed, so the next open starts from the member's current role.
+  const [pickerSession, setPickerSession] = useState(0);
   const [deleteMemberState, deleteMember] = useMutation(OrganizationMemberRow_DeleteMember);
 
-  if (!member.viewerCanRemove) {
-    return null;
-  }
+  const isProvisioned =
+    member.user.provisionInfo?.provisioningStatus === GraphQLSchema.ProvisioningStatus.Active;
+  const canChangeRole = organization.viewerCanAssignUserRoles && !member.isOwner && !isProvisioned;
```
jdolle, line 386: also unexpected... but looks like this fixed a bug?

packages/web/app/src/components/organization/settings/personal-access-tokens/create-personal-access-token-sheet-content.tsx
```diff
@@ -186,340 +188,279 @@ export function CreatePersonalAccessTokenSheetContent(
       });
       return;
     }
+    if (result.data?.createPersonalAccessToken.ok) {
+      props.onSuccess(result.data.createPersonalAccessToken.ok.privateAccessKey);
```
jdolle, line 190: i think i preferred calling onSuccess in the hook instead of watching a prop. This could have strange interactions if the component is un-rendered

packages/web/app/src/components/organization/settings/access-tokens/create-access-token-sheet-content.tsx
```diff
@@ -194,339 +196,281 @@ export function CreateAccessTokenSheetContent(
       });
       return;
     }
+    if (result.data?.createOrganizationAccessToken.ok) {
+      props.onSuccess(result.data.createOrganizationAccessToken.ok.privateAccessKey);
```
jdolle, line 198: ^2

jdolle (commented):
made it about 1/3 through the review so far

packages/web/app/src/components/base/overlays/dialog/dialog.spec.tsx
jonathanawesome: Means "step on the ladder/size scale/variant". Confusing, will rename.

packages/web/app/src/components/organization/members/list.tsx
jonathanawesome: Yep. Provisioned users can't be modified, the previous "change" link never worked...so we gate on it here.

jdolle (approved):
I tried fixing a few very subtle and minor bugs here https://github.com/graphql-hive/console/pull/8519
everything else is looking good to me now

packages/web/app/src/components/organization/settings/personal-access-tokens/create-personal-access-token-sheet-content.tsx
jonathanawesome: Which hook?

The older version of this used to keep the Sheet open while the ConfirmationDialog displayed on top of it. Now, the Sheet slides out and the Dialog fade in `onSuccess`.

PR #8509, opened 2026-09-17
Consolidate Toast

jonathanawesome:
This PR migrates `ui/toast`, `ui/toaster`, `ui/use-toast` and the react-toastify `useNotifications` helper over to a new base Toast on Base UI. The call shape stays `toast({ title, description, variant })`, so the 174 existing calls are an import-path change, and the ten react-toastify callers become the same call. 

Toasts draw with a status icon, stack up to three in the corner behind the newest and expand on hover, leave after five seconds unless they report a failure, which stays until closed, and can be swiped away. The provider is mounted once in the router in place of both old containers.

The three old files and the helper are deleted along with `@radix-ui/react-toast` and `react-toastify`. 

**Found along the way**
- The registry token dialog fired a "Token created" toast and then opened the created-token dialog; the dialog is the confirmation, so the toast is gone.
- The theme provider imported the whole hooks barrel for one hook, which became an import cycle through `lib/utils` the moment a hook needed a base component; it imports the hook it uses.
- The support ticket reply reported failures as one "Failed to reply: message" string; the message is the description now.


packages/web/app/src/components/base/toast/toast.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+    expect(within(toast).getByText('Failed to save')).toBeTruthy();
+    expect(within(toast).getByText('Name is taken.')).toBeTruthy();
+
+    act(() => {
+      fireEvent.click(toast.querySelector('button[aria-label="Close"]')!);
+    });
+    expect(screen.queryByRole('alertdialog', { hidden: true })).toBeNull();
+  });
+
+  it('lets a default toast go on its own', () => {
```
jdolle, line 38:
```suggestion
  it('hides after 5 seconds if not interacted with', () => {
```
I think this is what you mean?

packages/web/app/src/components/base/toast/toast.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <ToastProvider>
+        <Fire title="Saved" />
+      </ToastProvider>,
+    );
+    act(() => {
+      fireEvent.click(screen.getByRole('button', { name: 'Fire' }));
+    });
+    expect(screen.getByRole('dialog', { hidden: true })).toBeTruthy();
+    act(() => {
+      vi.advanceTimersByTime(5500);
```
jdolle, line 50:
```suggestion
      vi.advanceTimersByTime(5001);
```

packages/web/app/package.json
jdolle, line 1: i like this :)

jdolle (approved):

PR #8511, opened 2026-09-18
Consolidate tabs

jonathanawesome:
This PR migrates `ui/tabs` onto three base components on Base UI. 

Valid uses of `tabs` were migrated 1:1 with only slight visual changes (ie animated active indicator).

`TabbedView` is a new wrapper component for the previously-nested tab views on checks and history. 

`SecondaryNavigation` is a `<nav>` of router links replacing the old bits, which were navigation styled as tabs.

`ui/tabs.tsx`, `navigation/secondary-nav-link.tsx` and `@radix-ui/react-tabs` are deleted.

**Found along the way**
- A global Monaco reset, `[role='presentation'] { background: transparent !important }`, cleared the background of every presentation-role element on the page.
- The new-proposal page kept its Submit Proposal button inside the tab list. It moved to the page heading's side slot.
- The proposal editor let you rename a new service through an input inside the tab button. Renaming moved to the settings panel next to the service URL, with a name conflict shown on the field and as a tooltip on the tab. This settings panel is also now a proper popover, not handrolled.
- The checks and version pages kept a contract selection from the previous check or version, and a view choice carried across contracts even where that view was disabled. Both reset now.
- The layout bars' links were marked current by the router's prefix match as well as by the page value, so Overview and Schema read as current on every page in their area. Links match exactly now.
- Tab strips scroll sideways once their tabs outgrow them, as the old editor strip did. Tabs get a visible focus ring again.
- Every Select in the app has an accessible name now: a visible label pointing at the trigger where one exists, `aria-label` where none does. The trigger is a combobox, which takes no name from its text.
- Keyboard focus is settled by shape. Controls with an edge keep the 2px solid accent ring; tabs and the layout bars' links share a quiet 1px dotted one; the legacy `--ring` token follows the accent so the old button matches until it goes. 

**Checks & History updates**
- The meta cards at the top of both pages are `DescriptionList` now, which gained a `title` term style (12px medium terms over 13px values) and auto-fit columns.
- The check and version pages now name the failed contracts in a new base `FailureCard`.
- A new base `Legend` explains the status icons
- The content under the tabs on both pages, and the shared changes list, use the `critical` color token instead of raw red shades. 
- New seed for locally verifying checks and history: `TARGET=<target id> pnpm seed:history` replays a schema history against any target: versions and checks over six weeks, and on a federation project four contracts with a failed contract version and checks that fail only on a contract. 
- A contract whose baseline failed to compose now shows as failed in the picker, not passed, and a service can be removed from a proposal from the keyboard, through the settings popover.


jdolle (approved):
Looking good overall.

I've added some recommended changes to https://github.com/graphql-hive/console/pull/8525

Most are small bugs around state or very minor visual things. Nothing major, so I'm approving this PR.

n1ru4l (approved):

PR #8512, opened 2026-09-18
Make superadmin's foreign-org debug actions configurable

### LINKED ISSUE #8506: hive platform superadmin permissions/role
For hive platform operators we have a hidden feature that grants read-only access to other organizations. It is enabled by setting a boolean value on the users Postgres database table.

https://github.com/graphql-hive/console/blob/16f59a6b947b2f2f66234ab033747427691be837/packages/services/api/src/modules/auth/lib/supertokens-strategy.ts#L102-L127

Today, the super admin cannot modify foreign organizations. In the context of Hive Cloud that is the desired state. But other self-hosters might model their departments as dedicated hive organizations and want their superadmins be able to perform administrative tasks in all organizations without assigning multiple dedicated accounts for each.

We need to think about how we can have those platform operators specify permissions (or a role) that is used for organizations accessed through the superadmin feature.

Abhist17:
Fixes #8506.

## What was broken

`SuperTokensCookieBasedSession.loadPolicyStatementsForOrganization` hardcoded the policy granted to admins debugging an organization they aren't a member of: always exactly `*:describe` + `alert:modify`, with no way for a self-hoster to widen or narrow it.

## What changed

Per the discussion on #8506 (approach 1, with the maintainer's follow-up that the default should be `*:describe` only — `alert:modify` was only ever added for testing):

- Added `SUPERADMIN_FOREIGN_ORGANIZATION_ACTIONS` to `packages/services/server/src/environment.ts`: a comma-separated list of action names, defaulting to `*:describe` when unset, parsed into `env.auth.superadminForeignOrganizationActions`.
- Threaded it through `SuperTokensUserAuthNStrategy` → `SuperTokensCookieBasedSession` as a constructor dep, replacing the hardcoded two-entry array with `this.superadminForeignOrganizationActions.map(...)`.
- Self-hosted behavior is unchanged unless the env var is explicitly set (previous default included `alert:modify`; new default is `*:describe`-only per the maintainer's comment).

## Tests

Added `supertokens-strategy.spec.ts` covering the changed policy-resolution branch directly: default single-action grant, multi-action grant when configured, and that a non-admin with no membership still resolves to no permissions regardless of configuration (existing behavior, unchanged).

## Verification

- `vitest run` on the new spec + the sibling `authz.spec.ts`: 12/12 pass.
- `tsc --noEmit` clean on both `packages/services/api` and `packages/services/server`.
- `eslint` clean on all four changed/added files.


Abhist17:
Pushed two updates:
- Fixed a Prettier formatting issue in `supertokens-strategy.spec.ts` that was failing `lint:prettier` (import order + one line-length wrap).
- Rebased onto latest `main` to rule out stale-branch drift.

The `test / integration` and `test / e2e` failures on the previous run trace to a TypeScript error in `packages/libraries/laboratory/src/lib/monaco-graphql.spec.ts:47` — a file this PR doesn't touch (this PR only changes `supertokens-strategy.{ts,spec.ts}`, `environment.ts`, and `index.ts`). `main`'s own latest CI is green and didn't touch that file either, so this looks like an unrelated/flaky failure rather than something this change caused. Flagging in case it needs a rerun once CI is approved — happy to dig further if it's actually related.

n1ru4l (changes requested):
@Abhist17 can you also incroporate the requirements here:
- https://github.com/graphql-hive/console/issues/8506#issuecomment-5731318118
- https://github.com/graphql-hive/console/issues/8506#issuecomment-5732219201

Basically add `SUPERADMIN_ORGANIZATION_ID`, and only resolve those permissions for foreign organizations for all users that were provisioned by that organization?

PR #8518, opened 2026-09-18
feat: graph entity

n1ru4l:
TBD


packages/services/api/src/modules/schema/providers/schema-manager.ts
```diff
@@ -469,97 +462,6 @@ export class SchemaManager {
...
 
-  @traceFn('SchemaManager.createVersion', {
-    initAttributes: input => ({
-      'hive.target.id': input.targetId,
-      'hive.organization.id': input.organizationId,
-      'hive.project.id': input.projectId,
-      'hive.version.commit': input.commit,
-      'hive.version.valid': input.valid,
-      'hive.version.service': input.service?.name || '',
-    }),
```
n1ru4l, line 480: this proxy method is not needed, we can call the store directly from the publisher

PR #8520, opened 2026-09-19
Consolidate table review

jdolle:
Follow-ups from the DataTable review

Fixes and small additions found while reviewing #8504.

### DataTable
- Link, copy and link-out cells stop click propagation, so clicking one inside a clickable row no longer also triggers the row.
- Stripes are counted over data rows, so expanding a row no longer shifts the stripes below it.
- Rows can take a `critical` state, which tints the row. Avatar cells can strike through the name. (This adds back some functionality that was lost)
- Number cells gain `format="compact"` ("1.2M"), and percent values always show two decimals. (This adds back existing formatting)
- Pressing Enter in the "Go to page" field no longer submits an enclosing form.
- Badge cells key by index, and a link-out cell with no targets renders only its label. (edge case rendering bug if we have repeat badges)

### Elsewhere
- **Members list:** restores the disabled and provisioned user icons, the row tint and struck-through name for disabled users, and makes the owner tooltip keyboard reachable.
- **Insights:** Requests and Impact use the compact number format again.
- **Permission tables** truncate long titles instead of clipping them.
- **Roles and the check page** usage panels have their own empty messages.
- **Alert rules:** clicking the destinations info icon no longer opens the rule.
- **Misc Memoization:** Traces, members, roles, invitations, OIDC and the token tables memoize their columns and page rows so list memoization works.
- Removes the unused `Status` and `Priority` support components.

**Tests**
- Specs for link clicks inside clickable rows, compact and percent formats, stripe parity across an expanded row, the critical tint, and the empty link-out case.


packages/web/app/src/components/base/data-table/data-table-components.tsx
```diff
@@ -49,13 +49,13 @@ const surface = {
...
     wrapper: 'bg-neutral-1 dark:bg-neutral-2',
     head: 'bg-neutral-2 dark:bg-neutral-3',
-    stripe: 'even:bg-neutral-2/60 dark:even:bg-neutral-3/60',
+    stripe: 'bg-neutral-2/60 dark:bg-neutral-3/60',
     hover: 'hover:bg-neutral-3 dark:hover:bg-neutral-4',
   },
   raised: {
     wrapper: 'bg-neutral-2 dark:bg-neutral-3',
     head: 'bg-neutral-3 dark:bg-neutral-4',
-    stripe: 'even:bg-neutral-3/60 dark:even:bg-neutral-4/60',
```
jdolle, line 58: fixed using logic below to avoid a case where if a cell is added (e.g. "click and expand this row") it would cause all rows after it to flip the background color

PR #8523, opened 2026-09-20
Consolidate form

jonathanawesome:
This PR moves every form in the app onto one base `Form` on react-hook-form, gives base `Label` a text-only API the form label builds on, and deletes `ui/form`, `ui/label`, `formik`, `yup`, `json-schema-yup-transformer` and `@radix-ui/react-label`.

Every form is now a presentational component in its own file, taking `form` and `onSubmit`, with its zod schema and values type exported beside it and a spec of its own. The page keeps the query, the mutation and the toasts. The nine Formik forms were converted with their spec written and passing against the Formik version first, then kept green through the conversion.

The form call-sites preview under Base/FormControls/Form/Component Examples mounts every real form as the regression fixture, sized to the overlay each one ships in.

**Found along the way**
- The schema policy form's JSON Schema validation never ran. The policy service wraps each rule's config schema in an array, the yup transformer returns nothing for that root, and yup treats the missing field schema as no check. Rule configs are validated by the policy service with Ajv on save, as they always were, and the form now carries a plain zod schema.
- Boolean rule options showed unchecked when their default was true: the checkbox rendered before the default was seeded and Base UI locks it uncontrolled on first render.
- The conditional breaking changes form never displayed its "pick at least one target" error under Formik; it does now.
- A tooltip trigger inside a dialog was portaled as a flex item of the popup and shifted it by the gap on hover. Dialogs and sheets now provide a zero-height mount out of the flow.


jdolle (approved):
Reviewed with suggestions here https://github.com/graphql-hive/console/pull/8526

n1ru4l (approved):

PR #8524, opened 2026-09-20
Consolidate button

jonathanawesome:
This PR moves every button in the app onto base `Button`, on Base UI's render hook, and deletes `ui/button`, `@radix-ui/react-slot`, `@radix-ui/react-icons` and `react-icons`.

Every Radix and react-icons icon is on lucide. The brand marks lucide does not carry, or has deprecated, live in `ui/brand-icon.tsx`: GraphQL, Hive, GitHub, Google, Okta, OpenID and Slack. The twenty hand-rolled icons in the old `ui/icon.tsx` that duplicated lucide glyphs are gone with it.

Four things rendered the old Button with most of its defaults undone and are plain elements with their own classes now: the vertical sub-page nav item (also used by the target alerts page), the stepper's step circles, the calendar's nav arrows and day cells, and the explorer's A–Z letter strips.

**Found along the way**
- The "last used" shimmer on the sign-in page merged a gradient class onto the button through a Slot. It is an overlay on a wrapper now, so it works on any fill.
- The changelog trigger's pulse dot and the `CopyText` hover button both relied on classes the base Button does not accept; both sit on wrappers now.
- The stepper's four `asChild` slots had no caller and are gone.

**Design changes to review**
- Every button is a rung smaller: 36px where it was 40px, and dialog footers lose their 44px.
- The accent orange fill is gone; the loud action in a form is the neutral-12 primary.
- Destructive is the tinted, bordered base style rather than a solid red fill.


packages/web/app/src/components/base/floating/tooltip/tooltip.call-sites.preview.tsx
```diff
@@ -198,18 +199,19 @@ function LetterStrip() {
         <Tooltip
           key={letter}
           trigger={
-            <Button
+            <button
```
jdolle, line 202: did you mean for this to be a lowercase button?

jdolle (approved):
Added regression for https://github.com/graphql-hive/console/pull/8527

Otherwise this is looking good

packages/web/app/src/components/base/floating/tooltip/tooltip.call-sites.preview.tsx
jonathanawesome: Not for this preview, specifically. This is the letter strip on Explorer unused/deprecated. The letters are buttons, but they don't fit our Button component variants so these are styled independently.

n1ru4l (approved):

PR #8529, opened 2026-09-22
Routing and layout improvements

jonathanawesome:
This PR reshapes routing in `console` around three ideas: (1) the organization, project and target routes own their layouts and render them around an `<Outlet/>`, so pages are content only and the header and nav mount once instead of on every "tab" switch; (2) navigation is a list of router `Link`s and the router marks the current one from the URL (one `Navigation` component, no `page`/`value` props, no pathname parsing); (3) the tertiary navs (target/org/project settings, members, alerts) are child routes instead of `?page=` switching a render branch, with the default section at the bare URL (`/settings` is General, `/view/members` is the list, `/alerts` is Activity). 

Patterns and recipes are in `packages/web/app/docs/ROUTER.md`; two ESLint import bans enforce them. `src/router.tsx` is split into `src/routes/*` with a `tree.ts` mirroring `getParentRoute`.

Also upgrades `@tanstack/react-router` 1.34.9 → 1.170.38 via the pnpm catalog, with `zod-adapter` 1.167.0; drops the legacy `router-devtools` and the unused `react-form`.

#### URL changes (old URLs redirect, cataloged in `src/routes/legacy.ts`)

- `/trace/$traceId` → `/traces/$traceId`
- `/view/manage-subscription` → `/view/subscription/manage`
- `/alerts/activity` → `/alerts`
- `?page=<section>` on the three settings screens and members → `/<section>` paths (`?page=general` and `?page=list` → the bare URL; members keeps its search params)
- the History item links to `/history`, whose index redirects to the latest version. 
- Every entry carries `since`/`why` and an example that `legacy.spec.ts` drives
- redirects run in `beforeLoad` with `replace`, so old URLs never enter history.


jdolle (approved):
make sure tests pass

